PR #6, opened 2020-06-23
Adds ActionCable, AnyCable and React

iHiD:
This links ActionCable (Rails websockets framework), AnyCable (a websockets server that (sort of) sits in front of Rails), and React. It then tests things with a IterationSummary component (which will be the first one I actually build) which can be updated by firing:
```
IterationsSummaryTableChannel.broadcast! Solution.first
```

I've opened this for you to see how things are developing before it gets so deep to be impenetrable.

A bazillion credits to @SleeplessByte for telling me the various things I needed to google re React, providing explanations, and telling me about AnyCable.

(cc @dougal too as I discussed this with him)


Gemfile
```diff
@@ -9,6 +9,7 @@ gem 'mysql2', '>= 0.4.4'
 gem 'puma', '~> 4.1'
 gem 'redis', '~> 4.0'
 gem 'aws-sdk-s3', '~> 1'
+gem "anycable-rails", "~> 0.6"#, "~> 1.0.0.rc1"
```
iHiD, line 12: @ErikSchierboom @kntsoriano Read about AnyCable.

Procfile.dev
```diff
@@ -0,0 +1,3 @@
+server: bin/rails server
+assets: bin/webpack-dev-server
+anycable: bundle exec anycable --server-command "anycable-go --host=localhost --port 3334"
```
iHiD, line 3: The Procfile runs these in parallel

app/channels/iteration_channel.rb
```diff
@@ -0,0 +1,10 @@
+class IterationChannel < ApplicationCable::Channel
+  def subscribed
+    iteration = current_user.iterations.find(params[:iteration_id])
+    stream_for iteration
```
iHiD: One channel per iteration. This will probably get 1-1'd to the React component, so `IterationSummary`

app/views/layouts/application.html.haml
```diff
@@ -5,6 +5,7 @@
     %title Exercism
     = csrf_meta_tags
     = csp_meta_tag
+    = action_cable_meta_tag
```
iHiD, line 8: This sets the websockets server url (in this case `ws://localhost:3000/cable`)

app/views/pages/index.html.haml
```diff
@@ -1 +1,3 @@
 .hello hello 
+
+%div{data: {react: {"iteration-summary": true, data: {id: 1}}}}
```
iHiD: I'll add a Rails helper for this and standardise things.

config/application.rb
```diff
@@ -1,6 +1,21 @@
...
+require "action_controller/railtie"
+require "action_mailer/railtie"
+require "action_mailbox/engine"
+require "action_text/engine"
+require "action_view/railtie"
+require "action_cable/engine"
+require "rails/test_unit/railtie"
+
+# Remove Rails things we don't use
+# require "sprockets/railtie"
```
iHiD, line 18: We're going do everything through webpacker, so no sprockets.

config/environments/development.rb
```diff
@@ -1,4 +1,9 @@
 Rails.application.configure do
+  # Specify AnyCable WebSocket server URL to use by JS client
+  config.after_initialize do
+    config.action_cable.url = ActionCable.server.config.url = ENV.fetch("CABLE_URL", "ws://localhost:3334/cable") #if AnyCable::Rails.enabled?
```
iHiD, line 4: The commented out bit will be re-added when we move to AnyCable v1.

SleeplessByte (approved):
Is this approving myself?

README.md
```diff
@@ -43,13 +47,15 @@ GRANT ALL PRIVILEGES ON `exercism_v3_test-0`.* TO 'exercism_v3'@'localhost';
...
+
+We have a Procfile which executes the various commands need to run Exercism locally.
+On MacOSX we recommend using `hivemind` to manage this, which can be installed via `brew install hivemind`.
 
-Run both a webpack server and a Rails server
+To get everything started you can then run:
 
 ```bash
-bundle exec bin/webpack-dev-server 
-bundle exec bin/rails server
```
SleeplessByte, line 52: Maybe still keep this if people don't have / can't have hivemind :)

app/channels/iteration_channel.rb
SleeplessByte: They're pretty cheap, but you could consider per student+exercise slug 🤷‍♂️ . Saves you from reconnecting per iteration, then again, this works fine :)

README.md
iHiD: Yeah. I might add a bit that says "read the Procfile.dev and do the steps manually".

kntsoriano (approved):
LGTM! The last time I've used React was years ago, and there were no React Hooks at that time yet. We were just using Redux. Really excited to see how this works out.

When I was doing something like this, too, it was hard for me to set up E2E tests via selenium which I think we would probably need in the future. We also need to set up unit tests for our components so excited to see how we'll manage that.

Gemfile
kntsoriano: I've been using AnyCable for my virtual club!

ErikSchierboom: Sounds great! AnyCable is used to optimize the performance of WebSockets, is that correct?

app/channels/iteration_channel.rb
iHiD: Yeah. So I started there with an `IterationsList` component and `IterationsChannel`. I'm sort of 50/50 on it. My plan was still to have an `IterationsList` component+channel for if new iterations get submitted. But I guess I could hoist the channel so it's shared between the list, and just put logic in the `data`.

config/environments/development.rb
iHiD: See https://github.com/anycable/anycable-rails/issues/132 for context.

iHiD:
@kntsoriano Yeah - I'll get tests etc added to this PR. I intend for this branch to get to a "complete" state in terms of all these bits before I merge it. I'll then follow up with another minimal PR focussed on adding a new component, which we can use as a reference.

app/channels/iteration_channel.rb
iHiD: Changed to a solution-level component as per this.

test/system/react_components/iterations_summary_table_test.rb
```diff
@@ -0,0 +1,19 @@
+require "application_system_test_case"
```
iHiD, line 1: For each component, we can have a system test like this which tests functionality.

config/routes.rb
```diff
@@ -1,3 +1,13 @@
...
+  
+  mount ActionCable.server => '/cable'
+
+  unless Rails.env.production?
+    namespace :test do
+      namespace :components do
+        get 'iterations_summary_table/:id', to: 'iterations_summary_table#index', as: 'iterations_summary_table'
+      end
+    end
+  end
```
iHiD, line 12: These will be routes that are used by the tests. The idea is that each view will just render the relevant component and then the relevant test can test interactions with that test. I've also opened them to dev so that you can see and play with the components in dev mode, without having to find them in context.

This should all encourage good decoupling of components from each other and other view-level dependencies.

app/views/test/components/iterations_summary_table/index.html.haml
```diff
@@ -0,0 +1 @@
+%div{data: {react: {"iterations-summary-table": true, data: {solution_id: @solution.id, iterations: @solution.serialized_iterations.to_json}}}}
```
iHiD: See note in the routes below. This will be refactored into a component helper next.

Gemfile
iHiD: @ErikSchierboom Yes. It basically removes Ruby from the server/connection bit and handles that with something more speedy (Go). It then connects to the Rails app via Redis pub/sub to work out what needs sending where.

test/javascript/components/iteration_summary.test.js
```diff
@@ -0,0 +1,20 @@
...
+test("renders component", () => {
+  const { container, getByText } = render(
+    <IterationsSummaryTable
+      solutionId={1}
+      iterations={[{ id: 1, testsStatus: 'passed' }]}
+    />
+  );
+
+  expect(getByText("1: passed")).toBeInTheDocument();
+});
```
iHiD, line 20: This is really just a stub to have jest working up. We can use jest to check any unit level things (for fast iteration), but my preference is to just do end-to-end testing of components via Rails system tests (below).

iHiD:
I added a couple of extra commits (and a little more commentary). Thanks some more to DJ for insight :) I was working on getting websockets testing working via the Jest test but in the end after many hours have given up for now.

config/routes.rb
```diff
@@ -1,3 +1,13 @@
 Rails.application.routes.draw do
   root to: "pages#index"
+  
+  mount ActionCable.server => '/cable'
```
iHiD, line 4: @kntsoriano Do I need this if I use `AnyCable`?

test/javascript/components/iteration_summary.test.js
SleeplessByte: >  end-to-end testing of components

I recommend.

PR #9, opened 2020-06-28
Add husky

iHiD:
@kntsoriano @ErikSchierboom This adds pre-commit support (via husky) to run RuboCop on Ruby files and Prettier on everything else.


app/controllers/pages_controller.rb
```diff
@@ -1,6 +1,7 @@
 class PagesController < ApplicationController
   def index
     solution = Solution.first
+
     IterationsChannel.broadcast!(solution)
   end
-end
+end
```
kotp, line 7: Trailing white space introduced here.

kotp:
Well, I was in the middle of the review… too late now. :rofl: Missed it by seconds.

iHiD:
@kotp Sorry :) Feel free to leave comments :)

app/controllers/pages_controller.rb
iHiD: Hmmm. I don't understand why this didn't get picked up by CI. More debugging needed!

app/controllers/pages_controller.rb
```diff
@@ -1,6 +1,7 @@
 class PagesController < ApplicationController
   def index
     solution = Solution.first
+
     IterationsChannel.broadcast!(solution)
```
kotp, line 5: Unless `solution` is used more than once, this is a documenting variable.

`IterationsChannel.broadcast!(Solution.first)` is very clear without it.

PR #12, opened 2020-07-01
Run the website in Docker

ErikSchierboom:


Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ruby:2.6.6
```
iHiD, line 1: Is this Ubuntu?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    gunzip > /usr/local/bin/hivemind && \
+    chmod u+x /usr/local/bin/hivemind
+
+WORKDIR /usr/src/app
+
+COPY Gemfile Gemfile.lock package.json yarn.lock ./
+
+RUN yarn install && \
+    gem install bundler && \
+    bundle install
```
iHiD:
```suggestion
    bundle install && \
    yarn install
```

docker-compose.yml
```diff
@@ -16,10 +14,27 @@ services:
...
+  adminer:
+    image: adminer:4.7.7-standalone
+    depends_on:
+      - mysql
+    environment:
+      ADMINER_DEFAULT_SERVER: mysql
+    ports:
+      - 8080:8080
+
+  website:
```
iHiD: How does this know what "webiste" is.

Dockerfile
ErikSchierboom: Debian I believe

docker-compose.yml
ErikSchierboom: This is just a label used internally.

ErikSchierboom: You can think of this as the internal DNS name for the container.

config/anycable.yml
```diff
@@ -9,8 +9,7 @@
 #
 # Read more about AnyCable configuration here: https://docs.anycable.io/v1/#/ruby/configuration
 #
-default: &default
-  # Turn on/off access logs ("Started..." and "Finished...")
+default: &default # Turn on/off access logs ("Started..." and "Finished...")
```
iHiD, line 12: Revert

Procfile.dev
```diff
@@ -1,4 +1,4 @@
-server: bin/rails server
+server: MYSQL_HOST=mysql MYSQL_SOCKET= bin/rails server -b '0.0.0.0'
 assets: bin/webpack-dev-server
 anycable: bundle exec anycable --server-command "anycable-go --host=localhost --port 3334"
 js-tests: yarn test-watch
```
iHiD: Add `Procfile.docker.dev` with these changes in, and leave this file untouched.

docker/init.sh
```diff
@@ -0,0 +1,3 @@
+bundle install &&
+yarn install &&
+hivemind -p 3000 ./Procfile.dev
```
iHiD:
```suggestion
hivemind -p 3000 ./Procfile.docker.dev
```

Dockerfile
iHiD: Could you try switching this to `2.6.6-alpine3.12` pls :)

docker/init.sh
```diff
@@ -1,4 +1,3 @@
-yarn install --check-files
```
iHiD: I wonder if we still need this here. Because if you rerun the machine after it's built, but the package.json has changed upstream on git, this will need to run else the things below will fail.

docker-compose.yml
```diff
@@ -29,6 +29,7 @@ services:
       - 3000:3000
     volumes:
       - ./:/usr/src/app
+      - /usr/src/app/node_modules
```
iHiD: What does this line do?

Gemfile
```diff
@@ -4,7 +4,8 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby '2.6.6'
 
 # Service/framework dependencies
-gem 'rails', github: "rails/rails", branch: "master" # , '~> 6.0.2', '>= 6.0.2.2'
+# gem 'rails', github: "rails/rails", branch: "master" # , '~> 6.0.2', '>= 6.0.2.2'
+gem 'rails', '~> 6.0.2', '>= 6.0.2.2'
```
iHiD: Can you also see if this works on master again now pls? I don't massively want to downgrade to `6.0`

db/schema.rb
```diff
@@ -2,8 +2,8 @@
 # of editing this file, please use the migrations feature of Active Record to
```
iHiD: This file should be reverted too. This is an old schema from rails 6.0, not the new 6.1 one.

docker-compose.yml
```diff
@@ -2,24 +2,50 @@ version: "3"
...
+    restart: always
+    environment:
+      MYSQL_HOST: mysql
+      MYSQL_SOCKET:
+      MYSQL_PORT: 3306
+      MYSQL_USERNAME: root
+      MYSQL_PASSWORD: password
+      REDIS_HOST: redis
+    ports:
+      - 3000:3000
```
iHiD:
```suggestion
      - 5000:5000
```

docker/init.sh
```diff
@@ -0,0 +1,3 @@
+bin/rails db:create
+bin/rails db:migrate
+hivemind -p 3000 ./Procfile.docker.dev
```
iHiD, line 5:
```suggestion
hivemind -p 5000 ./Procfile.docker.dev
```

Procfile.docker.dev
```diff
@@ -0,0 +1,4 @@
+server: bin/rails server -b '0.0.0.0'
```
iHiD, line 1:
```suggestion
server: bin/rails server -b '0.0.0.0' -p 5000
```

iHiD: It would be nice if this ran on a different port to a local rails server

ErikSchierboom: We don't need to change it here as this is all Docker internals. The only exposed port is defined in the `docker-compose.yml` file, which I've changed to expose the website at port 5000

docker/init.sh
ErikSchierboom: See comment below about exposing ports in `docker-compose.yml`

docker-compose.yml
ErikSchierboom: I've changed this to `5000:3000`, to keep the website internally at the default port. This is usually how this is done.

Gemfile
ErikSchierboom: The problem was not with it not working, but with the download being horribly slow and flaky on my Docker for Windows setup. I'll revert.

docker-compose.yml
ErikSchierboom: It prevents the local `node_modules` folder from being mapped into the container, which causes issues

docker/init.sh
ErikSchierboom: We don't. If you pass `--build`, it will automatically rebuild if the `package.json` file changes.

config/anycable.yml
ErikSchierboom: I can't, prettier changes this automatically

PR #14, opened 2020-07-21
Use website on 3020 locally and via Docker

iHiD:
I'm starting to map out how everything fits together. I'm going to run things from port 3020-3040 if no-one tells me that's a bad idea. I've also decided that the Dockerfile and local systems should both end up on the same port so that other local systems can reliably chat to them.

### DIFF
--- a/DEV_ENVIRONMENT.md
+++ b/DEV_ENVIRONMENT.md
@@ -1,0 +1,23 @@
+# Development Environment
+
+Exercism is a large open-source project. It is made up of over 250 repositories. Creating a full local setup is entirely manageable, but will involve some effort. This guide talks you through the steps needed. In many cases it will defer to sub-project repos.
+
+## Exercism Website
+
+The world-facing website.
+
+- **Repo:** [exercism/v3-website](https://github.com/exercism/v3-website)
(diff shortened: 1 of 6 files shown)


SleeplessByte:
If all you need is a port range, which will be automated, and not necessarily hit by you (lower numbers are just better DX), then there is only one reason why you would not want `3020-3040`.

If we look at the [List of TCP and UDP port numbers](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers), we can see one collision. If you look at [speedguide](https://www.speedguide.net/port.php?port=3030), there is at least one collision for each port in this range, altho I am pretty certain **you don't run any of those applications**.

User ports are 1024-49151. Pick anything that's free. Just [`netstat`](https://www.howtogeek.com/513003/how-to-use-netstat-on-linux/) and make sure you're safe, and then it's all fine.

ErikSchierboom (approved):
LGTM

ErikSchierboom:
As @SleeplessByte, port collisions are the only possible reason not to use the low range, but it does indeed look like the range is not used by any applications we'll be running.

iHiD:
Sweet. Thanks y'all! :)

PR #15, opened 2020-07-22
Install dependencies at build-time in Dockerfile

ErikSchierboom:
This helps the startup time when none of the dependencies have changed.


iHiD:
Will this mean the Dockerfile needs building every time before it's used?

ErikSchierboom:
> Will this mean the Dockerfile needs building every time before it's used?

Nope. Docker will recognize if there are changes in the files, and if not, just re-use the old version. And this makes sure that booting is quick after building.

Dockerfile
```diff
@@ -13,6 +13,12 @@ RUN curl -L -o /usr/local/bin/anycable-go https://github.com/anycable/anycable-g
 
 RUN gem install bundler
 
+COPY Gemfile Gemfile.lock ./
+RUN bundle install
+
+COPY package.json yarn.lock ./
+RUN yarn install --check-files
+
```
iHiD: This is now on master.

ErikSchierboom: Nice

Dockerfile.dev
```diff
@@ -13,12 +13,12 @@ RUN curl -L -o /usr/local/bin/anycable-go https://github.com/anycable/anycable-g
 
 RUN gem install bundler
 
+WORKDIR /usr/src/app
```
ErikSchierboom, line 16: This caused the "jest not found" message I was referring to yesterday, simply because the `yarn install` would create the `node_modules` directory at the wrong place

config/anycable.yml
```diff
@@ -29,11 +29,17 @@ default: &default # Turn on/off access logs ("Started..." and "Finished...")
 
 development:
   <<: *default
-  redis_url: "redis://<%= ENV.fetch('REDIS_HOST', '127.0.0.1') %>:6379/1"
+  redis_url: "redis://127.0.0.1:6379/1"
   rpc_host: "127.0.0.1:50051"
```
ErikSchierboom, line 33: This RPC host, is that running on the Redis server or locally?

Dockerfile
```diff
@@ -15,12 +15,12 @@ RUN gem install bundler
...
+WORKDIR /usr/src/app
+
 COPY Gemfile Gemfile.lock ./
 RUN bundle install
 
 COPY package.json yarn.lock ./
-RUN yarn install
-
-WORKDIR /usr/src/app
+RUN yarn install --check-files
```
iHiD: I don't think we want to check the files in production.

```suggestion
RUN yarn install
```

config/anycable.yml
iHiD: Locally.

config/database.yml
```diff
@@ -20,6 +20,16 @@ development:
     <<: *default
     database: exercism_v3_dj_development
     migrations_paths: "db/dj_migrate"
+
+development_docker:
```
iHiD, line 24: Is it worth just doing `<<: &development` in here, rather than duplicating?

iHiD (approved):
One suggestion on the duplicated config. Nice work digging through all this :)

config/database.yml
ErikSchierboom: Let me try this!

ErikSchierboom: Just doing `<<: &development` doesn't work, but for the `primary` and `dj` ones it does work

iHiD: Interesting. Cool :)

iHiD (approved):

Dockerfile
```diff
@@ -15,10 +15,14 @@ RUN gem install bundler
...
 WORKDIR /usr/src/app
+
+COPY Gemfile Gemfile.lock ./
 RUN bundle install
+
+COPY package.json yarn.lock ./
 RUN yarn install
 
+COPY . ./
+
```
iHiD, line 27: @ErikSchierboom For this change, does this not only matter if we cache the intermidiate dockerfile stages somewhere. Atm I think they're lost every time. So should we cache these on GitHub's cache somehow?

SleeplessByte (commented):
Always good to see my name

Dockerfile
ErikSchierboom: > Atm I think they're lost every time. So should we cache these on GitHub's cache somehow?

Ah yes, we should. There was mention that GitHub would be implementing this themselves, but they haven't done that so far. There is a bit of complexity involved. Want to do this at a later stage?

PR #17, opened 2020-07-24
Add maintaining/iterations page

ErikSchierboom:


app/javascript/components/maintaining_iterations_summary_table.jsx
```diff
@@ -0,0 +1,52 @@
+import React, { useState, useEffect } from "react";
```
iHiD: I suggest namespacing this file into `components/maintaining/iterations_summary_table`

app/javascript/components/test_iterations_summary_table.jsx
```diff
@@ -1,17 +1,17 @@
 import React, { useState, useEffect } from "react";
-import actionCable from "actioncable";
+import "actioncable";
```
iHiD: Do you need this LOC at all?

app/javascript/components/test_iterations_summary_table.jsx
```diff
@@ -1,17 +1,17 @@
...
 
 import consumer from "../application/action_cable_consumer";
 
-function IterationSummary({ id, testsStatus }) {
+function TestIterationSummary({ id, testsStatus }) {
```
iHiD: Could we rename this to `Example...` rather than `Test...`?

iHiD: (or `Sample...`)

test/system/react_components/maintaining_iterations_summary_table_test.rb
```diff
@@ -0,0 +1,17 @@
+require "application_system_test_case"
```
iHiD: Namespace this into `maintaning/` pls

test/javascript/components/maintaining_iteration_summary_table.test.js
```diff
@@ -0,0 +1,21 @@
+// Deps
```
iHiD: Namespace this one into a subfolder too pls.

docker/init.sql
```diff
@@ -7,6 +8,6 @@ CREATE DATABASE exercism_v3_dj_development;
 ALTER DATABASE exercism_v3_dj_development CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 GRANT ALL PRIVILEGES ON exercism_v3_dj_development.* TO 'exercism_v3';
 
-CREATE DATABASE `exercism_v3_test`;
-ALTER DATABASE `exercism_v3_test` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
-GRANT ALL PRIVILEGES ON `exercism_v3_test`.* TO 'exercism_v3';
+CREATE DATABASE exercism_v3_test;
+ALTER DATABASE exercism_v3_test CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+GRANT ALL PRIVILEGES ON exercism_v3_test.* TO 'exercism_v3';
```
iHiD: Why this change?

app/javascript/components/test_iterations_summary_table.jsx
ErikSchierboom: Yes you do, otherwise the actioncable stuff is not loaded and does not work. I've checked

app/javascript/components/maintaining_iterations_summary_table.jsx
ErikSchierboom: I can do that. I didn't do that because `application.jsx` said:

> // Add all react components here.
> // Each should map 1-1 to a compoenent in app/helpers/components

app/javascript/components/test_iterations_summary_table.jsx
ErikSchierboom: Sure

docker/init.sql
ErikSchierboom: Oh, consistency with the other statements in the file

app/javascript/components/maintaining_iterations_summary_table.jsx
iHiD: Sweet. I think it's the right thing to do, and then we need to work out how to do the actual components at the Ruby side. I think I want to make those as actually classes somehow, not just a million arbitary methods in that helper file (I tend to hate helpers)

app/javascript/components/test_iterations_summary_table.jsx
iHiD: Interesting. Thanks. I dislike that as magic!

docker/init.sql
iHiD: So, the background to this is that you can't have hyphens in db names unless they're surrounded with `. And the test databases get split one per processor, so get named `exercism_v3_test-1`. They used to be specified here but they're (correctly) not now. I think I'd actually rather go the other way and add them everywhere in that file, as it makes things safer to have them in case we ever rename things.

test/system/react_components/maintaining_iterations_summary_table_test.rb
```diff
@@ -0,0 +1,17 @@
...
+    visit maintaining_iterations_summary_table_url()
+    sleep(0.1) # Set the websocket connection
+
+    iteration.tests_passed!
+    iteration.broadcast!
+
+    sleep(0.1) # Wait for websocket to broadcast
+    assert_text "#{iteration.id}"
+  end
+end
```
iHiD: So, general comment on this file. I'd like to test components independently of the pages they are on. The test routes (`app/views/test/...`) aims to do this. So in these system component tests, I'd rather us call the test route and test that component in isolation.

We can then have a second test later that tests the component in the context of the actual eventual page it's on. This test would sit in `test/system/maintaining`. 

Does that make sense?

docker/init.sql
ErikSchierboom: Makes sense! Will do that

app/javascript/components/maintaining_iterations_summary_table.jsx
ErikSchierboom: I am not a fan of helpers either

app/javascript/components/test_iterations_summary_table.jsx
ErikSchierboom: Hah. It's just a way for webpack to know what the component depends on.

app/channels/iteration_channel.rb
```diff
@@ -0,0 +1,16 @@
+class IterationChannel < ApplicationCable::Channel
+  def subscribed
+    stream_from CHANNEL_NAME
```
iHiD, line 3: Why change from `stream_from` to `stream_for`?

app/controllers/maintaining/iterations_summary_table_controller.rb
```diff
@@ -0,0 +1,8 @@
+class Maintaining::IterationsSummaryTableController < ApplicationController
+  def index
+    @iterations = Iteration.order(created_at: :desc).take(NUMBER_OF_ITERATIONS)
+  end
+
+  private
+  NUMBER_OF_ITERATIONS = 50.freeze
```
iHiD: This is a noop. Integers are singleton instances of the Integer class.

```
50.object_id
50.freeze.object_id
```

It's important for Strings, because strings can be mutated, but ints are immutable so don't need freezing.

config/routes.rb
```diff
@@ -3,6 +3,10 @@
 
   mount ActionCable.server => '/cable'
 
+  namespace :maintaining do
+    get 'iterations', to: 'iterations_summary_table#index', as: 'iterations_summary_table'
```
iHiD: So, this should be:
```
resources :iterations, only: [:index]
```

That will then create a route called `maintaining_iterations_path`, which is `/maintaining/iterations`, which maps to the `index` method (and template) in the `maintaining/iterations` controller.

iHiD: You can then change it to `[:index, :show]` to add a `show method` which maps to `/maintaining/iterations/:id`, which will be called `maintaining_iteration_path` (singular) which is a method that expects an id of the object to route to.

iHiD: The code you've got here is good for arbitary one-off routes that don't follow the normal RESTful approach Rails takes (for example, the test routes)

app/channels/iteration_channel.rb
ErikSchierboom: I took that from https://guides.rubyonrails.org/action_cable_overview.html#client-server-interactions. The way I read the docs, was that `stream_for` related to objects. But I might have misread

app/controllers/maintaining/iterations_summary_table_controller.rb
ErikSchierboom: Okay. Will change

test/system/react_components/maintaining/iterations_summary_table_test.rb
```diff
@@ -0,0 +1,17 @@
+require "application_system_test_case"
+
+class Maintaining::IterationsSummaryTableTest < ApplicationSystemTestCase
+  test "visiting the index" do
+    solution = create :concept_solution
+    iteration = create :iteration, solution: solution
+
+    visit test_components_maintaining_iterations_summary_table_url()
```
iHiD:
```suggestion
    visit test_components_maintaining_iterations_summary_table_url
```

config/routes.rb
ErikSchierboom: Nice!

app/channels/iteration_channel.rb
ErikSchierboom: I've checked this locally, and `stream_for` does indeed _not_ work. `stream_from` does.

app/controllers/tmp/tracks_controller.rb
```diff
@@ -0,0 +1,7 @@
+class Tmp::TracksController < ApplicationController
+  def create
+    Track.create!(slug: params[:track_slug], title: params[:track_title], repo_url: "http://github.com/exercism/#{params[:track_slug]}")
```
iHiD, line 3:
```suggestion
    Track.create!(slug: params[:track_slug].underscore, title: params[:track_title], repo_url: "http://github.com/exercism/#{params[:track_slug].dasherize}")
```

PR #21, opened 2020-07-25
Add Tooling Jobs

iHiD:
This is a WIP implementation of adding a job type for our internal tooling (test runners etc), backed by DynamoDB. At the time of writing it integrates with tooling-orchestrator and tooling-invoker successfully.

There's things that are commented out that shouldn't be, some messy bits, and missing tests. All of that will get fixed before I merged, but it's usable for development purposes. Don't worry about reviewing yet.

@ErikSchierboom - I might not get further until mid-next week because I'm quite exhausted so I wanted to put this up so you would have something working end-to-end so you're not blocked. 

~For anyone else reading (esp @saschamann!) the "boffin" thing is temporary :)~ (@saschamann - Gone 🥳 )

This goes with:
- https://github.com/exercism/tooling-invoker/pull/3
- https://github.com/exercism/tooling-orchestrator/pull/3
- https://github.com/exercism/config/pull/1
- https://gist.github.com/iHiD/8496ca580afbca7f73315076235aa519


SaschaMann:
> For anyone else reading (esp @SaschaMann!) the "boffin" thing is temporary :)

Every boffin knows that nothing lasts longer than temporary fixes :P

config/initializers/exercism_config.rb
```diff
@@ -12,3 +12,10 @@
 Exercism.config.test_runner_orchestrator_url = services_config[:test_runner_orchestrator_url]
 Exercism.config.analyzer_orchestrator_url = services_config[:analyzer_orchestrator_url]
 Exercism.config.representer_orchestrator_url = services_config[:representer_orchestrator_url]
+
+Exercism.config.dynamodb_client =
+  Aws::DynamoDB::Client.new(
+    Exercism.config.aws_auth.merge({
+                                     endpoint: "http://localhost:3039"
```
ErikSchierboom: As a note: we should configure the hostname of this endpoint, as a different host name will be used when using docker-compose

scripts/setup_dynamodb_locally.rb
```diff
@@ -0,0 +1,65 @@
...
+      key_type: "HASH", 
+    }
+  ],
+  provisioned_throughput: {
+    read_capacity_units: 1, 
+    write_capacity_units: 1, 
+  }, 
+)
+
+Exercism.config.dynamodb_client.update_table(
```
ErikSchierboom: Interesting. You both create _and_ update the table. Is that because the table creation cannot take all the parameters the `update_table` can?

ErikSchierboom:
The query syntax _is_ weird :D

scripts/setup_dynamodb_locally.rb
iHiD: @ErikSchierboom Yep! Very weird isn't it! 

The "attributes" can only be for the main index or for the secondary, so to add a secondary that doesn't have the same attributes as the main, you have to do two things. I might, of course, be missing something, but this is the only way I could get it working!

config/initializers/exercism_config.rb
iHiD: This has now moved into the Exercism.config, which will be set via the bin script.

.github/workflows/tests.yml
```diff
@@ -2,9 +2,9 @@ name: Tests
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
```
iHiD, line 5: @ErikSchierboom Prettier or Rubocop or something did this. Is it ok?

bin/create_dynamodb_config
```diff
@@ -0,0 +1,79 @@
...
+else
+  settings.merge!(
+    tooling_orchestrator_url: "http://127.0.0.1:3021",
+    anycable_endpoint: '127.0.0.1',
+    anycable_rpc_host: '127.0.0.1',
+    dynamodb_endpoint: "http://localhost:3039",
+    rds_master_endpoint: "localhost",
+    mysql_socket: "/tmp/mysql.sock"
+  )
+end
```
iHiD: @ErikSchierboom This is the key bit. So any of these settings can be used in **any** Exercism application/microservice/etc by doing:
```
Exercism.config.tooling_orchestrator_url
```

The production versions are all set via terraform when the architecture is deployed. So there's no config or yaml or shared settings/secrets anywhere other than in this central db.

This script needs to be called to set-up the development environment. I've got it set up so it runs every time Docker starts, and it's a one-off job if you want to run locally.

The crucial thing to understand for now though, is that if you need to change any settings for Docker/local to work, this is the place to do it.

iHiD: (cc @kntsoriano to this too - I'll talk you through it all properly when it's done and you can put your head in Exercism-space - but nice for you to see now :))

.github/workflows/tests.yml
ErikSchierboom: Yep. Totally fine.

bin/create_dynamodb_config
```diff
@@ -0,0 +1,79 @@
...
+settings = {
+  aws_iterations_bucket: "exercism-staging-iterations",
+  dynamodb_tooling_jobs_table: "tooling_jobs"
+}
+if ENV["EXERCISM_DOCKER"]
+  settings.merge!(
+    tooling_orchestrator_url: "http://tooling-orchestrator:3021",
+    anycable_endpoint: 'redis',
+    anycable_rpc_host: '0.0.0.0',
+    dynamodb_endpoint: "http://localhost:8000",
```
ErikSchierboom: I think this will be:

```suggestion
    dynamodb_endpoint: "http://dynamodb:8000",
```

ErikSchierboom (commented):
Nice!

iHiD:
Pushing this down! Wahoo!

ErikSchierboom:
Nice!

PR #22, opened 2020-07-28
Working Dockerfile

iHiD:
This now almost works ~(see http://webservers-ecs-710327607.eu-west-2.elb.amazonaws.com/ 🦜)~

I don't understand why HiveMind doesn't work in production though. The `docker/init.sh` blows up with no useful info.
Any ideas @ErikSchierboom @SleeplessByte (anyone else)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,4 +25,5 @@ RUN yarn install
 
 COPY . ./
 
-CMD ["sh", "docker/init.sh"]
+#CMD ["sh", "docker/init.sh"]
+ENTRYPOINT RAILS_LOG_TO_STDOUT=true bundle exec bin/rails server -e production -b '0.0.0.0' -p 80
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
-server: bundle exec bin/rails server -e production
-anycable: bundle exec anycable --server-command "anycable-go --host=localhost --port 3334"
+server:      RAILS_LOG_TO_STDOUT=true bundle exec bin/rails server -e production -b '0.0.0.0' -p 80
+anycable-go: RAILS_LOG_TO_STDOUT=true bundle exec bin/rails runner -e production "exec(\"anycable-go --host=0.0.0.0 --port 3334 --redis_url='#{Exercism.config.anycable_endpoint}'\")"
+anycable:    bundle exec anycable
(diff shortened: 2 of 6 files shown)


ErikSchierboom:
> The docker/init.sh blows up with no useful info

Could that be due to the `init.sh` command not having a shebang header (like `#!/usr/bin/env sh`)?

Dockerfile
```diff
@@ -25,4 +25,5 @@ RUN yarn install
 
 COPY . ./
 
-CMD ["sh", "docker/init.sh"]
+#CMD ["sh", "docker/init.sh"]
+ENTRYPOINT RAILS_LOG_TO_STDOUT=true bundle exec bin/rails server -e production -b '0.0.0.0' -p 80
```
ErikSchierboom, line 29: So this is there because the `docker/init.sh` file blows up, right?

Procfile.dev
```diff
@@ -1,4 +1,4 @@
-server: bin/rails server
-assets: bin/webpack-dev-server
+server: bundle exec bin/rails server
```
ErikSchierboom, line 1: The `bundle exec` should probably also be added to `Procfile.docker.dev`.

Procfile
```diff
@@ -1,2 +1,3 @@
-server: bundle exec bin/rails server -e production
-anycable: bundle exec anycable --server-command "anycable-go --host=localhost --port 3334"
+server:      RAILS_LOG_TO_STDOUT=true bundle exec bin/rails server -e production -b '0.0.0.0' -p 80
+anycable-go: RAILS_LOG_TO_STDOUT=true bundle exec bin/rails runner -e production "exec(\"anycable-go --host=0.0.0.0 --port 3334 --redis_url='#{Exercism.config.anycable_endpoint}'\")"
```
ErikSchierboom, line 2: Where is `Exercism.config.anycable_endpoint` defined?

ErikSchierboom (approved):

Procfile
SleeplessByte: My guess is that command is hackingly injecting a ruby string. So it would be defined in one of the initializers.

Dockerfile
iHiD: Correct.

Procfile
iHiD: So this is the `ExercismConfig` gem. It preloads `Exercism.config` with everything you need in any environment, loaded from either local/docker-composed or production dynamodb. Coming in a funky PR shortly.

ErikSchierboom: Okay cool!

PR #23, opened 2020-07-30
Add S3 mock

iHiD:
@ErikSchierboom This adds S3 mocking via a docker image. This means we now need no AWS credentials passed in, and no external dependancies for developing v3. I'm pleased with this :)

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -3,10 +3,10 @@ FROM ruby:2.6.6
 RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
     echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
     curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
-    apt-get install -y cmake nodejs yarn
+    apt-get install -y cmake nodejs yarn tmux
 
-RUN curl -L -o - https://github.com/DarthSim/hivemind/releases/download/v1.0.6/hivemind-v1.0.6-linux-arm64.gz | gunzip > /usr/local/bin/hivemind && \	
-    chmod u+x /usr/local/bin/hivemind
+RUN curl -L -o - https://github.com/DarthSim/overmind/releases/download/v2.2.0/overmind-v2.2.0-linux-arm64.gz | gunzip > /usr/local/bin/overmind && \	
+    chmod u+x /usr/local/bin/overmind
 
 RUN curl -L -o /usr/local/bin/anycable-go https://github.com/anycable/anycable-go/releases/download/v1.0.0/anycable-go-linux-amd64 && \
     chmod u+x /usr/local/bin/anycable-go
@@ -16,10 +16,10 @@ RUN gem install bundler
 WORKDIR /usr/src/app
 
 COPY Gemfile Gemfile.lock ./
-RUN bundle install --full-index
+RUN bundle install
 
 COPY package.json yarn.lock ./
-RUN yarn install --check-files
+RUN yarn install
 
 CMD ["sh", "docker/init.dev.sh"]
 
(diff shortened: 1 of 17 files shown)


Dockerfile.dev
```diff
@@ -3,10 +3,10 @@ FROM ruby:2.6.6
 RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
     echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
     curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
-    apt-get install -y cmake nodejs yarn
+    apt-get install -y cmake nodejs yarn tmux
 
-RUN curl -L -o - https://github.com/DarthSim/hivemind/releases/download/v1.0.6/hivemind-v1.0.6-linux-arm64.gz | gunzip > /usr/local/bin/hivemind && \	
-    chmod u+x /usr/local/bin/hivemind
+RUN curl -L -o - https://github.com/DarthSim/overmind/releases/download/v2.2.0/overmind-v2.2.0-linux-arm64.gz | gunzip > /usr/local/bin/overmind && \	
+    chmod u+x /usr/local/bin/overmind
```
iHiD, line 9: Overmind is hivemind's big sibling.

This allows us to do `docker-compose exec website overmind restart xxx` (where xxx is one of the services that is run via overmind), rather than having to reboot the whole website container.

Dockerfile.dev
```diff
@@ -16,10 +16,10 @@ RUN gem install bundler
 WORKDIR /usr/src/app
 
 COPY Gemfile Gemfile.lock ./
-RUN bundle install --full-index
+RUN bundle install
```
iHiD, line 19: I've taken this out and instead now check for updates to the `exercism_config` gem at boot-time.

app/controllers/tmp/iterations_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Tmp::IterationsController < ApplicationController
```
iHiD, line 1: We can use controllers in this directory for development purposes. They can have zero test coverage and will be thrown away before we launch. But for things like the button I've added below they are very useful for quickly testing things.

app/controllers/tmp/iterations_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  def create
```
iHiD, line 2: This action does the same thing as the `submit_iteration` script but is runnable via clicking a button instead. As neither will ever get near production, I don't mind the duplication for now.

app/views/pages/index.html.haml
```diff
@@ -1 +1,6 @@
-Exercism!
+Hello, World!
+%hr/
+%h2 Temporary useful buttons
+%ul
+  %li
+    = link_to "Create Iteration", tmp_iterations_path, method: :post, remote: true
```
iHiD, line 6: This triggers the action above, which should get end-to-end through the test runner. it's pending one final callback in the orchestrator having the right URL atm. Then it'll be done 👍 

I'm thinking I talk you through adding that config tomorrow to familiarise you more with all this config stuff.

package.json
```diff
@@ -37,7 +37,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged && git diff --name-only --staged | grep '.rb$' | xargs bundle exec rubocop --except Metrics --auto-correct --format quiet app/controllers/application_controller.rb && git diff --name-only --staged | grep '.rb$' | xargs git add"
+      "pre-commit": "pretty-quick --staged && git diff --name-only --staged | grep '.rb$' | xargs bundle exec rubocop --except Metrics --auto-correct --format quiet --force-exclusion app/controllers/application_controller.rb && git diff --name-only --staged | grep '.rb$' | xargs git add"
```
iHiD, line 40: This forces rubocop to honour the Excludes list rather than honouring the files passed on the command line.

Dockerfile.dev
ErikSchierboom: Nice!

app/views/pages/index.html.haml
ErikSchierboom: 👍

PR #26, opened 2020-07-30
Add start of caching support

iHiD:


Gemfile
```diff
@@ -1,10 +1,10 @@
 source 'https://rubygems.org'
-git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby '2.6.6'
 
 # Service/framework dependencies
-gem 'rails', github: 'rails/rails', branch: 'master' # , '~> 6.0.2', '>= 6.0.2.2'
+gem 'rails', git: 'http://github.com/rails/rails.git', ref: "f4bf62ee8e0f8762a82f85a789119b2211fdecb4"
```
iHiD, line 6: I've locked us to a commit for now. I'll probably run against master when I'm working, but I think it's better for not needing to constantly update the gems to fix a commit.

ErikSchierboom: Makes sense

db/dj_schema.rb
```diff
@@ -10,7 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20_200_721_220_344) do
+ActiveRecord::Schema.define(version: 2020_07_21_220344) do
```
ErikSchierboom, line 13: Yeah! I no longer have to exclude this file from my changed files list

PR #27, opened 2020-07-31
Fix cache path

ErikSchierboom:


Dockerfile.dev
```diff
@@ -25,6 +25,8 @@ RUN bundle install
 
 COPY package.json yarn.lock ./
 COPY .cache/yarn/ /usr/src/yarn_cache/
+COPY .cache/node_modules/ /usr/src/node_modules/
+RUN echo "--modules-folder /usr/src/node_modules" > .yarnrc
```
iHiD, line 29: If I have this, it times out on Mac, so I took it out.

PR #29, opened 2020-07-31
Exercised Daemons (Fixed executable architecture in Dockerfiles) 

TheLostLambda:


iHiD:
@ErikSchierboom This was the problem...

ErikSchierboom:
Oh my, so easy to miss! That makes total sense. I was almost starting to doubt the very foundation of Docker :D

iHiD:
QEMU means that this gets ignored (transposed?) on some docker rubs depending on a flag. @neenjaw can explain more :)

neenjaw:
There are some initiatives to run images from any arch on any arch (eg arm -> amd64), QEMU is a translation layer which is automatically enabled with Docker desktop (and docker with WSL2), but not docker CLI.  Which is why it worked for some, but not for all.

PR #30, opened 2020-08-03
Add multiple Dockerfiles for production

iHiD:
This starts the process of adding multiple containers for production


iHiD:
(FYI @joshgoebel)

PR #34, opened 2020-08-05
Add better rubocop check

iHiD:


.github/workflows/rubocop.yml
```diff
@@ -2,16 +2,28 @@ name: Rubocop
...
+
+      - name: Install gems
+        run: |
+          gem install rubocop
+          gem install rubocop-minitest
+          gem install rubocop-performance
+          gem install rubocop-rails
+
+      - name: Run Rubocop
+        run: rubocop --except Metrics
```
SleeplessByte, line 29: Is this different than `rubocop --rails`?

iHiD: Yes.

In the [rubocop.yml](https://github.com/exercism/v3-website/blob/master/.rubocop.yml#L2) we just the various things it should run (rails, performance, etc).

The `--except Metrics` says "don't fail if methods are too complex etc", only if there is something that is a lint violation.

SleeplessByte: I think it's worthwhile to make the unilateral decision to then move that to rubocop config file. It's weird if you get the errors when developing, but then not during CI.

With my comment I meant: I think `--rails` turns on some rails specific cops, that are different than `rubocop-rails`, which you might or might not want, as well.

PR #44, opened 2020-08-21
Fix repo URL

ErikSchierboom:


iHiD:
What exactly is this fixing?

It breaks all the tests because they use the fixture repos.

PR #47, opened 2020-08-22
Run representer automatically

ErikSchierboom:


app/commands/iteration/representation/process.rb
```diff
@@ -46,6 +46,10 @@ def call
         rescue StandardError
           iteration.analysis_exceptioned!
         end
+
+        # TODO: Mark iteration as represented and broadcast
```
iHiD: Can you remove the TODO's and add a test for this pls?

app/commands/iteration/analysis/process.rb
```diff
@@ -40,7 +40,7 @@ def call
 
         # TODO: Mark iteration as analyzed and broadcast
         # it here, when we've decided how that works
-        # iteration.broadcast!
+        iteration.broadcast!
```
iHiD, line 41: Can you remove the TODO's and add a test for this pls?

app/commands/iteration/create.rb
```diff
@@ -67,13 +67,13 @@ def init_services
       [
         Thread.new do
           Iteration::TestRun::Init.(iteration_uuid, solution.track.slug, solution.exercise.slug, s3_uri)
+        end,
+        Thread.new do
+          Iteration::Analysis::Init.(iteration_uuid, solution.track.slug, solution.exercise.slug, s3_uri)
```
iHiD, line 72: Test pls.

app/commands/iteration/create.rb
```diff
@@ -67,13 +67,13 @@ def init_services
...
+        end,
+        Thread.new do
+          Iteration::Representation::Init.(iteration_uuid, solution.track.slug, solution.exercise.slug, s3_uri)
```
iHiD, line 75: Test pls.

app/commands/tooling_job/process.rb
```diff
@@ -24,6 +24,20 @@ def call
           "Nothing to report", # TOOD
           attrs["result"]
         )
+      when "representer"
```
iHiD, line 27: Test pls.

app/commands/iteration/create.rb
ErikSchierboom: It looks like there already is a test for this: https://github.com/exercism/v3-website/blob/master/test/commands/iteration/create_test.rb#L22

app/models/iteration/analysis.rb
```diff
@@ -20,15 +20,15 @@ def ops_errored?
   end
 
   def approved?
-    status == :approved
+    status == :approve
```
ErikSchierboom, line 23: The statuses used did not match [the analyzer interface](https://github.com/exercism/automated-analysis/blob/master/docs/analyzers/interface.md#output-format).

test/commands/iteration/analysis/process_test.rb
```diff
@@ -60,4 +60,14 @@ class Iteration::Analysis::ProcessTest < ActiveSupport::TestCase
 
     assert iteration.reload.analysis_exceptioned?
   end
+
+  test "broadcast" do
+    iteration = create :iteration
+    data = { 'status' => :approve, 'comments' => [] }
+
+    IterationChannel.expects(:broadcast!).with(iteration)
+    IterationsChannel.expects(:broadcast!).with(iteration.solution)
```
ErikSchierboom, line 69: Normally I would have separate tests for the two broadcasts, but I've found that the Ruby error message is so clear (exact expectation that failed) that I think combining them is okay.

test/commands/iteration/create_test.rb
```diff
@@ -21,7 +20,6 @@ class Iteration::CreateTest < ActiveSupport::TestCase
     Iteration::TestRun::Init.expects(:call)
     Iteration::Analysis::Init.expects(:call)
     Iteration::Representation::Init.expects(:call)
-    ToolingJob::Create.expects(:call)
```
ErikSchierboom, line 24: I had to disable this to get the test passing, but I didn't understand at first why I had to do that. The `Iteration::TestRun::Init` call (and the other two) all call `ToolingJob::Create.` in their respective `.call` methods, but it looks like they were not registering. Turns out that `Iteration::TestRun::Init.expects(:call)` causes this. Is it okay to just remove this expectation?

test/commands/tooling_job/process_test.rb
```diff
@@ -39,4 +39,82 @@ class ToolingJob::ProcessTest < ActiveSupport::TestCase
 
     ToolingJob::Process.(id)
   end
+
+  test "proxies to analyzer" do
+    id = SecureRandom.uuid
```
ErikSchierboom, line 44: For now, I've just copy/pasted/modified the existing test runner test. Let me know if you'd like me to use a `setup` method.

app/models/iteration/analysis.rb
iHiD: I think my plan was to change the interface 🙂 

Let's chat about this at some point soon. Certainly `refer_to_mentor` doesn't make sense in v3, but I also wonder about how statuses and comments should work generally.

test/commands/iteration/analysis/process_test.rb
iHiD: Yeah - I agree entirely.

test/commands/iteration/create_test.rb
iHiD: Yeah, I think that's fine now :)

test/commands/tooling_job/process_test.rb
iHiD: No, this is fine for now.

iHiD:
Sweet. Nice work. This is great. I'll merge it now. I think we'll want to revert the analyzer statuses bit further down the line, but until that spec formally change, let's keep it all unified :)

PR #50, opened 2020-08-28
Don't store iteration file content in the db

iHiD:
- Retrieve files from s3 when rather than storing them locally.
- Use a hash of the filename for security (cc @SleeplessByte - we discussed this ages ago).

### DIFF
--- a/app/commands/iteration/representation/process.rb
+++ b/app/commands/iteration/representation/process.rb
@@ -29,7 +29,9 @@ class Iteration
           exercise_version: exercise_version,
           ast_digest: iteration_representation.ast_digest
         ) do |rep|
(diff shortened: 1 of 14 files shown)


app/commands/iteration/representation/process.rb
```diff
@@ -29,7 +29,9 @@ def call
           exercise_version: exercise_version,
           ast_digest: iteration_representation.ast_digest
         ) do |rep|
+          # rep.source_iteration = iteration
           rep.ast = ast
+          # rep.mapping = mapping
```
iHiD, line 34: These get added in my next PR.

SleeplessByte:
Yes good. Good bye injections.

PR #52, opened 2020-08-29
Add API endpoints and fully implement iteration file storage via S3

iHiD:
This allows you to use the CLI against the v3 website locally.

When running the db/seeds it outputs the command to use to reconfigure your client.

All files are stored locally in the s3 docker image.


iHiD:
(@ErikSchierboom @kntsoriano Not actually expecting you to review these btw - but tagging you so you can scan through it if you're interested 🙂)

PR #53, opened 2020-08-30
Add basic joining track flow

iHiD:
This does a few things:
- Adds haml linting (this can't autocorrect annoyingly)
- Adds friendly_id (so that urls can be `/tracks/ruby` not `/tracks/52`)
- Readds spring and bootsnap to speed up load time, especially for tests
- Adds a simple flow for starting a track
- Adds a pattern I'm trying for allowing authenticated and external views within the same controller. I might end up changing this, but I'm seeing how it goes for now (cc @kntsoriano and @SleeplessByte for thoughts on this with your Rails hats on)
- Renames `User::JoinTrack` to `UserTrack::Create`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,22 @@
 class ApplicationController < ActionController::Base
   # TODO: Remove when devise is added
-  attr_reader :current_user
+  def user_logged_in?
+    !!current_user
+  end
+
+  def current_user
+    User.first
+  end
+
+  %w[index show].each do |action|
+    define_method action do
+      if user_logged_in? && respond_to?("authenticated_#{action}")
+        send("authenticated_#{action}")
+        render action: "authenticated/#{action}"
+      elsif !user_logged_in? && respond_to?("external_#{action}")
+        send("external_#{action}")
+        render action: "external/#{action}"
+      end
+    end
+  end
 end
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -1,0 +1,21 @@
+class TracksController < ApplicationController
+  before_action :use_track, except: :index
+
+  def authenticated_index
+    @tracks = Track.all
+  end
+
+  def authenticated_show
+    @user_track = current_user.user_tracks.find_by(track: @track)
+  end
+
+  def join
+    UserTrack::Create.(current_user, @track)
+    redirect_to action: :show
+  end
+
+  private
+  def use_track
+    @track = Track.find(params[:id])
+  end
+end
--- a/db/migrate/20200830161328_create_friendly_id_slugs.rb
+++ b/db/migrate/20200830161328_create_friendly_id_slugs.rb
@@ -1,0 +1,21 @@
+MIGRATION_CLASS =
+  if ActiveRecord::VERSION::MAJOR >= 5
+    ActiveRecord::Migration["#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}"]
+  else
+    ActiveRecord::Migration
+  end
+
+class CreateFriendlyIdSlugs < MIGRATION_CLASS
+  def change
+    create_table :friendly_id_slugs do |t|
+      t.string   :slug,           :null => false
+      t.integer  :sluggable_id,   :null => false
+      t.string   :sluggable_type, :limit => 50
+      t.string   :scope
+      t.datetime :created_at
+    end
+    add_index :friendly_id_slugs, [:sluggable_type, :sluggable_id]
+    add_index :friendly_id_slugs, [:slug, :sluggable_type], length: { slug: 140, sluggable_type: 50 }
+    add_index :friendly_id_slugs, [:slug, :sluggable_type, :scope], length: { slug: 70, sluggable_type: 50, scope: 70 }, unique: true
+  end
+end
--- a/db/migrate/20200830161328_create_friendly_id_slugs.rb
+++ b/db/migrate/20200830161328_create_friendly_id_slugs.rb
@@ -1,11 +1,4 @@
-MIGRATION_CLASS =
-  if ActiveRecord::VERSION::MAJOR >= 5
-    ActiveRecord::Migration["#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}"]
-  else
-    ActiveRecord::Migration
-  end
-
-class CreateFriendlyIdSlugs < MIGRATION_CLASS
+class CreateFriendlyIdSlugs < ActiveRecord::Migration[6.0]
   def change
     create_table :friendly_id_slugs do |t|
       t.string   :slug,           :null => false
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -29,7 +29,7 @@ track_slugs.each do |track_slug|
   end
 
   p "Adding Track: #{track_slug}"
-  track = Track.create!(slug: track_slug, title: track_slug, repo_url: v3_url)
+  track = Track.create!(slug: track_slug, title: track_slug.titleize, repo_url: v3_url)
 
   begin
     #track.update(title: track.repo.config[:language])
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -20,15 +20,15 @@ track_slugs = []
 tree = repo.send(:repo).read_tree(repo.send(:repo).head_commit, "languages/")
 tree.each_tree { |obj| track_slugs << obj[:name] }
 
-p track_slugs
+puts track_slugs
 
 track_slugs.each do |track_slug|
   if Track.find_by(slug: track_slug)
-    p "Track already added: #{track_slug}"
+    puts "Track already added: #{track_slug}"
     next
   end
 
-  p "Adding Track: #{track_slug}"
+  puts "Adding Track: #{track_slug}"
   track = Track.create!(slug: track_slug, title: track_slug.titleize, repo_url: v3_url)
 
   begin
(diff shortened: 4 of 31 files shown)


db/migrate/20200830161328_create_friendly_id_slugs.rb
```diff
@@ -0,0 +1,21 @@
+MIGRATION_CLASS =
+  if ActiveRecord::VERSION::MAJOR >= 5
+    ActiveRecord::Migration["#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}"]
+  else
+    ActiveRecord::Migration
+  end
```
iHiD: This is auto-generated by friendly_id. We could probably remove it, and just use L3.

SleeplessByte: Yeah, you're not going to downgrade, ever.

app/controllers/tracks_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def authenticated_index
+    @tracks = Track.all
+  end
+
+  def authenticated_show
+    @user_track = current_user.user_tracks.find_by(track: @track)
+  end
+
+  def join
```
SleeplessByte, line 12: Don't you want to POST on `/tracks/ruby/membership` instead?

app/controllers/tracks_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    UserTrack::Create.(current_user, @track)
+    redirect_to action: :show
+  end
+
+  private
+  def use_track
+    @track = Track.find(params[:id])
```
SleeplessByte, line 19:
```suggestion
    @track = Track.find(params[:id])
```

I personally prefer to remove the magic and doing `Track.friendly.find` or whatever the syntax is.

db/seeds.rb
```diff
@@ -29,7 +29,7 @@ def rugged_repo
   end
 
   p "Adding Track: #{track_slug}"
-  track = Track.create!(slug: track_slug, title: track_slug, repo_url: v3_url)
+  track = Track.create!(slug: track_slug, title: track_slug.titleize, repo_url: v3_url)
```
SleeplessByte, line 32: The Track slug -> Name mapping is not always correct. Since you have the data, don't you want to read this from the config file and use this as fallback only?

SleeplessByte (commented):
Not a lot to comment on :) Overall is fine I think.

db/seeds.rb
iHiD: Yeah. This seeds is only temporary. Lots of the V3 repo data is very messy right now.

The actual db will be some sort of ETL from v2

app/controllers/tracks_controller.rb
iHiD: Maybe. I dunno :shrug:

I don't mind controllers having responsibility for multiple interactions as long as they don't have a lot of logic in each action. I find having lots of controllers with single actions (such as a membership action here) can get really noisy very quickly so I tend to start with just the add-on action and then refactor it out into a second controller once I know I need more actions (eg the leave track or pause track here).

The other consideration for me here is that the track isn't editable in this context. The actual controller is all about a track and the user's relationship with it. Which means I'd be tempted to fill a few actions on here because it won't have the other CRUD things. 

I guess from a philosophical standpoint I care less about the standard REST actions in websites than I do in APIs.

app/controllers/application_controller.rb
```diff
@@ -1,4 +1,22 @@
...
+
+  %w[index show].each do |action|
+    define_method action do
+      if user_logged_in? && respond_to?("authenticated_#{action}")
+        send("authenticated_#{action}")
+        render action: "authenticated/#{action}"
+      elsif !user_logged_in? && respond_to?("external_#{action}")
+        send("external_#{action}")
+        render action: "external/#{action}"
+      end
```
ErikSchierboom, line 19: Ruby meta-programming in action!

app/controllers/tracks_controller.rb
SleeplessByte: Oh yeah, as long as it's a conscious choice, that's fine 👍

I just figured you internally modelled it as "track membership".

db/seeds.rb
SleeplessByte: Nice.

app/controllers/tracks_controller.rb
iHiD: I'm sensing a theme in your feelings about Ruby magic here 😉

app/controllers/application_controller.rb
kntsoriano: Will this be removed once we add Devise in? If not, I'd rather have this hard coded since it's easier to read.

app/controllers/tracks_controller.rb
kntsoriano: In my experience, it's easier for me to model this as separate controllers because sometimes the `before_action`s do too much.  This is okay for now, but I see it as separate controllers especially if we take into errors when user joins a track.

test/controllers/tracks_controller_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
```
kntsoriano: Let's delete this file.

test/system/flows/join_track_test.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      # TODO: Add when adding devise
+      # sign_in(user)
+
+      visit tracks_url
+      click_link "Ruby"
+
+      click_on "Join Track"
+
+      assert_equal [track], user.tracks
```
kntsoriano, line 17: Can we instead assert on text instead of values in the database? I view system tests from a user perspective, and a user shouldn't have direct access to the database.

app/controllers/tracks_controller.rb
ErikSchierboom: Less Ruby magic is also easier for me to read as a Ruby n00b 😉

test/system/flows/join_track_test.rb
iHiD: I think we want two different things here. And I think probably we want a test for each.

- I want something that uses the app as a user does but where we verify that the correct things are happening under the hood. Without this test, there's nothing in our test-suite that actually checks that things work as expected. These are effectively integration tests, but I want them to use the Chrome for the testing, not me manually specifying the routes to traverse (which I find prone to bugs).
- You want something that tests the flow of the app specifically from how a user is perceiving it. That's really valuable too.

My issue is when we combine the two, making view-level changes, break the "functional" test that I want, which I find really frustrating. And that's even more acute at this early stage, as the actual views are effectively nonsense, but I still want to have proper functional tests that don't then break when we implement the views.

---

So maybe the key here is to have the tests in two different subdirectories where it's clear what the purpose of each is?

app/controllers/application_controller.rb
iHiD: By hard-coded, what do you mean? Duplicated in every controller?

iHiD: I'm merging this for now but will be refactoring it later.

test/system/flows/join_track_test.rb
iHiD: Moved to https://github.com/exercism/v3-website/issues/54

iHiD:
Thanks everyone 🙂 

I've tweaked a couple of things, merged and will continue to improve the things you commented on in future commits :)

app/controllers/application_controller.rb
kntsoriano: Not in every controller, just defined as separate `index` and `show` actions in `ApplicationController`.

I've been reading the code more and I understand what you want to get out of this. Maybe we could write a helper method for this that is called within the `index` and `show` actions rather than defining the actions programatically. It might be just personal preference that I want the crud actions, especially in controllers, as transparent as possible.

iHiD: I chatted quite a bit 1-1 with @SleeplessByte too. He suggested having a DSL for this, which I liked.

So rather than having this global block of code in application_controller, we might instead add the following to any controller where we want to have an external index and show page:

```
external :index, :show
```

Which would then do something like add a `before_action` that checks if the person is logged in, and if not, uses the `external_index` or `external_show` action rather than the `index` action.

So sort of somewhere between what I've got here (full automated) and what you suggest (quite manual).

The main reason for this, btw, is that I want to have consistent URLs across use cases. So someone logged out would go to `/tracks/ruby/exercises/bob` for the SEO page and someone logged in would go to the **same** URL for their solution to Ruby/Bob (whereas at the moment, everything internal gets namespaced to `my/...` and we have things like `/my/solution/xxxx` which is a non-sharable link.

So because you can't pivot the controller based on state in the routes (without writing custom middleware, which **might** be an option, but isn't a rabbit hole I want to go into now), pivoting the actions in the controller feels like a better option.

test/controllers/tracks_controller_test.rb
iHiD: Done. Thanks. I removed the helper too.

app/controllers/application_controller.rb
kntsoriano: Great idea!!!!!!

SleeplessByte: Thanks :)

app/controllers/tracks_controller.rb
SleeplessByte: I love Ruby magic, but working with ruby non-seniors has learned me to embrace the features whilst still being explicit :)

iHiD: So, I agree. I'm also going to put out there the reminder for the four of us as this process continues that we don't accept PRs to this repo, and that juniors aren't getting near it 🙂 

But the principle of least surpirses is something I'm passionate about.

PR #55, opened 2020-09-03
Add CONTRIBUTING doc

iHiD:
**This is WIP and will change lots so don't worry about reviewing it yet. If you have any general comments about things it should include, please add those below though.**

---

This PR covers the addition of a CONTRIBUTING.md document that outlines the technical philosophy and the rules/standards we want to develop the codebase against. I intend to write down everything I feel firmly about as a first pass, and then we can discuss any bits that are unclear etc. Then as the next couple of months evolve we can flesh out the other bits that are maybe less clear at this stage etc. I want this to be a living document that evolves throughout the process, and that by the time of launch is quite firm and immutable, which then makes accepting any contributions from others an easier thing.

PR #56, opened 2020-09-04
Add <NotificationIcon /> component

kntsoriano:
https://github.com/exercism/v3-website/issues/37

## Description
This adds the `<NotificationIcon />` component which will be seen on the header bar. 

It has two states:
1) the number of unread notifications
2) a boolean that states whether you have any unread notifications

## Preview
Sep-04-2020 15-10-28

## Questions
1. Did I understand the states correctly?
2. When do we turn off the state where you have any unread notifications? Is it when you click on the icon? Is it when all of your notifications have been read?


test/system/react_components/notifications/icon_test.rb
```diff
@@ -0,0 +1,23 @@
+require "application_system_test_case"
+
+module Notifications
```
kntsoriano: Maybe we should namespace this under `ReactComponents`. I just copied how we have it right now.

app/commands/notification/create.rb
```diff
@@ -6,10 +6,10 @@ class Create
 
     def call
       klass = "notifications/#{type}_notification".camelize.constantize
-      klass.create!(
-        user: user,
-        params: params
-      )
+
+      klass.create!(user: user, params: params).tap do
```
kntsoriano, line 10: For another discussion, but should we enforce whether commands have a return value or not?

iHiD: Yeah - I was thinking about this. 

Two options:
1. All `create` commands return something
2. We rely on the `on_success`/`on_failure` states.

If we do (1) then errors should raise Exceptions. If we do (2) then errors can be passed bac.

The main downside to (2) is that it makes both the internal and consuming code a bit more verbose:

```
  res = Notification::Create.(...) 
  handle_error unless res.success?
  notification = res.result
```

But that verbosity is also maybe slightly clearer. Thoughts?

test/system/react_components/notifications/icon_test.rb
iHiD: Yeah, namespacing sounds good.

In #55 I've renamed this folder to be `Components` FYI.

app/helpers/components_helper.rb
```diff
@@ -12,6 +12,10 @@ def maintaining_iterations_summary_table(iterations)
                     })
   end
 
+  def notification_icon(user)
```
iHiD, line 15: What do you think about rather than this behind a helper module, these being actual classes that are called? I dislike that we pollute the view context with some many magic helper methods. Feels brittle.

So we'd have a `app/view_components/` folder with `notification_icon.rb`, which inherits from `ViewComponent.rb` that has the `react_component` in.

```
Then we'd call `ViewComponents::NotificationIcon.render(user)`, which outputs the relevant haml.
```

That means it's a little less magic than the view helper methods, and it makes them easier to test functionally. The only thing I'm unclear of is the ease of using the view context, but I reckon it would be ok.

---

This wouldn't block this PR, but it could be a good refactor if y'all agreed?

test/system/react_components/notifications/icon_test.rb
```diff
@@ -0,0 +1,23 @@
...
+      assert_text "Count\n1", wait: 1
+    end
+
+    test "shows unread state when a new notification is created" do
+      create(:user)
+
+      visit test_components_notifications_icon_path
+      click_on "Create notification"
+
+      assert_text "Unread\ntrue", wait: 1
```
iHiD: `wait: 1` is nice. Could you refactor the other websockets tests to have that too pls?

iHiD: (can be a seperate PR)

test/system/react_components/notifications/icon_test.rb
```diff
@@ -0,0 +1,23 @@
+require "application_system_test_case"
+
+module Notifications
+  class IconTest < ApplicationSystemTestCase
+    test "increments number when a notification is created" do
+      create(:user)
+
+      visit test_components_notifications_icon_path
+      click_on "Create notification"
```
iHiD: We need to assert that Count is 0 before we click_on "Create notification"

test/system/react_components/notifications/icon_test.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      assert_text "Count\n1", wait: 1
+    end
+
+    test "shows unread state when a new notification is created" do
+      create(:user)
+
+      visit test_components_notifications_icon_path
+      click_on "Create notification"
```
iHiD: We need to assert that Unread is false before we click_on "Create notification"

iHiD:
@kntsoriano Nice work :) So I've left a few little comments - most of which are general to the codebase not this PR.

One overarching comment though. I think Rails should be pushing an update out with the serialized notifications count, rather than a message that there is a new notification. The main reason for this is that we're not also pushing a websocket event out  that says when a notification is read, which means that if someone reads a notification in another tab (or another devise) the count will be wrong. We might want to do that, but then we're going to end up with lots of logic in the component (events created, removed, marked as unread, etc) that we don't need. If instead we just pass an event with `type: notifications.changed, notification_count: current_user.notifications.unread.size` we just have one codepath. We can then use exactly the same code for initializing the react component as we use for updating it, which is my general hope for these components, and what React seems to be very good at handling.

I also think we can kick the unread state into the ReactComponent, as it's a function of the notification count, which saves us an extra piece of data to pass around.

test/system/react_components/notifications/icon_test.rb
SleeplessByte: Components I like better too, especially if you're going Vue or whatever at some time.

app/commands/notification/create.rb
kntsoriano: I personally like (2) as it's the most unsurprising option. Running the `call` method for me implies doing an action and not returning anything. I ran into this problem when I added the ActionCable broadcast to the end of `Notification::Create` and it failed as the test expected to get something back. I think we should have the flexibility to just add actions to the `call` method, not thinking about placing the return value at the end of it.

app/helpers/components_helper.rb
kntsoriano: I like the idea. It's also okay to pass the view_context into the component as it makes it explicit what the component is rendered for. It also helps us look at all our components at a glance, since we have separate files for each.

test/system/react_components/notifications/icon_test.rb
kntsoriano: I tried this and it's a bit more complicated. The fact that we click a button here allows the websockets to connect just in time. The other tests don't have that and would still need the sleep at the start of the test to establish the websocket connection. The second sleep that we have in those tests could also be deleted since assertions wait for a certain amount of time by default.

I will keep the existing tests as they are for now and will refactor later.

kntsoriano:
Thanks for the comments. I've updated the code. I've also updated the test page to this:
Sep-07-2020 09-00-42

ErikSchierboom:
@kntsoriano [Rubocop is failing](https://github.com/exercism/v3-website/pull/56/checks?check_run_id=1079282027#step:5:11)

app/helpers/components_helper.rb
ErikSchierboom: I like the suggested change. I can already imagine this helper file becoming huge at some point, which would make editing it slightly painful. And as kntsoriano said, it also help with the visibility.

kntsoriano:
@ErikSchierboom Rubocop should be green now.

Something to watch out for, though. When we merge this into master and the built image isn't pushed to DockerHub yet or you haven't pulled the latest website image yet, the website container won't startup as I've upgraded a gem. In order to work around this, I've opened https://github.com/exercism/v3-website/pull/59.

test/system/react_components/notifications/icon_test.rb
ErikSchierboom: Let's do this in a follow-up PR!

ErikSchierboom (approved):
LGTM!

app/javascript/components/notifications/icon.jsx
```diff
@@ -0,0 +1,38 @@
...
+  switch (action.type) {
+    case 'notifications.changed':
+      return { count: action.payload.count }
+  }
+}
+
+export function NotificationIcon({ count }) {
+  const [state, dispatch] = useReducer(reducer, { count: count })
+
+  function isUnread() {
```
SleeplessByte: There is no point for this function to exist, at the moment. Do you have different ideas with it?

```jsx
const isUnread = state.count > 0
```

app/javascript/components/notifications/icon.jsx
```diff
@@ -0,0 +1,38 @@
...
+    return state.count > 0
+  }
+
+  useEffect(() => {
```
SleeplessByte, line 15: Absolutely fine for now, probably want to extract this into a custom hook at some point, when you start getting more of these subscriptions.

test/support/websockets_helpers.rb
```diff
@@ -0,0 +1,5 @@
+module WebsocketsHelpers
+  def wait_for_websockets
+    sleep(0.5)
```
SleeplessByte, line 3: In some future iteration, it's probably less brittle to listen for the connection itself!

SleeplessByte (approved):

app/javascript/components/notifications/icon.jsx
kntsoriano: I don't actually. Your solution is better.

ErikSchierboom:
@kntsoriano The tests are failing. Could you look into that?

kntsoriano:
@ErikSchierboom Should be fixed now. Sorry about that!

ErikSchierboom (approved):

test/support/websockets_helpers.rb
iHiD: @SleeplessByte How? :)

PR #60, opened 2020-09-09
Add <MentorConversationsList />

kntsoriano:
## Description
This adds the skeleton for the `<MentorConversationsList />` component. We can now work on the pagination, filtering, and sorting separately. We might encounter some merge issues, but it should be relatively easy to fix.

I chose to keep it simple and not think about the API response for now. I don't have enough expertise on this to make a decision.

*EDIT: The failed rubocop run isn't in any way related to the changes here. `master` is broken and I've added a fix in https://github.com/exercism/v3-website/pull/56.*

## Screenshot
Screen Shot 2020-09-09 at 10 32 30 AM

## Questions
1. For the relative time display, I chose to use [dayjs](https://github.com/iamkun/dayjs). I'm pretty out of touch with the JS ecosystem so I'm unsure whether this was a good decision. (cc @SleeplessByte)
2. Did I miss any fields?

### DIFF
--- a/app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
+++ b/app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
@@ -15,7 +15,7 @@ class Test::Components::Mentoring::MentorConversationsListsController < Applicat
         haveMentoredPreviously: true,
         isNewIteration: true,
         postsCount: 15,
-        updatedAt: 1.year.ago,
+        updatedAt: 1.year.ago.iso8601,
         url: "https://exercism.io/conversations/1"
       }
     ]
--- a/app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
+++ b/app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
@@ -1,0 +1,23 @@
+class Test::Components::Mentoring::MentorConversationsListsController < ApplicationController
+  def show
+    @mentor = OpenStruct.new(conversations: conversations)
+  end
+
+  private
+  def conversations
+    [
+      {
+        trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
+        menteeAvatarUrl: "https://robohash.org/exercism",
+        menteeHandle: "mentee",
+        exerciseTitle: "Series",
+        isStarred: true,
+        haveMentoredPreviously: true,
+        isNewIteration: true,
+        postsCount: 15,
+        updatedAt: 1.year.ago,
+        url: "https://exercism.io/conversations/1"
+      }
+    ]
+  end
+end
--- a/app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
+++ b/app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
@@ -7,6 +7,7 @@ class Test::Components::Mentoring::MentorConversationsListsController < Applicat
   def conversations
     [
       {
+        trackTitle: "Ruby",
         trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
         menteeAvatarUrl: "https://robohash.org/exercism",
         menteeHandle: "mentee",
--- a/app/javascript/components/mentoring/mentor_conversations_list.jsx
+++ b/app/javascript/components/mentoring/mentor_conversations_list.jsx
@@ -1,0 +1,62 @@
+import React from 'react'
+import dayjs from 'dayjs'
+import RelativeTime from 'dayjs/plugin/relativeTime'
+dayjs.extend(RelativeTime)
+
+function MentorConversationListRow({
+  trackIconUrl,
+  menteeAvatarUrl,
+  menteeHandle,
+  exerciseTitle,
+  isStarred,
+  haveMentoredPreviously,
+  isNewIteration,
+  postsCount,
+  updatedAt,
+  url,
+}) {
+  return (
+    <tr>
+      <td>
+        <img style={{ width: 100 }} src={trackIconUrl} />
+      </td>
+      <td>
+        <img style={{ width: 100 }} src={menteeAvatarUrl} />
+      </td>
+      <td>{menteeHandle}</td>
+      <td>{exerciseTitle}</td>
+      <td>{isStarred.toString()}</td>
+      <td>{haveMentoredPreviously.toString()}</td>
+      <td>{isNewIteration.toString()}</td>
+      <td>{postsCount}</td>
+      <td>{dayjs(updatedAt).from(dayjs())}</td>
+      <td>{url}</td>
+    </tr>
+  )
+}
+
+export function MentorConversationsList({ conversations }) {
+  return (
+    <table>
+      <thead>
+        <tr>
+          <th>Track icon</th>
+          <th>Mentee avatar</th>
+          <th>Mentee handle</th>
+          <th>Exercise title</th>
+          <th>Starred?</th>
+          <th>Mentored previously?</th>
+          <th>New iteration?</th>
+          <th>Posts count</th>
+          <th>Updated at</th>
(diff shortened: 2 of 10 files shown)


app/javascript/components/mentoring/mentor_conversations_list.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import dayjs from 'dayjs'
+import RelativeTime from 'dayjs/plugin/relativeTime'
+dayjs.extend(RelativeTime)
+
+function MentorConversationListRow({
```
SleeplessByte, line 6: There was probably a decision to use JS instead of TS, but here you're going to benefit from type annotation this at least with jsdoc, in some editors, such as vscode and rubymine.

app/javascript/components/mentoring/mentor_conversations_list.jsx
```diff
@@ -0,0 +1,62 @@
...
+      <td>
+        <img style={{ width: 100 }} src={menteeAvatarUrl} />
+      </td>
+      <td>{menteeHandle}</td>
+      <td>{exerciseTitle}</td>
+      <td>{isStarred.toString()}</td>
+      <td>{haveMentoredPreviously.toString()}</td>
+      <td>{isNewIteration.toString()}</td>
+      <td>{postsCount}</td>
+      <td>{dayjs(updatedAt).from(dayjs())}</td>
```
SleeplessByte: Instead you probably want to use `.fromNow()`

app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      {
+        trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
+        menteeAvatarUrl: "https://robohash.org/exercism",
+        menteeHandle: "mentee",
+        exerciseTitle: "Series",
+        isStarred: true,
+        haveMentoredPreviously: true,
+        isNewIteration: true,
+        postsCount: 15,
+        updatedAt: 1.year.ago,
```
SleeplessByte: Convert to `.iso8601`, or pass a format when instantiating `dayJs` instance.

app/javascript/packs/application.jsx
```diff
@@ -3,33 +3,35 @@
...
 
-import React from "react";
-import { initReact } from "./react_bootloader.jsx";
-import { ExampleIterationsSummaryTable } from "../components/example/iterations_summary_table.jsx";
-import { MaintainingIterationsSummaryTable } from "../components/maintaining/iterations_summary_table.jsx";
+import React from 'react'
+import { initReact } from './react_bootloader.jsx'
+import { ExampleIterationsSummaryTable } from '../components/example/iterations_summary_table.jsx'
+import { MaintainingIterationsSummaryTable } from '../components/maintaining/iterations_summary_table.jsx'
+import { MentorConversationsList } from '../components/mentoring/mentor_conversations_list.jsx'
```
SleeplessByte, line 18: FYI, it's very unusual to see snake case for JavaScript files.

app/javascript/components/mentoring/mentor_conversations_list.jsx
```diff
@@ -0,0 +1,62 @@
...
+  isNewIteration,
+  postsCount,
+  updatedAt,
+  url,
+}) {
+  return (
+    <tr>
+      <td>
+        <img style={{ width: 100 }} src={trackIconUrl} />
+      </td>
```
SleeplessByte, line 27: I'm trying to respond on the line above this, but the UI won't let me.

Add `alt`, for example, "icon indicating <track name>". If you don't want it to announce anything (e.g. presentation only), use `alt=""`

app/javascript/components/mentoring/mentor_conversations_list.jsx
```diff
@@ -0,0 +1,62 @@
...
+      <td>
+        <img style={{ width: 100 }} src={menteeAvatarUrl} />
```
SleeplessByte: Probably want to add an alt "avatar for 'name'"

SleeplessByte (commented):
Looks good to me. Few minor inline comments.

DayJS is a pretty nice lib.

app/javascript/components/mentoring/mentor_conversations_list.jsx
ErikSchierboom: @iHiD will know why JS was chosen over TS.

app/javascript/packs/application.jsx
ErikSchierboom: I think that's probably a remnant of the Ruby naming convention. Maybe we should rename at some point.

test/system/components/mentoring/mentor_conversations_list_test.rb
```diff
@@ -0,0 +1,24 @@
...
+          within("td:nth-child(1)") { assert_css "img[src='https://assets.exercism.io/tracks/ruby-hex-white.png']" }
+          within("td:nth-child(2)") { assert_css "img[src='https://robohash.org/exercism']" }
+          within("td:nth-child(3)") { assert_text "mentee" }
+          within("td:nth-child(4)") { assert_text "Series" }
+          within("td:nth-child(5)") { assert_text "true" }
+          within("td:nth-child(6)") { assert_text "true" }
+          within("td:nth-child(7)") { assert_text "true" }
+          within("td:nth-child(8)") { assert_text "15" }
+          within("td:nth-child(9)") { assert_text "a year ago" }
+          within("td:nth-child(10)") { assert_text "https://exercism.io/conversations/1" }
```
ErikSchierboom: These tests all verify the exact position of the textual values. This makes changing this test a bit brittle as changing the HTML will cause these tests to fail. Maybe we could just assert that there is a `td` (without the position) with the specified text?

app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
ErikSchierboom: I like @SleeplessByte's suggestion about being explicit about the format. What do you think @kntsoriano?

ErikSchierboom (approved):

test/system/components/mentoring/mentor_conversations_list_test.rb
SleeplessByte: In case of a table, which this might not stay, I often first collect the headers, and then test that the value is in the right column, without hardcoding those columns.

app/javascript/components/mentoring/mentor_conversations_list.jsx
kntsoriano: Oh, yes. Thanks!

app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
kntsoriano: Good idea. We should enforce this rule when writing our serializers.

app/javascript/packs/application.jsx
kntsoriano: I agree with renaming too!

app/javascript/components/mentoring/mentor_conversations_list.jsx
kntsoriano: Good catch! I've added the `alt` attributes.

test/system/components/mentoring/mentor_conversations_list_test.rb
kntsoriano: This is really good. I will change this test to be less brittle.

kntsoriano: I've added a new `assert_table_row` matcher to make the test less brittle. It isn't perfect but it'll move this forward.

ErikSchierboom (approved):

PR #61, opened 2020-09-09
<MentorConversationsList /> pagination

kntsoriano:
## Description
This adds simple pagination for `<MentorConversationsList />`.

## Preview
Sep-09-2020 13-04-02

## Questions
1. Are we fine with just passing in the endpoint rather than the initial collection of conversations? I opted to do it this way because it was easier for me to structure the code.
2. Was it a good choice to use [react-query](https://github.com/tannerlinsley/react-query)? I decided to use this as we want an external library to handle caching when loading through the pages. I'm unsure though if this was the right choice.
3. Are my JS tests correct? And is it okay to use `userEvent`?
4. Have we decided on how the pagination widget behaves? I've set it up as Google-style right now where N number of pages are shown around the page. The last N pages aren't shown, but the design on the website shows otherwise. I decided to do it this way as this is the most simple option for now. (cc @taiyab)

### DIFF
--- a/app/javascript/components/mentoring/mentor_conversations_list.jsx
+++ b/app/javascript/components/mentoring/mentor_conversations_list.jsx
@@ -3,6 +3,7 @@ import { usePaginatedQuery } from 'react-query'
 import dayjs from 'dayjs'
 import RelativeTime from 'dayjs/plugin/relativeTime'
 import Pagination from './mentor_conversations_list/pagination'
+import { UrlParams } from '../../utils/url_params'
 dayjs.extend(RelativeTime)
 
 function MentorConversationListRow({
@@ -46,22 +47,28 @@ function MentorConversationListRow({
   )
 }
 
-async function fetchSolutions(key, url) {
+async function fetchSolutions(key, url, params) {
   const resp = await fetch(url)
 
   return resp.json()
 }
 
-export function MentorConversationsList({ endpoint }) {
+export function MentorConversationsList({ endpoint, ...props }) {
   const [page, setPage] = useState(1)
+  const urlParams = new UrlParams(
+    Object.assign({ page: page }, props.urlParams)
+  )
+  const retryParams = props.retryParams
   const { status, resolvedData, latestData } = usePaginatedQuery(
-    ['mentor-conversations-list', `${endpoint}?page=${page}`],
-    fetchSolutions
+    ['mentor-conversations-list', `${endpoint}?${urlParams.toString()}`],
+    fetchSolutions,
+    retryParams
   )
 
   return (
     <div>
       {status === 'loading' && <p>Loading</p>}
+      {status === 'error' && <p>Something went wrong</p>}
       {status === 'success' && (
         <>
           <table>
--- a/app/javascript/components/mentoring/mentor_conversations_list.jsx
+++ b/app/javascript/components/mentoring/mentor_conversations_list.jsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react'
-import { useQuery } from 'react-query'
+import { usePaginatedQuery } from 'react-query'
 import dayjs from 'dayjs'
 import RelativeTime from 'dayjs/plugin/relativeTime'
 dayjs.extend(RelativeTime)
@@ -45,6 +45,25 @@ function MentorConversationListRow({
   )
 }
 
+function Pagination({ current, total, setPage }) {
+  return (
+    <div>
+      {Array.from({ length: total }, (_, i) => i + 1).map((page) => {
+        return (
+          <button
+            key={page}
+            onClick={() => {
+              setPage(page)
+            }}
+          >
+            {page}
+          </button>
+        )
+      })}
+    </div>
+  )
+}
+
 async function fetchSolutions(key, url) {
   const resp = await fetch(url)
 
@@ -52,8 +71,9 @@ async function fetchSolutions(key, url) {
 }
 
 export function MentorConversationsList({ endpoint }) {
-  const { status, data } = useQuery(
-    ['mentor-conversations-list', endpoint],
+  const [page, setPage] = useState(1)
+  const { status, resolvedData, latestData } = usePaginatedQuery(
+    ['mentor-conversations-list', `${endpoint}?page=${page}`],
     fetchSolutions
   )
 
@@ -61,27 +81,36 @@ export function MentorConversationsList({ endpoint }) {
     <div>
       {status === 'loading' && <p>Loading</p>}
       {status === 'success' && (
-        <table>
-          <thead>
-            <tr>
-              <th>Track icon</th>
-              <th>Mentee avatar</th>
-              <th>Mentee handle</th>
-              <th>Exercise title</th>
-              <th>Starred?</th>
-              <th>Mentored previously?</th>
-              <th>New iteration?</th>
-              <th>Posts count</th>
-              <th>Updated at</th>
-              <th>URL</th>
-            </tr>
-          </thead>
-          <tbody>
-            {data.map((conversation, key) => (
-              <MentorConversationListRow key={key} {...conversation} />
-            ))}
-          </tbody>
-        </table>
+        <>
+          <table>
+            <thead>
+              <tr>
+                <th>Track icon</th>
+                <th>Mentee avatar</th>
+                <th>Mentee handle</th>
+                <th>Exercise title</th>
+                <th>Starred?</th>
+                <th>Mentored previously?</th>
+                <th>New iteration?</th>
+                <th>Posts count</th>
+                <th>Updated at</th>
+                <th>URL</th>
+              </tr>
+            </thead>
+            <tbody>
+              {resolvedData.results.map((conversation, key) => (
+                <MentorConversationListRow key={key} {...conversation} />
+              ))}
+            </tbody>
+          </table>
+          {latestData && (
+            <Pagination
+              current={page}
+              total={latestData.meta.total}
+              setPage={setPage}
+            />
+          )}
+        </>
       )}
     </div>
   )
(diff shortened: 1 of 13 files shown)


app/javascript/components/mentoring/mentor_conversations_list.jsx
```diff
@@ -35,28 +37,54 @@ function MentorConversationListRow({
...
-        </tr>
-      </thead>
-      <tbody>
-        {conversations.map((conversation, key) => (
-          <MentorConversationListRow key={key} {...conversation} />
-        ))}
-      </tbody>
-    </table>
+    <div>
+      {status === 'loading' && <p>Loading</p>}
```
ErikSchierboom, line 70: For future reference: at some point we probably want to extract the loading screen to a separate component

app/javascript/components/mentoring/mentor_conversations_list.jsx
```diff
@@ -35,28 +37,54 @@ function MentorConversationListRow({
...
+      {status === 'success' && (
```
ErikSchierboom, line 72: Do we also want to handle the error case?

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  return (
+    <div>
+      <button
+        onClick={() => {
+          setPage(1)
+        }}
+        disabled={current === 1}
+      >
```
ErikSchierboom: You could consider extracting the button to a separate component/function, as it is repeated three times. This is completely optional of course.

package.json
```diff
@@ -14,13 +14,16 @@
     "prop-types": "^15.7.2",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-query": "^2.15.4",
```
ErikSchierboom, line 17: This is a very nice library!

test/system/components/mentoring/mentor_conversations_list_test.rb
```diff
@@ -19,6 +19,15 @@ class MentorConversationsListTest < ApplicationSystemTestCase
           within("td:nth-child(10)") { assert_text "https://exercism.io/conversations/1" }
         end
       end
+
+      test "paginates results" do
+        visit test_components_mentoring_mentor_conversations_list_url
+        click_on "2"
+
+        within("tbody > tr:first-child") do
+          within("td:nth-child(4)") { assert_text "Tournament" }
```
ErikSchierboom: I have added a similar comment to another PR, but maybe consider not verifying the exact index of the td that has the text? Instead maybe do something like "there is a td that has that text"?

ErikSchierboom (approved):

app/javascript/components/mentoring/mentor_conversations_list.jsx
SleeplessByte: We do (want error handling)

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
```diff
@@ -0,0 +1,48 @@
...
+        First
+      </button>
+      {range.map((page) => {
+        return (
+          <button
+            key={page}
+            onClick={() => {
+              setPage(page)
+            }}
+            disabled={page === current}
```
SleeplessByte, line 34: Add `aria-current="page"` if the page is current.

We should add `aria-label` for each page:

```
aria-label={`Go to page ${page}`}
```

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
```diff
@@ -0,0 +1,48 @@
...
+      .map((_, i) => start + i)
+  }
+
+  return (
+    <div>
+      <button
+        onClick={() => {
+          setPage(1)
+        }}
+        disabled={current === 1}
```
SleeplessByte, line 21: We should add an `aria-label` here. We will read this as `First`, which isn't too bad, but users using a screen users won't have the same _visual_ context.

```
aria-label="Go to first page"
```

Add `aria-current="page"` if the page is current.

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
```diff
@@ -0,0 +1,48 @@
...
+          >
+            {page}
+          </button>
+        )
+      })}
+      <button
+        onClick={() => {
+          setPage(total)
+        }}
+        disabled={current === total}
```
SleeplessByte, line 46: Add `aria-current="page"` if the page is current.
Add `aria-label`

test/javascript/components/mentoring/mentor_conversations_list/pagination.test.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import '@testing-library/jest-dom/extend-expect'
+import Pagination from '../../../../../app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx'
```
SleeplessByte: I think it's worth setting up module mappers so we can do `from '~app/` or something like that (not this PR)

test/javascript/components/mentoring/mentor_conversations_list/pagination.test.js
```diff
@@ -0,0 +1,91 @@
...
+
+test('clicking on "First" sets page to 1', () => {
+  const setPage = jest.fn()
+  const { getByText } = render(
+    <Pagination current={2} total={10} setPage={setPage} />
+  )
+
+  userEvent.click(getByText('First'))
```
SleeplessByte: Usually this uses `fireEvent` so you don't need `userEvent`.

test/javascript/components/mentoring/mentor_conversations_list/pagination.test.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import '@testing-library/jest-dom/extend-expect'
+import Pagination from '../../../../../app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx'
+
+test('clicking on "First" sets page to 1', () => {
```
SleeplessByte, line 6: If you add the `aria` values, it is nice to _at least_ test the "current" (for announcing) and that the labels have more than just the page number.

SleeplessByte (commented):
Using `react-query` makes sense to me

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+
+export default function Pagination({ current, total, setPage, around = 3 }) {
```
SleeplessByte: Consider **not** using `default` exports. Named exports do _waaay_ better with intellisense.

SleeplessByte:
@kntsoriano I would probably, eventually, extract "getting" the conversations to a hook, which can be re-used. Passing the endpoint is exactly what I would do.

We **should** _still_ use `HTTP-Caching`, on top of `react-query`.

app/javascript/components/mentoring/mentor_conversations_list.jsx
kntsoriano: I've added a new commit that handles the error case.

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
kntsoriano: I've tried doing this, but I think we don't have enough info yet to do the extraction. They're currently just buttons right now and don't do anything special yet.

EDIT: Now that I've seen that we're adding a11y labels to it, it makes sense to extract this into a component.

test/system/components/mentoring/mentor_conversations_list_test.rb
kntsoriano: I've added a commit to the other PR that hopes to fix this issue.

app/javascript/components/mentoring/mentor_conversations_list.jsx
kntsoriano: Sounds good!

kntsoriano:
I've added a new commit to handle the error state. It involved changing the main component to have the following:
1. Be able to get separate query params as a prop alongside the endpoint. We need this in order to test the "error state". I also eventually figured that we want this if we want to render the table with initial settings on (filters on by default, etc.)
2. Be able to pass in retry params as a prop. With the test environment, we want to disable `react-query`'s automatic retries in order to make our tests run faster. In the production environment, we will use `react-query`'s defaults but we have the option in the future to tune them per component as we've made the params injectable.

Sep-10-2020 11-36-39

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
kntsoriano: WIll do. Thanks!

test/javascript/components/mentoring/mentor_conversations_list/pagination.test.js
kntsoriano: How do I set that up?

test/javascript/components/mentoring/mentor_conversations_list/pagination.test.js
kntsoriano: Cool. Thanks! I've updated the tests now.

ErikSchierboom (approved):

test/javascript/components/mentoring/mentor_conversations_list/pagination.test.js
SleeplessByte: That is a good question. I haven't looked at this project setup.

Do the tests run/transpile _through_ webpacker? If yes, it's as easy as using [`additional_paths`](https://github.com/rails/webpacker#resolved). In fact, it might already be set-up. If no, let me know how it is ran.

PR #62, opened 2020-09-09
Install specific Rubocop version in CI

ErikSchierboom:


ErikSchierboom:
I've noticed in https://github.com/exercism/v3-website/pull/58/checks?check_run_id=1084711502 that the Rubocop CI check failed. After doing some digging, I've found that this is due to the [CI workflow](https://github.com/exercism/v3-website/blob/master/.github/workflows/rubocop.yml) installing the rubocop gems manually without using Bundler. This means that the version installed in the CI check can be different from the version in the Gemfile.lock, and that indeed turned out to be the case.

To verify this, I've manually added the versions from the Gemfile to the CI install bit and this does indeed work.

I'm no Ruby expert by any means, so I don't know what the best solution is here. I've checked to see if I could have Bundler install just one gem at a time, but that does not seem to be possible. The current solution is also a bit fragile, as we'd have to update it each time Rubocop is updated in the Gemfile. Maybe we should update Rubocop to the latest version in the Gemfile and do that whenever a CI check fails?

ErikSchierboom:
Doh. I just saw https://github.com/exercism/v3-website/pull/56, in which you simply updated to the latest version.

kntsoriano:
I tried this solution as well, but ended up just upgrading our version because it is good to always check our syntax against the latest stable version.

PR #63, opened 2020-09-14
<MentorInbox /> track filtering

kntsoriano:
https://github.com/exercism/v3-website/issues/57

## Description
This PR adds filtering by track. This first gets all the tracks available via a GET request and then shows the dropdown.

This PR is a bit huge as I needed to deal with the following:
1. Being able to pass two endpoints into a component. As we're using `react-query`, I decided to create a `Request` struct that has the following parameters: `endpoint`, `query`, and `retry`. When passing in endpoints to a component, we need to use this struct. Only the endpoint is the required parameter, so I think there shouldn't be any problems when using this in our code. Plus, we get the capability of tuning our retries for each component.
2. As we're dealing now with two sub-components, I decided to rename the whole component `<MentorInbox />` as we've previously decided. Retaining the `<MentorConversationsList />` name will introduce name clashes, thus I've decided to do the renaming already.

## Preview
Sep-14-2020 13-58-18

## Questions
1. I needed a way to test that the page resets to 1 when switching the track filter. In order to do this, I exposed the reducer. Is this an okay approach? Or should I be testing the component as a whole?


app/javascript/components/mentoring/mentor_inbox.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+export function reducer(state, action) {
+  switch (action.type) {
+    case 'page.changed':
+      return { ...state, query: { ...state.query, page: action.payload.page } }
+    case 'track.changed':
+      return {
+        ...state,
+        query: { ...state.query, track: action.payload.track, page: 1 },
+      }
```
SleeplessByte, line 13: You might want to add `return state` at the bottom _or_ throw an error (or throw in development, return `state` at the bottom), so that you will gracefully "upgrade" when  there are more actions.

app/javascript/components/mentoring/mentor_inbox.jsx
```diff
@@ -0,0 +1,37 @@
...
+    props.conversationsRequest
+  )
+
+  function setPage(page) {
+    dispatch({ type: 'page.changed', payload: { page: page } })
+  }
+
+  function setTrack(track) {
+    dispatch({ type: 'track.changed', payload: { track: track } })
+  }
```
SleeplessByte: You can consider optimising re-renders here by using `useCallback` (as you didn't want to pass `dispatch` down). It would look like this:

```typescript
const setTrack = useCallback((track) => {
  dispatch({ type: 'track.changed', payload: { track: track } })
}, [dispatch])
```

Not a necessity (ie. optimise for fast renders and not for "not rendering".), but something to consider.

SleeplessByte (approved):

app/javascript/components/mentoring/mentor_inbox.jsx
ErikSchierboom: I like the above suggestion!

app/javascript/components/mentoring/mentor_inbox/conversation_list/conversation.jsx
```diff
@@ -0,0 +1,45 @@
...
+  updatedAt,
+  url,
+}) {
+  return (
+    <tr>
+      <td>
+        <img
+          style={{ width: 100 }}
+          src={trackIconUrl}
+          alt={`icon indicating ${trackTitle}`}
```
ErikSchierboom: Just a small textual suggestion:

```suggestion
          alt={`icon for ${trackTitle} track`}
```

app/javascript/hooks/request_query.js
```diff
@@ -0,0 +1,24 @@
+import { usePaginatedQuery, useQuery } from 'react-query'
+import { UrlParams } from '../utils/url_params'
+
+async function handleFetch(key, url, query) {
+  const resp = await fetch(`${url}?${new UrlParams(query).toString()}`)
+
+  return resp.json()
+}
+
+export function usePaginatedRequestQuery(key, request, fetch) {
```
ErikSchierboom: The `fetch` parameter is unused. Did you plan on using it somewhere? If not:

```suggestion
export function usePaginatedRequestQuery(key, request) {
```

app/javascript/hooks/request_query.js
```diff
@@ -0,0 +1,24 @@
...
+  return usePaginatedQuery(
+    [key, request.endpoint, request.query],
+    handleFetch,
+    request.retry
+  )
+}
+
+export function useRequestQuery(key, request, fetch) {
```
ErikSchierboom: The `fetch` parameter is unused. Did you plan on using it somewhere? If not:

```suggestion
export function useRequestQuery(key, request) {
```

test/javascript/components/mentoring/mentor_inbox.test.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import { reducer } from '../../../../app/javascript/components/mentoring/mentor_inbox'
```
ErikSchierboom: This feels like an implementation detail leaking out of the component. Is there perhaps a way this can be tested without exposing the reducer? CC @SleeplessByte

ErikSchierboom (approved):
Great work!

test/javascript/components/mentoring/mentor_inbox.test.js
SleeplessByte: Yeah, I wouldn't do this. Instead unmount and remount the component (if you are testing in isolation).

app/javascript/hooks/request_query.js
kntsoriano: Thanks!

test/javascript/components/mentoring/mentor_inbox.test.js
kntsoriano: Thanks! I'll be testing this better :)

app/javascript/components/mentoring/mentor_inbox.jsx
kntsoriano: Thanks! I'll do this.

test/javascript/components/mentoring/mentor_inbox.test.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
```
kntsoriano, line 1: @ErikSchierboom @SleeplessByte I've rewritten the test in this way in order to not test implementation details. Let me know what you think.

This also brings me to a bigger discussion as to whether we want to move to tests like these when dealing with JS components. The Rails system tests that we have right now are OK, but I'm now seeing an overlap between these JS tests and Rails system tests. I think the Rails system tests should just check that the components are rendered correctly (which tests that the correct files have been required by webpacker, etc),  but not the component's more specific behavior such as pagination.

The only good thing that I see with Rails system tests is that we could walk through them in our browser and it allows other developers to skin it much easier. But, we could do that by using something like [react-storybook](https://storybook.js.org/).

SleeplessByte: Yeah, if that is the end goal, storybook is a good choice

ErikSchierboom: I like it!

PR #64, opened 2020-09-15
Sort mentor inbox

ErikSchierboom:
Note: this PR builds on https://github.com/exercism/v3-website/pull/63, so only merged this  PR _after_ merging #63

### DIFF
--- a/app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
+++ b/app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
@@ -48,7 +48,7 @@ class Test::Components::Mentoring::MentorInboxes::ConversationsController < Appl
     ]
 
     results = results.select { |c| c[:trackId] == params[:track].to_i } if params[:track].present?
-    results = results.sort_by { |c| c[params[:sort].to_sym] } if params[:sort].present?
+    results = sort(results)
 
     page = params.fetch(:page, 1).to_i
     per = params.fetch(:per, 1).to_i
@@ -57,5 +57,16 @@ class Test::Components::Mentoring::MentorInboxes::ConversationsController < Appl
       results: results[page - 1, per],
       meta: { current: page, total: results.size }
     }
+  end
+
+  def sort(results)
+    case params[:sort]
+    when 'exercise'
+      results.sort_by { |c| c[:exerciseTitle] }
+    when 'recent'
+      results.sort_by { |c| c[:updatedAt] }.reverse
+    else
+      results.sort_by { |c| c[:menteeHandle] }
+    end
   end
 end
--- a/app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
+++ b/app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
@@ -8,7 +8,7 @@ class Test::Components::Mentoring::MentorInboxes::ConversationsController < Appl
         trackTitle: "Ruby",
         trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
         menteeAvatarUrl: "https://robohash.org/exercism",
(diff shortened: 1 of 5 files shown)


app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
```diff
@@ -8,7 +8,7 @@ def index
         trackTitle: "Ruby",
         trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
         menteeAvatarUrl: "https://robohash.org/exercism",
-        menteeHandle: "mentee",
+        menteeHandle: "Mentee",
```
ErikSchierboom, line 11: I've done this just to have the sorting make sense without having to do case-insensitive sorting :D

app/javascript/components/mentoring/mentor_inbox.jsx
```diff
@@ -28,9 +34,14 @@ export function MentorInbox({ tracksRequest, ...props }) {
...
   }
 
+  function setSort(sort) {
+    dispatch({ type: 'sort.changed', payload: { sort: sort } })
+  }
+
   return (
     <div>
       <TrackFilter request={tracksRequest} setTrack={setTrack} />
+      <Sorting sort={conversationsRequest.query.sort} setSort={setSort} />
```
ErikSchierboom: I'm not sure about the naming of this component. Let me know what you think.

app/javascript/components/mentoring/mentor_inbox/sorting.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+const sortingOptions = [
+  { value: 'exerciseTitle', description: 'Sort by Exercise' },
+  { value: 'menteeHandle', description: 'Sort by Student' },
+]
```
ErikSchierboom: I've hardcoded the sorting options here. We could pass them as parameters, although I probably wouldn't want to pass their description too as that is a UI responsibility.

app/javascript/components/mentoring/mentor_inbox/conversation_sorter.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+
+const sortingOptions = [
```
kntsoriano, line 3: Might be for another PR, but should we send this via the API as well? It allows us to introduce more sorting options server side without worrying about our JS.

app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
kntsoriano: Ha. Yes!

kntsoriano (approved):
Looks great!

app/javascript/components/mentoring/mentor_inbox/conversation_sorter.jsx
ErikSchierboom: Yeah that would be a good idea for a follow-up PR. Would you include both the value and the description in the API?

kntsoriano: Yes. I think we should include both so that the client side code is really flexible.

ErikSchierboom: Woudl you include it in the existing API call that retrieves the conversation list or create a new API endpoint?

kntsoriano: Oh man. Good question. I don't know. Should we go with whats easier right now?

ErikSchierboom: I'm fine with either option :)

PR #65, opened 2020-09-15
Filter mentor inbox

ErikSchierboom:
This builds on #64, which itself builds on #63.

### DIFF
--- a/app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
+++ b/app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
@@ -48,6 +48,12 @@ class Test::Components::Mentoring::MentorInboxes::ConversationsController < Appl
     ]
 
     results = results.select { |c| c[:trackId] == params[:track].to_i } if params[:track].present?
+    if params[:filter].present?
+      results = results.select do |c|
+        c[:exerciseTitle].downcase.include?(params[:filter].downcase) ||
+          c[:menteeHandle].downcase.include?(params[:filter].downcase)
+      end
+    end
     results = sort(results)
 
     page = params.fetch(:page, 1).to_i
--- a/app/javascript/components/mentoring/mentor_inbox.jsx
+++ b/app/javascript/components/mentoring/mentor_inbox.jsx
@@ -1,5 +1,6 @@
 import React, { useReducer, useCallback } from 'react'
 import { ConversationList } from './mentor_inbox/conversation_list'
+import { ConversationFilter } from './mentor_inbox/conversation_filter'
 import { ConversationSorter } from './mentor_inbox/conversation_sorter'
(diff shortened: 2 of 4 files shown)


app/javascript/components/mentoring/mentor_inbox.jsx
```diff
@@ -11,6 +13,16 @@ export function reducer(state, action) {
...
       }
+    case 'sort.changed':
+      return {
+        ...state,
+        query: { ...state.query, sort: action.payload.sort },
+      }
+    case 'filter.changed':
+      return {
+        ...state,
+        query: { ...state.query, filter: action.payload.filter },
```
kntsoriano: Should filtering reset the page?

kntsoriano (approved):
Looks great!

app/javascript/components/mentoring/mentor_inbox.jsx
ErikSchierboom: Absolutely! Good catch

app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
```diff
@@ -48,6 +48,12 @@ def index
     ]
 
     results = results.select { |c| c[:trackId] == params[:track].to_i } if params[:track].present?
+    if params[:filter].present?
+      results = results.select do |c|
+        c[:exerciseTitle].downcase.include?(params[:filter].downcase) ||
```
SleeplessByte, line 53: FYI, you might want to have done something like:

```ruby
parts = params[:filter].downcase.split(' ').map(&:strip)

parts.all? do |part|
  c[:exerciseTitle].downcase.include?(part) || c[:menteeHandle].downcase.include?(part)
end
```

Which is more in line with how people "think" search works.

app/javascript/components/mentoring/mentor_inbox.jsx
```diff
@@ -17,6 +18,11 @@ function reducer(state, action) {
...
         query: { ...state.query, sort: action.payload.sort },
       }
+    case 'filter.changed':
+      return {
+        ...state,
+        query: { ...state.query, filter: action.payload.filter, page: 1 },
+      }
     default:
       if (process.env.NODE_ENV === 'development') {
         throw 'Unknown action type'
```
SleeplessByte, line 28: Always throw errors, and add debuggable information to the error (preferably actionable).

```js
new Error(`Unknown action type: ${action.type}`)
```

ErikSchierboom: I've fixed this in #78

app/controllers/test/components/mentoring/mentor_inboxes/conversations_controller.rb
ErikSchierboom: I've fixed this in #78

PR #71, opened 2020-09-16
Parse representation file

ErikSchierboom:
See https://github.com/exercism/development-environment/issues/34

### DIFF
--- a/app/commands/tooling_job/process.rb
+++ b/app/commands/tooling_job/process.rb
@@ -29,7 +29,7 @@ module ToolingJob
           attrs["iteration_uuid"],
           attrs["execution_status"],
           "Nothing to report", # TODO
-          download_file(attrs["output"]["representation.txt"]),
+          JSON.parse(download_file(attrs["output"]["representation.txt"])),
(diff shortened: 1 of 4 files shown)


app/commands/tooling_job/process.rb
```diff
@@ -29,7 +29,7 @@ def call
           attrs["iteration_uuid"],
           attrs["execution_status"],
           "Nothing to report", # TODO
-          download_file(attrs["output"]["representation.txt"]),
+          JSON.parse(download_file(attrs["output"]["representation.txt"])),
```
ErikSchierboom, line 32: Without this, the data in the DB is stored as a string with double quotes around it.

iHiD: Hmmm, does that not seem wrong to you? I suspect this means that the representation.txt is getting uploaded with the extra quotes incorrectly in the first place?

ErikSchierboom: It does. It is on my "todo" list, but I wanted to get the PR's up for Mark (mjs) to be able to continue. I'll dig a bit deeper into this.

ErikSchierboom:
This can be closed

PR #73, opened 2020-09-16
Merged install dependencies on startup

ErikSchierboom:
This is the second PR in a series of PR's. https://github.com/exercism/v3-website/pull/72 should be merged first to get the diff down to just the actual changes. Once this is merged, #74 is up for review.

### DIFF
--- a/docker/dev/init.dev.sh
+++ b/docker/dev/init.dev.sh
@@ -3,6 +3,9 @@ while ! mysqladmin ping -h"$DB_HOST" --silent; do
     sleep 2
 done
 
+echo "Install dependencies"
+bundle install
+
 echo "Create and migrate DBs"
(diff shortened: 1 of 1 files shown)


docker/dev/init.dev.sh
```diff
@@ -3,6 +3,9 @@ while ! mysqladmin ping -h"$DB_HOST" --silent; do
     sleep 2
 done
 
+echo "Install dependencies"
+bundle install
```
iHiD: This is much quicker:

```suggestion
bundle check || bundle install
```

iHiD: It also has the effect of not inadvertently updating the `Gemfile.lock` and putting the Docker image out of sync.

PR #74, opened 2020-09-16
Add notifications component

ErikSchierboom:
This is the third PR in a series of PR's. https://github.com/exercism/v3-website/pull/73 should be merged first to get the diff down to just the actual changes. Once this is merged, #75 is up for review.


app/helpers/components_helper.rb
```diff
@@ -12,6 +12,10 @@ def maintaining_iterations_summary_table(iterations)
                     })
   end
 
+  def notification_icon(user)
+    react_component("notification-icon", { count: user.notifications.count })
```
iHiD: I think this should be `user.notifications.unread.count` and the tests etc updated accordingly.

app/commands/notification/create.rb
```diff
@@ -6,10 +6,13 @@ class Create
...
-      klass.create!(
-        user: user,
-        params: params
-      )
+
+      klass.create!(user: user, params: params).tap do
+        NotificationsChannel.broadcast_to(user, {
+                                            type: "notifications.changed",
+                                            payload: { count: user.notifications.unread.count }
+                                          })
```
iHiD: I think this can be refactored slightly later: https://github.com/exercism/v3-website/issues/84

app/commands/notification/create.rb
```diff
@@ -6,10 +6,13 @@ class Create
 
     def call
       klass = "notifications/#{type}_notification".camelize.constantize
-      klass.create!(
-        user: user,
-        params: params
-      )
+
+      klass.create!(user: user, params: params).tap do
```
iHiD, line 10: Just a note: Pending the decision on how whether we'd rather standardise on using the success handler, we might want to refactor this tap to be `on_success` instead.

app/commands/notification/create.rb
kntsoriano: I'll add this in now.

app/commands/notification/create.rb
kntsoriano: How do we do this? Re-wrap this into another command?

app/helpers/components_helper.rb
kntsoriano: Thanks. Done.

app/commands/notification/create.rb
iHiD: Ah - I actually misread this slightly. Let's discuss this one on the call. I'll think about whether I want to use those callbacks or not today.

app/controllers/test/components/notifications/icons_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Test::Components::Notifications::IconsController < ApplicationController
+  def show
+    @user = User.first
+  end
+
+  def update
+    NotificationsChannel.broadcast_to(
```
iHiD, line 7: Can this now use `NotificationsChannel .broadcast_changed`?

test/commands/notification/create_test.rb
```diff
@@ -16,4 +18,15 @@ class Notifications::CreateTest < ActiveSupport::TestCase
     assert_equal "#{user.id}-mentor_started_discussion-Discussion##{discussion.id}", notification.anti_duplicate_key
     assert_equal params, notification.send(:params)
   end
+
+  test "broadcasts message" do
+    user = create :user
+    type = :mentor_started_discussion
+    discussion = create(:solution_mentor_discussion)
```
iHiD, line 25: Let's not use brackets for factory creation. Maybe there's a rubocop rule we could add or something? And something to add to the style guide.

test/helpers/components_helper_test.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class ComponentsHelperTest < ActionView::TestCase
+  include ComponentsHelper
+
+  test "#notification_icon passes the correct parameters" do
+    user = create(:user)
+    create(:notification, user: user, read_at: nil)
+    create(:notification, user: user, read_at: 1.day.ago)
+
+    assert_component_equal notification_icon(user), { id: "notification-icon", props: { count: 1 } }
```
iHiD, line 11: Nice.

app/commands/notification/create.rb
kntsoriano: I'll resolve this now. I've added this to our tracking issue.

app/controllers/test/components/notifications/icons_controller.rb
kntsoriano: Good idea. Adding this to our tracking issue.

PR #75, opened 2020-09-16
Add mentor inbox component

ErikSchierboom:
This is the fourth PR in a series of PR's. https://github.com/exercism/v3-website/pull/74 should be merged first to get the diff down to just the actual changes. Once this is merged, #76 is up for review.

### DIFF
--- a/app/controllers/test/components/mentoring/mentor_conversations_lists/conversations_controller.rb
+++ b/app/controllers/test/components/mentoring/mentor_conversations_lists/conversations_controller.rb
@@ -1,0 +1,41 @@
+class Test::Components::Mentoring::MentorConversationsLists::ConversationsController < ApplicationController
+  def index
+    return head :internal_server_error if params[:state] == "Error" || params[:state] == "Loading"
+
+    results = [
+      {
+        trackTitle: "Ruby",
+        trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
+        menteeAvatarUrl: "https://robohash.org/exercism",
+        menteeHandle: "mentee",
+        exerciseTitle: "Series",
+        isStarred: true,
+        haveMentoredPreviously: true,
+        isNewIteration: true,
+        postsCount: 15,
+        updatedAt: 1.year.ago.iso8601,
+        url: "https://exercism.io/conversations/1"
+      },
+      {
+        trackTitle: "Go",
+        trackIconUrl: "https://assets.exercism.io/tracks/go-hex-white.png",
+        menteeAvatarUrl: "https://robohash.org/exercism_2",
+        menteeHandle: "User 2",
+        exerciseTitle: "Tournament",
+        isStarred: false,
+        haveMentoredPreviously: true,
+        isNewIteration: false,
+        postsCount: 22,
+        updatedAt: 1.week.ago.iso8601,
+        url: "https://exercism.io/conversations/2"
+      }
+    ]
+    page = params.fetch(:page, 1).to_i
+    per = params.fetch(:per, 1).to_i
+
+    render json: {
+      results: results[page - 1, per],
+      meta: { current: page, total: results.size }
+    }
+  end
+end
--- a/app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
+++ b/app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
@@ -1,0 +1,3 @@
+class Test::Components::Mentoring::MentorConversationsListsController < ApplicationController
+  def show; end
+end
--- a/app/javascript/components/mentoring/mentor_conversations_list.jsx
+++ b/app/javascript/components/mentoring/mentor_conversations_list.jsx
@@ -1,0 +1,106 @@
+import React, { useState } from 'react'
+import { usePaginatedQuery } from 'react-query'
+import dayjs from 'dayjs'
+import RelativeTime from 'dayjs/plugin/relativeTime'
+import { Pagination } from './mentor_conversations_list/pagination'
+import { UrlParams } from '../../utils/url_params'
+dayjs.extend(RelativeTime)
+
+function MentorConversationListRow({
+  trackTitle,
+  trackIconUrl,
+  menteeAvatarUrl,
+  menteeHandle,
+  exerciseTitle,
+  isStarred,
+  haveMentoredPreviously,
+  isNewIteration,
+  postsCount,
+  updatedAt,
+  url,
+}) {
+  return (
+    <tr>
+      <td>
+        <img
+          style={{ width: 100 }}
+          src={trackIconUrl}
+          alt={`icon indicating ${trackTitle}`}
+        />
+      </td>
+      <td>
+        <img
+          style={{ width: 100 }}
+          src={menteeAvatarUrl}
+          alt={`avatar for ${menteeHandle}`}
+        />
+      </td>
+      <td>{menteeHandle}</td>
+      <td>{exerciseTitle}</td>
+      <td>{isStarred.toString()}</td>
+      <td>{haveMentoredPreviously.toString()}</td>
+      <td>{isNewIteration.toString()}</td>
+      <td>{postsCount}</td>
+      <td>{dayjs(updatedAt).fromNow()}</td>
+      <td>{url}</td>
+    </tr>
+  )
+}
+
+async function fetchSolutions(key, url, params) {
+  const resp = await fetch(url)
+
+  return resp.json()
+}
+
+export function MentorConversationsList({ endpoint, ...props }) {
+  const [page, setPage] = useState(1)
+  const urlParams = new UrlParams(
+    Object.assign({ page: page }, props.urlParams)
+  )
+  const retryParams = props.retryParams
+  const { status, resolvedData, latestData } = usePaginatedQuery(
+    ['mentor-conversations-list', `${endpoint}?${urlParams.toString()}`],
+    fetchSolutions,
+    retryParams
+  )
+
+  return (
+    <div>
+      {status === 'loading' && <p>Loading</p>}
+      {status === 'error' && <p>Something went wrong</p>}
+      {status === 'success' && (
+        <>
+          <table>
+            <thead>
+              <tr>
+                <th>Track icon</th>
+                <th>Mentee avatar</th>
+                <th>Mentee handle</th>
+                <th>Exercise title</th>
+                <th>Starred?</th>
+                <th>Mentored previously?</th>
+                <th>New iteration?</th>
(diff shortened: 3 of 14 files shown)


config/routes.rb
```diff
@@ -61,6 +61,8 @@
         end
         namespace :notifications do
           resource :icon, only: %i[show update]
+        namespace :mentoring do
```
iHiD, line 65: It looks like we're missing an `end` above this?

app/controllers/test/components/mentoring/mentor_conversations_lists/conversations_controller.rb
```diff
@@ -0,0 +1,41 @@
+class Test::Components::Mentoring::MentorConversationsLists::ConversationsController < ApplicationController
```
iHiD, line 1: Thoughts on renaming `MentorConversationsLists` to `ConversationsList` as it's already scoped?

app/controllers/test/components/mentoring/mentor_conversations_lists/conversations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  def index
+    return head :internal_server_error if params[:state] == "Error" || params[:state] == "Loading"
```
iHiD, line 3: What does this do? Can we add a comment pls?

app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
```diff
@@ -0,0 +1,3 @@
+class Test::Components::Mentoring::MentorConversationsListsController < ApplicationController
+  def show; end
```
iHiD, line 2: @kntsoriano Two questions:
1. Do you like this style or would you prefer it over two lines? 
2. Do we treat empty methods in controllers as a smell? (ie they should be deleted or implemented)?

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
```
iHiD, line 1: Should this be descoped and live in `components/` or `utils/`?

app/javascript/components/mentoring/mentor_conversations_list.jsx
```diff
@@ -0,0 +1,106 @@
+import React, { useState } from 'react'
+import { usePaginatedQuery } from 'react-query'
+import dayjs from 'dayjs'
+import RelativeTime from 'dayjs/plugin/relativeTime'
+import { Pagination } from './mentor_conversations_list/pagination'
+import { UrlParams } from '../../utils/url_params'
+dayjs.extend(RelativeTime)
```
iHiD, line 7: Is this something we can do globally or is it desirable to happen on a per component basis?

app/javascript/components/mentoring/mentor_conversations_list.jsx
```diff
@@ -0,0 +1,106 @@
...
+  const { status, resolvedData, latestData } = usePaginatedQuery(
+    ['mentor-conversations-list', `${endpoint}?${urlParams.toString()}`],
+    fetchSolutions,
+    retryParams
+  )
+
+  return (
+    <div>
+      {status === 'loading' && <p>Loading</p>}
+      {status === 'error' && <p>Something went wrong</p>}
```
iHiD, line 71: Should we add a retry button if we get this? (Not for this PR, but as a future thing)

app/javascript/components/mentoring/mentor_conversations_list.jsx
```diff
@@ -0,0 +1,106 @@
...
+      {status === 'success' && (
+        <>
```
iHiD, line 73: What does this do?

app/views/test/components/mentoring/mentor_conversations_lists/show.html.haml
```diff
@@ -0,0 +1,7 @@
+%h1 Mentor conversations list
+= form_tag nil, method: :get do
+  = label_tag :state
+  = select_tag :state, options_for_select(["Success", "Error", "Loading"], params[:state])
+  = submit_tag "Submit", name: nil
+%hr/
+= mentor_conversations_list(test_components_mentoring_mentor_conversations_list_conversations_path, { state: params[:state] }, { retry: params[:state] === "Loading" })
```
iHiD, line 7: Why `===` and not `==`?

config/routes.rb
```diff
@@ -59,6 +59,14 @@
         namespace :example do
           get 'iterations_summary_table/:id', to: 'iterations_summary_table#index', as: 'iterations_summary_table'
         end
+        namespace :notifications do
+          resource :icon, only: %i[show update]
+        end
+        namespace :mentoring do
+          resource :mentor_conversations_list, only: [:show] do
+            resources :conversations, only: [:index], controller: "mentor_conversations_lists/conversations"
```
iHiD, line 67: Can you merge these two controllers and have an extra action on the one please. I don't want to have two controllers, each with one action, for something here - it makes it harder to work out what's happening, and I don't mind about keeping this "Rails Restful". 

So:
```
resource :mentor_conversations_list, only: [:show] do
  get :conversations, on: :collection
end
```

and move the conversations action onto the `MentorConversationsListController`.

app/controllers/test/components/mentoring/mentor_conversations_lists/conversations_controller.rb
kntsoriano: I'm not entirely sure. The component is named as `<MentorConversationsLists />` still. Maybe we could rename the component to `<Mentoring.ConversationsList />`?

We won't need to think about this though if we go with your suggestion below to make this a controller action.

app/controllers/test/components/mentoring/mentor_conversations_lists/conversations_controller.rb
kntsoriano: This allows us to mock that the controller is returning an error. We want to be able to see our component in the different states that it is in.

app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
kntsoriano: 1. Prefer it like this.
2. Not really. I like having empty controllers on a method in order to have visibility on what actions are implemented.

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
kntsoriano: I'm trying to keep it conservative for now. Let's descope once we have another component that needs this.

app/javascript/components/mentoring/mentor_conversations_list.jsx
kntsoriano: We can do globally, but let's hold off until we have the need to do it.

app/javascript/components/mentoring/mentor_conversations_list.jsx
kntsoriano: Yeah. Good idea. FYI, our react library already retries 3 times before showing the error state. Still a good idea to add a retry button though.

app/javascript/components/mentoring/mentor_conversations_list.jsx
kntsoriano: https://reactjs.org/docs/fragments.html

app/views/test/components/mentoring/mentor_conversations_lists/show.html.haml
kntsoriano: Should be `==`. Too much JS :)

config/routes.rb
kntsoriano: What makes it harder to work out?

app/controllers/test/components/mentoring/mentor_conversations_lists/conversations_controller.rb
iHiD: OK nice. Could you add a comment to this for future reference please.

app/controllers/test/components/mentoring/mentor_conversations_lists_controller.rb
iHiD: OK, cool. Can you add to the standards document for controllers pls.

app/javascript/components/mentoring/mentor_conversations_list/pagination.jsx
iHiD: OK, cool.

app/javascript/components/mentoring/mentor_conversations_list.jsx
iHiD: OK, fine.

app/javascript/components/mentoring/mentor_conversations_list.jsx
iHiD: Nice. That's good to know. Can you add a low-priority issue for it pls.

app/javascript/components/mentoring/mentor_conversations_list.jsx
iHiD: Sweet! Thanks :)

app/views/test/components/mentoring/mentor_conversations_lists/show.html.haml
iHiD: Hahaha! 😁

config/routes.rb
iHiD: Quite simply, I have to look in two files to see what's going on, rather than keeping all the things in one file together.

app/controllers/test/components/mentoring/mentor_conversations_lists/conversations_controller.rb
iHiD: Yeah - let's go with this: `<Mentoring.ConversationsList />`

app/views/test/components/mentoring/mentor_conversations_lists/show.html.haml
iHiD: Can you add an issue to fix it up pls.

config/routes.rb
iHiD: I basically have a strong preference against having lots of controllers with relatively shared/overlapping functionality. It makes it much harder to get a feel for what is going on when you come to the code cold, than if you can look in one place where all the code is together. I would rather encapsulate the functionality of all the relevant things in one controller, than having multiple controllers for the conformity with the REST-ideals.

I'll add this to the standards document.

PR #77, opened 2020-09-16
Allow sorting of mentor inbox

ErikSchierboom:
This is the sixth PR in a series of PR's. https://github.com/exercism/v3-website/pull/76 should be merged first to get the diff down to just the actual changes. Once this is merged, #78 is up for review.


app/controllers/test/components/mentoring/mentor_inboxes_controller.rb
```diff
@@ -1,3 +1,9 @@
 class Test::Components::Mentoring::MentorInboxesController < ApplicationController
-  def show; end
+  def show
+    @sort_options = [
+      { value: 'recent', text: 'Sort by Most Recent' },
```
SleeplessByte: `label` instead of `text` is more idiomatic, I think.

You can consider making the values live on some constant, so that you don't need to "match strings" but it's fine for now.

SleeplessByte (approved):

app/controllers/test/components/mentoring/mentor_inboxes_controller.rb
ErikSchierboom: I've changed it to `label`.

Regarding the constants, good idea! I'll leave it up to @kntsoriano or @iHiD to find a suitable place to define them :)

SleeplessByte: (My suggestion would be a Query class, that encapsulates the logic for the mentor inbox query)

kntsoriano:
Merged to #78.

PR #79, opened 2020-09-17
Rename JavaScript files

ErikSchierboom:
This PR renames the JavaScript files to follow the customary naming convention of using dashes: https://github.com/exercism/v3-website/pull/60#discussion_r485315647.

This is the eight PR in a series of PR's. #78 should be merged first to get the diff down to just the actual changes.

PR #82, opened 2020-09-17
Use classes for JavaScript components

ErikSchierboom:
This implements the suggestion from https://github.com/exercism/v3-website/pull/56#discussion_r483580349. My Rails skills are weak, as I couldn't get the `ViewComponents::` namespacing working.

This is the ninth PR in a series of PR's. #79 should be merged first to get the diff down to just the actual changes.


SleeplessByte:
Didn't
```ruby
module ViewComponents
  class ExampleIterationsSummaryTable < ViewComponent
   ...
```

work?

ErikSchierboom:
It didn't. Maybe I did something wrong. I then called it using `ViewComponents::ExampleIterationsSummaryTable` but I got some uninitialized constant error.

SleeplessByte:
🤷 this works tho, soooo let's leaf/leave it for now 🥇 ?

kntsoriano (approved):
This is Rails conventions being weird.

Any folder one level below `app/` (such as `app/models/ or app/view_components`) will be autoloaded without appending the module name to it. Notice how we can just load models without any namespacing whatsoever.

If we want to stick with this, we'd need to create a `app/view_components/view_components` model and put everything in there, alternatively we could move the folder  to `lib/view_components`.

Another alternative is to name it the rails way which is to append a `ViewComponent` suffix just as how we name controllers (i.e. `MentorInboxViewComponent`)

iHiD:
Nice. I like this a lot.

I would probably rather it move to the command pattern though. So rather than:
```ruby
NotificationIcon.new.render(@user)
```

we could do

```ruby
ViewComponents::RenderNotificationIcon.(@user)
```

That would involve:

1. Move into `app/commands`
2. Add a subdirectory there for `ViewComponents`
3. Rename from `XYZ` to `RenderXYZ`
4. Add `include Mandate` to each
5. Rename `render` to `call`
6. Update the places they're used from `NotificationIcon.new.render(@user)` to `ViewComponents::RenderNotificationIcon.(@user)`

I'd also be happy to get rid of step 3, and just have `ViewComponents::NotificationIcon.(@user)`, or maybe `ViewComponentRenderers::NotificationIcon.(@user)`. Thoughts?

kntsoriano:
@iHiD might hate my OOP.

What do think of this approach?

Component

```ruby
class NotificationIcon < ViewComponent
  attr_reader :user
  def initialize(user)
    @user = user
  end

  def to_s
    react_component("notification-icon", { count: user.notifications.count })
  end
end
```

View

```tuby
%h1 Notification icon
= form_with method: :patch do |f|
  = f.label :count
  = f.number_field :count
  = f.submit "Update"
= NotificationIcon.new(@user)
```

If we need to use any helpers, we can pass in the view context in the initializer as well.

iHiD:
I don't hate it - I think it's good 🙂 

But I'd still use Mandate to avoid the boilerplate initializer.

```ruby
class NotificationIcon < ViewComponent
  attr_reader :user
  def initialize(user)
    @user = user
  end

  def to_s
    react_component("notification-icon", { count: user.notifications.count })
  end
end
```

vs this (presuming that ViewComponent has Mandate included):
```ruby
class NotificationIcon < ViewComponent
  initialize_with :user

  def to_s
    react_component("notification-icon", { count: user.notifications.count })
  end
end
```

Then you can do `NotificationIcon.new(user)` and it should work magically via the `to_s` thing (is that html_safe @kntsoriano?)

I would like these not to be in the global scope though - I want them namespaced under `ViewComponents`. @ErikSchierboom If you try putting it into `app/helpers/view_components/notification_icon.rb` and have 
```ruby
module ViewComponents
  class NotificationIcon < ViewComponent
    ...
  end
end
```

does that work?

ErikSchierboom:
@iHiD That worked!

app/helpers/view_components/example_iterations_summary_table.rb
```diff
@@ -0,0 +1,14 @@
+module ViewComponents
+  class ExampleIterationsSummaryTable < ViewComponent
+    include Mandate
```
iHiD: You should just be able to do this in ViewComponent and not in each individual one I think.

kntsoriano:
@iHiD Yes. It should be html_safe as we're using the tag helper in the `ViewComponent` base class.

Happy with these changes :)

app/helpers/view_components/example_iterations_summary_table.rb
ErikSchierboom: Yep. Updated.

app/helpers/view_components/mentor_inbox.rb
```diff
@@ -0,0 +1,13 @@
+module ViewComponents
+  class MentorInbox < ViewComponent
```
iHiD: Should this be scoped the same as the component, so `ViewComponents::Mentoring::Inbox`?

test/helpers/view_components/notification_icon_test.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class NotificationIconTest < ViewComponentTestCase
+  test "#notification_icon passes the correct parameters" do
+    user = create(:user)
+    create(:notification, user: user, read_at: nil)
+
+    assert_component_equal ViewComponents::NotificationIcon.new(user).to_s,
+                           { id: "notification-icon", props: { count: 1 } }
+  end
+end
```
iHiD: Can you wizz these up for the other view components pls.

app/helpers/view_components/mentor_inbox.rb
iHiD: (I'm not sure what conclusions we've come to on the scoping so far)

ErikSchierboom: I like that. I was unsure about the scoping resolution too, but I _think_ we decided we wanted to do add the additional layer.

ErikSchierboom: Done

test/helpers/view_components/notification_icon_test.rb
ErikSchierboom: I've added tests for the other components. There is a bug in the test verification though. It has to do with how the `assert_component_equal` method is implemented:

```ruby
def assert_component_equal(component, params)
    assert_dom_equal(
      %(<div data-react-#{params[:id]}="true" data-react-data=#{params[:props].to_json} />),
      component
    )
  end
```

The new tests have a `props` argument where there are strings in the values that are quoted in the output. I've briefly looked into this, but figured it is probably much more efficient for one of you to fix this (or point me to the fix) :)

```
Expected: <div data-react-maintaining-iterations-summary-table="true" data-react-data={"solution_id":1,"iterations":[{"id":1,"testsStatus":"pending"},{"id":2,"testsStatus":"passed"}]} />
Actual: <div data-react-example-iterations-summary-table="true" data-react-data="{&quot;solution_id&quot;:90,&quot;iterations&quot;:[{&quot;id&quot;:53,&quot;testsStatus&quot;:&quot;pending&quot;},{&quot;id&quot;:54,&quot;testsStatus&quot;:&quot;passed&quot;}]}" />
```

app/helpers/view_components/mentor_inbox.rb
iHiD: See https://github.com/exercism/v3-project-mangement/issues/21 for the wider discussion of this. We should decide there and then document this in the CONTRIBUTING.md

test/helpers/view_components/notification_icon_test.rb
iHiD: @kntsoriano One for you :) If the `j` method (alias for `escape_javascript`) is available here that might work?

iHiD:
Assigned @kntsoriano to fix the build.

kntsoriano (approved):

PR #92, opened 2020-09-22
Rename JavaScript files

ErikSchierboom:
Superseded #79 which could not be reopened 🤔  

I've also applied @kntsoriano's suggestion about the naming convention for React components.


app/javascript/application/action-cable-consumer.js
```diff
@@ -0,0 +1,2 @@
+import { createConsumer } from '@rails/actioncable'
+export default createConsumer()
```
iHiD: Should this be in `utils/`?

iHiD (approved):

kntsoriano (approved):
Looks good!

app/javascript/application/action-cable-consumer.js
ErikSchierboom: Moved

PR #93, opened 2020-09-23
<Mentoring.Queue />

### LINKED ISSUE #66: Add another mentoring table
## Description
As discussed in our previous meeting, we also want to implement this other mentoring table. We're sticking to the table for now, not including the tabs on the top and the filters on the right side.

## Preview
image

## Questions
1. What do we call this component? Is `<MentorSolutionsList />` okay?
2. What are the "badge-like" things on the third column? cc @taiyab

kntsoriano:
Closes https://github.com/exercism/v3-website/issues/66.

## Description
This adds the other mentoring table in the issue above.

In order to DRY this up, I introduced a new hook named `useList`, added generic `<Sorter />` and `<TextFilter />` components, and moved the `<Pagination />` compnent out.

The `useList` hook returns the necessary functions to update the page, filter, and sort query. I've also exposed a function that allows to add a query parameter to the request in cases we need it, such as having the extra `<TrackFilter />` on `<MentorInbox />`

I decided to go with a hook instead of a generic table component as this makes it more transparent what is being rendered on the page. You can look through each component and see what is rendered. A downside, though, is that there is some setup needed to create tables like these. Let me know what you think about this.


app/helpers/components_helper.rb
```diff
@@ -24,6 +24,10 @@ def mentor_inbox(conversations_request, tracks_request, sort_options)
                     })
   end
 
+  def mentor_solutions_list(request, sort_options)
```
ErikSchierboom: Note: this will need to be change when https://github.com/exercism/v3-website/pull/82 is merged (which has separate classes for the view components).

app/javascript/components/mentoring/sorter.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
```
ErikSchierboom: This component is not specific to mentoring I think and I assume there will be more places where it will be used. When that happens, we should probably move this to the parent `components` directory. The same goes for the pagination and textfilter components. For now, this is perfectly fine.

app/javascript/components/mentoring/mentor_inbox.jsx
```diff
@@ -1,75 +1,33 @@
...
 
   return (
-    <div>
+    <div className="mentor-inbox">
       <TrackFilter request={tracksRequest} setTrack={setTrack} />
-      <ConversationFilter
+      <TextFilter
         filter={conversationsRequest.query.filter}
         setFilter={setFilter}
+        id="conversation-filter"
```
ErikSchierboom: Slightly off-topic: do we have a naming strategy for IDs and classes?

app/javascript/components/mentoring/mentor_solutions_list/solution_list.jsx
```diff
@@ -0,0 +1,47 @@
+import React, { useReducer, useCallback } from 'react'
```
ErikSchierboom: `useReducer` and `useCallback` are unused:

```suggestion
import React from 'react'
```

app/javascript/hooks/use_list.js
```diff
@@ -0,0 +1,62 @@
...
+  )
+
+  const setSort = useCallback(
+    (sort) => {
+      dispatch({ type: 'sort.changed', payload: { sort: sort } })
+    },
+    [dispatch]
+  )
+
+  const addQuery = useCallback(
```
ErikSchierboom: Was there any particular reason to name this `addQuery` instead of `setQuery`?

ErikSchierboom (approved):
Nice work! Maybe also rename the JavaScript files to follow the pattern introduced in https://github.com/exercism/v3-website/pull/92?

app/javascript/hooks/use_list.js
```diff
@@ -0,0 +1,62 @@
...
+      }
+    default:
+      if (process.env.NODE_ENV === 'development') {
+        throw new Error(`Unknown action type: ${action.type}`)
+      }
+      return state
+  }
+}
+
+export function useList(initialRequest) {
```
ErikSchierboom: To me, it wasn't immediately obvious what `useList` does when I saw it included in a component. That said, I t tried coming up with a more descriptive name for this hook but failed.

app/javascript/components/mentoring/mentor_inbox.jsx
iHiD: Please can we use classes everywhere rather than ids if possible (I think I mentioned this in another PR too). We should always presume we might get multiple of these components on a page, and ids should be unique.

In terms of naming strategy, we use hyphens not underscores. I'm very big on scoping with CSS and JS. So that a component's class should have a prefix (eg `co-...`) and then everything that every acts on it should be scoped to be `.co-mentoring-inbox .conversation-filter`. This guarantees that there is no leakage of styles etc between components.

iHiD: - For pages we'd then use `#page-...` for example. 
- For things that are global I tend to use the `.lo-` prefix (for layout)

Happy to change the prefixes, but would like them there.

ErikSchierboom: Cool. That's very convenient. This should probably be added to https://github.com/exercism/v3-website/pull/55/files

app/controllers/test/components/mentoring/mentor_solutions_list/solutions_controller.rb
```diff
@@ -0,0 +1,70 @@
+class Test::Components::Mentoring::MentorSolutionsList::SolutionsController < ApplicationController
```
iHiD, line 1: As per my comment on the other PR, please change this to be a `solutions` action on `Test::Components::Mentoring::MentorSolutionsList`:)

iHiD: See https://github.com/exercism/v3-project-mangement/issues/21

app/controllers/test/components/mentoring/mentor_solutions_list/solutions_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    render json: {
+      results: results[page - 1, per],
+      meta: { current: page, total: results.size }
+    }
+  end
+
+  def sort(results)
+    case params[:sort]
+    when 'exercise'
+      results.sort_by { |c| c[:exerciseTitle] }
```
iHiD: This is a good example of the camelCase that makes my heart weeps and resulted in https://github.com/exercism/v3-project-mangement/issues/20 😉

app/helpers/components_helper.rb
```diff
@@ -24,6 +24,10 @@ def mentor_inbox(conversations_request, tracks_request, sort_options)
...
+    react_component("mentor-solutions-list", { request: request, sort_options: sort_options })
```
iHiD: I feel like `sort_options` should be hard coded into this helper (and class when #82 is merged). I don't think it's the responsibility of the controller and the test controller to pass it it. The sort options are part of the actual component.

app/javascript/components/mentoring/mentor_inbox.jsx
iHiD: Merged that so that we can PR directly to that file and have discussions (like this one) there.

iHiD:
Is there a nicer name for this (akin to Inbox)? I'm thinking maybe Queue. So `Mentor::Queue`. Thoughts?

ErikSchierboom:
I like `Mentor:Queue`.

app/javascript/components/mentoring/mentor_inbox.jsx
kntsoriano: This has a label and thus needs an ID. Thanks for the reminder above, though.

app/javascript/hooks/use_list.js
kntsoriano: It's so hard!

app/javascript/components/mentoring/sorter.jsx
kntsoriano: Agreed. Now that I'm working on the `<TrackList />`, I will extract this to a more generic component.

app/javascript/hooks/use_list.js
kntsoriano: I didn't notice! I will use `setQuery`.

kntsoriano:
I'll rename it to `Mentor::Queue`.

app/helpers/components_helper.rb
kntsoriano: Let's merge #82 today when I fix it. I will then apply the changes here.

app/helpers/components_helper.rb
kntsoriano: I think the sort options should be passed in by the controller as a list of available options to sort by. This makes our client side code easier to change in cases where we want to remove a sorting option. This also keeps the list of sorting options in one place as we also need the sorting options on the server (to determine what option is selected, etc.)

kntsoriano:
@iHiD @ErikSchierboom I've updated this PR to follow our new coding conventions.

app/controllers/test/components/mentoring/queue_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      { value: "recent", label: "Sort by Most Recent" },
+      { value: "exercise", label: "Sort by Exercise" },
+      { value: "student", label: "Sort by Student" }
+    ]
+  end
+
+  def solutions
+    return head :internal_server_error if params[:state] == "Error" || params[:state] == "Loading"
+
+    results = [
```
ErikSchierboom, line 13: This will probably be converted to a serializer (see https://github.com/exercism/v3-website/pull/101) later, right?

ErikSchierboom (approved):
Nice!

app/helpers/components_helper.rb
iHiD: Yeah - I agree with that. But I think they should live in the component class not the controller class on the server.

iHiD: I'll PR it seperately.

iHiD:
Sweet. Nice work! 🙂

iHiD:
@kntsoriano This seems to be missing two of the test files:
- `test/helpers/view_components/mentoring/queue_test.rb`
- `test/javascript/components/mentoring/Queue.test.js`

Can you add them please?

---

@ErikSchierboom I missed this in reviewing too. Let's try and catch these in future.

ErikSchierboom:
Doh. I'll remember

iHiD:
It would be lovely at some point in the future to have a GHA that checks this for each component. I think we have enough standardisation among names of things that this would be possible 🙂

PR #95, opened 2020-09-23
<TrackList />

kntsoriano:
https://github.com/exercism/v3-project-mangement/issues/11

## Description
This is a first pass for our `<TrackList />` component. This PR is based on https://github.com/exercism/v3-website/pull/93 as this allows me to use the `useList` custom hook that I've created on there.

## Preview
Sep-23-2020 14-30-21

## Questions
1. Did I choose the right name?
2. Where should I namespace this? Something like `<Mentee.TrackList />`? `<Learner.TrackList />`?
3. Will this always hold true? (# of concept exercises + # of practice exercises = # of total exercises)
4. I opted to defer the calculation of the progress percentage to the server. I feel like it is easier to change this way and it clears up the responsibilities of the server and client side.
5. Contrary to the description in the issue, I decided to stick with just one `<Track />` component and have them be differentiated with a `isJoined` attribute. I feel that it is cleaner this way and allows us to make filtering and sorting easier. Is this okay? Would the two components diverge so much that they deserve separate components?
6. Along with the point above, I feel that we should defer sorting to the server side. The server should just pass the joined tracks first and the unjoined tracks next. If we decide to add a sorting option on top of this, we can easily pass a query to the server and the server will easily determine the sort order.
7. I've introduced a generic component named `<TextFilter />` (adapted from the tables on the mentoring section). Where should we put this? Create something like `app/javascripts/components/generic`?
8. I've pretty much assumed the JSON response  based on the points I've said above, but if you've firmed on a JSON response already, just paste it here and I'll change the code. @iHiD 

## Changes / Refactors
1. As I tried to pass an initial set of data, notice that I have renamed the `retry` key to `options` for our requests. The `retry` key was the wrong key name all along.
2. I needed to change our table matchers in order to work when more than 1 row is rendered on the table. You now need to pass in a row number for the row you want to match. I've set a default of it to be the first row in order to not change our tests so much.
3. I just realised that `react-query` has different states for loading and fetching. The `loading` state only happens on the initial load of the data, while the `fetching` state happens everytime we try and fetch the data. We might need to change our existing components to this, but we can save that for another PR.


app/helpers/components_helper.rb
```diff
@@ -28,6 +28,12 @@ def mentor_solutions_list(request, sort_options)
     react_component("mentor-solutions-list", { request: request, sort_options: sort_options })
   end
 
+  def track_list(tracks, request)
```
kntsoriano, line 31: This is an example of us passing initial data. We can adapt this pattern for our other components.

app/controllers/test/components/track_lists/tracks_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
+        exerciseCount: 20,
+        conceptExerciseCount: 10,
+        practiceExerciseCount: 10,
+        studentCount: 10,
+        isNew: true,
+        isJoined: true,
+        tags: ["OOP", "Web Dev"],
+        completedExerciseCount: 15,
+        trackProgressPercentage: "66.67",
```
ErikSchierboom, line 18: Any reason why this is a string and not a number?

app/controllers/test/components/track_lists/tracks_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Test::Components::TrackLists::TracksController < ApplicationController
+  def index
+    return head :internal_server_error if params[:state] == "Error" || params[:state] == "Loading"
+
+    results = [
+      {
+        id: 2,
+        trackTitle: "Ruby",
+        trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
+        exerciseCount: 20,
```
ErikSchierboom, line 10: This could also be calculated by doing `conceptExerciseCount + practiceExerciseCount`. Might be worth considering when this is converted to an actual API.

app/controllers/test/components/track_lists/tracks_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        conceptExerciseCount: 10,
+        practiceExerciseCount: 10,
+        studentCount: 10,
+        isNew: true,
+        isJoined: true,
+        tags: ["OOP", "Web Dev"],
+        completedExerciseCount: 15,
+        trackProgressPercentage: "66.67",
+        url: "https://exercism.io/tracks/2"
```
ErikSchierboom, line 19: At first I wondered if we should be including the URL in the response, but having giving it some thought it definitely makes sense. Especially if we consider that we might be exposing the API to the outside world too.

app/controllers/test/components/track_lists_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+        trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
+        exerciseCount: 20,
+        conceptExerciseCount: 10,
+        practiceExerciseCount: 10,
+        studentCount: 10,
+        isNew: true,
+        isJoined: true,
+        tags: ["OOP", "Web Dev"],
+        completedExerciseCount: 15,
+        trackProgressPercentage: "66.67",
```
ErikSchierboom, line 16: Any reason why this is not a number?

app/javascript/components/TrackList/Track.jsx
```diff
@@ -0,0 +1,39 @@
...
+  url,
+}) {
+  return (
+    <tr>
+      <td>
+        <img
+          style={{ width: 100 }}
+          src={trackIconUrl}
+          alt={`icon for ${trackTitle} track`}
+        />
```
ErikSchierboom, line 24: I think this now makes it three different places where this is used. It might be worth considering extracting it to a separate component?

app/javascript/components/TrackList.jsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <div className="track-list">
+      <TextFilter
+        filter={request.query.filter}
+        setFilter={setFilter}
+        id="track-list-filter"
+        placeholder="Search language tracks"
+      />
+      {status === 'error' && <p>Something went wrong</p>}
+      {isFetching && <p>Loading</p>}
```
ErikSchierboom, line 20: I like the use of `isFetching`, which is more declarative than `status == "loading"`. Looking at [the documentation](https://react-query.tanstack.com/docs/api#usequery), it looks like there are also `isError` and `isSuccess` properties being returned. Maybe we could use them to replace the status? 

If you think this is a nice improvement, let me know and I'll create an issue to update them elsewhere too.

ErikSchierboom, line 20: We now have several places where we have this loading element. Maybe we should extract it to a separate component?

app/javascript/components/TrackList.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <div className="track-list">
+      <TextFilter
+        filter={request.query.filter}
+        setFilter={setFilter}
+        id="track-list-filter"
+        placeholder="Search language tracks"
+      />
+      {status === 'error' && <p>Something went wrong</p>}
```
ErikSchierboom, line 19: We now have several places where we have this error element. Maybe we should extract it to a separate component?

app/javascript/components/TrackList.jsx
```diff
@@ -0,0 +1,48 @@
...
+export function TrackList(props) {
+  const [request, setFilter, setSort, setPage] = useList(props.request)
+  const { status, data, isFetching } = useRequestQuery('track-list', request)
+
+  return (
+    <div className="track-list">
+      <TextFilter
+        filter={request.query.filter}
+        setFilter={setFilter}
+        id="track-list-filter"
```
ErikSchierboom, line 16: See [this comment](https://github.com/exercism/v3-website/pull/93/files#r493478915) about preferring class names over IDs.

app/helpers/components_helper.rb
ErikSchierboom: I don't see the `initialData` field being used anywhere? Is that correct?

ErikSchierboom: This needs to be changed when https://github.com/exercism/v3-website/pull/82 is merged.

app/views/test/components/mentoring/mentor_inboxes/show.html.haml
```diff
@@ -6,4 +6,4 @@
   = select_tag :tracks_endpoint_state, options_for_select(%w[Success Error Loading], params[:tracks_endpoint_state])
   = submit_tag "Submit", name: nil
 %hr/
-= mentor_inbox({ endpoint: test_components_mentoring_mentor_inbox_conversations_path, query: { state: params[:conversations_endpoint_state], sort: :exercise }, retry: { retry: params[:conversations_endpoint_state] == "Loading" } }, { endpoint: test_components_mentoring_mentor_inbox_tracks_path, query: { state: params[:tracks_endpoint_state] }, retry: { retry: params[:tracks_endpoint_state] == "Loading" } }, @sort_options)
+= mentor_inbox({ endpoint: test_components_mentoring_mentor_inbox_conversations_path, query: { state: params[:conversations_endpoint_state], sort: :exercise }, options: { retry: params[:conversations_endpoint_state] == "Loading" } }, { endpoint: test_components_mentoring_mentor_inbox_tracks_path, query: { state: params[:tracks_endpoint_state] }, options: { retry: params[:tracks_endpoint_state] == "Loading" } }, @sort_options)
```
ErikSchierboom, line 9: When https://github.com/exercism/v3-website/pull/82 is merged, this needs to be updated.

app/views/test/components/mentoring/mentor_solutions_lists/show.html.haml
```diff
@@ -4,4 +4,4 @@
   = select_tag :state, options_for_select(%w[Success Error Loading], params[:state])
   = submit_tag "Submit", name: nil
 %hr/
-= mentor_solutions_list({ endpoint: test_components_mentoring_mentor_solutions_list_solutions_path, query: { state: params[:state] }, retry: { retry: params[:state] == "Loading" } }, @sort_options)
+= mentor_solutions_list({ endpoint: test_components_mentoring_mentor_solutions_list_solutions_path, query: { state: params[:state] }, options: { retry: params[:state] == "Loading" } }, @sort_options)
```
ErikSchierboom, line 7: When https://github.com/exercism/v3-website/pull/82 is merged, this needs to be updated.

app/views/test/components/track_lists/show.html.haml
```diff
@@ -0,0 +1,11 @@
...
+= form_tag nil, method: :get do
+  = label_tag :state
+  = select_tag :state, options_for_select(%w[Success Error Loading], params[:state])
+  = submit_tag "Submit", name: nil
+
+%br
+Error / Loading states are only seen when filtering
+%hr/
+
+= track_list(@tracks, { endpoint: test_components_track_list_tracks_path, query: { state: params[:state] }, options: { retry: params[:state] == "Loading" } })
```
ErikSchierboom, line 11: When https://github.com/exercism/v3-website/pull/82 is merged, this needs to be updated.

ErikSchierboom (commented):
Great work!

> Did I choose the right name?

I like the name.

> Where should I namespace this? Something like <Mentee.TrackList />? <Learner.TrackList />?

I think we usually refer to our users as Students. A Student does not have to be a Mentee (as in the student is only a mentee in the context of a mentoring). My vote would be for `Student.TrackList` if we namespace this. It is important to consider if we also want to use this component for rendering the tracks when the user is _not_ logged-in.

> Will this always hold true? (# of concept exercises + # of practice exercises = # of total exercises)

I think so.

> I opted to defer the calculation of the progress percentage to the server. I feel like it is easier to change this way and it clears up the responsibilities of the server and client side.

👍 

> Contrary to the description in the issue, I decided to stick with just one <Track /> component and have them be differentiated with a isJoined attribute. I feel that it is cleaner this way and allows us to make filtering and sorting easier. Is this okay? Would the two components diverge so much that they deserve separate components?

They could still diverge a bit. Looking at the design, I see two things that are different: the progress percentage and the completion count. If we were to have this in a single `<Track />` there would have to be two conditional statements. This is not bad necessarily, but I do have a preference for having two separate components. 

> Along with the point above, I feel that we should defer sorting to the server side. The server should just pass the joined tracks first and the unjoined tracks next. If we decide to add a sorting option on top of this, we can easily pass a query to the server and the server will easily determine the sort order.

I like that.

> I've introduced a generic component named <TextFilter /> (adapted from the tables on the mentoring section). Where should we put this? Create something like app/javascripts/components/generic?

The [React docs](https://reactjs.org/docs/faq-structure.html) use `common`, so maybe  `app/javascripts/components/common`?

app/controllers/test/components/track_lists/tracks_controller.rb
iHiD: We're going to have ConceptCount and PracticeExerciseCount and no ExerciseCount.

Also do we prefer XyzCount or NumXyz?

ErikSchierboom: I prefer count personally.

app/controllers/test/components/track_lists/tracks_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Test::Components::TrackLists::TracksController < ApplicationController
+  def index
+    return head :internal_server_error if params[:state] == "Error" || params[:state] == "Loading"
+
+    results = [
+      {
+        id: 2,
+        trackTitle: "Ruby",
```
iHiD, line 8:
```suggestion
        title: "Ruby",
```

app/controllers/test/components/track_lists/tracks_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        trackIconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
```
iHiD, line 9:
```suggestion
        iconUrl: "https://assets.exercism.io/tracks/ruby-hex-white.png",
```

iHiD:
@kntsoriano Great work 🙂 

Note that I've updated the spec a little [here](https://github.com/exercism/v3-project-mangement/issues/11#issuecomment-697957562). This shouldn't change again before launch.

---

> Did I choose the right name?
I like it!

> Where should I namespace this? Something like <Mentee.TrackList />? <Learner.TrackList />?
I think `Student::TrackList` is probably best.

>  Will this always hold true? (# of concept exercises + # of practice exercises = # of total exercises)

We no longer show total exercises

> I opted to defer the calculation of the progress percentage to the server. I feel like it is easier to change this way and it clears up the responsibilities of the server and client side.

See the update on the new exercise. I think the server should return the data (num concepts completed, num total concepts) and then the JS turn that into percentages for the progress bar. The main reason I think this is because this API might also be consumed for other purposes where having the absolute numbers is important. So I'd rather let the JS use it to display using percentages.

> Contrary to the description in the issue, I decided to stick with just one <Track /> component and have them be differentiated with a isJoined attribute. I feel that it is cleaner this way and allows us to make filtering and sorting easier. Is this okay? Would the two components diverge so much that they deserve separate components?

I think they're different components. They'll potentially link to different pages and Only now have the icon+title in common. However, I have no strong preference. so if it's easier/less-code to do it the way you've done it, then I'm personally 100% happy to pragmatically say that that is fine.

> Along with the point above, I feel that we should defer sorting to the server side. The server should just pass the joined tracks first and the unjoined tracks next. If we decide to add a sorting option on top of this, we can easily pass a query to the server and the server will easily determine the sort order.

I agree with this.

> I've introduced a generic component named <TextFilter /> (adapted from the tables on the mentoring section). Where should we put this? Create something like app/javascripts/components/generic?

I don't have an opinion on this.

> I've pretty much assumed the JSON response based on the points I've said above, but if you've firmed on a JSON response already, just paste it here and I'll change the code. @iHiD

I've made a general comment about the JSON serialization [here](https://github.com/exercism/v3-project-mangement/issues/20)

---

As a general note, I'm not going to review the JS on these PRs at all. Mainly because I'm already the furthest behind out of you, me, @ErikSchierboom and @SleeplessByte, and secondly because JS doesn't come naturally it takes quite a lot of energy for me to read things. So it's **incredibly** helpful for you to ask questions in the way you have done here (as well as the Loom's as ever!) as I can get to the heart of the issues quickly. Thank you 🙂

app/controllers/test/components/track_lists/tracks_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Test::Components::TrackLists::TracksController < ApplicationController
```
iHiD, line 1: See https://github.com/exercism/v3-project-mangement/issues/21

app/javascript/components/TrackList/Track.jsx
kntsoriano: Will make this extraction part of the new PR I'm doing.

app/javascript/components/TrackList.jsx
kntsoriano: Cool! I think it makes sense to do that :)

app/javascript/components/TrackList.jsx
kntsoriano: Cool. I will make that part of the new PR I'm doing.

kntsoriano:
@iHiD @ErikSchierboom I will open a new PR for this as we've firmed up more on the details of this component.

app/javascript/components/mentoring/mentor_solutions_list/solution.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import dayjs from 'dayjs'
+import RelativeTime from 'dayjs/plugin/relativeTime'
+dayjs.extend(RelativeTime)
```
SleeplessByte, line 4: You'll want to extract this to a new utils file so that you may replace the library with something else later. I recommend:

```javascript
// time.js

import dayjs from 'dayjs'
import RelativeTime from 'dayjs/plugin/relativeTime'
dayjs.extend(RelativeTime)

export { dayjs }
```

It would be even easier to replace, if you don't expose the entire export, and instead do:

```javascript
// time.js

import dayjs from 'dayjs'
import RelativeTime from 'dayjs/plugin/relativeTime'
dayjs.extend(RelativeTime)

export function fromNow(dateTime) {
  return dayjs(updatedAt).fromNow()
}
```

SleeplessByte (commented):
Code looks sane to me :)

app/javascript/components/mentoring/mentor_solutions_list/solution.jsx
kntsoriano: Sounds good! Will open a PR for this.

PR #97, opened 2020-09-24
Namespace components and standardize controllers

### LINKED ISSUE #87: Add namespace to component name
https://github.com/exercism/v3-website/pull/75#discussion_r492058014

Should we then name it to `<Mentoring.MentorInbox />` or `<Mentoring.Inbox />`?

kntsoriano:
Closes https://github.com/exercism/v3-website/issues/87.
https://github.com/exercism/v3-project-mangement/issues/21

If we're okay with this approach, I will then fix the rest of our components.


iHiD (approved):

app/helpers/components_helper.rb
```diff
@@ -0,0 +1,34 @@
+module ComponentsHelper
```
iHiD: Did we not get rid of this file?

kntsoriano: Ah yes! Bad merge :)

ErikSchierboom (approved):

PR #98, opened 2020-09-24
Convert snake_case from API to camelCase

kntsoriano:
## Description
It seems that we only need this from API calls. When rendering initial data via `react-bootloader`, the keys are converted automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
     "dayjs": "^1.8.35",
     "global": "^4.4.0",
+    "humps": "^2.0.1",
     "isomorphic-fetch": "^2.2.1",
     "prop-types": "^15.7.2",
     "react": "^16.13.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
     "dayjs": "^1.8.35",
     "global": "^4.4.0",
+    "humps": "^2.0.1",
```
ErikSchierboom, line 14: While using a library is nice, it does come at the cost of increased bundle/download size. As we're only doing a relatively simple string transformation, what do you think about just writing it ourselves? Or even better, just copy-pasting it from SO: https://stackoverflow.com/a/2970588/2071395 ;)

ErikSchierboom: The cost of the library is estimated at 1.5Kb: https://bundlephobia.com/result?p=humps@2.0.1

ErikSchierboom (approved):

package.json
kntsoriano: We will also need the transformation back to snake case when passing back the data to the API. Would those two functions now equal 1.5kb? If not, I'm happy to just copy paste our own :)

ErikSchierboom: Right. They might actually. Let's go with the library for now, and we can always revisit later.

ErikSchierboom (approved):

iHiD:
Nice. Thanks 🙂

PR #99, opened 2020-09-24
Use PascalCase for JS directories

kntsoriano:


iHiD:
@kntsoriano I think you've not added the deleted files, so it's just adding rather than renaming. This might be because git is case insensitive.

You may need to do two commits. One which removes the files, and another that readds them with the correct case.

kntsoriano:
@iHiD Oh man. Good catch!

ErikSchierboom:
What was the reason for this change? I forgot. Looking at the [React file structure example](https://reactjs.org/docs/faq-structure.html), they only have the actual components in PascalCase. For some reason I like the directories to be lowercase :)

kntsoriano:
@ErikSchierboom I think the doc above has the first-level folders as "features" and not "sub components", thus they are in lower case. Having that distinction on our end would make it clearer as well. Maybe we could do `mentoring/` instead of `Mentoring/`, and `inbox/` instead of `Inbox/`. Let me know what you think about that.

ErikSchierboom:
Okay. Let's ask our React expert: @SleeplessByte would you mind chipping in?

SleeplessByte:
I think grouping by feature is the right thing to do. I don't care for the naming that much, as long as it's consistent.

Directories `snake_case` or `hypen-case` and component files `LikeThis`, is an easy enough rule to follow. It also seems that that is what I usually do:

image

iHiD:
> Maybe we could do mentoring/ instead of Mentoring/, and Inbox/ instead of inbox/.

I'm not clear on what the difference is between Mentoring and Inbox here. They're both namespaces, right? The fact that Inbox is also a component, is sort of incidental?

ErikSchierboom:
> I'm not clear on what the difference is between Mentoring and Inbox here. They're both namespaces, right? The fact that Inbox is also a component, is sort of incidental?

The second example was incorrect. I meant to use lowercase for both. The only thing that should be uppercase IMHO is the actual component.

kntsoriano:
Cool. Sounds good! Let's just make the components uppercase :)

ErikSchierboom:
Great!

PR #100, opened 2020-09-24
Remove controller nesting

kntsoriano:


app/controllers/test/components/mentoring/inbox_controller.rb
```diff
@@ -6,4 +6,109 @@ def show
       { value: 'student', label: 'Sort by Student' }
     ]
   end
+
+  def tracks
```
kntsoriano, line 10: I'd argue this is less readable as the file is too long. Unless we move the results to a YAML file or something like that?

app/controllers/test/components/mentoring/inbox_controller.rb
```diff
@@ -6,4 +6,109 @@ def show
...
+    per = params.fetch(:per, 1).to_i
+
+    render json: {
+      results: results[page - 1, per],
+      meta: { current: page, total: results.size }
+    }
+  end
+
+  private
+  def sort_conversations(results)
```
kntsoriano, line 104: This could've been without the `conversations` suffix if it were in a different controller.

iHiD: 🙂

app/controllers/test/components/mentoring/inbox_controller.rb
iHiD: I think it would make most sense for this controller to use the actual serializer. Having all this mocked data here means that we could change the serializer and this test would still pass but the actual component would then break. Using the real serializer here (and setting the data up via a factory) gives us much safer test coverage.

I think when we do this, this controller will then become much much smaller.

ErikSchierboom: I like the above suggestion!

iHiD:
Needs a rebase :)

app/controllers/test/components/mentoring/inbox_controller.rb
kntsoriano: Yes. That'll be the best :)

ErikSchierboom: So we'll add the serializer support later once Jeremy's PR has been merged, right?

ErikSchierboom (approved):

app/controllers/test/components/mentoring/inbox_controller.rb
iHiD: Yes, I'll do that later today :)

PR #101, opened 2020-09-24
Add tracks serializer

iHiD:
This adds a serializer for the tracks data. This will power both the public tracks endpoint in the API and the JS.

## Questions
- Is the data correct? Anything missing?
- Any comments on the pattern?
- Are you both happy with the `# TODO` pattern for where the data isn't yet in the schema. Doing this allows me to do smaller PRs for these situations.

## Notes

I'm deliberately removing n+1 queries by doing memoized groups/count etc at the bottom. Any time we serialize data we should be thinking of how we can do things like this to reduce the number of DB queries.

### DIFF
--- a/app/serializers/api/tracks_serializer.rb
+++ b/app/serializers/api/tracks_serializer.rb
@@ -1,0 +1,98 @@
+class API::TracksSerializer
+  include Rails.application.routes.url_helpers
+  include Mandate
+
+  def initialize(tracks, user = nil)
+    @tracks = tracks
+    @user = user
+  end
+
+  def to_hash
+    {
+      tracks: tracks.map do |track|
+        data_for_track(track).merge(user_data_for_track(track))
+      end
+    }
+  end
+
+  private
+  attr_reader :tracks, :user
+
+  def data_for_track(track)
+    {
+      id: track.id,
+      title: track.title,
+      num_concept_exercises: concept_exercise_counts[track.id].to_i,
+      num_practice_exercises: practice_exercise_counts[track.id].to_i,
+      web_url: routes.track_url(track),
+
+      # TODO: Set all three of these
+      icon_url: "https://assets.exercism.io/tracks/ruby-hex-white.png",
+      is_new: true,
+      tags: ["OOP", "Web Dev"]
+    }
+  end
+
+  def user_data_for_track(track)
+    return {} unless user
+
+    {
+      is_joined: joined_track_ids.include?(track.id),
+      num_completed_concept_exercises: completed_concept_exercise_counts[track.id].to_i,
+      num_completed_practice_exercises: completed_practice_exercise_counts[track.id].to_i
+    }
+  end
+
+  memoize
+  def routes
+    Rails.application.routes.url_helpers
+  end
+
+  memoize
+  def concept_exercise_counts
+    ConceptExercise.
+      where(track: tracks).
+      group(:track_id).
+      count
+  end
+
+  memoize
+  def practice_exercise_counts
+    PracticeExercise.
+      where(track: tracks).
+      group(:track_id).
+      count
+  end
+
+  memoize
+  def joined_track_ids
+    UserTrack.
+      where(user: user).
+      where(track: tracks).
+      map(&:track_id)
+  end
+
+  memoize
+  def completed_concept_exercise_counts
+    # TODO: This is currently exercises started. Once we've added
+    # the completed flags to the db we should change it to completed
+    ConceptSolution.
+      joins(:exercise).
+      where(user: user).
+      where('exercises.track_id': tracks).
+      group('exercises.track_id').
+      count
+  end
+
+  memoize
+  def completed_practice_exercise_counts
+    # TODO: This is currently exercises started. Once we've added
+    # the completed flags to the db we should change it to completed
+    PracticeSolution.
+      joins(:exercise).
+      where(user: user).
+      where('exercises.track_id': tracks).
+      group('exercises.track_id').
+      count
+  end
+end
--- a/test/serializers/api/tracks_serializer.rb
+++ b/test/serializers/api/tracks_serializer.rb
@@ -1,0 +1,67 @@
+require 'test_helper'
+
+class API::TracksSerializerTest < ActiveSupport::TestCase
+  test "without user" do
+    track = create :track
+
+    num_concept_exercises = 3
+    num_practice_exercises = 7
+
+    num_concept_exercises.times { create :concept_exercise, track: track }
+    num_practice_exercises.times { create :practice_exercise, track: track }
+
+    serializer = API::TracksSerializer.new([track])
+    expected = {
+      tracks: [
+        {
+          id: track.id,
+          title: track.title,
+          num_concept_exercises: num_concept_exercises,
+          num_practice_exercises: num_practice_exercises,
+          web_url: "https://test.exercism.io/tracks/#{track.slug}",
+
+          # TODO: Set all three of these
+          icon_url: "https://assets.exercism.io/tracks/ruby-hex-white.png",
+          is_new: true,
+          tags: ["OOP", "Web Dev"]
+        }
+      ]
+    }
+
+    assert_equal expected, serializer.to_hash
+  end
+
+  test "with user not joined" do
+    track = create :track
+
+    serializer = API::TracksSerializer.new([track], create(:user))
+
+    track_data = serializer.to_hash[:tracks].first
+    refute track_data[:is_joined]
+    assert_equal 0, track_data[:num_completed_concept_exercises]
+    assert_equal 0, track_data[:num_completed_practice_exercises]
+  end
+
+  test "with user joined and progressed" do
+    track = create :track
+    ces = Array.new(3).map { create :concept_exercise, track: track }
+    pes = Array.new(3).map { create :practice_exercise, track: track }
+
+    user = create :user
+    create :user_track, user: user, track: track
+
+    # TODO: Change to be completed when that is in the db schema
+    # and add a case where it's not completed to check the flag is
+    # being used correctly.
+    create :concept_solution, exercise: ces[0], user: user
+    create :practice_solution, exercise: pes[0], user: user
+    create :practice_solution, exercise: pes[1], user: user
+
+    serializer = API::TracksSerializer.new([track], user)
+
+    track_data = serializer.to_hash[:tracks].first
+    assert track_data[:is_joined]
+    assert_equal 1, track_data[:num_completed_concept_exercises]
+    assert_equal 2, track_data[:num_completed_practice_exercises]
+  end
+end


app/serializers/api/tracks_serializer.rb
```diff
@@ -0,0 +1,98 @@
+class API::TracksSerializer
+  include Rails.application.routes.url_helpers
+  include Mandate
+
+  def initialize(tracks, user = nil)
```
ErikSchierboom, line 5: Is the reason this `initialize` method has to be defined the fact that `user` can be `nil`? I thought `Mandate` generated this code

app/serializers/api/tracks_serializer.rb
```diff
@@ -0,0 +1,98 @@
...
+      num_completed_practice_exercises: completed_practice_exercise_counts[track.id].to_i
+    }
+  end
+
+  memoize
+  def routes
+    Rails.application.routes.url_helpers
+  end
+
+  memoize
```
ErikSchierboom, line 51: How does the memoize works, as in: what is the scope in which the memoized value is cached?

app/serializers/api/tracks_serializer.rb
```diff
@@ -0,0 +1,98 @@
...
+      tracks: tracks.map do |track|
+        data_for_track(track).merge(user_data_for_track(track))
+      end
+    }
+  end
+
+  private
+  attr_reader :tracks, :user
+
+  def data_for_track(track)
```
ErikSchierboom, line 21: What determines when something gets memoized? This information is also quite static.

app/serializers/api/tracks_serializer.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  def initialize(tracks, user = nil)
+    @tracks = tracks
+    @user = user
+  end
+
+  def to_hash
+    {
+      tracks: tracks.map do |track|
+        data_for_track(track).merge(user_data_for_track(track))
```
ErikSchierboom, line 13: I like this pattern of separating non-user and user-specific data.

ErikSchierboom (approved):
Looks good!

app/serializers/api/tracks_serializer.rb
iHiD: Mandate doesn't have support (yet) for `nil` constructor params, and I don't want to have to pass `nil` into it in the places where it's not used. PRs to Mandate welcome to fix this 😉

app/serializers/api/tracks_serializer.rb
iHiD: It effectively does `@#{method_name} ||= the_result_of_the_function`. So it's setting an instance variable (so scoped to the instance of the object - or serializer in this case).

The code is [here](https://github.com/iHiD/mandate/blob/master/lib/mandate/memoize.rb) - I'll happily chat you through it if you'd like to take a walk through some Ruby metaprogramming.

app/serializers/api/tracks_serializer.rb
iHiD: memoize doesn't work on methods that take params, as it memoizes the result of the method the first time it comes across it.

The result of this method is only used once in the class, so it wouldn't be appropriate to memoize it.

ErikSchierboom:
> Is the data correct? Anything missing?

The number of enrolled students?

> Any comments on the pattern?

I like the pattern, but @kntsoriano probably has insight here.

> Are you both happy with the # TODO pattern for where the data isn't yet in the schema. Doing this allows me to do smaller PRs for these situations.

I'm all in favor or smaller PR's.

app/serializers/api/tracks_serializer.rb
ErikSchierboom: Okay!

app/serializers/api/tracks_serializer.rb
ErikSchierboom: Hah!

app/serializers/api/tracks_serializer.rb
```diff
@@ -0,0 +1,98 @@
+class API::TracksSerializer
+  include Rails.application.routes.url_helpers
+  include Mandate
```
kntsoriano, line 3: How do we use `Mandate`? Is it like a swiss army knife of short cuts? Or should it just be used for commands?

app/serializers/api/tracks_serializer.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  def initialize(tracks, user = nil)
+    @tracks = tracks
+    @user = user
+  end
+
+  def to_hash
```
kntsoriano, line 10: Calling `#to_hash` on the serializer doesn't seem a good name since we are actually transforming the object to hash, and not the serializer. `#call` might be better.

Serializers are really awkward objects for me,  as they don't feel like an actual thing in our system. I also don't like naming models that have the `-er` suffix, as it doesn't feel concrete to me. As an alternative, I usually reach for read models instead (example being calling this `StudentTrackList` and calling `to_hash` on it).

That being said though,  we're still starting out and searching for things that work for us. I don't really mind using this pattern to see how it will affect how we use the code.

test/serializers/api/tracks_serializer.rb
```diff
@@ -0,0 +1,67 @@
+require 'test_helper'
+
+class API::TracksSerializerTest < ActiveSupport::TestCase
+  test "without user" do
```
kntsoriano, line 4: Any thoughts on structuring our tests to follow the [Four-Phase Test](https://thoughtbot.com/blog/four-phase-test)?

test/serializers/api/tracks_serializer.rb
```diff
@@ -0,0 +1,67 @@
...
+          is_new: true,
+          tags: ["OOP", "Web Dev"]
+        }
+      ]
+    }
+
+    assert_equal expected, serializer.to_hash
+  end
+
+  test "with user not joined" do
```
kntsoriano, line 34: As I was reading this test, it's somewhat telling me that we could split this into two objects. One dealing with the context of not having a user, and one dealing with the context of having a user. It seems that this is testing too much as we're testing two cases within the user context.

We don't need to change it right now. It's just something that I noticed.

kntsoriano:
With regards the TODO pattern, maybe we could just not include the assertion then. I think it's a lot cleaner this way since we might get to a point where we'll be stuck with lots of TODO comments and we don't know what are finished or not.

app/serializers/api/tracks_serializer.rb
iHiD: Bit of both. It does `memoize` and `initializer` shortcuts when we need those (e.g. in these situations) and then the `.call` magic for the commands. So swiss-army-knife in lots of places then also commands in others.

app/serializers/api/tracks_serializer.rb
iHiD: So the purpose of this is purely to serialize data for outputs. It's quite deliberately not a "thing" in its own right. It shouldn't have any behaviour etc, and just has methods for `to_json` and `to_hash` for when that data is serialized. It's really just an object that outputs a view of something.

I've modelled it as a serializer Object, but we could equally have it as a command (`SerializeTracks`) instead - I'd be quite happy with that.

test/serializers/api/tracks_serializer.rb
iHiD: I think it is. Setup (L5-11), exercise (13), verify (14-31). And then no teardown.

In what way do you think it doesn't conform to that?

test/serializers/api/tracks_serializer.rb
iHiD: Yeah, maybe. My reluctance to do that was that then we have to encode the same logic in multiple places to determine which serializer to call, and so the responsibility moves to multiple non-DRY controllers. Whereas I think what's actually changing is the output of the serialization based on various criteria (e.g. whether there is a user) - so it feels to me like the logic should be encapsualted within that serialization logic.

app/serializers/api/tracks_serializer.rb
ErikSchierboom: > I've modelled it as a serializer Object, but we could equally have it as a command (SerializeTracks) instead - I'd be quite happy with that.

I like that!

iHiD: OK, I can do that as a seperate PR.

iHiD:
> With regards the TODO pattern, maybe we could just not include the assertion then

If we don't include the test, I wouldn't want to include the code either. Else we have untested code. The problem then is that we couldn't test the component end-to-end. It's a tricky one. Fundamentally something has to get stubbed somewhere - either in the JS, or the serializer or the model.

> we might get to a point where we'll be stuck with lots of TODO comments and we don't know what are finished or not.

I think the rule has to be that we are not finished (as in ready to launch) until all TODOs are implemented.

iHiD:
Closing this in favour of #103 but please feel free to continue conversations 🙂

test/serializers/api/tracks_serializer.rb
kntsoriano: Just the spacing between the steps. I think we should have each step as one paragraph.

kntsoriano:
> If we don't include the test, I wouldn't want to include the code either. Else we have untested code. The problem then is that we couldn't test the component end-to-end. It's a tricky one. Fundamentally something has to get stubbed somewhere - either in the JS, or the serializer or the model.

Cool. Sounds good to stub then :)

app/serializers/api/tracks_serializer.rb
kntsoriano: This I guess is what makes me confused. I feel like we should have this in the docs somewhere, as `Mandate` implies it being a command always.

app/serializers/api/tracks_serializer.rb
kntsoriano: I'd argue it's actually a thing, but wouldn't want to press that right now. I'm happy with having this as a command to serialize tracks, but would argue in the future that it would make our code less flexible. Having explicit classes for commands is less flexible than running `to_hash` on the object you want to serialize.

test/serializers/api/tracks_serializer.rb
kntsoriano: Exactly why I think this isn't just a plain serializer.

app/serializers/api/tracks_serializer.rb
iHiD: OK, cool. Maybe we should `include Mandate::Memoize` here, rather than the whole thing?

test/serializers/api/tracks_serializer.rb
iHiD: Ah, ok. Cool.

test/serializers/api/tracks_serializer.rb
iHiD: Yeah - I agree. We need a better name I think.

app/serializers/api/tracks_serializer.rb
iHiD: @kntsoriano Continued this discussion [here](https://github.com/exercism/v3-website/issues/115) as a stand-alone issue 🙂

app/serializers/api/tracks_serializer.rb
kntsoriano: Yeah. That's a really good idea. If that should be easy to change, let's do it.

iHiD: Great. I'll update things where I come across them and do it that way from now on 🙂

PR #102, opened 2020-09-24
Tracks searching

iHiD:
This PR adds a command for searching tracks. It's pretty straight forward 🙂 

Note that I'm **not** adding new migrations for changes. I want to end up with one migration per table to make ETL easier. So you will need to reset your DB your use this. This will be a regular thing to do 🙂 You can do this with: 
```bash
bundle exec db:migrate:reset db:seed
```

### DIFF
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -10,6 +10,8 @@ class Track < ApplicationRecord
 
   delegate :head_sha, to: :repo, prefix: "git"
 
+  scope :active, -> { where(active: true) }
+
   def self.for!(param)
     return param if param.is_a?(Track)
     return find_by!(id: param) if param.is_a?(Numeric)
--- a/test/commands/track/search_test.rb
+++ b/test/commands/track/search_test.rb
@@ -1,0 +1,47 @@
+require "test_helper"
+
+class Badge::CreateTest < ActiveSupport::TestCase
+  test "only include active tracks" do
+    # Create one active and one inactive track
+    track = create :track, active: true
+    create :track, active: false
+
+    assert_equal [track], Track::Search.()
+  end
+
+  test "wildcard search with criteria" do
+    # Create one track with Ruby in the title
+    track = create :track, title: "A Ruby Track"
+    create :track, title: "A JS Track"
+
+    assert_equal [track], Track::Search.(criteria: "ruby")
+  end
+
+  test "status raises without a user" do
+    assert_raises TrackSearchStatusWithoutUserError do
+      Track::Search.(status: :all)
+    end
+  end
+
+  test "status raises unless its valid" do
+    assert_raises TrackSearchInvalidStatusError do
+      Track::Search.(status: :foobar, user: create(:user))
+    end
+  end
+
+  test "status pivots correctly" do
+    user = create :user
+    joined = create :track
+    create :user_track, user: user, track: joined
+    unjoined = create :track
+
+    assert_equal [joined, unjoined],
+                 Track::Search.(status: :all, user: user)
+
+    assert_equal [joined],
+                 Track::Search.(status: :joined, user: user)
+
+    assert_equal [unjoined],
+                 Track::Search.(status: :unjoined, user: user)
+  end
+end
--- a/test/models/track_test.rb
+++ b/test/models/track_test.rb
@@ -16,7 +16,7 @@ class TrackTest < ActiveSupport::TestCase
     assert_equal track, Track.for!(track.slug)
   end
 
-  test "#active scope" do
+  test ".active scope" do
     # Create one active and one inactive track
     track = create :track, active: true
     create :track, active: false
--- a/test/models/track_test.rb
+++ b/test/models/track_test.rb
@@ -16,6 +16,14 @@ class TrackTest < ActiveSupport::TestCase
     assert_equal track, Track.for!(track.slug)
   end
 
+  test "#active scope" do
+    # Create one active and one inactive track
+    track = create :track, active: true
+    create :track, active: false
+
+    assert_equal [track], Track.active
+  end
+
   %w[head_sha].each do |delegate|
     test "delegates git_#{delegate}" do
       slug = SecureRandom.uuid
--- a/app/commands/track/search.rb
+++ b/app/commands/track/search.rb
@@ -1,0 +1,51 @@
+class Track
+  class Search
+    STATUSES = %i[all joined unjoined].freeze
+
+    include Mandate
+
+    def initialize(criteria: nil, tags: nil, status: nil, user: nil)
+      @criteria = criteria
+      @tags = tags
+      @status = status.try(&:to_sym)
+      @user = user
+    end
+
+    def call
+      @tracks = Track.active
+      filter_criteria
+      filter_tags
+      filter_status
+      @tracks
+    end
+
+    private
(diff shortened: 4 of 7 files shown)


test/models/track_test.rb
```diff
@@ -16,6 +16,14 @@ class TrackTest < ActiveSupport::TestCase
     assert_equal track, Track.for!(track.slug)
   end
 
+  test "#active scope" do
+    # Create one active and one inactive track
```
kntsoriano, line 20: I think we can remove this comment.

test/models/track_test.rb
```diff
@@ -16,6 +16,14 @@ class TrackTest < ActiveSupport::TestCase
     assert_equal track, Track.for!(track.slug)
   end
 
+  test "#active scope" do
```
kntsoriano: What do you think about using `.active` instead? It's a class method.

test/commands/track/search_test.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  test "status raises unless its valid" do
+    assert_raises TrackSearchInvalidStatusError do
+      Track::Search.(status: :foobar, user: create(:user))
+    end
+  end
+
+  test "status pivots correctly" do
```
kntsoriano, line 32: Maybe have a separate test for each status? Might be clearer that way.

test/commands/track/search_test.rb
```diff
@@ -0,0 +1,47 @@
+require "test_helper"
+
+class Badge::CreateTest < ActiveSupport::TestCase
```
ErikSchierboom, line 3: Should this be:

```suggestion
class Track::SearchTest < ActiveSupport::TestCase
```

test/commands/track/search_test.rb
ErikSchierboom: I like separating them.

test/models/track_test.rb
ErikSchierboom: Ah, so the convention is to start with a `#` for module methods and with a `.` for class methods?

test/models/track_test.rb
```diff
@@ -16,6 +16,14 @@ class TrackTest < ActiveSupport::TestCase
...
+    # Create one active and one inactive track
+    track = create :track, active: true
+    create :track, active: false
```
ErikSchierboom, line 22: Is the result of this call used anywhere?

app/models/track.rb
```diff
@@ -10,6 +10,8 @@ class Track < ApplicationRecord
 
   delegate :head_sha, to: :repo, prefix: "git"
 
+  scope :active, -> { where(active: true) }
```
ErikSchierboom, line 13: I just read up on what this does. Great feature!

ErikSchierboom (approved):
> Note that I'm not adding new migrations for changes. I want to end up with one migration per table to make ETL easier. So you will need to reset your DB your use this. This will be a regular thing to do 🙂 You can do this with:
>
> bundle exec db:migrate:reset db:seed

If we're expected to do this occasionally, maybe add a script to `scripts/docker`?

ErikSchierboom:
> bundle exec db:migrate:reset db:seed

This didn't work for me.

> bundle exec rake db:reset db:migrate db:seed

did

test/models/track_test.rb
iHiD:
```suggestion
  test ".active scope" do
```

test/models/track_test.rb
iHiD: The reason this comment is there is that L22 creates a track but doesn't do anything with it. Someone could accidentally delete that line, the test would still pass, but the actual usefulness of the test would then be lost. So by adding the comment I wanted to emphasise that two tracks should be created, and the second line isn't some sort of mistake etc.

Is there a better way to achieve this without a comment?

test/models/track_test.rb
iHiD: Ha - And Erik perfectly emphasises my point re the comment. It would be really easy for Erik to remove this line because it seemingly does nothing (although in reality, it's crucial to this test working).

test/commands/track/search_test.rb
iHiD: So, I disagree 🙂 

I like the fact we have one setup that we can then test works as expected. This links back to the other comment about deleting the comment. If we split this into three tests, then we have to have the setup for each test with models that don't get used, which makes the test more brittle.

In my thinking, having the superset of states, and then checking they filter correctly, is the safest way to check that everything works as expected.

iHiD:
Closing this in favour of #103 but please feel free to continue conversations 🙂

iHiD:
> If we're expected to do this occasionally, maybe add a script to scripts/docker?

@ErikSchierboom Yep! Want to do it? 🙂

test/models/track_test.rb
ErikSchierboom: Would assigning the result to an unused variable be better?

ErikSchierboom:
@iHiD Yep!

ErikSchierboom:
https://github.com/exercism/v3-website/pull/110 opened

test/models/track_test.rb
kntsoriano: Assert that the other track is not included in the scope.

test/commands/track/search_test.rb
kntsoriano: This means that this should be a separate object then. It seems that we're testing things too deep.

iHiD: > This means that this should be a separate object then. It seems that we're testing things too deep.

I don't understand what you mean. 

This is just a test that a method is working correctly. The method can take three options and outputs three values as a result. I don't think we're testing anything too deeply here? Could you expand.

I see two options. One is what I have with one test. The second is to replace it with three tests:

```ruby
  test "status with status all" do
    user = create :user
    joined = create :track
    create :user_track, user: user, track: joined
    unjoined = create :track

    assert_equal [joined, unjoined],
                 Track::Search.(status: :all, user: user)
  end

  test "status with status joined" do
    user = create :user
    joined = create :track
    create :user_track, user: user, track: joined
    create :track

    assert_equal [joined],
                 Track::Search.(status: :joined, user: user)
  end

  test "status with status unjoined" do
    user = create :user
    create :track
    create :user_track, user: user, track: joined
    unjoined = create :track

    assert_equal [unjoined],
                 Track::Search.(status: :unjoined, user: user)
  end
```

I don't see how just duplicating the same setup three times makes this test and more robust or clear. Plus adding a fourth state would mean having to remember to add the extra line in 4 places.

kntsoriano: I think what I'm saying that this should be a separate object, something like a `StatusFilter`. We can add the relevant tests to that just concerning the status. Having 3 different assertions for 3 different states is telling me that it should be on it's own object. This ensures that the setup is robust across them all. I just felt that it may be too deep as when we add a new status (a good possibility), having more than 3 assertions on 1 test seems to be too much.

I might be looking too far ahead in the future though. I'm completely okay with what we have right now :) Just thinking out loud here.

iHiD: OK, sweet. I think I'm ok with it - but if it gets problematic then definitely agree that we could split it out. Thanks :)

PR #103, opened 2020-09-24
Integrated tracks

iHiD:
This integrates #101 and #102 with the API and normal controller. The final two commits are new. The first is a little refactoring and the second is new functionality.

This whole PR can be summed up with:
```ruby
    def index
      tracks = Track::Search.(
        criteria: params[:criteria],
        tags: params[:tags],
        status: params[:status],
        user: current_user
      )

      serializer = API::TracksSerializer.new(tracks, current_user)
      render json: serializer.to_hash, status: :ok
    end
```

### DIFF
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -2,7 +2,16 @@ class TracksController < ApplicationController
   before_action :use_track, except: :index
 
   def authenticated_index
-    @tracks = Track.all
+    # TODO: This should cease to be an instance variable
+    # once the view is using the tracks_json to render
+    @tracks = Track::Search.(
+      criteria: params[:criteria],
+      tags: params[:tags],
+      status: params[:status],
+      user: current_user
+    )
+
+    @tracks_json = API::TracksSerializer.new(@tracks, current_user).to_json
   end
 
   def authenticated_show
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -59,3 +59,7 @@ Rails.application.configure do
 
   config.hosts << "website" if ENV['EXERCISM_DOCKER']
 end
+
+Rails.application.routes.default_url_options = {
+  host: "http://lvh.me:3020"
+}
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -89,3 +89,24 @@ class ActiveSupport::TestCase
     ).body.read
   end
 end
+
+class ActionDispatch::IntegrationTest
+  # TODO: Add this when adding devise
+  # include Devise::Test::IntegrationHelpers
+
+  # TODO: Add this implmentation back when devise
+  # is added.
+  def sign_in!(user = nil)
+    #  @current_user = user || create(:user, :onboarded)
+    #  @current_user.confirm
+    #  sign_in @current_user
+  end
+
+  def assert_correct_page(page)
+    assert_includes @response.body, "<div id='#{page}'>"
+  end
+end
+
+ActionDispatch::IntegrationTest.register_encoder :js,
+                                                 param_encoder: ->(params) { params },
+                                                 response_parser: ->(body) { body }
--- a/app/commands/track/search.rb
+++ b/app/commands/track/search.rb
@@ -1,0 +1,51 @@
+class Track
+  class Search
+    STATUSES = %i[all joined unjoined].freeze
+
+    include Mandate
+
+    def initialize(criteria: nil, tags: nil, status: nil, user: nil)
+      @criteria = criteria
+      @tags = tags
+      @status = status.try(&:to_sym)
+      @user = user
+    end
+
+    def call
+      @tracks = Track.active
+      filter_criteria
+      filter_tags
+      filter_status
+      @tracks
+    end
+
+    private
+    attr_reader :criteria, :tags, :status, :user
+    attr_reader :tracks
+
+    def filter_criteria
+      return if criteria.blank?
+
+      @tracks = tracks.where(
+        "title like ?", "%#{criteria}%"
+      )
+    end
+
+    # TODO: Decide how to model tags filter
+    # and add it here.
+    def filter_tags; end
+
+    def filter_status
+      return if status.blank?
+      raise TrackSearchStatusWithoutUserError unless user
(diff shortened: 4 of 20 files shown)


test/test_helper.rb
```diff
@@ -89,3 +89,24 @@ def download_s3_file(bucket, key)
...
+
+  # TODO: Add this implmentation back when devise
+  # is added.
+  def sign_in!(user = nil)
+    #  @current_user = user || create(:user, :onboarded)
+    #  @current_user.confirm
+    #  sign_in @current_user
+  end
+
+  def assert_correct_page(page)
```
kntsoriano, line 105: We don't seem to be using this.

app/controllers/tracks_controller.rb
```diff
@@ -2,7 +2,16 @@ class TracksController < ApplicationController
   before_action :use_track, except: :index
 
   def authenticated_index
-    @tracks = Track.all
+    # TODO: This should cease to be an instance variable
+    # once the view is using the tracks_json to render
+    @tracks = Track::Search.(
```
ErikSchierboom, line 7: Show I be able to navigate to this page using `/tracks/`? I get the following error:

```
Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true

Extracted source (around line #25):

Application Trace | Framework Trace | Full Trace
app/serializers/api/tracks_serializer.rb:25:in `data_for_track'
app/serializers/api/tracks_serializer.rb:11:in `block in to_hash'
app/serializers/api/tracks_serializer.rb:10:in `map'
app/serializers/api/tracks_serializer.rb:10:in `to_hash'
app/serializers/api/serializer.rb:6:in `to_json'
app/controllers/tracks_controller.rb:14:in `authenticated_index'
app/controllers/application_controller.rb:14:in `block (2 levels) in <class:ApplicationController>'
```

config/environments/development.rb
```diff
@@ -59,3 +59,7 @@
 
   config.hosts << "website" if ENV['EXERCISM_DOCKER']
 end
+
+Rails.application.routes.default_url_options = {
+  host: "http://lvh.me:3020"
```
ErikSchierboom, line 64: Does this mean that I should be using this URL to load things locally? It doesn't appear to work with the Docker setup. Do I need to setup something (like a hosts file)?

app/controllers/tracks_controller.rb
iHiD: @ErikSchierboom You need to restart your server. Rails doesn't hot reload to changes in `config/`

config/environments/development.rb
iHiD: Ah - Interesting/odd that it doesn't work with docker. It should just resolve to localhost (https://www.quora.com/What-is-lvh-me).

The reason I/we use it is because you can't do subdomains on localhost: .e.g `api.localhost`, but you can do `api.lvh.me`.

test/test_helper.rb
iHiD: No, it's copied from `website`. I'm slowly moving helper methods etc over as I touch each file.

app/controllers/tracks_controller.rb
ErikSchierboom: Found it. 

```
config.hosts << "lvh.me"
```

needs to be added to the `development.rb` file

ErikSchierboom: [PR submitted](https://github.com/exercism/v3-website/pull/106)

SleeplessByte:
> ```ruby
>       serializer = API::TracksSerializer.new(tracks, current_user)
>       render json: serializer.to_hash, status: :ok
> ```

The amount of magic that goes on in these two lines -- especially the bottom one -- makes the naming even more confusing imo. The serializer isn't a serializer by any conventional means; it's _maybe_ a formatter. Sure you can argue it's "serializing" the data to a rails-JSON-compatible-ruby-object, but none of that is apparent from the naming.

Also not a fan of making these all `application/json` -- and we don't have any schema to validate generated responses against. 

It's probably fine, but right now it's still possible to change what you use or how you write it. If this API is going to be publicly available, at some point, I _strongly_ recommend we media type it ánd validate it 🤷 

---

What I do like are solutions such as [MediaTypes::Serialization](https://github.com/XPBytes/media_types-serialization). Might go "far", but we've never had an instance where a response had a type, field, or value that was unexpected / expected but missing.

```ruby
class Api::TracksController< ActionController::API
  include MediaTypes::Serialization

  allow_output_serializer(Api::TracksSerializer, only: %i[index])
  freeze_io!
      
  def show 
    tracks = Track::Search.(
      criteria: params[:criteria],
      tags: params[:tags],
      status: params[:status],
      user: current_user
    )

    render_media tracks
  end
end
```

iHiD:
> The serializer isn't a serializer by any conventional means; it's maybe a formatter. Sure you can argue it's "serializing" the data to a rails-JSON-compatible-ruby-object, but none of that is apparent from the naming.

In #111 we change this to be a command, not a serializer, to separate the concern slightly. I agree that this isn't serialzing in terms of converting the data from one format to another - it's collecting the data and organising it into an object for serialization.

If you have a suggestion for a better name, I'm open to suggestions. I'm in no way fixed on them being called Serializers. I guess it's more like `Track::CollateDataForAPIRendering` but I'm not calling it that :wink:

> Also not a fan of making these all application/json -- and we don't have any schema to validate generated responses against. It's probably fine, but right now it's still possible to change what you use or how you write it. If this API is going to be publicly available, at some point, I strongly recommend we media type it ánd validate it 🤷

Yeah - happy to do this is a "right" way but for now I'm happy with it just being JSON objects and refactoring it later. The refactor won't be hard as these things are all in one place, and it something I don't want us to get bogged down with while trying to get this over the line.

SleeplessByte:
All good with that. As long as that's a conscious decision 👍

iHiD:
@SleeplessByte Thanks 🙂  Continued the discussion on naming in a stand-alone issue [here](https://github.com/exercism/v3-website/issues/115) and opened an internal issue to remind us to do the media type stuff before launch. When we get to there, I'll come back to you for help 🙂

PR #107, opened 2020-09-25
Allow manually refetching conversations list

ErikSchierboom:
I have some questions:

Do we want to add a test if this works?
Do we want to update the test component to have this working there? At the moment, this button is shown when the state is explicitly set to the Error state, but that changes the URL which means that doing a refetch will include the error state URL and thus the same erronous data is sent back.

### DIFF
--- a/app/javascript/hooks/request-query.js
+++ b/app/javascript/hooks/request-query.js
@@ -3,7 +3,7 @@ import { UrlParams } from '../utils/url-params'
 import fetch from 'isomorphic-fetch'
 import { camelizeKeys } from 'humps'
 
-async function handleFetch(key, url, query) {
+function handleFetch(key, url, query) {
   return fetch(`${url}?${new UrlParams(query).toString()}`)
     .then((response) => response.json())
     .then((json) => camelizeKeys(json))
--- a/test/javascript/components/mentoring/inbox/ConversationList.test.js
+++ b/test/javascript/components/mentoring/inbox/ConversationList.test.js
@@ -1,0 +1,32 @@
+import React from 'react'
+import { render, screen, fireEvent, waitFor } from '@testing-library/react'
+import { rest } from 'msw'
+import { setupServer } from 'msw/node'
+import '@testing-library/jest-dom/extend-expect'
+import { ConversationList } from '../../../../../app/javascript/components/mentoring/inbox/ConversationList.jsx'
+
+const server = setupServer(
+  rest.get('https://exercism.test/conversations', (req, res, ctx) => {
+    return res(ctx.status(500, 'Internal server error'))
+  })
+)
+
+beforeAll(() => server.listen())
+afterEach(() => server.resetHandlers())
+afterAll(() => server.close())
+
+test('allow retry after loading error', async () => {
+  const setPage = jest.fn()
+
+  render(
+    <ConversationList
+      request={{
+        endpoint: 'https://exercism.test/conversations',
+        query: { page: 2 },
+      }}
+      setPage={setPage}
+    />
+  )
+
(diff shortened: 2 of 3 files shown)


ErikSchierboom:
I have some questions:

- Do we want to add a test if this works? 
- Do we want to update the test component to have this working there? At the moment, this button is shown when the state is explicitly set to the Error state, but that changes the URL which means that doing a refetch will include the error state URL and thus the same erronous data is sent back.

iHiD:
> Do we want to add a test if this works?

Yes 😄 

> but that changes the URL which means that doing a refetch will include the error state URL and thus the same erronous data is sent back.

Will this happen in production or just in the test mode? I'm guessing just the test mode?

ErikSchierboom:
> Will this happen in production or just in the test mode? I'm guessing just the test mode?

Just the test mode (because the URL contains the error state which influences the JSON response)

I'll add the test.

iHiD:
@kntsoriano Can you review this one pls?

ErikSchierboom:
Note: I still have to add the test :)

app/javascript/hooks/request-query.js
```diff
@@ -3,7 +3,7 @@ import { UrlParams } from '../utils/url-params'
 import fetch from 'isomorphic-fetch'
 import { camelizeKeys } from 'humps'
 
-async function handleFetch(key, url, query) {
+function handleFetch(key, url, query) {
```
ErikSchierboom, line 6: We didn't use the `await` keyword, so the `async` modifier doesn't actually do anything.

test/javascript/components/mentoring/inbox/ConversationList.test.js
```diff
@@ -0,0 +1,32 @@
...
+    <ConversationList
+      request={{
+        endpoint: 'https://exercism.test/conversations',
+        query: { page: 2 },
+      }}
+      setPage={setPage}
+    />
+  )
+
+  await waitFor(() => expect(screen.getByText('Retry')).toBeInTheDocument())
```
ErikSchierboom, line 32: @kntsoriano For some reason this does not work. I've dug into this a bit and even though I've added the following server setup:

```js
rest.get('https://exercism.test/conversations', (req, res, ctx) => {
    return res(ctx.status(500, 'Internal server error'))
  })
```

The `status` key returned by the `usePaginatedRequestQuery` call in the `ConversationList` component does not change from `loading` to `error`. Passing valid JSON data does change that status from `loading` to `success`. Any ideas?

iHiD:
Tis red.

app/javascript/hooks/request-query.js
kntsoriano: Ah yes!

test/javascript/components/mentoring/inbox/ConversationList.test.js
kntsoriano:
```
    <ConversationList
      request={{
        endpoint: 'https://exercism.test/conversations',
        query: { page: 2 },
        options: { retry: false }
      }}
```

Can you try that?

ErikSchierboom: Ha, found it! There was a mistake in how options were passed to the `usePaginatedQuery` and `useQuery` functions. I've fixed it in this PR.

ErikSchierboom:
Tis green!

iHiD:
I'm going to try and not review this to check you two can work on purely JS changes without me. Let me know if you can't merge once Karlo has :+1:'d it.

iHiD:
1. @ErikSchierboom It needs a rebase
2. @kntsoriano Pls review 🙂

kntsoriano (approved):
Looks great!

PR #108, opened 2020-09-25
Refactor sort options into the relevant view component class

iHiD:
As I commented [here](https://github.com/exercism/v3-website/pull/93#discussion_r493896871), I think sort_options are a property of the component, not the controller. So I've moved them in here. This still keeps everything serverside, but removes the duplication between identical data in app and test that could cause out-of-sync-bugs.

### DIFF
--- a/app/helpers/view_components/mentoring/inbox.rb
+++ b/app/helpers/view_components/mentoring/inbox.rb
@@ -1,15 +1,24 @@
 module ViewComponents
   module Mentoring
     class Inbox < ViewComponent
-      initialize_with :conversations_request, :tracks_request, :sort_options
+      initialize_with :conversations_request, :tracks_request
 
       def to_s
-        react_component("mentoring-inbox", {
-                          conversations_request: conversations_request,
-                          tracks_request: tracks_request,
-                          sort_options: sort_options
-                        })
+        react_component(
+          "mentoring-inbox",
+          {
+            conversations_request: conversations_request,
+            tracks_request: tracks_request,
+            sort_options: SORT_OPTIONS
+          }
+        )
       end
+
+      SORT_OPTIONS = [
+        { value: 'recent', label: 'Sort by Most Recent' },
+        { value: 'exercise', label: 'Sort by Exercise' },
+        { value: 'student', label: 'Sort by Student' }
+      ].freeze
     end
   end
 end
--- a/app/helpers/view_components/mentoring/inbox.rb
+++ b/app/helpers/view_components/mentoring/inbox.rb
@@ -19,6 +19,7 @@ module ViewComponents
         { value: 'exercise', label: 'Sort by Exercise' },
         { value: 'student', label: 'Sort by Student' }
       ].freeze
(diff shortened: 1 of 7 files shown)


app/helpers/view_components/mentoring/inbox.rb
```diff
@@ -1,15 +1,24 @@
...
+            sort_options: SORT_OPTIONS
+          }
+        )
       end
+
+      SORT_OPTIONS = [
+        { value: 'recent', label: 'Sort by Most Recent' },
+        { value: 'exercise', label: 'Sort by Exercise' },
+        { value: 'student', label: 'Sort by Student' }
+      ].freeze
```
iHiD, line 21: I initially set this to be a `private_constant` but then used it in the tests so removed that.

test/helpers/view_components/mentoring/inbox_test.rb
```diff
@@ -4,23 +4,18 @@ class MentoringInboxTest < ViewComponentTestCase
...
-                             tracks_request: { endpoint: "tracks-endpoint" },
-                             sort_options: [
-                               { value: 'recent', label: 'Sort by Most Recent' },
-                               { value: 'exercise', label: 'Sort by Exercise' },
-                               { value: 'student', label: 'Sort by Student' }
-                             ]
-                           } }
+    component = ViewComponents::Mentoring::Inbox.new(
+      conversations_request, tracks_request
+    ).to_s
```
iHiD, line 10: I pulled this out of the assert bit below as I found it really hard to read having that code and the hash there as well with the slightly strange indenting that Rubocop is insisting on.

test/helpers/view_components/mentoring/inbox_test.rb
```diff
@@ -4,23 +4,21 @@ class MentoringInboxTest < ViewComponentTestCase
...
+
+    assert_component_equal(
+      component,
+      {
+        id: "mentoring-inbox",
+        props: {
+          conversations_request: { endpoint: "conversations-endpoint" },
+          tracks_request: { endpoint: "tracks-endpoint" },
+          sort_options: ViewComponents::Mentoring::Inbox::SORT_OPTIONS
```
ErikSchierboom: I'm not a big fan of tests directly using values in the component being tested for the expected value. While I can see the advantage (changing it in the source still has the test passing), I feel that the test is now somehow less valuable, as any typos in the sort options would not get caught. It also limits refactoring of the sort options as there is this direct link. My personal preference would be to duplicate the sort options here and make it an internal details of the component.

ErikSchierboom (approved):
I totally agree with the fact that the sort options should be moved to the view component. I've left one comment on the test.

test/helpers/view_components/mentoring/inbox_test.rb
iHiD: I agree with that. I'll update it.

ErikSchierboom (approved):
Great!

kntsoriano (approved):

PR #111, opened 2020-09-25
Change serializers to be commands

iHiD:
Fixes the [issue]( https://github.com/exercism/v3-website/pull/101#discussion_r494743757) that @kntsoriano raised and changes our serializers to be commands.

- I also descoped them from `API::` for now as that's premature.
- I am also now using Exercism::Routes for the routing, which is a richer proxy to the normal routing. This effectively encapsulates that in our own component rather than requiring a Rails internal everywhere.


config/initializers/routes.rb
```diff
@@ -1,4 +1,12 @@
 class Exercism::Routes
+  def self.method_missing(name, *args)
+    routes.send(name, *args)
+  end
+
+  def self.respond_to_missing?(name, include_all = false)
+    routes.respond_to?(name, include_all)
+  end
```
ErikSchierboom, line 8: What does this new functionality do?

ErikSchierboom (approved):

config/initializers/routes.rb
iHiD: Basically `Exercism::Routes` can have some methods defined on it (e.g. `personal_solution_url` below). For all methods that aren't defined on it, it proxies those methods to the routes object (which is basically the application's internal routing table). So we can call `Exercism::Routes.tracks_url` and that returns the actual `tracks_url`.

Internally in Ruby:
- `method_missing` is what gets called on any Ruby object if the method that's called isn't defined. It takes the name of the method that was called and any args.
- `respond_to?` is defined on every object and returns whether a method is defined. `respond_to_missing?` is a layer that pairs with method_missing to respond sensibly in the same way.

Does that make sense?

ErikSchierboom: Right. So `Exercism::Routes` is like a proxy.

iHiD: Yes. A proxy that can have extra routes defined on it, that can't be easily defined in config/routes. e.g. conditional routes, or routes where the params are complex, and we want to encapsulate the route generation in one place. We're still always calling the underline routes defined in `config/routes` but this allows us to have some extra useful routes defined, but abstract that away from the rest of the app.

kntsoriano (approved):

config/initializers/routes.rb
```diff
@@ -1,4 +1,12 @@
 class Exercism::Routes
```
kntsoriano, line 1: I like this.

iHiD: Great. I thought you would! 🙂 You're a good influence on me 😉

kntsoriano: It's the other way around ;)

PR #112, opened 2020-09-27
[Feature] Add ConceptGraph

neenjaw:
This is a translation of [neenjaw/ruby-exercism-graph](https://github.com/neenjaw/ruby-exercism-graph) according to neenjaw/ruby-exercism-graph#1.

Purpose: The Concept Graph of the v3-dashboard created an important visualization for the progression of the track as it is to be presented to students.  Non-final versions of this are being worked on for the student-facing website.  This is a small class library to be able to transform a list of concept-exercise hashes (not ActiveRecords) into a directed graph, and also to compute a presentation layout.

Four files are being added here (2 source with 2 test):
- `app/models/concept_graph.rb`
  - The object creates the graph from the list of concept-exercise hashes
- `lib/graph_utils.rb`
  - provides module methods for working with graph data structures
- `test/app/models/concept_graph_test.rb`
- `test/lib/graph_utils.rb`

Things to work on:
- how the concept graph obtains its data & when, how it is triggered
- test coverage
  - Being newer to rails, I am unsure how to test this feature in isolation from the other tests (I don't have mariadb set up locally and my attempts to run tests have failed because I think it can't connect to a test database)
  - How are test fixtures defined in rails tests?
    - Just so I can test a known list of exercises and determine the expected results


app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+    @edges = make_edges
+    @adjacency = GraphUtils.make_adjacency_list(self)
+    @has_cycle = GraphUtils.contains_cycle?(adjacency)
+  end
+
+  # Layout computes the ordering of the exercises in a track by creating
+  # a graph then traversing it to discover and memoize the progression
+  def to_layout
+    return [] if nodes.empty?
+    return :error if has_cycle
```
iHiD, line 28: Let's inline this. We only use the variable in one place so we don't need it extracted. This way we also can do it JIT.

```suggestion
    return :error if GraphUtils.contains_cycle?(adjacency)
```

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
+# ConceptGraph transforms a track configuration to a form compatible for manipulating for
+# displaying the progression of the track
+class ConceptsGraph
```
iHiD, line 3: Mandate will give us some useful helper methods - not least `memoize`.

```suggestion
class ConceptsGraph
  include Mandate
```

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+        index: i,
+        slug: exercise['slug'],
+        uuid: exercise['uuid'],
+        concepts: exercise['concepts'],
+        prerequisites: exercise['prerequisites']
+      )
+    end.freeze
+  end
+
+  def make_concept_lookup
```
iHiD, line 53: Rather than setting the instance variable, let's just pop this in a memoized method.

```suggestion
  memoize
  def node_lookup_by_concept
```

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+    :node_lookup_by_index,
+    :adjacency,
+    :has_cycle
+  )
+
+  def initialize(track_config)
+    @language = track_config['language']
+    @nodes = make_nodes(track_config['exercises'])
+    @node_lookup_by_concept = make_concept_lookup
+    @node_lookup_by_index = make_index_lookup
```
iHiD, line 18: These can be deleted if extracted into the memoized method.

```suggestion
    @node_lookup_by_concept = make_concept_lookup
    @node_lookup_by_index = make_index_lookup
```

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+    @edges = make_edges
+    @adjacency = GraphUtils.make_adjacency_list(self)
+    @has_cycle = GraphUtils.contains_cycle?(adjacency)
```
iHiD, line 21: This can be inlined and deleted (see below)

```suggestion
```

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+    :adjacency,
+    :has_cycle
+  )
+
+  def initialize(track_config)
+    @language = track_config['language']
+    @nodes = make_nodes(track_config['exercises'])
+    @node_lookup_by_concept = make_concept_lookup
+    @node_lookup_by_index = make_index_lookup
+    @edges = make_edges
```
iHiD, line 19: And this into a memoized method too.

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+    :language,
+    :nodes,
+    :edges,
+    :node_lookup_by_concept,
+    :node_lookup_by_index,
+    :adjacency,
+    :has_cycle
+  )
+
+  def initialize(track_config)
```
iHiD, line 14: This will take a `track` object.

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+    @language = track_config['language']
+    @nodes = make_nodes(track_config['exercises'])
```
iHiD, line 16: Pop this as a memoized method. You can directly refer to `track.concepts` and `track.practice_exercises` from in that method.

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
+# ConceptGraph transforms a track configuration to a form compatible for manipulating for
+# displaying the progression of the track
+class ConceptsGraph
+  attr_reader(
+    :language,
```
iHiD, line 5: This can be removed.

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+    :nodes,
+    :edges,
+    :node_lookup_by_concept,
+    :node_lookup_by_index,
+    :adjacency,
+    :has_cycle
+  )
+
+  def initialize(track_config)
+    @language = track_config['language']
```
iHiD, line 15: This can be removed.

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+        concepts: exercise['concepts'],
+        prerequisites: exercise['prerequisites']
+      )
+    end.freeze
+  end
+
+  def make_concept_lookup
+    nodes.each_with_object({}) do |node, lookup|
+      node.concepts.each { |concept| lookup[concept] = node }
+    end.freeze
```
iHiD, line 56: This is another way to do this:

```suggestion
  nodes.flat_map do |node|
    node.concepts.map { |concept| [concept, node] }
  end.to_h.freeze
```

I don't have a preference here, but in other places it's helpful (see below)

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+    nodes.index_by(&:index).freeze
+  end
+
+  def make_edges
+    nodes.each_with_object([]) do |node, aggregate|
+      node.prerequisites.each_with_object(aggregate) do |prereq, a|
+        prereq_node = node_lookup_by_concept[prereq]
+        a << Edge.new(from: prereq_node, to: node) unless prereq_node.nil?
+      end
+    end.freeze
```
iHiD, line 69: An example of where flat_map can help 🙂 

```suggestion
    nodes.flat_map do |node|
      node.prerequisites.map do |prereq|
        prereq_node = node_lookup_by_concept[prereq]
        Edge.new(from: prereq_node, to: node) unless prereq_node.nil?
      end
    end.freeze
```

app/models/concepts_graph.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  # Node for representing an exercise within a track
+  class Node
+    attr_reader :index, :slug, :concepts, :prerequisites
+
+    def initialize(index:, slug:, uuid:, concepts:, prerequisites:)
+      @index = index
+      @slug = slug
+      @uuid = uuid
+      @concepts = concepts.reduce([]) { |a, concept| a << concept.dup }
```
iHiD, line 94:
```suggestion
      @concepts = concepts.map { |concept| concept.dup }
```

I'm not sure I understand why we're duplicating the things here. What does it add?

iHiD:
Excellent work.

I've put together a PR with some suggestions and comments [here](https://github.com/neenjaw/exercism-v3-website/pull/1).

iHiD:
> how the concept graph obtains its data & when, how it is triggered

So, my thinking is:
- The levels are stored as a JSON field on the tracks table. Let's call it `concept_levels`.

Then we have something like

```
class Track
...
  def concept_levels
    super || begin
      Track::DetermineConceptLevels.(self).tap do |levels|
        update(levels: levels)
      end
    end
  end
...
end
```

> Being newer to rails, I am unsure how to test this feature in isolation from the other tests (I don't have mariadb set up locally and my attempts to run tests have failed because I think it can't connect to a test database)

To test one file:
```
rails test test/lib/graph_utils_test.rb
```

However, you probably need the `development_environment` setup to do anything. There are then instructions in the README here.

iHiD:
> How are test fixtures defined in rails tests?
> Just so I can test a known list of exercises and determine the expected results

We use FactoryBot to give us factories rather than fixtures for general situations. So far we've not had any data that's not in the db or git (see `test/repos`) as the source.

Having a `test/fixtures/...` directory is idiomatic. So creating that and putting it there would be fine.

neenjaw:
> [In regard to the data format](https://github.com/exercism/v3-website/pull/112#issuecomment-699694331)

There are a few things that I am unfamiliar with in your code snippet (`def.concept_levels` (is the period intentional?), `begin..end`, but what I assume this only determines the concept levels if they previously were not determined or if they needed to be updated (in the case of a track change)?

neenjaw:
I will get the development_environment working on my machine, I don't think I've kept it up to date so will go through the readme instructions again.

iHiD:
> def.concept_levels

Dot was a typo (I've correct it now). 

> begin..end, but what I assume this only determines the concept levels if they previously were not determined or if they needed to be updated (in the case of a track change)?

Yeah so you can wrap anything in a `begin...end` block and then treat it as a single thing. So `super || begin...end` says call `super` (get the data from the db) and if it's `nil`, then do `begin...end`. The result of that will get returned as the result of the function.

> I will get the development_environment working on my machine, I don't think I've kept it up to date so will go through the readme instructions again.

Cool. It should now be a bazillion times easier than the last time you probably tried it 🙂 

Also - README updates for any unclear bits are appreciated.

app/commands/track/determine_concepts_layout.rb
```diff
@@ -0,0 +1,121 @@
...
+      include Mandate
+
+      # Edge representing a directed edge
+      Edge = Struct.new(:from, :to, keyword_init: true)
+
+      # Node for representing an exercise within a track
+      Node = Struct.new(:index, :slug, :uuid, :concepts, :prerequisites, keyword_init: true)
+
+      # Add public readers
+      attr_reader :nodes, :edges
```
iHiD: Switch this to be private now.

app/commands/track/determine_concepts_layout.rb
```diff
@@ -0,0 +1,121 @@
...
+      memoize
+      def has_cycle?
+        GraphUtils.contains_cycle?(adjacencies)
+      end
+
+      def node_for_concept(concept)
+        nodes_keyed_by_concept[concept]
+      end
+
+      def node_for_index(index)
```
iHiD, line 118: We don't publically use this, so we could make it private or just use `nodes_keyed_by_index` directly internally.

app/commands/track/determine_concepts_layout.rb
```diff
@@ -0,0 +1,121 @@
...
+        exercises.map.with_index do |exercise, idx|
+          Node.new(
+            index: idx,
+            slug: exercise['slug'],
+            uuid: exercise['uuid'],
+            concepts: exercise['concepts'],
+            prerequisites: exercise['prerequisites']
+          )
+        end.freeze
+      end
```
iHiD, line 151: Just a future note: We won't actually need this method when we pass real data in, as the exercises will already be things that have all of these methods defined on them. So the constructor will just be able to have `@nodes = exercises`

test/commands/track/determine_concepts_layout_test.rb
```diff
@@ -0,0 +1,45 @@
...
+            'prerequisites' => %w[basics]
+          },
+          {
+            'slug' => 'basics',
+            'uuid' => 'c29c6092-9d44-4f21-8138-b873384fd90b',
+            'concepts' => %w[basics],
+            'prerequisites' => []
+          }
+        ]
+      }.freeze
```
iHiD: I wouldn't bother changing this, as we'll run this of factories before we merge anyway, but this would be more Rubyish to use symbols for the keys, rather than strings and fat arrows. e.g.:

```
            slug: 'basics',
            uuid: 'c29c6092-9d44-4f21-8138-b873384fd90b',
            concepts: %w[basics],
            prerequisites: []
          }
```

test/lib/graph_utils_test.rb
```diff
@@ -0,0 +1,30 @@
+require "test_helper"
+require 'graph_utils'
+
+class GraphUtilsTest < ActiveSupport::TestCase
+  def test_detect_cycles
+    assert GraphUtils.contains_cycle?([[1], [0]])
+    assert GraphUtils.contains_cycle?([[1], [2], [0]])
+    assert GraphUtils.contains_cycle?([[1], [2], [3], [1]])
```
iHiD, line 8: Can we have a `refute` or two in here too pls. Atm the test would pass if it `contains_cycle?` just returned `true`.

app/commands/track/determine_concepts_layout.rb
neenjaw: `#node_for_index(idx)` is used in `DetermineConceptsLayout#call` on line 23, did you mean `#node_for_concept(concept)`?

app/commands/track/determine_concepts_layout.rb
neenjaw: I'm uncertain which segment of the code you're referring to here by the snippet shown, but I think what you are saying is that the `Node` class may be redundant at some point in the future.  One limitation would be that an arbitrary index is assigned to the node in this implementation to facilitate graph manipulation since the graph functions use the index of the array to represent the node in the graph.

app/commands/track/determine_concepts_layout.rb
neenjaw: Done, see next commit

test/lib/graph_utils_test.rb
neenjaw: done

app/commands/track/determine_concepts_layout.rb
iHiD: Yes. Sorry 🙂

app/commands/track/determine_concepts_layout.rb
iHiD: OK, cool. Yeah - that's good to know 🙂

iHiD: (For clarity, I meant the whole of this method above).

ErikSchierboom:
Thanks for working on this @neenjaw! Much appreciated.

neenjaw:
I think also implementing something like a [transitive reduction pass](https://en.wikipedia.org/wiki/Transitive_reduction) might be worthwhile as well to reduce path noise.

ErikSchierboom:
That (the transitive reduction pass) _definitely_ looks like a great idea.

iHiD:
> I think also implementing something like a transitive reduction pass might be worthwhile as well to reduce path noise.


Yes. Agreed!

config/initializers/exceptions.rb
```diff
@@ -6,3 +6,4 @@ class SolutionLockedByAnotherMentorError < RuntimeError; end
 class TooManyIterationsError < RuntimeError; end
 class TrackSearchStatusWithoutUserError < RuntimeError; end
 class TrackSearchInvalidStatusError < RuntimeError; end
+class TrackHasCyclicPrerequisiteError < RuntimeError; end
```
iHiD: Good name. Could you put this in alphabetical order pls :)

app/commands/track/determine_concepts_layout.rb
```diff
@@ -0,0 +1,117 @@
+require 'graph_utils'
+
+# ConceptGraph transforms a track configuration to a form compatible for manipulating for
+# displaying the progression of the track
+class Track
+  class DetermineConceptsLayout
+    include Mandate
+
+    def initialize(track_config)
```
iHiD: So, presuming we pass a `track` into this, we have:

```ruby
track.concept_exercises #=> [Track::ConceptExercise]

# Then for each concept exercise
concept_exercise.taught_concepts #=> [Track::Concept]
concept_exercise.prerequisites #=> [Track::Concept]
```

You can get these set up in the tests using something like:

```ruby
track = create track
strings = create :track_concept, slug: "strings", track: track
booleans = create :track_concept, slug: "booleans", track: track

e1 = create :concept_exercise
e1.taught_concepts << strings

e2 = create :concept_exercise
e2.taught_concepts << booleans

e3 = create :concept_exercise
e3.prerequisites << strings
e3.prerequisites << booleans
```

This currently relies on #152 being merged.

app/commands/track/determine_concepts_layout.rb
```diff
@@ -0,0 +1,142 @@
...
+          end
+        end
+      end
+
+      def determine_nodes
+        exercises.map.with_index do |exercise, idx|
+          Node.new(
+            index: idx,
+            slug: exercise['slug'],
+            uuid: exercise['uuid'],
```
iHiD: Two thoughts:
1. I think it would be worth passing in the Concept object (which you'll get from the code above: https://github.com/exercism/v3-website/pull/112/files#r499166532). We can then use that to generate URLs, get their names, etc.
2. We don't use this uuid field anywhere as far as I can see.

app/commands/track/determine_concepts_layout.rb
```diff
@@ -0,0 +1,142 @@
...
+      end
+    end
+
+    memoize
+    def connections
+      levels.drop(1).with_index.flat_map do |(level, level_idx)|
+        level.flat_map do |node|
+          node.prerequisites.
+            map { |prerequisite| graph.node_for_concept(prerequisite) }.
+            reject(&:nil?).
```
iHiD:
```suggestion
            compact.
```

app/commands/track/determine_concepts_layout.rb
```diff
@@ -0,0 +1,142 @@
...
+    attr_reader :graph
+
+    memoize
+    def levels
+      return [] if graph.empty?
+
+      raise TrackHasCyclicPrerequisiteError if graph.has_cycle?
+
+      Array.new(graph.levels.max + 1) { [] }.tap do |level|
+        graph.levels.each_with_index do |level_idx, node_idx|
```
iHiD:
```suggestion
        graph.levels.each.with_index do |level_idx, node_idx|
```

app/commands/track/determine_concepts_layout.rb
```diff
@@ -0,0 +1,142 @@
...
+          node.level = level_idx
+
+          level[level_idx] << node
+        end
+      end
+    end
+
+    memoize
+    def connections
+      levels.drop(1).with_index.flat_map do |(level, level_idx)|
```
iHiD:
```suggestion
      levels.drop(1).each.with_index.flat_map do |level, level_idx|
```

app/commands/track/determine_concepts_layout.rb
```diff
@@ -0,0 +1,142 @@
...
+        level.flat_map do |node|
+          node.prerequisites.
+            map { |prerequisite| graph.node_for_concept(prerequisite) }.
+            reject(&:nil?).
+            uniq.
+            select { |prerequisite_node| prerequisite_node.level + 1 == level_idx }.
```
iHiD:
```suggestion
            select { |prerequisite_node| prerequisite_node.level == level_idx }.
```

app/commands/track/determine_concepts_layout.rb
iHiD: (This is purely aesthetic)

app/commands/track/determine_concepts_layout.rb
```diff
@@ -7,20 +7,56 @@ class DetermineConceptsLayout
     include Mandate
 
     def initialize(track)
+      @track = track
```
iHiD, line 10: I think currently you don't need to store this as an instance variable.

app/commands/track/determine_concepts_layout.rb
```diff
@@ -7,20 +7,56 @@ class DetermineConceptsLayout
...
       }
     end
 
     private
-    attr_reader :graph
+    attr_reader :graph, :track
+
+    memoize
+    def concepts
+      counter = 0 # TODO: remove this if the component doesn't actually need an 'index'
```
iHiD: Does it? :)

neenjaw: It requires a refactor of the component, maybe in a refinement PR?

neenjaw: Nevermind, I need to add the display name to the component. Will do it in next commit.

iHiD: It's not a blocker at all. I just wondered from the comment what was driving that decision.

app/commands/track/determine_concepts_layout.rb
```diff
@@ -26,13 +26,9 @@ def call
 
     memoize
     def concepts
-      counter = 0 # TODO: remove this if the component doesn't actually need an 'index'
```
iHiD: Nice commit! :)

neenjaw: :smile:

app/commands/track/determine_concepts_layout.rb
neenjaw: I think just a remnant of code from a time when putting together the component. So was writing this around it.  It has now been removed in the component, so removed here.

iHiD:
This is fantastic. Very exciting and really great work. Thank you!!! 💙

PR #116, opened 2020-09-29
Extract fromNow into utils

kntsoriano:
This will allow us to change our time library easily if we need to.


ErikSchierboom (approved):
Nice!

iHiD:
Yeah - this is good. Nice 🙂

PR #117, opened 2020-09-29
<Student.TrackList />

kntsoriano:
## Description
This is a rewrite of the `<Student.TrackList />` component implementing the several things we've agreed upon the past few days. This doesn't include tag filtering yet as I'm waiting for it to be finished on the back end.


app/controllers/test/components/student/track_list_controller.rb
```diff
@@ -0,0 +1,18 @@
+class Test::Components::Student::TrackListController < ApplicationController
+  def show
```
kntsoriano: I've started use of our serializers and search commands in our test controller.

ErikSchierboom: Great!

app/javascript/components/mentoring/Queue/Solution.jsx
```diff
@@ -18,11 +19,7 @@ export function Solution({
   return (
     <tr>
       <td>
-        <img
-          style={{ width: 100 }}
-          src={trackIconUrl}
-          alt={`icon indicating ${trackTitle}`}
-        />
+        <TrackIcon track={{ title: trackTitle, iconUrl: trackIconUrl }} />
```
ErikSchierboom, line 20: Nice!

app/javascript/components/mentoring/inbox/ConversationList.jsx
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import { Pagination } from '../Pagination'
 import { Conversation } from './conversation-list/Conversation'
 import { usePaginatedRequestQuery } from '../../../hooks/request-query'
+import { Loading } from '../../common/Loading'
 
 export function ConversationList({ request, setPage }) {
   const { status, resolvedData, latestData } = usePaginatedRequestQuery(
```
ErikSchierboom: Maybe also use the `isSuccess, isError, isFetching` fields here instead of status?

app/javascript/components/mentoring/Queue/SolutionList.jsx
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import { Solution } from './Solution'
 import { Pagination } from '../Pagination'
 import { usePaginatedRequestQuery } from '../../../hooks/request-query'
+import { Loading } from '../../common/Loading'
 
 export function SolutionList({ request, setPage }) {
   const { status, resolvedData, latestData } = usePaginatedRequestQuery(
```
ErikSchierboom: Maybe also use the `isSuccess, isError, isFetching` fields here instead of status?

ErikSchierboom (approved):
Nice! Could you also add tests for the view component and the `TrackList` component?

app/controllers/test/components/student/track_list_controller.rb
iHiD: Nice! 🙂

app/helpers/view_components/student/track_list.rb
```diff
@@ -0,0 +1,22 @@
...
+    class TrackList < ViewComponent
+      initialize_with :data, :request
+
+      def status_options
+        [
+          { value: "all", label: "All" },
+          { value: "joined", label: "Joined" },
+          { value: "unjoined", label: "Unjoined" }
+        ]
+      end
```
iHiD: Could we make this a constant (STATUS_OPTIONS) and `.freeze` it pls?

app/javascript/components/common/TrackIcon.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+export function TrackIcon({ track }) {
+  return (
+    <img
+      style={{ width: 100 }}
+      src={track.iconUrl}
+      alt={`icon for ${track.title} track`}
```
iHiD, line 8: Could we add a (CSS) class of `c-track-icon` to this pls. Maybe this but I don't know if it's correct or if `class` is reserved:
```suggestion
      alt={`icon for ${track.title} track`}
      class="c-track-icon"
```

All React Components should have a css class starting with `c-` and describing them. (See https://github.com/exercism/v3-website/pull/121)

app/javascript/components/common/TrackIcon.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+export function TrackIcon({ track }) {
+  return (
+    <img
+      style={{ width: 100 }}
```
iHiD: If this is a common component, it shouldn't have a width. I don't mind it going in temporarily, but could you add a `TODO` comment for it to be removed once the css class is implemented please.

test/system/components/mentoring/queue_test.rb
```diff
@@ -11,7 +11,7 @@ class QueueTest < ApplicationSystemTestCase
 
         row = {
           "Track icon" => lambda {
-            assert_css "img[src='https://assets.exercism.io/tracks/ruby-hex-white.png'][alt='icon indicating Ruby']"
+            assert_css "img[src='https://assets.exercism.io/tracks/ruby-hex-white.png'][alt='icon for Ruby track']"
```
iHiD, line 14:
```suggestion
            assert_css "img[src='https://assets.exercism.io/tracks/ruby-hex-white.png'][alt='icon for the Ruby track']"
```

iHiD:
Nice work 🙂 

I think this should be `Student.TracksList`. Thoughts?

ErikSchierboom:
> I think this should be `Student.TracksList`. Thoughts?

I'm fine with either option. We're currently using `<Singular>List` for the other components. I think this stems from the fact that one usually has a singular component and a list version of that. So you'd have `Track` and `TrackList`. @SleeplessByte any thoughts?

SleeplessByte:
I normally don't namespace components, because they're namespaced by their import path / directory structure.

I would personally import `Tracks` or `TrackList` or `JoinedTracks` from a student folder in order to indicate this were the student track list _if student_ was a folder / feature. If, instead, the "feature" is "Tracks", then I'd call the component `StudentTracks` or `StudentTrackList`, and have it be that.

Almost never will I namespace using nested objects.

iHiD:
(@kntsoriano Regardless of the nesting conversation - please rename to the plural `TracksList`. We'll keep singluar for where the object is singular, and plural for where it's a group of objects - similar to Rails)

app/javascript/components/mentoring/inbox/ConversationList.jsx
kntsoriano: Okay. I will add that to this PR.

app/views/test/components/mentoring/inbox/show.html.haml
```diff
@@ -7,4 +7,4 @@
   = select_tag :tracks_endpoint_state, options_for_select(%w[Success Error Loading], params[:tracks_endpoint_state])
   = submit_tag "Submit", name: nil
 %hr/
-= ViewComponents::Mentoring::Inbox.new({ endpoint: conversations_test_components_mentoring_inbox_path, query: { state: params[:conversations_endpoint_state], sort: :exercise }, retry: { retry: params[:conversations_endpoint_state] == "Loading" } }, { endpoint: tracks_test_components_mentoring_inbox_path, query: { state: params[:tracks_endpoint_state] }, retry: { retry: params[:tracks_endpoint_state] == "Loading" } })
+= ViewComponents::Mentoring::Inbox.new({ endpoint: conversations_test_components_mentoring_inbox_path, query: { state: params[:conversations_endpoint_state], sort: :exercise }, options: { retry: params[:conversations_endpoint_state] == "Loading" } }, { endpoint: tracks_test_components_mentoring_inbox_path, query: { state: params[:tracks_endpoint_state] }, options: { retry: params[:tracks_endpoint_state] == "Loading" } })
```
iHiD, line 10: I don't want to block this PR, but I think this can be improved lots. I've written up an issue: https://github.com/exercism/v3-website/issues/139

app/views/test/components/student/tracks_list/show.html.haml
```diff
@@ -0,0 +1,11 @@
...
+%h2 Tracks list
+= form_tag nil, method: :get do
+  = label_tag :state
+  = select_tag :state, options_for_select(%w[Success Error Loading], params[:state])
+  = submit_tag "Submit", name: nil
+%br
+Error / Loading states are only seen when filtering
+%hr/
+
+= ViewComponents::Student::TracksList.new(@data, { endpoint: tracks_test_components_student_tracks_list_path, query: { state: params[:state] }, options: { retry: params[:state] == "Loading", initialStale: true } })
```
iHiD, line 11: I don't want to block this PR, but I think this can be improved lots. I've written up an issue: https://github.com/exercism/v3-project-management/issues/37

iHiD:
Nice work! 🙂

test/system/components/student/tracks_list_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        visit test_components_student_tracks_list_url
+
+        row = {
+          "Icon" => lambda {
+            assert_css "img[src='https://assets.exercism.io/tracks/ruby-hex-white.png'][alt='icon for Ruby track']"
+          },
+          "Title" => "Ruby",
+          "Num concept exercises" => 10,
+          "Num practice exercises" => 10,
```
iHiD, line 25: @kntsoriano Try and keep everything in tests as different values to ensure we're not silently calling the wrong method. Both of these could be calling num_concepts right now and we wouldn't know. I've fixed this in my PR by changing the number of concept exercises to `8`.

iHiD: Also on this, the tests will be faster the less objects we create, so using `2` and `3` would be a better option than `8` and `10`.

PR #118, opened 2020-09-29
Add popup to <Mentoring.Inbox />

kntsoriano:
https://github.com/exercism/v3-project-mangement/issues/22

This is a simple implementation of Popper.

### DIFF
--- a/app/javascript/components/mentoring/inbox/conversation-list/Conversation.jsx
+++ b/app/javascript/components/mentoring/inbox/conversation-list/Conversation.jsx
@@ -1,7 +1,8 @@
-import React from 'react'
+import React, { useState } from 'react'
 import dayjs from 'dayjs'
 import RelativeTime from 'dayjs/plugin/relativeTime'
 dayjs.extend(RelativeTime)
+import { usePopper } from 'react-popper'
 
 export function Conversation({
   trackTitle,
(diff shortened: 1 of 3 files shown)


app/javascript/components/mentoring/inbox/conversation-list/Conversation.jsx
```diff
@@ -16,30 +17,54 @@ export function Conversation({
   updatedAt,
   url,
 }) {
+  const [referenceElement, setReferenceElement] = useState(null)
+  const [popperElement, setPopperElement] = useState(null)
+  const [showPopper, setShowPopper] = useState(null)
```
ErikSchierboom, line 22: Maybe we could initialize this state with a boolean `false` value? Makes it slightly more explicit that we're dealing with state containing a value that is interpreted as a boolean. Or is there something special about passing in `null`?

```suggestion
  const [showPopper, setShowPopper] = useState(false)
```

app/javascript/components/mentoring/inbox/conversation-list/Conversation.jsx
```diff
@@ -16,30 +17,54 @@ export function Conversation({
...
-    </tr>
+    <>
+      <tr
+        ref={setReferenceElement}
+        onMouseEnter={() => {
+          setShowPopper(true)
+        }}
+        onMouseLeave={() => {
+          setShowPopper(null)
+        }}
```
ErikSchierboom, line 34: As we only have a single line and `setShowPopper` returns `void`, this could also be written as:

```suggestion
        onMouseEnter={() => setShowPopper(true)}
        onMouseLeave={() => setShowPopper(false)}
```

app/javascript/components/mentoring/inbox/conversation-list/Conversation.jsx
```diff
@@ -16,30 +17,54 @@ export function Conversation({
...
+        <td>{url}</td>
+      </tr>
+      {showPopper && (
+        <div
+          ref={setPopperElement}
+          style={styles.popper}
+          {...attributes.popper}
+        >
+          Lorem ipsum
+        </div>
```
ErikSchierboom, line 66: Once we'll have multiple of these popups, we should probably extract it to a separate component. It is perfectly fine for now.

ErikSchierboom (approved):

kntsoriano:
@ErikSchierboom I will close this and reopen on another PR as I've added it to the wrong component.

PR #122, opened 2020-09-29
Add basic default setup for Tailwind

iHiD:
We're going to try using Tailwind for some CSS stuff. This is the basic setup for it. Not sure if it'll stay long term but let's see 🙂 

This also adds some very basic styling for the track component as an example. 

Naming of classes/tailwind variables are just placeholders for now - no need to comment on those.

Thoughts welcome 🙂


postcss.config.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
   plugins: [
     require('postcss-import'),
+    require('tailwindcss'),
+    //require('autoprefixer'),
```
iHiD: Everything says we should have this uncommented but I'm not sure why yet. I'm leaving it commented for now so if something doesn't work later I can remember to try uncommenting this, but I'll remove it if we don't use it once we're going.

tailwind.config.js
```diff
@@ -0,0 +1,12 @@
...
+    // removeDeprecatedGapUtilities: true,
+    // purgeLayersByDefault: true,
+  },
+  purge: [],
+  theme: {
+    extend: {},
+  },
+  variants: {},
+  plugins: [],
+}
```
iHiD, line 12: This is the default generated config.

postcss.config.js
neenjaw: From what I understand, tailwind doesn't need autoprefixer to work, but it does create css rules that may require vendor prefixes in some browsers. [This is the best I can find right now for proof of this](https://github.com/tailwindlabs/tailwindcss/issues/291)

iHiD: @neenjaw Thanks 🙂  Turns out that `postcss-preset-env` already has support for auto-prefixing so this isn't required at all.

app/javascript/css/application.css
```diff
@@ -14,4 +14,23 @@
...
+            @apply font-bold;
+        }
+    }
+}
+
+input {
+    &::placeholder {
+        color: #123456;
+    }
+}
```
iHiD: I've done this manually to test that auto-prefixes get added (they do). Any thoughts on how we could write a regression test to ensure this works (ie, we don't accidently lose the prefixing later down the line). (cc @SleeplessByte too)

iHiD: This in general is just me checking that things are working.

app/views/pages/index.html.haml
```diff
@@ -1,4 +1,15 @@
...
+.text-purple-700 Some purple text
+.hello 
+  Hello, 
+  .world World!
+
+.tw-text-green-400 
+  Hello
+  .tw-inline-block.md:tw-font-bold World!
+
+%input.tw-border-solid.tw-border-4.tw-border-green-500{placeholder: "foobar"}
```
iHiD: This is also just me POC'ing.

app/javascript/css/application.css
neenjaw: I safari still uses a lot of vendor tags, you could write a test on a flex-box enable page to look for that css vendor prefix.

- e.g. http://shouldiprefix.com/#flexbox

iHiD: I'm going to move this to a separate issue: https://github.com/exercism/v3-website/issues/123

tailwind.config.js
```diff
@@ -0,0 +1,45 @@
+module.exports = {
+  future: {
+    // removeDeprecatedGapUtilities: true,
+    // purgeLayersByDefault: true,
```
iHiD, line 4: These are added by default - I've not checked what they mean yet.

app/javascript/css/application.css
```diff
@@ -14,4 +14,10 @@
  *= require_self
  */
 
-.hello { color: green; }
+@import "tailwindcss/base";
+@import "tailwindcss/components";
+@import "tailwindcss/utilities";
+
+body {
+    @apply font-body bg-bgGray antialiased;
```
iHiD, line 22: Re antialiased: https://github.com/exercism/v3-website/issues/124

app/javascript/css/components/track.css
```diff
@@ -0,0 +1,17 @@
+@import "../styles";
```
kntsoriano, line 1: What does `&` mean here?

kntsoriano (approved):
Looks good! Does this work outside React? Just thinking we might need to enforce consistency between those two (component styles vs non-component styles) if ever it's needed.

ErikSchierboom (approved):
Looks good!

app/javascript/css/components/track.css
iHiD: `&` is the CSS3 nesting operator. Whereas in sass it's optional (and normally not added), in CSS3 it's compulsory, so every nested style has to have it.

iHiD:
> Does this work outside React?

Yeah - this isn't react at all. It's postcss (which is basically babel for css - so you write css3 and it downgrades to css2) and then tailwind as a library on it. All our CSS will go through this (rather than the asset pipeline) regardless of whether it's for JS components or normal things.

taiyab:
Big fan of Tailwind 🚀

PR #126, opened 2020-09-30
Make authenticated pivot explicit

iHiD:
This provides a less "magic" approach to providing authenticated and unauthenticated (which I've called `external` as it makes it easier to parse than the same word twice, one just with `un` in front of it).

The CONTRIBUTING.md changes should explain how it works. If its not clear from that, let's get that clear as part of this PR 🙂


app/controllers/application_controller.rb
```diff
@@ -1,21 +1,31 @@
...
   # TODO: Remove when devise is added
-  def user_logged_in?
-    !!current_user
+  def authenticate_user!
+    raise "Not logged in" unless user_logged_in?
   end
 
   def current_user
-    User.first
+    @__current_user__ ||= User.first # rubocop:disable Naming/MemoizedInstanceVariableName
```
iHiD, line 10: I want to ensure we don't accidentally use `@current_user` rather than `current_user` anywhere atm, which is why I'm doing this. All of this gets removed once I add Devise, but I'm defering that until later in October.

ErikSchierboom (approved):
LGTM

app/controllers/application_controller.rb
```diff
@@ -1,21 +1,31 @@
...
 
-  %w[index show].each do |action|
-    define_method action do
-      if user_logged_in? && respond_to?("authenticated_#{action}")
-        send("authenticated_#{action}")
-        render action: "authenticated/#{action}"
-      elsif !user_logged_in? && respond_to?("external_#{action}")
-        send("external_#{action}")
-        render action: "external/#{action}"
+  def self.allow_unauthenticated!(*actions)
```
kntsoriano, line 17: Any particular reason why we used this instead of a mixin? I know there really the same under the hood, but just asking for reading purposes :)

kntsoriano (approved):

app/controllers/application_controller.rb
iHiD: I figured as it might evolve more etc, I'd wait until we were sure it was stable then we could refactor it.

I did consider using a Concern for it, but figured I'd have to remember how they worked and that sounded like effort :wink:

Once we're happy everything is stable, I'll refactor it out 👍

PR #131, opened 2020-09-30
Loosen Ruby version

SleeplessByte:
This allows patch level differences for the Ruby version, so that systems that _require_ certain patches can apply them, without having to pin a specific patch level on their system (rbenv, uru, etc.)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '2.6.6'
+ruby '~> 2.6.0'
 
 # Service/framework dependencies
 gem 'rails', git: 'http://github.com/rails/rails.git', ref: "7101489293ece071013417aeed2da3aa1b0927c9"


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '2.6.6'
+ruby '~> 2.6.0'
```
kotp, line 3:
```suggestion
ruby '~> 2.6'
```

maybe?

kotp (approved):
I think this syntax is valid, and communicates the right thing?

Otherwise, what is here should work as well.

Either way, looks good to me.

Gemfile
SleeplessByte: I would be okay with that -- albeit very loose, but it doesn't mean the same thing.

iHiD:
So my initial reaction is that I'm not a fan of this because it means two people running bundle on the same `Gemfile` get different `Gemfile.lock` files (as the Ruby version is stored in there). And I want the Gemfile.lock to match the version of the Ruby that's on the Dockerfiles and in production.

I presume the advantage is so that you don't need to install an extra version on your OS?

SleeplessByte:
> I presume the advantage is so that you don't need to install an extra version on your OS?

I keep running into this because whatever project I contribute to chooses the "latest" at the moment this file is written, without this _actually_ giving you a guarantee that whatever Gemfile.lock decides to list will install correctly.

> ... Ruby that's on the Dockerfiles and in production.

I think this is a good reason to want to lock the ruby version, but I feel that that's better left to the Dockerfile than the Gemfile.

> (as the Ruby version is stored in there)

I don't think that's right. It will store the `bundler` version (as `BUNDLED_WITH`), and the `platform` under `PLATFORMS`, and those will _still_ be able to differ, regardless of ruby version.

**Edit**: See TypeScript PR where it has not changed `RUBY VERSION`

iHiD:
> I think this is a good reason to want to lock the ruby version, but I feel that that's better left to the Dockerfile than the Gemfile.

This issue is when the Ruby version in the Gemfile.lock is out of sync with the Ruby version installed on the Docker machine, when bundle is run with `--freeze`. This mismatch can cause `bundle exec` to fail. So someone accidently committing 2.6.3 to the Gemfile.lock, but it being having 2.6.6 in the Dockerfile, will work all the way until it tries to run in production when it then fails.

---

As the officially supported way of dev'ing exercism is `development-environment`, I'm going to leave this to be as "safe" for me in production, which is to specify the hard-version in the Gemfile. But I'll update the README to specify how to change this for people locally if they run into problems with a missing version when installing locally :)

iHiD:
README change here: https://github.com/exercism/v3-website/pull/134

Gemfile
kotp: Thanks @SleeplessByte yeah, the proposed would be 2.6 or 2.7 or further, rather than restricting to within 2.6.x?  It has been a little bit since I had to know this.

PR #133, opened 2020-09-30
Add TypeScript support

SleeplessByte:
This adds TypeScript support to webpacker; **without** the requirement to use it. 

It also has an example of converting an existing component to TypeScript enabled React. We instantly catch a problem: `IterationsSummaryTable` is returning an `Array` of `Element`s, instead of a single `React.ReactNode` element. Arrays at the top level should be wrapped inside a `Fragment`, and TypeScript has told us so.

I have _not_ tested this code (because I couldn't be bothered to go through the whole docker setup on this machine), so someone should checkout this branch and try it out, but it should be straightforward.


SleeplessByte:
(Someone please see why the tests fail; I probably forgot to update a reference somewhere)

Gemfile
```diff
@@ -40,6 +40,9 @@ platforms :mingw, :mswin, :x64_mingw, :jruby do
   gem 'tzinfo-data'
 end
 
+# To avoid polling for changes on Windows MRI, it needs this opt-in gem
+gem 'wdm', '>= 0.1.0' if Gem.win_platform?
```
iHiD, line 44: Can this move into the `group :development, :test do` pls.

SleeplessByte: Probably!

iHiD:
Thanks @SleeplessByte. One minor comment by me. Could whoever fixes the tests pls make that change too.

@ErikSchierboom @kntsoriano So there's a question here of whether you want to use `tsx` and TypeScript in general. Obviously, making this decision now will mean we don't have to change lots later, but it might be more effort for you now. I am a :+1: on doing this because it'll make things most solid, but I'm also not the one writing it so I'm leaving the two of you to make the decision 🙂 Can you have a chat (either here or 1-1 is fine too) and let us know pls.

Either way, having the support in webpacker etc seems sensible?

SleeplessByte:
If you put `strict: false` in `tsconfig.json` , you will probably not notice it, but you can still benefit -- I'm happy to be available a few hours a day if there is trouble with TypeScript, but I think Erik has the gist, and he was onboarded quickly!

kntsoriano:
Looks good! Want to try out TypeScript myself so I'm okay with this.

kntsoriano (approved):

app/javascript/components/example/IterationsSummaryTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const received = (data: { iterations: readonly Iteration[] }) =>
+      setIterations(data.iterations);
+
+    const subscription = consumer.subscriptions.create(
+      { channel: "IterationsChannel", solution_id: solutionId },
+      {
+        received,
+      }
+    );
+    //console.log(subscription.consumer)
```
ErikSchierboom:
```suggestion
```

app/javascript/components/example/IterationsSummaryTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+type Iteration = {
+  id: string;
+  testsStatus: unknown;
+};
+
+function IterationSummary({
+  id,
+  testsStatus,
+  idx,
+}: Iteration & { idx: number }) {
```
ErikSchierboom, line 15: It looks like the `idx` isn't used?

ErikSchierboom (approved):
I've worked with TypeScript quite a bit and I really, really like it. I like it for:

- Catching mistakes (wrong argument types, etc.)
- Code-completion
- Documentation

ErikSchierboom:
I get a different error when running `yarn test`:

```
SyntaxError: /usr/src/app/app/javascript/components/example/IterationsSummaryTable.tsx: Unexpected token, expected ";" (6:5)

      4 | import consumer from "../../utils/action-cable-consumer";
      5 |
    > 6 | type Iteration = {
        |      ^
      7 |   id: string;
      8 |   testsStatus: unknown;
      9 | };
```

This looks like Jest is not able to work with TypeScript. @SleeplessByte could you add support for that?

SleeplessByte:
On it

app/javascript/components/example/IterationsSummaryTable.tsx
SleeplessByte: Exactly!

iHiD:
We should probably add [this file](https://github.com/exercism/v3-website/pull/151/files#diff-ff25e0d5bd5e29e4c0a514d47835cc16) to this too.

SleeplessByte:
I haven't re-setup the dev environment on my PC yet, hence me "being on it" is taking a while. If, in the meanwhile, someone must work with this, what we need to do is make `jest` understand typescript, which comes down to very little:

First we must ensure `babel` is used for `jest`:

```shell
yarn add --dev babel-jest @babel/core @babel/preset-env
```

Then configure Babel to target your current version of Node by creating a babel.config.js file in the root of your project:

```js
// babel.config.js
module.exports = {
  presets: [
    [
      '@babel/preset-env',
      {
        targets: {
          node: 'current',
        },
      },
    ],
  ],
};
```

Jest supports TypeScript, via Babel. Next, install the `@babel/preset-typescript` via yarn:

```shell
yarn add --dev @babel/preset-typescript
```

Then add `@babel/preset-typescript` to the list of presets in the `babel.config.js`.

```diff
// babel.config.js
module.exports = {
  presets: [
    [
      '@babel/preset-env',
      {
        targets: {
          node: 'current',
        },
      },
    ],
+   '@babel/preset-typescript',
  ],
};
```

Naturally, we are using `react`, so it's likely `'@babel/preset-react'` needs to be in there (or is already in there).

kntsoriano:
I've made the JS tests pass, but am unable to make the system tests pass now. It seems that we need to add something to make `ActionCable` compatible with Typescript. Also, I think we need to make `react-bootloader` compatible with Typescript.

I'm not sure if that is correct, but I've spent the past few hours trying to fix it, but haven't found a solution yet. Hahaha.

iHiD:
@kntsoriano The system tests pass on CI. Sure it wasn't something locally cached on your setup?

iHiD:
Screenshot 2020-10-06 at 12 11 46

@SleeplessByte You might find it useful to read a tutorial introducing you to the idea of branches. [This](https://www.toolsqa.com/git/branch-in-git/) looks like a good resource 🙂

ErikSchierboom:
It also works on my machine!

ErikSchierboom (approved):

iHiD (approved):
I'll approve so @ErikSchierboom can merge, but three things to do before merging please:
- Move Gem
- Fix idx if it needs fixing?
- Add ts and tsx to prettier

Then wahooooo 🎉

SleeplessByte:
@kntsoriano in general, adding allowJs allows JavaScript to be imported into typescript files, but it will.complain about type definitions not being present.

Many can be installed from npm, but I'll show ya how you can do it yourself if you're ever stuck again.

SleeplessByte:
> Screenshot 2020-10-06 at 12 11 46
> 
> @SleeplessByte You might find it useful to read a tutorial introducing you to the idea of branches. [This](https://www.toolsqa.com/git/branch-in-git/) looks like a good resource 🙂

PR #135, opened 2020-10-01
Add popup to <Mentoring.Queue />

kntsoriano:
@ErikSchierboom This is the new PR with Popper added to the right component. I've also applied your suggestions here. Thanks :)

### DIFF
--- a/app/javascript/components/mentoring/Queue/Solution.jsx
+++ b/app/javascript/components/mentoring/Queue/Solution.jsx
@@ -1,6 +1,8 @@
 import React from 'react'
+import React, { useState } from 'react'
 import { fromNow } from '../../../utils/time'
 import { TrackIcon } from '../../common/TrackIcon'
+import { usePopper } from 'react-popper'
 
 export function Solution({
   trackTitle,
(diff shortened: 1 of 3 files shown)


app/javascript/components/mentoring/Queue/Solution.jsx
```diff
@@ -1,5 +1,7 @@
 import React from 'react'
```
ErikSchierboom: There is now a double React import which leads to a runtime error. Could you remove the first line?

```suggestion
```

ErikSchierboom (commented):
>  I've also applied your suggestions here

I don't seem them applied in this PR? Am I missing something?

kntsoriano:
@ErikSchierboom I've set `null` to `false`.

app/javascript/components/mentoring/Queue/Solution.jsx
```diff
@@ -13,29 +14,53 @@ export function Solution({
...
-    </tr>
+    <>
+      <tr
+        ref={setReferenceElement}
+        onMouseEnter={() => {
+          setShowPopper(true)
+        }}
+        onMouseLeave={() => {
+          setShowPopper(false)
+        }}
```
ErikSchierboom:
```suggestion
        onMouseEnter={() => setShowPopper(true)}
        onMouseLeave={() => setShowPopper(false)}
```

ErikSchierboom (approved):

iHiD:
Tis red.

PR #138, opened 2020-10-01
Broadcast test runs

ErikSchierboom:
- Add serialization for test run
- Broadcast test run

### DIFF
--- a/test/channels/test_run_channel_test.rb
+++ b/test/channels/test_run_channel_test.rb
@@ -1,0 +1,16 @@
+require "test_helper"
+
+class TestRunChannelTest < ActionCable::Channel::TestCase
+  test ".broadcast! broadcasts test run" do
+    test_run = create :iteration_test_run,
+                      ops_status: 403,
+                      raw_results: { message: nil }
+
+    assert_broadcast_on(
+      'test_run',
+      test_run: { id: test_run.id, status: "ops_error", message: "Some error occurred", tests: nil }
+    ) do
+      TestRunChannel.broadcast!(test_run)
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


test/commands/iteration/test_run/process_test.rb
```diff
@@ -79,6 +79,7 @@ class Iteration::TestRun::ProcessTest < ActiveSupport::TestCase
 
     IterationChannel.expects(:broadcast!).with(iteration)
     IterationsChannel.expects(:broadcast!).with(iteration.solution)
+    TestRunChannel.expects(:broadcast!).with { |test_run| test_run == iteration.test_runs.last }
```
ErikSchierboom: Is this the way in which I can verify an expected value _after_ it has run? I couldn't just use `.with(iteration.test_runs.last)` like with the other calls, as the actual test run is created in the `Iteration::TestRun::Process.` call.

test/channels/test_run_channel_test.rb
```diff
@@ -0,0 +1,16 @@
+require "test_helper"
+
+class TestRunChannelTest < ActionCable::Channel::TestCase
+  test ".broadcast! broadcasts test run" do
```
ErikSchierboom, line 4: If this is okay, I can also add tests for the `IterationChannel` and `IterationsChannel` classes.

iHiD: This looks good to me.

test/commands/iteration/test_run/process_test.rb
iHiD: This is absolutely fine. You could also do 
```
TestRunChannel.expects(:broadcast!).with(kind_of(Iteration::TestRun))
```

As you're not validating that the test run is particularly the "correct" test run, the `kind_of` check is probably just as safe. I don't have a preference over either. You could firm up the test by asserting that  `iteration.test_runs.size == 1` too

test/channels/test_run_channel_test.rb
ErikSchierboom: I've added tests for the other two untested channels too.

iHiD (approved):
Great. Thanks 🙂 

Tis red though 😢

ErikSchierboom:
Should become green again :)

PR #140, opened 2020-10-01
Integrated styled tracks

iHiD:
This PR places the the `TracksList` component into the actual view and adds the correct styling to it.

I recommend reading commit-by-commit.

Screenshot 2020-10-02 at 01 01 01

---

@kntsoriano Can you push a couple of commits to this please:
- [x] Fix the React warning.
- [x] In between the `h3` and `.counts`, conditionally add `<div class="joined">Joined</div>` if the user is joined the track.

I'll then be happy to get this merged, and I can continue to build on this.


app/javascript/components/common/TrackIcon.jsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react'
 
-export function TrackIcon({ track, width = 100 }) {
+export function TrackIcon({ track }) {
   return (
     <img
-      style={{ width: width }}
+      className="c-track-icon"
```
iHiD, line 6: We can add a default width/height to this component if we want to in the CSS. I've not for now though.

app/javascript/components/student/TracksList/List/Track.jsx
```diff
@@ -4,29 +4,27 @@ import { ProgressBar } from './Track/ProgressBar'
...
-      <td>
-        {track.isJoined && (
-          <ProgressBar
-            numConceptExercises={track.numConceptExercises}
-            numPracticeExercises={track.numPracticeExercises}
-            numCompletedConceptExercises={track.numCompletedConceptExercises}
-            numCompletedPracticeExercises={track.numCompletedPracticeExercises}
-          />
-        )}
-      </td>
```
iHiD: I'll add this back in in the next PR.

app/javascript/components/student/TracksList/List/Track.jsx
```diff
@@ -4,29 +4,27 @@ import { ProgressBar } from './Track/ProgressBar'
...
+            exercises
+          </li>
+        </ul>
+        <ul className="tags">
+          {track.tags.slice(0, 3).map((tag) => {
+            return <li>{tag.split(':')[1]}</li>
+          })}
+        </ul>
+      </div>
+      <i>›</i>
```
iHiD: This will be a proper icon but that involves working out how to add the icon library via NPM so it's a job for a different PR.

app/javascript/css/components/track.css
```diff
@@ -1,17 +1,36 @@
...
+        & .title {
+            @apply mb-8;
+            @apply text-h3-mono mb-8 text-black;
+        }
 
-    & .counts {
-        @apply flex font-medium text-16 gap-12 text-textLight;
+        & .counts {
+            @apply flex gap-12 mb-12;
+            @apply font-medium text-16;
```
iHiD, line 21: I'm following the principle of having one line for layout and a separate line for styling.

db/seeds.rb
```diff
@@ -16,6 +16,55 @@ def rugged_repo
...
+    "Domain:Embedded systems",
+    "Domain:Games",
+    "Domain:GUIs",
+    "Domain:Logic",
+    "Domain:Maths",
+    "Domain:Mobile",
+    "Domain:Scripting",
+    "Domain:Web development",
+  ]
+]
```
iHiD, line 66: I've not 100% decided how to model this, so I'm leaving this as stub data for now. We need to think about it from the OSS POV, which I'll want @ErikSchierboom's expertise on. This is ok for now though I think.

tailwind.config.js
```diff
@@ -16,7 +21,9 @@ module.exports = {
       transparent: 'transparent',
       current: 'currentColor',
       bgGray: '#FBFCFE',
-      textLight: '#343741',
+      textBase: '#343741',
+      textLight: '#6D6986',
+      borderLight: '#CBC9D9',
```
iHiD, line 26: @taiyab - This is where we'll put all the colours and their names later btw.

test/application_system_test_case.rb
```diff
@@ -18,4 +18,52 @@ def sign_in!(user = nil)
...
+    # Create a useful and readable error message in case of no match
+    error_msg = lambda {
+      pretty_html = Nokogiri::XML(html).to_xhtml(indent: 2, save_with: nokogiri_save_options)
+      pretty_context = Nokogiri::XML(context).to_xhtml(indent: 2, save_with: nokogiri_save_options)
+      "Could not find:\n#{pretty_html}\n...in...#{pretty_context}"
+    }
+
+    # Run the assertion
+    assert_includes formatted_context, formatted_html, error_msg
+  end
```
iHiD, line 68: This is a complex method, but it works really well! Happy to move these into mixins etc too, but let's do that as a separate job in a separate PR 🙂

test/system/components/student/tracks_list_test.rb
```diff
@@ -3,54 +3,56 @@
...
+                <li>2/4 exercises</li>
+              </ul>
+              <ul class="tags">
+                <li>Bar</li>
+                <li>Xyz</li>
+              </ul>
+            </div>
+            <i>›</i>
+          </a>
+        ', within: ".student-track-list"
```
iHiD: I started off by asserting each bit and descended into madness. I rethought it based on `assert_table_row` pattern, and decided having one assertion for everything made much more sense. 

So having one "happy-path" assertion of the rendered HTML, then using other methods to check bits that change. We'll probably want a second full-check for the joined track too.

test/system/components/student/tracks_list_test.rb
```diff
@@ -61,12 +63,13 @@ def setup
...
 
-        row = { "Title" => "Go" }
-        assert_table_row first("table"), row
-        assert_selector(".student-track-list tbody tr", count: 1)
+        assert_selector(".student-track-list .c-track", count: 1)
+        assert_text "Go", within: ".c-track"
       end
 
       test "shows empty state" do
+        create :user
```
iHiD: These might be better as `sign_in!(user)`.

app/javascript/css/components/track.css
```diff
@@ -1,17 +1,36 @@
 @import "../styles";
 
 .c-track {
-    @apply border-black p-24 bg-white shadow-1 rounded-8 flex;
+    @apply flex items-start;
```
kntsoriano, line 4: Clean ✨

test/application_system_test_case.rb
kntsoriano: Is this different from `assert_dom_equal`? I'm unsure though if that method works on system tests. Maybe we could import the module?

test/system/components/student/tracks_list_test.rb
```diff
@@ -3,54 +3,56 @@
...
-          "New?" => "true",
-          "Tags" => "OOP, Web Dev",
-          "Joined?" => "true",
-          "Num completed concept exercises" => 5,
-          "Num completed practice exercises" => 5,
-          "Progress" => "Completed concept exercises: 25%, Uncompleted concept exercises: 25%, "\
-                        "Completed practice exercises: 25%, Uncompleted practice exercises: 25%"
-        }
-        assert_table_row first("table"), row
+        assert_html '
```
kntsoriano, line 25: I don't know how to feel about this. I feel like it may be brittle, but at the same time I like it because it's accurate. Every time we change the components, we might really want to change the system test. I want to see where this approach takes us.

app/javascript/components/student/TracksList/List/Track.jsx
ErikSchierboom: Aha, so _this_ is the weird character being displayed :)

test/system/components/student/tracks_list_test.rb
ErikSchierboom: I'm also not sure. Is the goal to verify the exact HTML, or that the right information is displayed? I have no idea how system tests are usually used in Ruby though. @SleeplessByte?

db/seeds.rb
ErikSchierboom: Yeah let's postpone this.

app/javascript/components/student/TracksList/List/Track.jsx
```diff
@@ -4,29 +4,28 @@ import { ProgressBar } from './Track/ProgressBar'
...
+            {track.numCompletedConceptExercises}/{track.numConceptExercises}{' '}
+            concepts
+          </li>
+          <li>
+            {track.numCompletedPracticeExercises}/{track.numPracticeExercises}{' '}
+            exercises
+          </li>
+        </ul>
+        <ul className="tags">
+          {track.tags.slice(0, 3).map((tag) => {
```
ErikSchierboom: We probably want to have some sort of ordering of the tags at some point, with the most important tags being displayed first and such.

ErikSchierboom (approved):

test/application_system_test_case.rb
iHiD: I think the `assert_dom_equal` expects the two things to be identical, whereas this expects one to include the other. So you say "this block of HTML should appear somewhere" without having to worry about the wider structure of the page around it.

test/system/components/student/tracks_list_test.rb
iHiD: > Every time we change the components, we might really want to change the system test

Yeah, so this was the conclusion I came to.

>  Is the goal to verify the exact HTM

And yes, I think it is.

---

I think what this gives us is a very firm test that our components are rendering correctly, which protects us from inadvertently changing something, or React etc changing which changes something unexpectedly. Having the firmness of this test here then means I would happily just assert `.c-track` exists anywhere else in the tests, knowing that we have this guarantee that the track component renders properly.

That was my thinking anyway. I spent a lot of time wondering whether this was good or stupid, but settled on good 🙂

app/javascript/components/student/TracksList/List/Track.jsx
iHiD: Yes, good shout. This will actually display the tags in the order they're written to the db, which is probably going to be the order written in the tracks' `config.json` files, so maybe making the order in there meaningful is the best way?

test/system/components/student/tracks_list_test.rb
SleeplessByte: I'm not a fan of this either, but YMMV. 

It's fine tho; just go with it.

ErikSchierboom: So three non-fans, but none of us feel that strongly about it :)

app/javascript/components/student/TracksList/List/Track.jsx
ErikSchierboom: Maybe, I don't know. I feel like it might be easy for maintainers to re-order things there. We could perhaps make this explicit in the track's config.json with a separate field?

iHiD: I've created an issue where we can work on this: https://github.com/exercism/v3-project-management/issues/40

PR #145, opened 2020-10-02
Add default request values to view components

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/37.

Our test files are verbose as we need to pass in extra parameters in order to easily stub errors, etc. When this PR is merged, a view component can simply be rendered by `ViewComponents::Mentoring::Queue.new`. This view component already has the corrects defaults inside it.

### DIFF
--- a/app/helpers/view_components/mentoring/inbox.rb
+++ b/app/helpers/view_components/mentoring/inbox.rb
@@ -28,10 +28,12 @@ module ViewComponents
 
       private
       def default_conversations_request
+        # TODO: Change this to the actual endpoint, not the test endpoint
         { endpoint: Exercism::Routes.conversations_test_components_mentoring_inbox_path }
       end
 
       def default_tracks_request
+        # TODO: Change this to the actual endpoint, not the test endpoint
         { endpoint: Exercism::Routes.tracks_test_components_mentoring_inbox_path }
       end
     end
--- a/app/helpers/view_components/mentoring/inbox.rb
+++ b/app/helpers/view_components/mentoring/inbox.rb
@@ -1,8 +1,6 @@
 module ViewComponents
   module Mentoring
     class Inbox < ViewComponent
-      attr_reader :conversations_request, :tracks_request
-
       def initialize(conversations_request = default_conversations_request, tracks_request = default_tracks_request)
         @conversations_request = conversations_request
         @tracks_request = tracks_request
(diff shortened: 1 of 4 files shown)


app/helpers/view_components/mentoring/inbox.rb
```diff
@@ -20,6 +25,15 @@ def to_s
         { value: 'student', label: 'Sort by Student' }
       ].freeze
       private_constant :SORT_OPTIONS
+
+      private
+      def default_conversations_request
```
kntsoriano, line 30: For these default values, we need to make sure to set them to the correct endpoint. For now, we don't have those so I've routed them to the test endpoints.

ErikSchierboom: Would it make sense to add a TODO comment? As to not forget about it?

app/helpers/view_components/mentoring/queue.rb
```diff
@@ -1,7 +1,10 @@
 module ViewComponents
   module Mentoring
     class Queue < ViewComponent
-      initialize_with :request
+      attr_reader :request
+      def initialize(request = default_request)
```
ErikSchierboom, line 4:
```suggestion

      def initialize(request = default_request)
```

ErikSchierboom (approved):
I like it!

app/helpers/view_components/mentoring/inbox.rb
```diff
@@ -1,7 +1,12 @@
 module ViewComponents
   module Mentoring
     class Inbox < ViewComponent
-      initialize_with :conversations_request, :tracks_request
+      attr_reader :conversations_request, :tracks_request
```
iHiD: I think should be a private `attr_reader` (move it below the `private` LOC)

app/helpers/view_components/mentoring/inbox.rb
iHiD: Yes please. For both.

app/helpers/view_components/mentoring/queue.rb
```diff
@@ -1,7 +1,10 @@
 module ViewComponents
   module Mentoring
     class Queue < ViewComponent
-      initialize_with :request
+      attr_reader :request
```
iHiD: I think should be a private `attr_reader` (move it below the `private` LOC)

app/helpers/view_components/student/tracks_list.rb
```diff
@@ -1,7 +1,12 @@
 module ViewComponents
   module Student
     class TracksList < ViewComponent
-      initialize_with :data, :request
+      attr_reader :data, :request
```
iHiD: I think should be a private `attr_reader` (move it below the `private` LOC)

kntsoriano:
Thanks for the comments. I've applied them here. I'm merging this now.

PR #148, opened 2020-10-02
Fix <Student.TracksList /> caching problem

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/39.

This one was tricky. Since we pass the `initialData` parameter in, each change to the query will still have the `initialData` parameter set to what we first passed in. I decided to solve this by setting the `initialData` to undefined whenever we change the query. This might not be the best solution, but I hope to find a better pattern once I work on more components.

### DIFF
--- a/app/javascript/components/student/TracksList.jsx
+++ b/app/javascript/components/student/TracksList.jsx
@@ -11,11 +11,13 @@ function reducer(state, action) {
       return {
         ...state,
         query: { ...state.query, criteria: action.payload.criteria },
+        options: { ...state.options, initialData: undefined },
       }
     case 'status.changed':
       return {
         ...state,
         query: { ...state.query, status: action.payload.status },
+        options: { ...state.options, initialData: undefined },
       }
   }
 }


ErikSchierboom (approved):
This works for me. @SleeplessByte is this a good strategy?

SleeplessByte:
`initialData` set to undefined _should_ solve this. You might want to try instead `initialStale: true`, or changing the _query key_ when the query changes (and `initialData` should only be passed when there is NO query).

iHiD:
@kntsoriano: @SleeplessByte will hopefully have change to expand on this before you're back on it this evening (our time), but he mentioned to me 1-1 that "each query value should yield a new react-query key", which is maybe why things are going wrong. I don't fully understand what that means, but hopefully it's a pointer in the right direction for you 🙂

SleeplessByte:
I think Tanner has an example on this in the docs somewhere, but I can't recall or quickly find it.

Right now the query key is 'trackslist' (or something like that), which means that the results will be tied to this key. So far, it's fine. Initial data is also tied to this key, which is fine.

The call itself actually is scoping the results on some input variable (the query), and it's results will replace the cache for the 'trackslist' query key.

- nothing entered, initial data is full list, everything is fine
- entered "foo", initial data is full list (meh, but ok)
- entered "bar"… depending on the state, initial data is either the full list, or an intermediary value which is the result of foo. This is not good.

I'm reality, each request - response pair, should have it's own query cache key, because each result is disjunct from the other results. Having initial data doesn't really help us much, unless the query is empty!

iHiD:
I'm totally ignorant to how React is working but...

It seems to me that the fundamental problem here is that when I type, the page is changing before the results are returned from the server. I would expect the flow to be:

- All results show
- I type "A"
- Loading icon appears
- Response goes to server
- Results HTML hasn't changed yet
- Data come back from server
- New results HTML is rendered
- Loading icon disappears.

Then...

- I type "s" (so now "As")
- Loading icon appears
- Response goes to server
- Results HTML hasn't changed yet
- Data come back from server
- New results HTML is rendered
- Loading icon disappears.

Then...

- I press backspace (so now "A")
- Loading icon appears
- Cache is hit, and data is retrieved
- New results HTML is rendered
- Loading icon disappears.

---

We currently have eventual consistency but with this weird intermediate state where it shows seemingly either all the data, or just some of the data (a temporary cache hit?)

---

If this is what y'all are already saying then ignore me.

SleeplessByte:
What I was saying has nothing to do with react itself, but just the issue of binding "all results" to _all_ queries as initial data, and have _any_ query update the cache (initial data), meaning the next initial data is always out of sync, if you start anew.

What you're saying is how it should work, and will work, when each query has its own cache key.

kntsoriano:
I think I'm okay with this solution for now. I've spent some time exploring the other solutions, and it seems to me that this feels the cleanest right now.

I'm unsure if we can just pass `initialData` if query is empty, as there are cases where we would want to pass a query by default and at the same time pass the initial data for that specific query. The thing that I like about this solution is that it tracks the changes of the query, and removes the initial data when we see a change in the query.

iHiD (approved):
@kntsoriano If you're ok with it, then go with it. If we run into trouble we can dig in a little more to what @SleeplessByte has been explaining 🙂

PR #150, opened 2020-10-02
Further style tracks

iHiD:
The first commit does this:

Screenshot 2020-10-02 at 15 56 15

---

The second and third commits add this:
Screenshot 2020-10-02 at 16 42 11

### DIFF
--- a/app/javascript/components/student/TracksList.jsx
+++ b/app/javascript/components/student/TracksList.jsx
@@ -34,7 +34,7 @@ export function TracksList({ statusOptions, ...props }) {
           <Search dispatch={dispatch} />
         </div>
       </div>
-      <div class="md-container">
+      <div className="md-container">
         <StatusFilter dispatch={dispatch} options={statusOptions} />
 
         {isLoading && <Loading />}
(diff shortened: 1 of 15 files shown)


app/javascript/components/student/TracksList.jsx
```diff
@@ -28,12 +28,19 @@ export function TracksList({ statusOptions, ...props }) {
...
 
   return (
-    <div className="student-track-list">
-      <Search dispatch={dispatch} />
-      <StatusFilter dispatch={dispatch} options={statusOptions} />
-      {isLoading && <Loading />}
-      {isError && <p>Something went wrong</p>}
-      {isSuccess && <List data={data} />}
+    <div className="c-tracks-list">
+      <div className="search-bar">
```
kntsoriano, line 32: Do you think we should put this div within the `<Search />` component?

iHiD: This div will also have the filters etc within it, so it can't. It's potentially badly named though - as it covers search and filter, not just search. I might just inline the styles into it rather than have a named class for it. I'm trying to work out when to do which of those things still.

iHiD:
Merging this so the other PRs can be rebased onto it. Please still leave any comments if you have anything 🙂

app/javascript/components/student/TracksList.jsx
kntsoriano: Cool!

PR #152, opened 2020-10-03
Refactor Concepts and add controller

iHiD:
- Commits 1+2 are really a naming refactor as the shape of concepts has become clearer to me since I initially wrote this code.
- Commit 3 adds the concepts controller so that @neenjaw can build on it.\
- Commit 4 is a copy and paste of commit 3.

My aim for this week is to get the full flow through concepts and exercises done, so this is my laying scaffolding basically.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -21,10 +21,10 @@ class ApplicationController < ActionController::Base
       define_method action do
         if user_logged_in?
           send("authenticated_#{action}")
-          render action: "#{action}/authenticated"
+          render action: "#{action}/authenticated" unless performed?
         elsif !user_logged_in?
           send("external_#{action}")
-          render action: "#{action}/external"
+          render action: "#{action}/external" unless performed?
         end
       end
     end
--- a/app/controllers/tracks/concepts_controller.rb
+++ b/app/controllers/tracks/concepts_controller.rb
@@ -1,0 +1,47 @@
+class Tracks::ConceptsController < ApplicationController
+  before_action :use_track
+
+  allow_unauthenticated! :index, :show
+
+  def authenticated_index
+    use_concepts
+
+    if current_user.joined_track?(@track)
+      render action: "index/joined"
+    else
+      render action: "index/unjoined"
+    end
+  end
+
+  def external_index
+    use_concepts
+  end
+
+  def authenticated_show
+    use_concept
+
+    if current_user.joined_track?(@track)
+      render action: "show/joined"
+    else
+      render action: "show/unjoined"
+    end
+  end
+
+  def external_show
+    use_concept
+  end
+
+  private
+  def use_track
+    @track = Track.find(params[:track_id])
+    @user_track = current_user&.user_track_for(@track)
+  end
+
+  def use_concepts
+    @concepts = @track.concepts
+  end
+
+  def use_concept
+    @concept = @track.concepts.find(params[:id])
+  end
+end
--- a/app/models/concept_exercise.rb
+++ b/app/models/concept_exercise.rb
@@ -1,5 +1,5 @@
 class ConceptExercise < Exercise
-  # TODO: Add these
-  #  has_many :exercise_concepts, class_name: "Exercise::Prerequisite", dependent: :destroy
-  #  has_many :concepts, through: :exercise_concepts, source: :track_concept
+  # TODO: Add taught concepts here
+  # has_many :exercise_concepts, class_name: "Exercise::Prerequisite", foreign_key: :exercise_id, dependent: :destroy
+  # has_many :concepts, through: :exercise_concepts, source: :track_concept
 end
--- a/app/models/concept_exercise.rb
+++ b/app/models/concept_exercise.rb
@@ -1,5 +1,11 @@
 class ConceptExercise < Exercise
-  # TODO: Add taught concepts here
-  # has_many :exercise_concepts, class_name: "Exercise::Prerequisite", foreign_key: :exercise_id, dependent: :destroy
-  # has_many :concepts, through: :exercise_concepts, source: :track_concept
+  has_many :exercise_taught_concepts,
+    class_name: "Exercise::TaughtConcept",
+    foreign_key: :exercise_id,
+    inverse_of: :exercise,
+    dependent: :destroy
+
+  has_many :taught_concepts,
+    through: :exercise_taught_concepts,
+    source: :concept
 end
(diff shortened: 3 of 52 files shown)


app/models/concept_exercise.rb
```diff
@@ -1,5 +1,11 @@
 class ConceptExercise < Exercise
-  # TODO: Add these
-  #  has_many :exercise_concepts, class_name: "Exercise::Prerequisite", dependent: :destroy
-  #  has_many :concepts, through: :exercise_concepts, source: :track_concept
+  has_many :exercise_taught_concepts,
+    class_name: "Exercise::TaughtConcept",
+    foreign_key: :exercise_id,
+    inverse_of: :concept,
+    dependent: :destroy
```
iHiD: These might end up working fine just as straight-up join tables, but I prefer to materialise them now and then refactor them out later, rather than the other way around.

app/controllers/application_controller.rb
```diff
@@ -21,10 +21,10 @@ def self.allow_unauthenticated!(*actions)
       define_method action do
         if user_logged_in?
           send("authenticated_#{action}")
-          render action: "#{action}/authenticated"
+          render action: "#{action}/authenticated" unless performed?
```
iHiD, line 24: This allows us to override the render call in the individual controllers and only do **this** render if we've not rendered elsewhere.

app/controllers/tracks/concepts_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  private
+  def use_track
+    @track = Track.find(params[:track_id])
+    @user_track = current_user&.user_track_for(@track)
+  end
+
+  def use_concepts
+    @concepts = @track.concepts
```
iHiD, line 41: This will change to generate the Concepts Graph data (cc @neenjaw - this is your entry point for real rendering)

app/models/exercise/taught_concept.rb
```diff
@@ -0,0 +1,6 @@
+class Exercise::TaughtConcept < ApplicationRecord
+  belongs_to :exercise
+  belongs_to :concept, # rubocop:disable Rails/InverseOf
```
iHiD, line 3: We don't need this inverse_of as we don't have an association on the other end at the moment.

app/controllers/tracks/concepts_controller.rb
neenjaw: for rendering, how is the student's progress communicated from the back end to the front end?

app/controllers/application_controller.rb
kntsoriano: I'm a bit confused. Won't overriding this function in the controller not run this code?

Something like:
```
class Tracks::ConceptsController < ApplicationController
  before_action :use_track

  allow_unauthenticated! :index

  def external_index
    // This code runs, and not the method generated by the macro
  end
end
```

app/controllers/tracks/concepts_controller.rb
```diff
@@ -0,0 +1,47 @@
+class Tracks::ConceptsController < ApplicationController
+  before_action :use_track
+
+  allow_unauthenticated! :index, :show
+
+  def authenticated_index
+    use_concepts
+
+    if current_user.joined_track?(@track)
```
kntsoriano, line 9: Okay with this for now. But I'm thinking to redirect to another action when it's unjoined? Maybe something like `tracks/concepts/preview`. Just feel like it's more concrete that way.

app/controllers/tracks/concepts_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def external_show
+    use_concept
+  end
+
+  private
+  def use_track
+    @track = Track.find(params[:track_id])
+    @user_track = current_user&.user_track_for(@track)
```
kntsoriano, line 37: Any opinion in finding the right place for `#user_track_for`? I think if we continue doing methods like this, we'll end up with a very large `User` class. Maybe we could talk about this in a separate issue.

test/models/exercise/taught_concept_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
```
kntsoriano, line 1: Can we delete this?

app/controllers/application_controller.rb
iHiD: The method generated by the macro is `index`.

So you get:
```ruby
def index
  if user_logged_in?
    send("authenticated_index")
    render action: "index/authenticated" unless performed?
  else
    send("external_index")
    render action: "index/external" unless performed?
  end
end```

So this generated method calls `authenticated_index` (setting instance variables etc) then renders `index/authenticated`, unless the `authenticated_index` method itself calls render or redirect.

File: app/controllers/tracks/concepts_controller.rb (hunk 3 above)
Comment (iHiD) [reply app/controllers/tracks/concepts_controller.rb:41]: We'll expand out the node JSON to include that information. So it'll go either into the graph-data-generator thing you've written, or in a serializer/decorator around it.

File: app/controllers/tracks/concepts_controller.rb (hunk 5 above)
Comment (iHiD) [reply app/controllers/tracks/concepts_controller.rb:9]: Yeah - I agree with your instincts.

However, I don't want the URLs to be different, which is why this whole (slightly painful) setup is in place, otherwise I'd just use normal methods. I want people to have consistent URLs whether they're Google, joined, not joined, etc.

Same for people's solutions etc. If it all becomes too painful, I might change my mind, but I want to try it out for a bit first.

File: app/controllers/tracks/concepts_controller.rb (hunk 6 above)
Comment (iHiD) [reply app/controllers/tracks/concepts_controller.rb:37]: https://github.com/exercism/v3-website/issues/155 🙂

Comment (iHiD) [reply app/controllers/tracks/concepts_controller.rb:37]: I won't block merging on this for now, but will change it based on what we decide there.

File: test/models/exercise/taught_concept_test.rb (hunk 7 above)
Comment (iHiD) [reply test/models/exercise/taught_concept_test.rb:1]: I've added a test instead 🙂

**Comment by iHiD** [comment]
Also, merging this so we can move forward.

File: app/controllers/application_controller.rb (hunk 2 above)
Comment (kntsoriano) [reply app/controllers/application_controller.rb:24]: Ah!

File: app/models/user_track.rb [hunk 8]
```diff
@@ -54,7 +54,7 @@ def call
         FROM
           (
             SELECT prereqs.exercise_id, COUNT(*) AS num_concepts
-            FROM user_track_concepts utc
+            FROM user_track_learnt_concepts utc
```
ErikSchierboom, line 57: I like the new name!

PR #154, opened 2020-10-05
Tag filtering in <Student.TracksList />

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/49.
Closes https://github.com/exercism/v3-project-management/issues/48.

## Preview
Oct-05-2020 12-50-28


app/helpers/view_components/student/tracks_list.rb
```diff
@@ -20,6 +21,18 @@ def to_s
       ].freeze
       private_constant :STATUS_OPTIONS
 
+      TAG_OPTIONS = [
```
kntsoriano, line 24: Can you help me with this?

1. Is the structure okay?
2. Is the naming okay?

iHiD: Yeah - this is totally fine for now. We'll probably change it once we've thought it through further, but this works for now 🙂

app/controllers/test/components/student/tracks_list_controller.rb
```diff
@@ -6,9 +6,11 @@ def show
   def tracks
     return head :internal_server_error if params[:state] == "Loading" || params[:state] == "Error"
 
+    tags = params[:tags].to_s.split(",")
```
iHiD: Is this a pattern we're planning on using in general?

If so, I'd like to standardise it so we have a helper in JS land that serializes to this, and a helper in Rails land that deserializes. It feels a little too arbitary to me to have this LOC in multiple places without it being  formalised as a pattern.

That's not blocking this being merged, but if you agree, can you open an issue to do that when you merge please.

We should also consider what character to use - there's probably a better choice than a `,`, which might actually be part of the string in some places. Maybe a unicode character that we have confidence wouldn't be used in a string?

iHiD (approved):
This seems great to me. Let Erik have a review as well pls before merging.

app/controllers/test/components/student/tracks_list_controller.rb
SleeplessByte: Instead of using a `,`, use the HTML standard and push `tags` as an array: 

```
tags[]=foo&tags[]=bar

params[:tags]
# => ['foo', 'bar']
```

iHiD: @SleeplessByte Do you know of a widly used React (or JS or whatever) helper for this, which we should use?

SleeplessByte: For autocomplete? Something like this: https://github.com/alphagov/accessible-autocomplete, or maybe [this fork](https://bitzesty.com/2020/09/25/making-gov-uk-accessible-autocomplete-a-suitable-replacement-for-select-enhancement/?utm_content=buffer992c3&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer). Almost anything else, such as `selectize` is **not** accessible.

iHiD: Sorry, no, I just meant the serialization to the HTML array standard thing within React, before the data is passed to the server.

SleeplessByte: Ahhh. Got it. 

- If you want to push it to a URL: `URLSearchParams` (modern browsers, or [polyfill](https://github.com/ungap/url-search-params))
- If you want to push it as form data: `FormData` (built-in).

For URLs:  multiple values will have the same key, not an array key, so it will become `?tag=foo&tag=bar`, which is valid in browsers, dunno what Rails does, but this is correct behaviour.

But usually I just role my own. It's nothing more than `key[]={encodeUriComponent(value)}`

iHiD: @kntsoriano Shall we go with this instead to be more HTML compliant?

SleeplessByte: FWIW, comma separated works too, but yeah, then you have to deal with a comma never being a character inside a tag.

kntsoriano: Thanks! Applied the changes now.

kntsoriano:
@iHiD I've updated this PR to include the solution to https://github.com/exercism/v3-project-management/issues/48. Can you take a look at it again if you have a chance? Thanks.

iHiD:
Sweet. I'm happy with this. I'll let @ErikSchierboom review and merge if he's happy.

app/javascript/components/student/tracks-list/TagOption.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+export function TagOption({ value, label, checked, onChange }) {
+  const id = `tag-option-${value}`
```
ErikSchierboom, line 4: Is there any chance we'll have another component render tag options? If so, there might be ID duplicates.

app/javascript/components/student/tracks-list/Header.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+export function Header({ data, query = {} }) {
+  const isFiltering = query.criteria || (query.tags && query.tags.length !== 0)
```
ErikSchierboom: Just a tiny suggestion:

```suggestion
  const isFiltering = query.criteria || (query.tags && query.tags.length > 0)
```

I feel that `> 0` is slightly easier to read than `!== 0`.

app/javascript/components/student/tracks-list/TagOptionList.jsx
```diff
@@ -0,0 +1,36 @@
...
+      {options.map((option) => {
+        return (
+          <div key={option.category}>
+            <p>{option.category}</p>
+            {option.options.map((option) => {
+              return (
+                <TagOption
+                  key={option.value}
+                  onChange={handleChange}
+                  checked={value.indexOf(option.value) >= 0}
```
ErikSchierboom: What browsers are we supporting @iHiD? Are we supporting Internet Explorer?

If not, the [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) function could possibly be used here.

app/javascript/components/student/tracks-list/TagOptionList.jsx
```diff
@@ -0,0 +1,36 @@
+import React, { useState } from 'react'
+import { TagOption } from './TagOption'
+
+export function TagOptionList({ options, onSubmit, value, setValue }) {
```
ErikSchierboom: I didn't immediately understand what `value` represented. We do use it in several other places, so I can understand why we're using it. Just wanted to put out an alternative:

```suggestion
export function TagOptionList({ options, onSubmit, selectedTags, setSelectedTags }) {
```

app/javascript/components/student/tracks-list/TagsFilter.jsx
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from 'react'
+import { TagOptionList } from './TagOptionList'
+
+export function TagsFilter({ options, dispatch }) {
+  const [hidden, setHidden] = useState(true)
+  const [value, setValue] = useState([])
```
ErikSchierboom: I didn't immediately understand what `value` represented. We do use it in several other places, so I can understand why we're using it. Just wanted to put out an alternative:

```suggestion
  const [selectedTags, setSelectedTags] = useState([])
```

or

```suggestion
  const [filterTags, setFilterTags] = useState([])
```

app/javascript/components/student/tracks-list/TagsFilter.jsx
```diff
@@ -0,0 +1,50 @@
...
+    setValue([])
+    dispatch({ type: 'tags.changed', payload: { tags: [] } })
+  }
+
+  return (
+    <div>
+      <button
+        onClick={() => {
+          setHidden(!hidden)
+        }}
```
ErikSchierboom:
```suggestion
        onClick={() => setHidden(!hidden)}
```

app/javascript/components/student/tracks-list/TagsFilter.jsx
```diff
@@ -0,0 +1,50 @@
...
+          <TagOptionList
+            value={value}
+            options={options}
+            setValue={setValue}
+            onSubmit={handleSubmit}
+          />
+          <button
+            onClick={() => {
+              setHidden(true)
+            }}
```
ErikSchierboom:
```suggestion
            onClick={() => setHidden(true)}
```

ErikSchierboom (approved):
Nice!

app/javascript/utils/url-params.js
```diff
@@ -5,7 +5,19 @@ export class UrlParams {
...
+        const value = this.object[key]
+
+        if (Array.isArray(value)) {
+          return value
+            .map((item) => {
+              return key + '[]=' + item
+            })
+            .join('&')
+        } else {
+          return key + '=' + value
```
SleeplessByte:
```suggestion
          return key + '=' + encodeURIComponent(value)
```

app/javascript/utils/url-params.js
```diff
@@ -5,7 +5,19 @@ export class UrlParams {
...
   toString() {
     return Object.keys(this.object)
-      .map((key) => key + '=' + this.object[key])
+      .map((key) => {
+        const value = this.object[key]
+
+        if (Array.isArray(value)) {
+          return value
+            .map((item) => {
+              return key + '[]=' + item
```
SleeplessByte:
```suggestion
              return key + '[]=' + encodeURIComponent(item)
```

test/javascript/utils/url-params.test.js
```diff
@@ -7,3 +7,9 @@ test('toString() returns empty string when object is null', () => {
 test('toString() returns empty string when object is undefined', () => {
   expect(new UrlParams(undefined).toString()).toEqual('')
 })
+
+test('toString() converts arrays correctly', () => {
+  const object = { tags: ['oop', 'static'] }
```
SleeplessByte:
```suggestion
  const object = { tags: ['oop', 'static'], foo: 'bar' }
```

test/javascript/utils/url-params.test.js
```diff
@@ -7,3 +7,9 @@ test('toString() returns empty string when object is null', () => {
...
+
+  expect(new UrlParams(object).toString()).toEqual('tags[]=oop&tags[]=static')
```
SleeplessByte:
```suggestion
  expect(new UrlParams(object).toString()).toEqual('tags[]=oop&tags[]=static&foo=bar')
```

app/javascript/utils/url-params.js
```diff
@@ -5,7 +5,19 @@ export class UrlParams {
 
   toString() {
     return Object.keys(this.object)
-      .map((key) => key + '=' + this.object[key])
+      .map((key) => {
+        const value = this.object[key]
+
```
SleeplessByte, line 10: You might want to invert this:

```typescript
 return Object.keys(this.object)
   .map((key) => {
     const value = this.object[key]

     // TODO: should check the spec if these should be _left out_ or should be added
     //             like this. If they should be left out, instead use a filter above the .map
     // &foo
     if (value === null || value === undefined || value === '') {
       return key
     }
     
     // &foo=value
     if (!Array.isArray(value)) {
       return `${key}=${encodeURIComponent(value)}`
     }
     
     // &foo[]=value1&foo[]=value2
     return value
       .map((item) => `${key}[]=${encodeURIComponent(item)}`)
       .join('&')
  })
  .join('&')
```

app/javascript/components/student/tracks-list/TagOptionList.jsx
SleeplessByte: Transpilation of this can be configured in babel. I think it should be `includes` anyway.

app/javascript/components/student/tracks-list/TagOption.jsx
kntsoriano: Good point. This is something to remember in the future.

app/javascript/components/student/tracks-list/Header.jsx
kntsoriano: Sounds good!

kntsoriano:
Thanks everyone! Merging this in now.

PR #156, opened 2020-10-05
Add support for creating solutions not via CLI

iHiD:
This adds two things:
1. Support for uploading files via a mapped files has (`(filename: contents}`) which is how I presume the editor should work.
2. A major flag for iterations. 

I don't particularly like this approach or solution. I suspect a better other option is that we have a model for submitted code (a `submission`) and then an iteration is more like a pointer to a specific submission. I still think I might end up doing this, but it's a chunky rewrite to do it and I don't want to go into that rabbit hole right now, else we'll just get behind.

I'm trying to insulate the public API from knowing about this, and the downstream of tooling jobs is already insulated, so if I do  change it, it shouldn't affect the React side much.

### DIFF
--- a/app/commands/iteration/create.rb
+++ b/app/commands/iteration/create.rb
@@ -2,11 +2,13 @@ class Iteration
   class Create
     include Mandate
 
-    def initialize(solution, files, submitted_via)
+    def initialize(solution, files, submitted_via, major)
+      @iteration_uuid = SecureRandom.compact_uuid
+
       @solution = solution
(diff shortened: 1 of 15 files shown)


iHiD:
@kntsoriano @ErikSchierboom I'm going to merge this as I intend to rewrite it (so it's not worth your time reviewing) but you'll will need it on master to do the editor stuff, and I want to build on it to get stuff ready for you both :)

PR #157, opened 2020-10-05
Add flow to creation of iteration

iHiD:
This gives you the ability to go from the front page to the editor, where you can submit a solution. That editor page is where the React component should live, and the form is an example of the endpoint you can use. This can be merged then overriden.

This is the second PR I've written today that's more WIP than solid code! But I'll keep it going down to master so you both can progress, and then refactor/rework bits further down the line.


app/controllers/api/solutions_controller.rb
```diff
@@ -67,12 +67,16 @@ def update
...
       rescue DuplicateIterationError
         return render_error(400, :duplicate_iteration)
       end
 
-      render json: {}, status: :created
+      render json: {
+        iteration: {
+          uuid: iteration.uuid
+        }
+      }, status: :created
```
iHiD, line 79: This will get a proper serializer at some stage.

app/controllers/tracks/concepts_controller.rb
```diff
@@ -1,34 +1,42 @@
...
+    solution = Solution::Create.(@user_track, @exercise)
+    redirect_to edit_solution_path(solution.uuid)
+  end
+
+  def complete
+    ConceptExercise::Complete.(@exercise)
+    solution = Solution.for(current_user, @exercise)
+    solution.complete!
+    redirect_to action: :show
   end
```
iHiD: These are here for now, but I'll almost certainly move them at some stage to a different controller (ConceptExercise or ConceptSolutions or Solutions). But I want to work out some more paths first.

app/views/solutions/edit.html.haml
```diff
@@ -0,0 +1,8 @@
+.md-container
+  Really sexy editor for #{@track.title}: #{@exercise.title}
+
+  = form_with url: api_solution_path(@solution.uuid), method: :patch do
+    = hidden_field_tag :auth_token, current_user.auth_tokens.first.token
+    = hidden_field_tag :major, false
+    = text_area_tag "files[foobar.rb]", "class Code\nend", style: 'border:2px solid black;padding:20px;'
+    = button_tag "Submit"
```
iHiD: @kntsoriano This is where the editor component should be rendered from, and this form submits something, so you can replicate in JS.

app/views/tracks/concepts/show/joined.html.haml
```diff
@@ -1,2 +1,13 @@
...
+  - if @solution
+    - if @solution.completed?
+      Exercise Completed
+    - else
+      Exercise Started
+      = link_to "Continue", edit_solution_path(@solution.uuid)
+      = link_to "Complete", complete_track_concept_path(@track, @concept), method: :patch
+  - else
+    Exercise not started
+    = button_to "Start exercise", start_track_concept_url(@track, @concept), method: :patch
```
iHiD: This is all just scaffolding for enabling us to walk through everything. It'll all get done properly once the designs and URLs are finalised.

config/routes.rb
```diff
@@ -34,16 +34,31 @@
...
+  resources :solutions, only: %i[create edit]
+
   root to: "pages#index"
 
+  # ###############
+  # Legacy routes #
+  # ###############
+  get "solutions/:uuid" => "solutions#legacy_show"
+  get "my/solutions/:uuid" => "solutions#legacy_show"
+  get "mentor/solutions/:uuid" => "solutions#legacy_show"
```
iHiD: These are existing v2 routes that are changing for v3.

test/controllers/api/solutions_controller_test.rb
```diff
@@ -285,7 +305,7 @@ def setup
     http_files = [SecureRandom.uuid, SecureRandom.uuid]
     files = mock
     Iteration::PrepareHttpFiles.expects(:call).with(http_files).returns(files)
-    Iteration::Create.expects(:call).with(solution, files, :cli, true)
+    Iteration::Create.expects(:call).with(solution, files, :cli, true).returns(mock(uuid: "foobar"))
```
iHiD: Once we've got the iteration serializer worked out, these can use that instead.

app/controllers/tracks/concepts_controller.rb
```diff
@@ -43,5 +51,6 @@ def use_concepts
 
   def use_concept
     @concept = @track.concepts.find(params[:id])
+    @exercise = ConceptExercise.that_teaches!(@concept)
```
ErikSchierboom: Nice naming!

test/models/solution_test.rb
```diff
@@ -49,4 +41,48 @@ class SolutionTest < ActiveSupport::TestCase
...
+
+    assert_equal solution, Solution.for(user, exercise)
+  end
+
+  test "anonymised_user_handle" do
+    solution_1 = create :concept_solution
+    solution_2 = create :concept_solution
+
+    assert_equal solution_1.anonymised_user_handle,
+      "anonymous-#{Digest::SHA1.hexdigest("#{solution_1.id}-#{solution_1.uuid}")}"
```
ErikSchierboom, line 83: This is duplicating the actual implementation, but with the `id` and `uuid` not being stable, I don't think we can do much about that.

ErikSchierboom (approved):

app/controllers/tracks/concepts_controller.rb
iHiD: Thanks! 😁

test/models/solution_test.rb
iHiD: Yeah - I could stub them both I guess. But I figured it was at least safe.

kntsoriano (approved):
Happy for you to just merge this in.

iHiD:
Thanks. I'll add some more tests tomorrow, and merge, and then you should have the base point to work from 🙂

app/models/user_track.rb
```diff
@@ -51,7 +70,7 @@ class DetermineAvailableExercisesIds
 
     # Get two datasets.
     # The first is the exercises with their counts of MATCHING preqreq concepts
-    # The second is the execises with the count of all their preqreq concepts
+    # The second is the exercises with the count of all their preqreq concepts
```
ErikSchierboom:
```suggestion
    # The second is the exercises with the count of all their prereq concepts
```

app/models/user_track.rb
```diff
@@ -51,7 +70,7 @@ class DetermineAvailableExercisesIds
 
     # Get two datasets.
     # The first is the exercises with their counts of MATCHING preqreq concepts
```
ErikSchierboom:
```suggestion
    # The first is the exercises with their counts of MATCHING prereq concepts
```

app/models/user_track.rb
```diff
@@ -19,14 +19,37 @@ def self.for(user_param, track_param)
...
     available_exercises.select { |e| e.is_a?(ConceptExercise) }
   end
 
+  memoize
   def available_practice_exercises
     available_exercises.select { |e| e.is_a?(PracticeExercise) }
   end
 
+  def exercise_available?(exercise)
+    (exercise.prerequisites - learnt_concepts).empty?
```
ErikSchierboom, line 41: Being able to use the `-` is nice!

ErikSchierboom (approved):

app/models/user_track.rb
iHiD: Done in a separate commit on master. Thanks 👍

app/models/user_track.rb
iHiD: Done in a separate commit on master. Thanks 👍

iHiD:
Going to merge this in. There's loads that I think will change about it, so let's in no way consider this as a set of good practices yet, but it gives us some basis to get started with!

PR #159, opened 2020-10-05
Only use the bits of Mandate that are needed

iHiD:


test/models/user_track_test.rb
```diff
@@ -66,6 +66,9 @@ class UserTrackTest < ActiveSupport::TestCase
     assert_equal [concept_exercise_1], user_track.available_concept_exercises
     assert_equal [practice_exercise_1], user_track.available_practice_exercises
 
+    # Reload the user track to override memoizing
```
kntsoriano, line 69: This might be a smell.

kntsoriano (approved):
Nice!

ErikSchierboom (approved):

test/models/user_track_test.rb
iHiD: Yeah - the problem is that `track.reload` doesn't work with mandate's memoizing. I'm tempted to fix it (but not tempted enough to invest the time right now). So it's a question of whether `track.reload` would also be a smell I guess.

PR #161, opened 2020-10-06
Rename iteration to submission

iHiD:
**This PR does not need reviewing** _(else you'll be here all day)_

---

This PR renames `iterations` to `submissions`.

- All code submitted will now be in stored as a submission. 
- Submissions are a bundle of files, that have the tests run on them. They are what is submitted via the CLI etc.
- Submissions are essentially an internal construct - users wouldn't go back and look at previous submissions. It is expected most will fail the tests.
- Iterations are effectively breakpoints that we show to users. A future PR will add them in.

Comments etc will still be stored on iterations, not on submissions.


Gemfile
```diff
@@ -21,7 +21,7 @@ gem 'rest-client' # http
 gem 'mandate', '0.4.0.beta.1'
 
 # Setup dependencies
-gem 'exercism-config', '>= 0.43.0'
+gem 'exercism-config', '>= 0.44.0'
```
iHiD, line 25: This changes the config so the field is now `aws_submissions_bucket` not `aws_iterations_bucket`

Gemfile
```diff
@@ -5,8 +5,9 @@ ruby '2.6.6'
 # Service/framework dependencies
 gem 'rails',
   git: 'http://github.com/rails/rails.git',
-  ref: '7101489293ece071013417aeed2da3aa1b0927c9'
+  ref: '207f116234374d6414593246240b05261a96fa3a'
```
iHiD, line 8: Sporadic bumping of Rails that I thought I'd do while breaking everything else at the same time 😉

config/initializers/delayed_job.rb
```diff
@@ -1,3 +1,5 @@
 class Delayed::Job
+  self.abstract_class = true
```
iHiD, line 2: Needed because of https://github.com/rails/rails/pull/40219

iHiD:
I'm going to merge this, because everything else should be rebased onto it, and doing so should be straightforward. But if it gets more out of sync it'll become impossible 🙂

ErikSchierboom:
👍

PR #166, opened 2020-10-06
Downgrade webpacker

iHiD:
This issue that https://github.com/exercism/v3-website/pull/165 fixed is that webpacker is typescripting the babel output, not the typescript itself. Downgrading webpacker fixes this.

This does the downgrade and reverts the previous hotfix.


ErikSchierboom:
Weird!

PR #171, opened 2020-10-07
Install yarn dependencies at runtime

ErikSchierboom:
This is similar to how we install the Ruby bundles at runtime. In practice this addition means that new NPM packages can be installed and used with just a restart of the container, instead of having to do a full rebuild.﻿


SleeplessByte:
@neenjaw no more forced restarts 💯

neenjaw:
:party-parrot:

PR #177, opened 2020-10-08
Stop header from jumping around

### LINKED ISSUE #175: Content jumping around while loading tracks
When searching/filtering etc, there's a lot of flashing happening caused by page elements jumping around:

https://www.loom.com/share/026b7c0909c84fd7a62e9fbf22547138

It feels like the whole page is being rerendered each time, causing everything to jump around, rather than just the parts that actually need to be re-rendered.

If only the relevant bits of content were changing the the page wouldn't be jumping and we wouldn't get this flashing effect.

We should fix this before doing more component stuff.

(cc @SleeplessByte for advice 🙂)

kntsoriano:
Closes https://github.com/exercism/v3-website/issues/175.

I've also fixed the tests in this PR.

## Preview

Oct-08-2020 13-53-19

### DIFF
--- a/app/javascript/components/student/TracksList.jsx
+++ b/app/javascript/components/student/TracksList.jsx
@@ -1,11 +1,10 @@
 import React, { useReducer } from 'react'
-import { useRequestQuery } from '../../hooks/request-query'
+import { usePaginatedRequestQuery } from '../../hooks/request-query'
 import { Search } from './tracks-list/Search'
 import { StatusFilter } from './tracks-list/StatusFilter'
 import { TagsFilter } from './tracks-list/TagsFilter'
 import { List } from './tracks-list/List'
 import { Header } from './tracks-list/Header'
-import { Loading } from '../common/Loading'
 
 function reducer(state, action) {
   switch (action.type) {
@@ -32,10 +31,12 @@ function reducer(state, action) {
 
 export function TracksList({ statusOptions, tagOptions, ...props }) {
   const [request, dispatch] = useReducer(reducer, props.request)
-  const { data, isSuccess, isError, isLoading } = useRequestQuery(
-    'track-list',
-    request
-  )
+  const {
+    resolvedData,
+    latestData,
+    isError,
+    isFetching,
+  } = usePaginatedRequestQuery('track-list', request)
 
   return (
     <div className="c-tracks-list">
@@ -46,16 +47,15 @@ export function TracksList({ statusOptions, tagOptions, ...props }) {
         </div>
       </section>
       <section className="md-container">
-        {isSuccess && <Header data={data} query={request.query} />}
+        <Header latestData={latestData} query={request.query} />
         <StatusFilter
           value={request.query.status}
           dispatch={dispatch}
           options={statusOptions}
         />
 
-        {isLoading && <Loading />}
         {isError && <p>Something went wrong</p>}
-        {isSuccess && <List data={data} />}
+        {resolvedData && <List data={resolvedData} />}
       </section>
     </div>
   )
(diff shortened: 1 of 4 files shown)


app/javascript/components/student/TracksList.jsx
```diff
@@ -32,10 +31,12 @@ function reducer(state, action) {
...
-  const { data, isSuccess, isError, isLoading } = useRequestQuery(
-    'track-list',
-    request
-  )
+  const {
+    resolvedData,
+    latestData,
+    isError,
+    isFetching,
+  } = usePaginatedRequestQuery('track-list', request)
```
ErikSchierboom, line 39: It used to be `useRequestQuery`. Is there paging involved now?

app/javascript/components/student/TracksList.jsx
```diff
@@ -46,16 +47,15 @@ export function TracksList({ statusOptions, tagOptions, ...props }) {
         </div>
       </div>
       <div className="md-container">
-        {isSuccess && <Header data={data} query={request.query} />}
+        <Header latestData={latestData} query={request.query} />
```
ErikSchierboom: Ah, so this is the trick where we don't conditionally render the element but instead use the latest data. Nice!

Are there more places where we should be doing this?

app/javascript/components/student/tracks-list/Header.jsx
```diff
@@ -1,14 +1,28 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import pluralize from 'pluralize'
+import { Loading } from '../../common/Loading'
 
-export function Header({ data, query = {} }) {
+export function Header({ latestData, query = {} }) {
+  const [text, setText] = useState()
```
ErikSchierboom: I don't know if this makes sense, but would initializing this with a default value also help reduce a re-render?

```suggestion
  const [text, setText] = useState("Exercism's Language Tracks")
```

test/javascript/components/student/TracksList.test.js
```diff
@@ -12,9 +14,41 @@ test('disables the selected status filter', () => {
   const { getByText } = render(
     <TracksList
       statusOptions={options}
+      tagOptions={[]}
```
ErikSchierboom: We'll be able to catch these errors once we've converted everything to TypeScript!

ErikSchierboom (approved):

app/javascript/components/student/tracks-list/Header.jsx
iHiD: This is a good idea I think.

app/javascript/components/student/tracks-list/Header.jsx
```diff
@@ -1,14 +1,28 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import pluralize from 'pluralize'
+import { Loading } from '../../common/Loading'
 
-export function Header({ data, query = {} }) {
+export function Header({ latestData, query = {} }) {
```
iHiD: I don't get the reason for this rename inside this component. As far as the Header is concerned is this not always it's `data`?

iHiD (approved):
@kntsoriano I rebased this. I'll let you reply to the comments or update things as you see fit, and then merge

app/javascript/components/student/tracks-list/Header.jsx
kntsoriano: Ah yes. I'll rename it back to data.

app/javascript/components/student/tracks-list/Header.jsx
kntsoriano: Yeah. Good idea!

app/javascript/components/student/TracksList.jsx
kntsoriano: I think we will need to adapt this pattern more and more in the future. But let's leave that for later I think.

PR #184, opened 2020-10-08
Create codeql-analysis.yml

iHiD:
This is GitHub's auto code analyzer. It keeps bugging me to add it. Should we?

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '0 4 * * 1'
+
+jobs:
+  analyze:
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
As this seems to be a security feature, I'm in favor of enabling this. Individual tracks should definitely enable this at some point I feel.

I assume you just enabled JavaScript to get rid of the message? What would the disadvantage be of enabling the other support languages? That it takes a long time to run?

iHiD:
JS is the only language that this repo uses that it supports (ie it doesn't support Ruby), so I enabled it for that. The only thing I don't know is if it really only supports node, rather than client side JS, and then if we like it we can put it on for other repos. It might be valuable for the test-runners etc, where there's actual code being executed.

I'll turn it on and we can see what happens 🤷 🙂

ErikSchierboom:
Okay cool!

SleeplessByte:
I think it just scans _any_ .js/.ts.

> With ATM, your JavaScript and TypeScript queries will identify more security problems. For example, we used ATM in the GitHub Security Lab to find 118 new NoSQL injection vulnerabilities across 50 JavaScript projects.

iHiD:
Yeah, but sql injection is server side. I guess what I mean is does it look for client side vulnerabilities? As we don't use JS in production on the server it's not useful if not.

SleeplessByte:
I need a shrug GitHub réaction.

It doesn't cost us any way right?

iHiD:
Exactly.

PR #185, opened 2020-10-09
Add fast refresh

### LINKED ISSUE #169: Add Fast Refresh
Follow [these instructions](https://github.com/shakacode/webpacker/blob/ba51a95076731b90a9cda70c5216f57491a82433/docs/react.md#react-refresh-webpack-plugin).

Allows for refreshing the react component _without_ having the page to reload. Might not work with our bootloader, but then we can figure _that_ out.

kntsoriano:
It works!! 🥳 

Closes https://github.com/exercism/v3-website/issues/169.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,4 +1,4 @@
 server: bin/rails server
-assets: bin/webpack-dev-server
+assets: WEBPACK_DEV_SERVER_PORT=3035 bin/webpack-dev-server
 anycable: RAILS_ENV=development bundle exec anycable --server-command "anycable-go --host=localhost --port 3334"
 js-tests: yarn test-watch
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,4 +1,4 @@
 server: bin/rails server
-assets: WEBPACK_DEV_SERVER_PORT=3035 bin/webpack-dev-server
+assets: bin/webpack-dev-server
 anycable: RAILS_ENV=development bundle exec anycable --server-command "anycable-go --host=localhost --port 3334"
 js-tests: yarn test-watch
--- a/config/webpack/development.js
+++ b/config/webpack/development.js
@@ -21,11 +21,7 @@ const isWebpackDevServer = process.env.WEBPACK_DEV_SERVER
 if (isWebpackDevServer) {
   environment.plugins.append(
     'ReactRefreshWebpackPlugin',
-    new ReactRefreshWebpackPlugin({
-      overlay: {
-        sockPort: process.env.WEBPACK_DEV_SERVER_PORT,
-      },
-    })
+    new ReactRefreshWebpackPlugin({ overlay: false })
   )
 }
 
--- a/config/webpack/development.js
+++ b/config/webpack/development.js
@@ -23,7 +23,7 @@ if (isWebpackDevServer) {
     'ReactRefreshWebpackPlugin',
     new ReactRefreshWebpackPlugin({
       overlay: {
-        sockPort: 3035,
+        sockPort: process.env.WEBPACK_DEV_SERVER_PORT,
       },
     })
   )
--- a/config/webpack/development.js
+++ b/config/webpack/development.js
@@ -16,4 +16,17 @@ environment.plugins.append(
   })
 )
 
+const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin')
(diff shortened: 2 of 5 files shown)


SleeplessByte (approved):
Isn't that cool?

Should speed up development significantly

iHiD (approved):

iHiD (approved):

config/webpack/development.js
```diff
@@ -16,4 +16,17 @@ environment.plugins.append(
...
 )
 
+const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin')
+const isWebpackDevServer = process.env.WEBPACK_DEV_SERVER
+if (isWebpackDevServer) {
+  environment.plugins.append(
+    'ReactRefreshWebpackPlugin',
+    new ReactRefreshWebpackPlugin({
+      overlay: {
+        sockPort: 3035,
```
iHiD: This port is configurable via the development-environment (because not having that broke @SleeplessByte's world), so that config needs to get into here somehow.

SleeplessByte: Yes

kntsoriano: Currently, the port is also hard coded inside `config/webpacker.yml`. I think this is the file that should be configurable via the development-environment. And then this file would then get the port from `config/webpacker.yml`.

kntsoriano:
@iHiD @SleeplessByte Are you okay with the latest commit? The next thing we need to do is to make `WEBPACK_DEV_SERVER_PORT` configurable via the `development-environment`. When we use this environment variable, we override the port written in`config/webpacker.yml`.

SleeplessByte:
Yeah, as long as out tooling overrides it we're good.

For context, if this runs on 3035 in docker, and is forwarded as 3035 (by default), and you start a rails server outside of docker, and start webpacker outside of docker, the bind won't fail (docker bound on 0.0.0.0 and the new one on localhost). 

The result is that docker webpacker will server assets to your localhost rails, without it failing. So all the hashes are wrong, but you don't know/can't tell why.

Procfile.dev
```diff
@@ -1,4 +1,4 @@
 server: bin/rails server
-assets: bin/webpack-dev-server
+assets: WEBPACK_DEV_SERVER_PORT=3035 bin/webpack-dev-server
```
iHiD: At the moment the development-environment sets `WEBPACKER_DEV_SERVER_PUBLIC`. I guess we should change that to be `WEBPACKER_DEV_SERVER_PORT` instead?

https://github.com/exercism/development-environment/blob/99a722e84b37d298eb6451e3755b75e9a65e7524/stack.default.yml#L24

I think what you have here is correct, but I just want to check this all makes sense end-to-end before merging.

kntsoriano: Yeah. That sounds good. Is the `WEBPACKER_DEV_SERVER_PUBLIC` an actual env variable for webpacker configuration? Or is it something that we just created?

ErikSchierboom: @kntsoriano It looks like it is a default way to set variables: https://github.com/rails/webpacker/blob/227661b546701e107213aa9b6f83451bc89cf7d5/docs/cloud9.md

ErikSchierboom (commented):
@kntsoriano Have you tried to run this using the Docker-based `development-environment` setup? I think there is an issue:

```
server   | Started GET "/sockjs-node/info?t=1602579788052" for 172.18.0.1 at 2020-10-13 09:03:08 +0000
server   | Cannot render console from 172.18.0.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
server   |   
server   | ActionController::RoutingError (No route matches [GET] "/sockjs-node/info"):
server   |   
server   | rails (207f11623437) actionpack/lib/action_dispatch/middleware/debug_exceptions.rb:33:in `call'
```

Procfile.dev
kntsoriano: Cool. Sounds good! I will open an issue to set `WEBPACK_DEV_SERVER_PORT` in the development environment.

kntsoriano:
@ErikSchierboom Yes. I've tried it on the docker based setup. I'll check again.

kntsoriano:
@iHiD @SleeplessByte @ErikSchierboom I've gone back to this and noticed that we don't need the overlay option, as Rails already has this built in. We shouldn't worry about the ports anymore as well as the error message above.

I realized this as I was able to make the overlay option work, but it's redundant. I'll merge this once tests pass.

PR #186, opened 2020-10-10
Deploy Exercism v3 from main branch commits

iHiD:
This ridiculous workflow:
- Builds Nginx Docker image
- Builds AnyCable-Go Docker image
- Builds the Rails Docker image
- Pushes the Dockerfiles to ECR
- Uploads the compiled assets from the Rails image to S3
- Builds a new task-definition for ECS
- Deploys the new task-definition, thus activating a deployment, which
  - Recreates the database (temporarily!)
  - Runs the migrations with a concurrent guard for when multiple servers are running them at once.
  - Starts the server

I've tried this with both multiple (paralllel) jobs and a single job, but ended up on a single job because nginx and anycable are so fast that the duplication and multiple caches just slowing the overall thing down. This also means the Docker-layer caching can happen after **everything**, which saves us about 5 minutes.

It works and is ready for review 🙂


iHiD:
@kntsoriano @ErikSchierboom This is ready to review.

iHiD:
The two TODOs still need doing before merge.

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,205 @@
...
+  dockerfile_rails: "docker/rails.Dockerfile"
+
+  ecr_repository_nginx: "webserver-nginx"
+  dockerfile_nginx: "docker/nginx.Dockerfile"
+
+  ecr_repository_anycable_go: "webserver-anycable-go"
+  dockerfile_anycable_go: "docker/anycable-go.Dockerfile"
+
+  # TODO: Move this to be a secret
+  s3_bucket: "exercism-assets-staging"
```
iHiD, line 19: I'll do this in a separate commit before final launch.

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,205 @@
...
+          # Change these to rest teh cache
+          key: docker-layer-caching-deploy-{hash}
+          restore-keys: docker-layer-caching-deploy-
+          concurrency: 30
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_DEPLOY_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_DEPLOY_SECRET_ACCESS_KEY }}
```
iHiD, line 51: These are org-secrets that point to keys for a user called `github-deploy`. That user's permissions are defined in [these files](https://github.com/exercism/terraform/tree/master/terraform/github_deploy) in the terraform repo, are are effectively limited to reading and writing to the relevant ECR repositories, getting and pushing new task-definitions and spinning a new service up.

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,205 @@
...
+          DOCKERFILE: ${{ env.dockerfile_rails }}
+          IMAGE_NAME: "${{ steps.login_to_ecr.outputs.registry}}/${{ env.ecr_repository_rails }}:${{ github.sha }}"
+        run: |
+          # Build a docker container and push it to ECR
+          docker build -t $IMAGE_NAME -f $DOCKERFILE .
+          docker push $IMAGE_NAME
+
+          # Retag this as the latest
+          docker tag $IMAGE_NAME $ECR_REGISTRY/$ECR_REPOSITORY:latest
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:latest
```
iHiD, line 77: The reason for this is that if we override the task-definition in terraform, we can just use `latest` there, without having to do complex interpolation of the latest version. However, for general practice it's good use to use the actual shas so that we can avoid race-conditions etc. We therefore tag the same image twice with sha and with `latest`.

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,209 @@
...
+
+      ##################################
+      ##################################
+      # Extact and upload assets to s3 #
+      ##################################
+      ##################################
+      - name: Copy assets from Dockerfile
+        run: |
+          container_id=$(docker create website)
+          docker cp ${container_id}:/opt/exercism/website/current/public/packs packs
```
iHiD, line 93: These two lines create (?) a container, which I think is like exploding it without running it, then copy the compiled assets from inside it into the FS of the GHA for uploading to S3 below.

_(@neenjaw This was that bit of bash you helped me with 🙂)_

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,209 @@
...
+
+      - name: Upload assets to s3
+        run: |
+          aws s3 sync packs s3://${{ env.s3_bucket }}/packs/ --acl public-read --no-progress
```
iHiD, line 98: This pushes new files up to S3. These are the compiled CSS and JS files. They get unique shas (e.g. `application-123asd.css`), so they don't override existing files.

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,209 @@
...
+
+      #########################
+      #########################
+      # Build the Nginx image #
+      #########################
+      #########################
```
iHiD, line 104: Rinse and repeat the build and push steps for nginx and anycable, but without the assets hassle.

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,209 @@
...
+      ##############################
+
+      # We get the existing task definition to update
+      # Note that this is passed into the first update action
+      # but then chained from action to action for the others.
+      - name: Download task definition
+        run: |
+          aws ecs describe-task-definition --task-definition ${{ env.task_definition_name }} --query taskDefinition > task-definition.json
+
+      # Set puma to rails
```
iHiD, line 161: This (`puma`) and `anycable_rails` are two containers that use the same image, so we have two blocks to update in the file, although we've only created one image for both.

The task definition sets a custom ENTRYPOINT for the anycable one.

bin/start_webserver
```diff
@@ -0,0 +1,23 @@
...
+end
+
+FileUtils.chdir APP_ROOT do
+  # TODO: We're going to want to remove this evenutally!!
+  # Recreates the database on each deployment.
+  system! 'DISABLE_DATABASE_ENVIRONMENT_CHECK=1 bundle exec bin/rails db:migrate:reset db:seed'
+
+  # Firstly we're going to run the migrations.
+  # We do this via this helper script to sleep and retry on conurrent failures
+  system! 'bundle exec bin/rails runner lib/run_migrations_with_concurrent_guard.rb'
```
iHiD, line 19: So every time we deploy we run the migrations as the service starts. 

- If the migrations fail, then this script will abort, the server (below) won't start, and so the deploy will fail (based on the Fargate health checks). We will also raise to Bugsnag here eventually.
- If two servers run migrations at the same time (very likely) the first will lock it and the second will keep getting an exception until the first has finished. Once that's done, the second will run the migrations again (a no-op) then continue. There's a script below for this.

config/environments/production.rb
```diff
@@ -100,4 +100,10 @@
   # config.active_record.database_selector = { delay: 2.seconds }
   # config.active_record.database_resolver = ActiveRecord::Middleware::DatabaseSelector::Resolver
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
+
+  config.action_controller.asset_host = "exercism-assets-staging.s3.eu-west-2.amazonaws.com"
 end
+
+Rails.application.routes.default_url_options = {
+  host: "http://exercism.lol"
+}
```
iHiD, line 109: This will be our staging domain - at least for a while. If the nameservers ever propgate!

config/environments/production.rb
```diff
@@ -100,4 +100,10 @@
   # config.active_record.database_selector = { delay: 2.seconds }
   # config.active_record.database_resolver = ActiveRecord::Middleware::DatabaseSelector::Resolver
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
+
+  config.action_controller.asset_host = "exercism-assets-staging.s3.eu-west-2.amazonaws.com"
```
iHiD, line 104: This is the S3 bucket our resources are hosted in. I'll set it up as a Cloudfront distribution and access it via the Exercism Config gem eventually

tailwind.config.js
```diff
@@ -3,7 +3,10 @@ module.exports = {
     // removeDeprecatedGapUtilities: true,
     // purgeLayersByDefault: true,
   },
-  purge: [],
+  purge: [
+    './app/views/**/*.haml',
+    './app/javascript/**/*',
+  ],
```
iHiD, line 9: @ErikSchierboom @kntsoriano @SleeplessByte Are these the only places that will have CSS class names? I think so.

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,209 @@
...
+      # Setup Steps #
+      ###############
+      ###############
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - uses: satackey/action-docker-layer-caching@v0.0.5
+        continue-on-error: true
+        with:
+          # Change these to rest teh cache
```
ErikSchierboom:
```suggestion
          # Change these to rest the cache
```

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,209 @@
...
+          # Build a docker container and push it to ECR
+          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG -f $DOCKERFILE .
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
+
+          # Retag this as the latest
+          docker tag $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG $ECR_REGISTRY/$ECR_REPOSITORY:latest
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:latest
+
+          # Save the output for writing to the task definiton
+          echo "::set-output name=image::$ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG"
```
ErikSchierboom: Where is this output being used?

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,209 @@
...
```
ErikSchierboom: Where is this output being used?

.github/workflows/deploy.yml
ErikSchierboom: This is a well-known pattern to get data out of Docker container. Could you also remove the container at the end?

```suggestion
          docker cp ${container_id}:/opt/exercism/website/current/public/packs packs
          docker rm ${container_id}
```

bin/start_webserver
```diff
@@ -0,0 +1,23 @@
...
+# Path to the application root.
+APP_ROOT = File.expand_path('..', __dir__)
+
+# We want to guard against failures at each stage of this
+def system!(*args)
+  system(*args) || abort("\n== Command #{args} failed ==")
+end
+
+FileUtils.chdir APP_ROOT do
+  # TODO: We're going to want to remove this evenutally!!
```
ErikSchierboom:
```suggestion
  # TODO: We're going to want to remove this eventually!!
```

ErikSchierboom (approved):

iHiD:
It works!! 🦜 

Merging this monstrosity!!

.github/workflows/deploy.yml
iHiD: Added. Thanks 🙂

tailwind.config.js
SleeplessByte: If you ensure you don't write html on the controller, model, decorator, etc. side, and dont use the regular asset pipeline, then yes.

PR #190, opened 2020-10-12
Show passing test results

kntsoriano:
## Description
This is a first pass of showing the passing test results in the editor. I decided to open this PR right away as I want to get feedback as quick as possible, and be able to merge this is in so @ErikSchierboom and I could work at several parts of this component in parallel.

## Questions
1. What do we call this component?
2. I decided to make this component subscribe to `SolutionsChannel`. I did this as I understand this as creating multiple submissions for a specific solution. Is this okay? This would also help catch race conditions as we can instruct the `SolutionsChannel` to just always broadcast the last submission in the DB.

### DIFF
--- a/app/channels/test/solutions_channel.rb
+++ b/app/channels/test/solutions_channel.rb
@@ -1,0 +1,21 @@
+module Test
+  class SolutionsChannel < ApplicationCable::Channel
+    def subscribed
+      stream_for solution
+    end
+
+    def submit
+      Test::SolutionsChannel.broadcast_to(solution, {
+                                            status: "pass",
+                                            tests: [
+                                              { name: :test_a_name_given, status: :pass, output: "Hello" }
+                                            ]
+                                          })
+    end
+
+    private
+    def solution
+      @solution ||= Solution.first
+    end
+  end
+end
--- a/app/helpers/view_components/editor.rb
+++ b/app/helpers/view_components/editor.rb
@@ -1,0 +1,14 @@
+module ViewComponents
+  class Editor < ViewComponent
+    def initialize(solution)
+      @solution = solution
+    end
+
+    def to_s
+      react_component("editor", { solution_id: solution.uuid })
+    end
+
+    private
+    attr_reader :solution
+  end
+end
--- a/app/javascript/components/common/Editor.tsx
+++ b/app/javascript/components/common/Editor.tsx
@@ -1,0 +1,70 @@
+import React, { useEffect, useState } from 'react'
+import consumer from '../../utils/action-cable-consumer'
+
+type Submission = {
+  status: SubmissionStatus
+  tests: TestResult[]
+}
+
+enum SubmissionStatus {
+  PASS = 'pass',
+}
+
+type TestResult = {
+  name: string
+  status: TestResultStatus
+  output: string
+}
+
+enum TestResultStatus {
+  PASS = 'pass',
+}
+
+export function Editor({ solutionId }: { solutionId: string }) {
+  const [submission, setSubmission] = useState<Submission>()
+  const [subscription, setSubscription] = useState(
+    consumer.subscriptions.create(
+      { channel: 'Test::SolutionsChannel', id: solutionId },
+      {
+        received: (submission: Submission) => {
+          setSubmission(submission)
+        },
+      }
+    )
+  )
+  const [code, setCode] = useState('')
+
+  useEffect(() => {
+    return () => {
+      subscription.unsubscribe()
+    }
+  }, [])
+
+  function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+    e.preventDefault()
+
+    subscription.perform('submit', { code: code })
+  }
+
+  function handleChange(e: React.ChangeEvent<HTMLTextAreaElement>) {
+    setCode(e.target.value)
+  }
+
+  return (
+    <div>
+      <form onSubmit={handleSubmit}>
+        <label htmlFor="code">Code</label>
+        <textarea onChange={handleChange} id="code"></textarea>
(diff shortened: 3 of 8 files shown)


app/channels/test/solutions_channel.rb
```diff
@@ -0,0 +1,21 @@
+module Test
+  class SolutionsChannel < ApplicationCable::Channel
+    def subscribed
+      stream_for solution
+    end
+
+    def submit
+      Test::SolutionsChannel.broadcast_to(solution, {
```
kntsoriano, line 8: @iHiD I'm unsure of what serializer to use here. Should we return the whole serialized submission? Or just the test runs?

app/channels/test/solutions_channel.rb
```diff
@@ -0,0 +1,21 @@
+module Test
```
kntsoriano, line 1: Scoped under test as we'll need to create an actual `SolutionsChannel` which does everything correctly. This test channel will just return stubbed responses.

iHiD:
Good decision to put the WIP PR on. Gives us good stuff to talk about 🙂 

It feels like there are a few architectural -level decisions to make. I'll put a few thoughts down. Nothing I'm sane here is set in stone so shout if you disagree with anything 🙂 

### Seperation of components

I think we should separate out:
- The [write code + submit] component (Call this `Editor::CodeEditorComponent`?).
- Test test results component (call this `Editor::TestRunsComponent`?).
- An overarching component that links these together (`EditorComponent?`).

I think doing that now would start to give us a bit more clarity over roles + responsibilities.

### Submit via API or via Websockets

We're currently submitting via websockets. I think we should actually submit via API. That's probably a change from what I originally said, but it feels cleaner and easier to me to just hit an API endpoint with some files and get a response back. It also makes the websockets simpler and just used for broadcasting, which I think is a good thing until we feel really comfortable with our patterns.

I'll have the new API endpoints in place today to handle you submitting the submission.

### Workflow

Based on the above I think we then have the following workflow:

Setup:
- We init a `EditorComponent`.
- This inits a `CodeEditorComponent` and a `TestRunsComponent`.
- The `TestRunsComponent` subscribes to the `Submission::TestRuns` channel.

...

- Someone types code into the `CodeEditorComponent`
- Someone presses the "Submit code" in the `CodeEditorComponent`. This proxies a message up to the `EditorComponent`.
- The `EditorComponent` sends an API POST request with the `solution_id` and the code. The responses contains a serialized submission - which will have `submission_id` (amongst other things).
- This result is propagated up to `EditorComponent`, which updates the `TestRunsComponent` with the new `submission_id` and the changes the state to `pending` and  `TestRunsComponent` changes to the UI to a "In Progress" state (this would just be automatic via React when the submission data is updated). ~(I don't know if this should create a new `TestRunsComponent` or whether it's ok just to change the data - @SleeplessByte might know if either would be more idiomatic in this situation).~

...

- The test results finish on the server, and a `Submission::TestRun` object is created server-side.
- A websocket broadcast is sent out on the `Submission::TestRuns` channel with the `submission_id` and the serialized test run.
- This is picked up by the `TestRunsComponent`, which then updates its data to render the test run.

There are a few advantages of this.
1. Test Runs that come out of sync (which is a very real possibility) will be ignored by the TestRunComponent as it only cares about the latest submission it has submitted (ie a result that matches the `submission_id` it has stored).
2. The `TestRunComponent` is just responsible for rendering the test run. Any other actions (e.g. Pressing the "Retry" button) can just proxy up to the `EditorComponent`.
3. The `CodeEditorComponent` is just responsible for rendering the pretty (and later quite complicated text-editor).
4. The only websocket interaction is a single broadcast when TestRuns are completed, which is nice and straight-forward.

---

This is where I'd start, and then we can add more complications further down the line 🙂

SleeplessByte:
Changing the data is something different than creating a new component (rerender vs mount render unmount).

For dealing with a react component that should exist on the page, but, like in this case, should update when something external happens:

I usually try to hook up some event so I can handle the update inside the react ecosystem. The reason for that is that it makes it far less complex to have partial, internal state. Remounting a component means that it's state will reset. You will shoot yourself in the foot at some point with this.

So in this case, I'd try to update the data/props which should trigger a re-render, or we can force it.

For testing react components:

If you specifically want to test that the component visually updates (and I think you do, so I recommend this), it should rerender, not remount.

Usually we achieve this by

1. Mount the component
2. Check things
3. Fire event
4. Check things

In other words: no explicit rerender. 

Of course, if we're one level "too deep", this wont be possible, because it will be the props that need updating. In that case, [explicitly rerender](https://testing-library.com/docs/example-update-props) with the new props.

app/helpers/view_components/editor.rb
```diff
@@ -0,0 +1,14 @@
+module ViewComponents
+  class Editor < ViewComponent
+    def initialize(solution)
+      @solution = solution
+    end
```
ErikSchierboom, line 5: Could we use Mandate here?

app/javascript/components/common/Editor.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <button>Submit</button>
+      </form>
+      {submission && <p>Status: {submission.status}</p>}
+      {submission &&
+        submission.tests.map((result) => (
+          <p>
+            name: {result.name}, status: {result.status}, output:{' '}
+            {result.output}
+          </p>
+        ))}
```
ErikSchierboom, line 67: At some point I could imagine this being a separate component, nice work!

ErikSchierboom (commented):
Nice!

iHiD:
@kntsoriano Shall we close this for now?

kntsoriano:
Closing this now.

PR #191, opened 2020-10-14
Show test results

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/52.

## Description

1. We submit via API.
2. When the API returns the submission, we subscribe to the channel for that specific submission (referenced by the submission ID).
3. When the channel returns further data about the submission, we update the fields accordingly.

## Question
1. This approach creates a channel per submission. Is this okay?

### DIFF
--- a/app/javascript/components/common/Editor/CodeEditor.tsx
+++ b/app/javascript/components/common/Editor/CodeEditor.tsx
@@ -1,0 +1,23 @@
+import React, { useState } from 'react'
+
+export function CodeEditor({ onSubmit }: { onSubmit: (code: string) => void }) {
+  const [code, setCode] = useState('')
+
+  function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+    e.preventDefault()
+
+    onSubmit(code)
+  }
+
+  function handleChange(e: React.ChangeEvent<HTMLTextAreaElement>) {
+    setCode(e.target.value)
+  }
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <label htmlFor="code">Code</label>
+      <textarea onChange={handleChange} id="code"></textarea>
+      <button>Submit</button>
+    </form>
+  )
+}
--- a/app/javascript/components/common/Editor/TestRuns.tsx
+++ b/app/javascript/components/common/Editor/TestRuns.tsx
@@ -1,0 +1,35 @@
+import React, { useState, useEffect } from 'react'
+import { Submission } from '../Editor'
+import consumer from '../../../utils/action-cable-consumer'
+
+export function TestRuns({ submission }: { submission: Submission }) {
+  const [testsStatus, setTestsStatus] = useState(submission.testsStatus)
+  const [testRuns, setTestRuns] = useState(submission.testRuns)
+
+  useEffect(() => {
+    const subscription = consumer.subscriptions.create(
+      { channel: 'Test::Submissions::TestRunsChannel', id: submission.id },
+      {
+        received: (submission: any) => {
+          setTestsStatus(submission.tests_status)
+          setTestRuns(submission.test_runs)
+        },
+      }
+    )
+
+    return () => {
+      subscription.unsubscribe()
+    }
+  }, [submission.id])
+
+  return (
+    <div>
+      <p>Status: {testsStatus}</p>
+      {testRuns.map((run) => (
+        <p>
+          name: {run.name}, status: {run.status}, output: {run.output}
+        </p>
+      ))}
+    </div>
+  )
+}
--- a/app/javascript/components/common/Editor/TestRuns.tsx
+++ b/app/javascript/components/common/Editor/TestRuns.tsx
@@ -1,18 +1,21 @@
 import React, { useState, useEffect } from 'react'
-import { Submission } from '../Editor'
+import { Submission, SubmissionTestsStatus } from '../Editor'
 import consumer from '../../../utils/action-cable-consumer'
 
-export function TestRuns({ submission }: { submission: Submission }) {
-  const [testsStatus, setTestsStatus] = useState(submission.testsStatus)
-  const [testRuns, setTestRuns] = useState(submission.testRuns)
+export function TestRuns(props: { submission: Submission }) {
+  const [submission, setSubmission] = useState(props.submission)
 
   useEffect(() => {
     const subscription = consumer.subscriptions.create(
       { channel: 'Test::Submissions::TestRunsChannel', id: submission.id },
       {
-        received: (submission: any) => {
-          setTestsStatus(submission.tests_status)
-          setTestRuns(submission.test_runs)
+        received: (json: any) => {
+          setSubmission({
+            id: submission.id,
+            testsStatus: json.tests_status,
+            testRuns: json.test_runs,
+            message: json.message,
+          })
         },
       }
     )
@@ -22,14 +25,32 @@ export function TestRuns({ submission }: { submission: Submission }) {
     }
   }, [submission.id])
 
-  return (
-    <div>
-      <p>Status: {testsStatus}</p>
-      {testRuns.map((run) => (
+  let content
+
+  switch (submission.testsStatus) {
+    case SubmissionTestsStatus.PASS: {
+      content = submission.testRuns.map((run) => (
         <p key={run.name}>
           name: {run.name}, status: {run.status}, output: {run.output}
         </p>
-      ))}
+      ))
+    }
+    case SubmissionTestsStatus.FAIL: {
+      content = submission.testRuns.map((run) => (
+        <p key={run.name}>
+          name: {run.name}, status: {run.status}, output: {run.output}
+        </p>
+      ))
+    }
+    case SubmissionTestsStatus.ERROR: {
+      content = <p>{submission.message}</p>
+    }
+  }
+
+  return (
+    <div>
+      <p>Status: {submission.testsStatus}</p>
+      {content}
     </div>
   )
 }
--- a/app/javascript/components/common/Editor/TestRuns.tsx
+++ b/app/javascript/components/common/Editor/TestRuns.tsx
@@ -26,7 +26,7 @@ export function TestRuns({ submission }: { submission: Submission }) {
     <div>
       <p>Status: {testsStatus}</p>
       {testRuns.map((run) => (
-        <p>
+        <p key={run.name}>
           name: {run.name}, status: {run.status}, output: {run.output}
         </p>
       ))}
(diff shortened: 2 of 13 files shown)


test/javascript/components/common/Editor.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { render, fireEvent, waitFor } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import { rest } from 'msw'
+import { setupServer } from 'msw/node'
+import { Editor } from '../../../../app/javascript/components/common/Editor'
+
+const server = setupServer(
+  rest.post('https://exercism.test/submissions', (req, res, ctx) => {
+    return res(
+      ctx.json({ id: 2, tests_status: 'pending', test_runs: [], message: '' })
```
kntsoriano, line 11: @iHiD This is the JSON that I expect from the API. Let me know if we need to change this.

app/javascript/components/common/Editor/TestRuns.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export function TestRuns(props: { submission: Submission }) {
+  const [submission, setSubmission] = useState(props.submission)
+
+  useEffect(() => {
+    const subscription = consumer.subscriptions.create(
+      { channel: 'Test::Submissions::TestRunsChannel', id: submission.id },
+      {
+        received: (json: any) => {
```
kntsoriano, line 33: @SleeplessByte Is there a good way to do the ff:

1. Camelize the keys from the JSON response.
2. Convert the JSON response into one of our types.

I'm doing it manually here now, but I'm unsure if there's a better way.

app/javascript/components/common/Editor/TestRuns.tsx
```diff
@@ -0,0 +1,57 @@
...
+      return <></>
+  }
+}
+
+export function TestRuns(props: { submission: Submission }) {
+  const [submission, setSubmission] = useState(props.submission)
+
+  useEffect(() => {
+    const subscription = consumer.subscriptions.create(
+      { channel: 'Test::Submissions::TestRunsChannel', id: submission.id },
```
kntsoriano, line 31: We will make the channel name injectable once we move this out into the website.

app/javascript/components/common/Editor/CodeEditor.tsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  function handleChange(e: React.ChangeEvent<HTMLTextAreaElement>) {
+    setCode(e.target.value)
+  }
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <label htmlFor="code">Code</label>
+      <textarea onChange={handleChange} id="code"></textarea>
```
ErikSchierboom, line 19: The ID is relatively generic and there might be a small risk of a duplicate ID being used. Do we have a strategy to make sure ID's are unique? For example by using some sort of "namespacing"?

app/javascript/components/common/Editor/TestRuns.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useState, useEffect } from 'react'
+import { Submission, SubmissionTestsStatus } from '../Editor'
+import consumer from '../../../utils/action-cable-consumer'
+
+function Content({ submission }: { submission: Submission }) {
```
ErikSchierboom, line 5: The name `Content` is fairly non-descriptive. I don't know if I can come up with  better name though 🤷

ErikSchierboom (approved):

app/javascript/components/common/Editor/TestRuns.tsx
SleeplessByte: Yeah.

1. Run [`camel-case`](https://www.github.com/blakeembrey/change-case/tree/master/packages%2Fcamel-case%2FREADME.md) recursively.
2. We can assert then type cast:

```typescript
function typecheck<T>(json: any, key: string): T | never {
  // For simplicity, let's say that we trust the
  // JSON if it's a non-empty object. In reality
  // we'd want to use media types to ensure
  // this. I recommend the above regardless
  // so structure all JSON results with at least
  // a single top level key, e.g. exercise,
  // concepts, messages, user:
  //
  // {
  //    "exercise": ...
  // }
  
  if (json === null || typeof json !== 'object') {
    throw new Error(`Expexted non-null object, actual ${json === null ? null : typeof json}`)
  }

  if (!json.hasOwnProperty(key)) {
    const keys = Object.keys(json)
    throw new Error(`Expected top-level key ${key}, actual keys: ${keys}`)
  }
  
  // You can either decide to type the value of 
  // the top level key, and return the value,
  // or if you want more than one top-level key
  // type the entire thing.
  return json[key] as T
}
```

Typed this on my phone, apologies for errors

app/javascript/components/common/Editor/CodeEditor.tsx
iHiD: Yeah - I don't want us using `id` in React at all really. Is there ever a reason to? If it's for testing, can we use `data-xyz` attributes instead pls?

app/javascript/components/common/Editor/TestRuns.tsx
```diff
@@ -0,0 +1,57 @@
...
+        },
+      }
+    )
+
+    return () => {
+      subscription.unsubscribe()
+    }
+  }, [submission.id])
+
+  let content
```
iHiD, line 49: Does this do anything?

app/javascript/packs/application.tsx
```diff
@@ -17,6 +17,7 @@ import 'components/concepts-map/ConceptsMap.css'
 
 import React from 'react'
 import { initReact } from './react-bootloader.jsx'
+import { Editor } from '../components/common/Editor'
```
iHiD, line 20: I think this should probably be in `student`, not `common`, or not namespaced at all.

test/javascript/components/common/Editor.test.tsx
iHiD: What is `message` here? I don't think a submission has a message.

iHiD: Other than that it's good. We could just have a single `test_run` though, rather than `test_runs`. I think we only ever want the JS to expect a single one. The serailizer should probably just choose the latest.

app/javascript/components/common/Editor/TestRuns.tsx
iHiD: I think this should be the actual channel - I don't think we need any test channels.

I've implemented more of this in https://github.com/exercism/v3-website/pull/193 so if you rebase this onto that, you should have access to `Submission::TestRunsChannel` (note that `Submission` is singular), which can be subscribed to using the `submission.uuid` as you've done here.

I think we're both thinking the same way here 🙂

app/javascript/components/common/Editor/TestRuns.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const subscription = consumer.subscriptions.create(
+      { channel: 'Test::Submissions::TestRunsChannel', id: submission.id },
+      {
+        received: (json: any) => {
+          setSubmission({
+            id: submission.id,
+            testsStatus: json.tests_status,
+            testRuns: json.test_runs,
+            message: json.message,
+          })
```
iHiD, line 39: @kntsoriano So I think the bit you've misunderstood here (misunderstood from what I meant when explaining) is that this won't be a submission coming back, it'll be a serialized test run.

This is why I'm confused about the `message` bit in Submission as you've sort of merged two serailizations into one.

app/javascript/components/common/Editor/TestRuns.tsx
iHiD: I think this whole thing can be inlined in to the bit below. I don't think we need this separate `Content` block

app/javascript/components/common/Editor/TestRuns.tsx
```diff
@@ -0,0 +1,57 @@
...
+  useEffect(() => {
+    const subscription = consumer.subscriptions.create(
+      { channel: 'Test::Submissions::TestRunsChannel', id: submission.id },
+      {
+        received: (json: any) => {
+          setSubmission({
+            id: submission.id,
+            testsStatus: json.tests_status,
+            testRuns: json.test_runs,
+            message: json.message,
```
iHiD, line 38: So these keys in the seralized response will be:
```
id, status, message, tests
```

test/javascript/components/common/Editor.test.tsx
kntsoriano: I based this from here: https://github.com/exercism/v3-project-management/issues/52#issue-715234576. Is this outdated now?

app/javascript/components/common/Editor/CodeEditor.tsx
kntsoriano: We need IDs for labels. I think we need to resolve how we want to namespace this.

app/javascript/components/common/Editor/TestRuns.tsx
kntsoriano: Cool. I'll put it back! I decided to put it away as I was thinking of refactoring to objects that render themselves: Like calling `submission.render()` or something like that, but let's save that for later.

app/javascript/components/common/Editor/TestRuns.tsx
kntsoriano: I will remove this.

app/javascript/packs/application.tsx
kntsoriano: I'll namespace it to student for now.

app/javascript/components/common/Editor/TestRuns.tsx
kntsoriano: Just one test run? I based the JSON from here: https://github.com/exercism/v3-project-management/issues/52#issue-715234576

kntsoriano:
@iHiD Ah. I now get what a test run is! Hahaha. I was really confused. I'm going to close this for now and reopen with the change applied. I will also get this to work with https://github.com/exercism/v3-website/pull/193.

app/javascript/components/common/Editor/CodeEditor.tsx
iHiD: Ah yeah. I always forget this. so yeah, they should follow the class name thing then. So this would be `c-editor-code-textarea`

iHiD:
Great. It's hard to jump into all this cold! Keep asking questions 💙

PR #192, opened 2020-10-14
simplify bezier curve control point math

neenjaw:
Issue: when zoomed or even with varying computers there was notable difference in the apperance of the bezier curves used to draw each path. The mathematical determination of control points is simplified to remove variable inputs and create a consistent appearance.

New appearance:
new-curve

Old (problematic) appearance:
chrome-125-ubuntu


iHiD:
Lovely. This should deploy itself shortly :)

PR #194, opened 2020-10-15
Rewrite in-browser editor component

kntsoriano:
I've rewritten the editor component based on the previous comments.

I also was confused between a test run and a submission, but hopefully I understand it now.

I've held off on using the actual endpoints as we might want to agree on what the specs are.

UPDATE:
Screencast: https://www.loom.com/share/90cfa81628734908a4a6b592f0b7c8a0


app/channels/test/submission/test_runs_channel.rb
```diff
@@ -0,0 +1,14 @@
+module Test
+  module Submission
+    class TestRunsChannel < ApplicationCable::Channel
+      def subscribed
+        stream_for submission
+      end
+
+      private
+      def submission
+        @submission ||= ::Submission.find_by(uuid: params[:uuid])
```
ErikSchierboom: The `||=` pattern is used to ensure that the calculation is only done once, right? At other places we use Mandate's `memoize` support for that. It might be good to use the same pattern everywhere, or is there a benefit to this method over using Mandate?

app/helpers/view_components/student/editor.rb
```diff
@@ -0,0 +1,16 @@
+module ViewComponents
+  module Student
+    class Editor < ViewComponent
+      def initialize(endpoint)
+        @endpoint = endpoint
+      end
```
ErikSchierboom: Could this be replaced with a Mandate initialize?

test/javascript/components/student/Editor.test.js
```diff
@@ -0,0 +1,37 @@
...
+      ctx.json({ id: 2, tests_status: 'pending', test_runs: [], message: '' })
+    )
+  })
+)
+
+beforeAll(() => server.listen())
+afterEach(() => server.resetHandlers())
+afterAll(() => server.close())
+
+test('clears current submission when resubmitting', async () => {
```
ErikSchierboom, line 34: Are we testing that the current submission is cleared or that a new submission is pending?

ErikSchierboom (approved):
Nice!

app/channels/test/submission/test_runs_channel.rb
kntsoriano: Ah yes! I always forget.

app/helpers/view_components/student/editor.rb
kntsoriano: I also forget this one! Thanks.

test/javascript/components/student/Editor.test.js
kntsoriano: We're testing that the previous submission was cleared. How can we rewrite this test better?

app/helpers/view_components/student/editor.rb
```diff
@@ -0,0 +1,16 @@
...
+    class Editor < ViewComponent
+      initialize_with :solution
+
+      def to_s
+        react_component("student-editor", {
+                          endpoint: Exercism::Routes.api_solution_submissions_path(
+                            solution.uuid,
+                            auth_token: solution.user.auth_tokens.first.to_s
+                          )
+                        })
```
iHiD: Could you drop "student-editor" onto a new line, and then the `{` that follows, then run lint this file pls. Just so it's less long lines and more readable.

app/helpers/view_components/student/editor.rb
```diff
@@ -0,0 +1,19 @@
...
+      def to_s
+        react_component(
+          "student-editor",
+          {
+            endpoint: Exercism::Routes.api_solution_submissions_path(
+              solution.uuid,
+              auth_token: solution.user.auth_tokens.first.to_s
+            )
+          }
+        )
```
iHiD, line 15: Much cleaner!!

app/controllers/test/components/student/editor_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Test::Components::Student::EditorController < Test::BaseController
+  def show
+    @solution = Solution.first
```
SleeplessByte: Probably want a solution under your control, such as a fixture, or one you create here.

Otherwise, when the "test seed data" changes, this can potentially break.

app/javascript/components/student/Editor.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useState } from 'react'
+import { CodeEditor } from './editor/CodeEditor'
+import { TestRunSummary } from './editor/TestRunSummary'
+import fetch from 'isomorphic-fetch'
```
SleeplessByte: We don't SSR, so we don't need isomorphic-fetch, right?

If you just wanted this for polyfilling, I recommend maybe this file I wrote and pulled from another project:

```typescript
import 'yet-another-abortcontroller-polyfill';
import { fetch } from 'whatwg-fetch';

export { default as URLSearchParams } from '@ungap/url-search-params';

// Use native browser implementation if it supports aborting 
const abortableFetch = 'signal' in new Request('') ? window.fetch : fetch; 

// Will either be the native one, or the polyfilled one. const AbortController = window.AbortController; 

export { AbortController, abortableFetch as fetch }; 
```

And then import from here instead. You can also make your life easier and assign the exported constants to window, so that you literally can use it without imports.

app/javascript/components/student/Editor.tsx
```diff
@@ -0,0 +1,45 @@
...
+export enum TestRunStatus {
+  PASS = 'pass',
+  FAIL = 'fail',
+  ERROR = 'error',
+  PENDING = 'pending',
+  OPS_ERROR = 'ops_error',
+}
+
+export function Editor({ endpoint }: { endpoint: string }) {
+  const [submission, setSubmission] = useState<Submission>()
```
SleeplessByte, line 27: In strict typescript, this may fail, unless they changed the typing (realistically possible).

The type of submission isn't `Submission` but `Submission | undefined`.

Check if `strict` is true, and if yes, and this doesn't complain, ignore me. They changed the typing to automatically add `undefined` unless an initial value is given.

app/javascript/components/student/Editor.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  function submit(code: string) {
+    setSubmission(undefined)
+
+    fetch(endpoint, {
```
SleeplessByte: Consider making this cancellable if a new submission is started whilst the request still runs.

app/javascript/components/student/Editor.tsx
```diff
@@ -0,0 +1,45 @@
...
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
```
SleeplessByte: Add the accept header as well.

app/javascript/components/student/Editor.tsx
```diff
@@ -0,0 +1,45 @@
...
+      body: JSON.stringify({ files: { file: code } }),
+    })
+      .then((response) => response.json())
```
SleeplessByte: Consider:

1. First check if the request is in the 1xx or 2xx range:

```typescript
if (!response.ok) {
  throw response
}

// More here
```

Some people frown upon throwing non promise values, but I don't. It's a perfectly valid way to separate error cases from the happy path, without wrapping it in a custom error, which you can do. I'll leave that up to you.

2. Check the response content type to determine if you can JSONfy this:

```typescript
const contentType = response.headers.get('content-type')

if (!contentType.includes('+json') && !contentType.includes('application/json')) {
  // Unexpected content type. Optionally wrap
  // in a custom error, for example "Unexpected
  // 'content-type'. Expected ..., actual ...
  throw response
}

return response.json()
```

3. Handle the error _response_
```typescript
.catch((errorOrResponse) => {

  // Probably want to invert this statement, but
  // I am on my phone and was too much work.
  // Inverting leads to early return and less
  // nested code.
  if (errorOrResponse instanceof Response) {

    // Convert response to error. I usually check
    // the 'content-type' here and determine
    // how to deserialize the error. I then have
    // custom error classes per type:
    //
    // - StructuredError where I try to extract the
    //   error message from the JSON.
    // - TextError where I just forward the text in
    //   the response as the error message
    // - Client side 415 meaning Unsupported 
    //   MediaType. 
    //
    // I'd prefix the error message with response
    // status code and maybe statusText.
  }

   return Promise.reject(errorOrResponse)
})
```

You can look at my npm package `fetch-media` for inspiration. I recommend for now not using that and extracting this in a "roll your own" utility, because we need to explain to people how this works and otherwise it will be "abstracted away" in a package.

app/javascript/components/student/Editor.tsx
```diff
@@ -0,0 +1,45 @@
...
+      .then(({ submission }: any) => {
+        setSubmission({
```
SleeplessByte: There are a few issues with this:

1. By the time you get here, the component may be unmounted and thus this will lead to a "unmounted component tried to render" error.

You can use my npm package `use-is-mounted` or roll your own:

```typescript
function Component() {
  const isMountedRef = useIsMounted()

  // ...

  .then((...) => {
    if (!isMountedRef.current) {
      return
    }

    setSubmission(...)
  })
```

2. Race condition! You must check inside the then if the submission hasn't changed in-between. For example you could check if the code has changed.

Far easier is to keep track with a nonce/state/timestamp/whatever in a custom ref, and check if it's still the same as before you started fetch. If so, yay. Otherwise ignore.

3. Both issues above can (partially) be mitigated by cancelling the request when the component unmounts or a new request comes in. Do this by using an abortcontroller, and adding a signal to the `fetch`. When a new submissions request is made, before the fetch, cancel the existing abortcontroller. Do the same on unmount:

```typescript
useEffect(() => () => {
  // Check if there is an existing abortcontroller and cancel it
}, [])
```

app/javascript/components/student/editor/CodeEditor.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const [code, setCode] = useState('')
+
+  function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+    e.preventDefault()
+
+    onSubmit(code)
+  }
+
+  function handleChange(e: React.ChangeEvent<HTMLTextAreaElement>) {
+    setCode(e.target.value)
```
SleeplessByte: Causes a re-render on each key stroke. Use an uncontrolled component with a ref to mitigate this.

Re-renders are not bad, but it's unnecessary and this will probably be slow on low end devices because the code will be big 😁

SleeplessByte (changes requested):
Some tips and tricks.

Also: are you both sending the response to the fetch request ánd via actioncable? If yes: why? (Fine with me, just want to know)

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -0,0 +1,76 @@
...
+        },
+      }
+    )
+
+    return () => {
+      subscription.unsubscribe()
+    }
+  }, [submission.uuid])
+
+  let content
```
SleeplessByte: I'd extract this switch to a function and use return instead of assignment. Will make it easier if it gets more complex.

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -0,0 +1,76 @@
...
+  })
+
+  useEffect(() => {
+    const subscription = consumer.subscriptions.create(
+      {
+        channel: 'Submission::TestRunsChannel',
+        submission_uuid: submission.uuid,
+      },
+      {
+        received: ({ test_run: testRun }: any) => {
```
SleeplessByte: You probably want to _type_ this event as something else than `any`

SleeplessByte (commented):
And two more.

app/javascript/components/student/Editor.tsx
iHiD: @SleeplessByte Good shout.

@kntsoriano This should happen server-side. Can you add a `# TODO: ...` at https://github.com/exercism/v3-website/blob/master/app/commands/submission/create.rb#L34 for us to cancel any other submissions to the same user's solution, and an issue about this in the project-management repo, please.

SleeplessByte: This should *also happen server-side.

app/javascript/components/student/Editor.tsx
kntsoriano: We need `isomorphic-fetch` to have fetch run in the test environment.

SleeplessByte: Then I would suggest instead we mock fetch to use `isomorphic-fetch`, `cross-fetch` or `fetch-ponyfill` or something in test. `jest` has many tools to accomplish this.

kntsoriano: Good idea!

app/javascript/components/student/Editor.tsx
kntsoriano: It's set to strict in `tsconfig.json`.

app/javascript/components/student/Editor.tsx
kntsoriano: I've created an issue for this. I'll add this feature in a future PR.

Edit: I've already the client-side part of this one. I cancel the current submission if a new submission comes in.

app/javascript/components/student/Editor.tsx
kntsoriano: I went with 3 for now.

test/javascript/components/student/Editor.test.js
ErikSchierboom: Oh I was not commenting on the actual test, but only on the name of the test :) Maybe something like:

```suggestion
test('resubmitting creates pending submission', async () => {
```

or something like that

app/javascript/channels/testRunChannel.ts
```diff
@@ -0,0 +1,29 @@
...
+import { TestRun } from '../components/student/editor/TestRunSummary'
+import { typecheck } from '../utils/typecheck'
+
+export class TestRunChannel {
+  subscription: ActionCable.Channel
+
+  constructor(submission: Submission, received: (testRun: TestRun) => void) {
+    this.subscription = consumer.subscriptions.create(
+      {
+        channel: 'Submission::TestRunsChannel',
```
iHiD, line 13: Can we change this to the real channel and remove this stubbed channel pls?

app/javascript/channels/testRunChannel.ts
```diff
@@ -0,0 +1,29 @@
...
+          received(typecheck<TestRun>(formattedResponse, 'testRun'))
+        },
+      }
+    )
+  }
+
+  disconnect() {
+    this.subscription.unsubscribe()
+  }
+}
```
iHiD, line 29: Why the decision here to have a channel class as opposed to inlining like [we've done elsewhere](https://github.com/exercism/v3-website/blob/master/app/javascript/components/notifications/Icon.jsx#L15-L25)?

How do we decide which pattern to follow at different times?

package.json
```diff
@@ -6,6 +6,9 @@
     "@rails/actioncable": "^6.0.0",
     "@rails/activestorage": "^6.0.0",
     "@rails/ujs": "^6.0.0",
+    "@types/actioncable": "^5.2.3",
+    "@types/humps": "^2.0.0",
+    "@types/isomorphic-fetch": "^0.0.35",
```
iHiD: Should this be a devDependency if it's only now used for testing?

app/channels/submission/test_runs_channel.rb
```diff
@@ -1,6 +1,6 @@
 class Submission::TestRunsChannel < ApplicationCable::Channel
   def subscribed
-    submission = current_user.solutions.find_by!(uuid: params[:submission_uuid])
+    submission = current_user.submissions.find_by!(uuid: params[:submission_uuid])
     stream_for submission
   end
```
iHiD, line 5: Let's remove this file and use the real Channel.

app/javascript/utils/fetch-json.ts
```diff
@@ -0,0 +1,27 @@
+import { camelizeKeys } from 'humps'
+
+export async function fetchJSON(endpoint: string, options: any) {
```
SleeplessByte, line 3: You can type `Options` here! Check the type on `fetch` if you need a pointer.

app/javascript/utils/fetch-json.ts
```diff
@@ -0,0 +1,27 @@
...
+  const headers = {
```
SleeplessByte, line 4: In HTTP/2+, headers _should_ be downcased before they're sent. In HTTP/1, HTTP/2 and HTTP/3, headers are case-insensitive. I recommend using lowercase to be compliant (`fetch` will probably transform it for you, but then there is really no need for them to have capitals here).

app/javascript/utils/fetch-json.ts
```diff
@@ -0,0 +1,27 @@
+import { camelizeKeys } from 'humps'
```
SleeplessByte, line 1: Maybe [camelcase-keys](https://github.com/sindresorhus/camelcase-keys#readme) is a better choice given the author / maintenance-state.

app/javascript/utils/typecheck.ts
```diff
@@ -0,0 +1,14 @@
+export function typecheck<T>(json: any, key: string): T | never {
```
SleeplessByte, line 5:
```suggestion
/**
 * Checks if the json has a top-level key. If it does, types the value at
 * that top-level key as T.
 */
export function typecheck<T>(json: unknown, key: string): T | never {
```

Probably more accurate.

SleeplessByte (approved):

app/javascript/components/student/Editor.tsx
SleeplessByte: (These were not options, but 3 issues!)

app/javascript/channels/testRunChannel.ts
kntsoriano: This is not the stubbed channel. You might've been confused :)

app/javascript/channels/testRunChannel.ts
kntsoriano: I decided to do it this way as we want a separate layer to make sure that the message we receive back follows a type and that we change the keys to camel case. In the future, we might just have one channel class that does camelizing of keys, but I didn't want to generalize it yet.

app/channels/submission/test_runs_channel.rb
kntsoriano: This is the real channel.

app/javascript/utils/fetch-json.ts
kntsoriano: Where do I find this?

app/javascript/utils/fetch-json.ts
kntsoriano: I tried using this, but it doesn't work.

app/javascript/channels/testRunChannel.ts
ErikSchierboom: Probably by the word "Test" :)

app/javascript/channels/testRunChannel.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.subscription = consumer.subscriptions.create(
+      {
+        channel: 'Submission::TestRunsChannel',
+        submission_uuid: submission.uuid,
+      },
+      {
+        received: (response: any) => {
+          const formattedResponse = camelizeKeys(response)
+
+          received(typecheck<TestRun>(formattedResponse, 'testRun'))
```
ErikSchierboom: I was slightly confused by the duplicate use of `received`, once for the callback and one for the object key. It might be worth considering changing the parameter to a different name. Maybe something like `onReceived`?

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const haveTestsResolved = RESOLVED_TEST_STATUSES.includes(testRun.status)
+
+  useEffect(() => {
+    const channel = new TestRunChannel(submission, (testRun: TestRun) => {
+      dispatch({ type: 'testRun.received', payload: testRun })
+    })
+
+    return () => {
+      channel.disconnect()
+    }
```
ErikSchierboom, line 81: I think you could simplify this to:

```suggestion
    return () => channel.disconnect()
```

But feel free to ignore this.

app/javascript/channels/testRunChannel.ts
ErikSchierboom: I quite like having the handling of the channel be in a separate class, as it feels like components should not necessarily be aware of the details of how to connect to a channel. I do agree that it is inconsistent with the other code, so maybe we should also introduce classes for the other channels?

app/javascript/utils/fetch-json.ts
ErikSchierboom: If you use VS Code, you could do Ctrl+Click IIRC.

ErikSchierboom (approved):

app/javascript/channels/testRunChannel.ts
iHiD: lol. Sorry. Thanks!

app/channels/submission/test_runs_channel.rb
iHiD: :D

app/javascript/channels/testRunChannel.ts
kntsoriano: Cool. Sounds good!

app/javascript/components/student/editor/TestRunSummary.tsx
kntsoriano: I want to do this, too, but our linter says otherwise.

ErikSchierboom: Ah okay, then please ignore!

iHiD: Fix the linter? :)

app/javascript/utils/fetch-json.ts
SleeplessByte: Did you give it `{ deep: true }` as option?

app/javascript/components/student/editor/TestRunSummary.tsx
SleeplessByte: The linter is not broken. What @ErikSchierboom suggests isn't the same as what's written,

```typescript
() => { channel.disconnect() }

// equals to
type fn = (): void
```

```typescript
() => channel.disconnect()}

// equals to
type fn = (): ReturnType<typeof channel['disconnect']>
```

Hooks must return the former.

app/javascript/channels/testRunChannel.ts
SleeplessByte: (This would break with linting: no-shadow, so changing now = good idea)

kntsoriano:
@iHiD Can we merge this in now?

iHiD:
Sure! I've not read the JS, but it's got two approves, so yes :)

Thanks @kntsoriano and @ErikSchierboom/ @SleeplessByte for reviews!

PR #195, opened 2020-10-16
Show timeout status when test run channel doesn't respond in time

kntsoriano:


ErikSchierboom (approved):
Nice!

iHiD (approved):

SleeplessByte:
It _might_ start to be interesting to extract all the hardcoded endpoints in the tests only to a single file, so changing the domain or a route is isolated to a single place.

PR #196, opened 2020-10-17
[UserTrack] ConceptsMap Data Command

neenjaw:
@iHiD, 

I think this will generate all of the layout data that would be required by the react component concept map to be able to style the nodes and lines as much as we have discussed so far.  I haven't changed the component yet, because I thought it would be better to get feedback on this implementation of the data side first.


app/commands/user_track/generate_concepts_map_data.rb
```diff
@@ -0,0 +1,24 @@
...
+    def concepts_status
+      all_concepts = user_track.track.concepts.map(&:slug)
+      available_concepts = user_track.available_concepts.map(&:slug)
+      learnt_concepts = user_track.learnt_concepts.map(&:slug)
+
+      {}.
+        tap { |h| all_concepts.each { |concept| h[concept] = :locked } }.
+        tap { |h| available_concepts.each { |concept| h[concept] = :unlocked } }.
+        tap { |h| learnt_concepts.each { |concept| h[concept] = :complete } }
+    end
```
iHiD: Nice. You could also DRY this up with a helper lambda.

```ruby
{}.tap do |output|
  set_status = ->(slugs, status) { slugs.each { |slug| output[slug] = status } }

  set_status.(all_concepts, :locked)
  set_status.(available_concepts, :unlocked)
  set_status.(learnt_concepts, :available)
end
```

Another suggestion, as this is a command, that you could pull out that lambda into a method, and then call the method each time, using an instance variable. As the aim of commands is that they're only called once, and can only be accessed through that command, having that DRY lambda as a helper function makes sense.

I'd merge this as it is though - I'm nitpicking and exploring code ideas only here 🙂

app/commands/user_track/generate_concepts_map_data.rb
```diff
@@ -0,0 +1,24 @@
+class UserTrack
+  class GenerateConceptsMapData
+    include Mandate
+
+    initialize_with :user_track
+
+    def call
+      Track::DetermineConceptsLayout.(user_track.track).tap { |data| data[:status] = concepts_status }
```
iHiD: I think we probably want to do this separately at the controller level and just return the `concepts_status` data here.

We'll be caching the `Track::DetermineConceptsLayout.(user_track.track)` data in the database (or maybe in the cache) so we'll want to  do the merging elsewhere I think. 

By having them as separate data sets, we can then be conditional in the controller about whether we merge it at all, or whether we show the track without the user data in the "logged out" version.

iHiD (commented):
Solid first cut. You'll want to add a test too.

app/commands/user_track/generate_concepts_map_data.rb
neenjaw: Ok, I can simplify this to just apply the status to he concepts. Then, as you said, it can be joined in a controller

neenjaw:
~Is there way to define a setup block for just some test cases?~

Got it, thanks!

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+
+    create :user_track_learnt_concept, user_track: user_track, concept: basics
+
+    assert_equal(
+      {
+        'basics' => :complete,
+        'booleans' => :unlocked,
+        'atoms' => :locked
+      },
+      ::UserTrack::GenerateConceptsStatus.(user_track)
```
iHiD: You shouldn't need the `::` at the start of any of these.

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
+require "test_helper"
+
+class UserTrack::GenerateConceptsStatusTest < ActiveSupport::TestCase
+  test "generates concepts for empty track" do
+    track = create :track
+    user_track = create :user_track, track: track
```
iHiD:
```suggestion
    user_track = create :user_track
```

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+      {},
+      ::UserTrack::GenerateConceptsStatus.(user_track)
+    )
+  end
+
+  test "concepts without prerequisite are :unlocked" do
+    track = create :track
+    user_track = create :user_track, track: track
+
+    create :track_concept, slug: 'basics', track: track
```
iHiD: Where possible I like avoiding setting the params on factories. This means if we want to change defaults etc, it becomes easier. 

So as a general rule I try and do this in situations where I'd only use track once or twice.

```suggestion
    user_track = create :user_track
    create :track_concept, slug: 'basics', track: user_track.track
```

---

If I was using track lots, I'd either keep it as you had it (probably 50% of the time) or do:
```ruby
    user_track = create :user_track
    track = user_track.track

    ...
```

I can't articulate when I'd choose which, but I'll think on it.

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+
+    assert_equal(
+      { 'basics' => :unlocked },
+      ::UserTrack::GenerateConceptsStatus.(user_track)
+    )
+  end
+
+  test "concepts following incomplete concepts are :locked" do
+    track = create :track
+    user_track = create :user_track, track: track
```
iHiD: This is a good example of where I'd definitely define `track = ...` at the top, because its used so often and is more the core model in this test. (So leave this exactly as is)

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+      ::UserTrack::GenerateConceptsStatus.(user_track)
+    )
+  end
+
+  test "concepts following complete concepts are :unlocked" do
+    track = create :track
+    user_track = create :user_track, track: track
+
+    # Set up concepts
+
```
iHiD:
```suggestion
```

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+    basics = create :track_concept, slug: 'basics', track: track
+    booleans = create :track_concept, slug: 'booleans', track: track
+    atoms = create :track_concept, slug: 'atoms', track: track
+
+    # Set up exercises
+
```
iHiD:
```suggestion
```

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+    pacman = create :concept_exercise, track: track
+    pacman.taught_concepts << booleans
+    pacman.prerequisites << basics
+
+    logger = create :concept_exercise, track: track
+    logger.taught_concepts << atoms
+    logger.prerequisites << booleans
+
+    # learn concepts
+
```
iHiD:
```suggestion
    # Simulate learning a concept
```

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+      ::UserTrack::GenerateConceptsStatus.(user_track)
+    )
+  end
+
+  test "concepts with multiple first-level pre-reqs are locked" do
+    track = create :track
+    user_track = create :user_track, track: track
+
+    # Set up concepts
+
```
iHiD:
```suggestion
```

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+    basics = create :track_concept, slug: 'basics', track: track
+    booleans = create :track_concept, slug: 'booleans', track: track
+    atoms = create :track_concept, slug: 'atoms', track: track
+
+    # Set up exercises
+
```
iHiD:
```suggestion
```

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+    pacman = create :concept_exercise, track: track
+    pacman.taught_concepts << booleans
+
+    logger = create :concept_exercise, track: track
+    logger.taught_concepts << atoms
+    logger.prerequisites << basics
+    logger.prerequisites << booleans
+
+    # Before any are completed
+
```
iHiD:
```suggestion
```

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+      ::UserTrack::GenerateConceptsStatus.(user_track)
+    )
+  end
+
+  test "concepts with two prereqs are :locked unless both are complete" do
+    track = create :track
+    user_track = create :user_track, track: track
+
+    # Set up concepts
+
```
iHiD:
```suggestion
```

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+    basics = create :track_concept, slug: 'basics', track: track
+    booleans = create :track_concept, slug: 'booleans', track: track
+    atoms = create :track_concept, slug: 'atoms', track: track
+
+    # Set up exercises
+
```
iHiD:
```suggestion
```

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,201 @@
...
+    pacman = create :concept_exercise, track: track
+    pacman.taught_concepts << booleans
+
+    logger = create :concept_exercise, track: track
+    logger.taught_concepts << atoms
+    logger.prerequisites << basics
+    logger.prerequisites << booleans
+
+    # One parent is completed
+
```
iHiD:
```suggestion
```

test/commands/user_track/generate_concepts_status_test.rb
neenjaw: Ok, I saw it elsewhere in another test file so though that was the codebase style, can be removed

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,177 @@
...
+    slugs.map { |slug| create type, slug: slug, track: track }
+  end
+
+  def setup_concepts(track, *slugs)
+    setup_from_factory(track, :track_concept, slugs)
+  end
+
+  def setup_concept_exercises(track, *slugs)
+    setup_from_factory(track, :concept_exercise, slugs)
+  end
```
iHiD: I'd pop all these methods to below the tests and make them private.

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,177 @@
+require "test_helper"
+
+class UserTrack::GenerateConceptsStatusTest < ActiveSupport::TestCase
+  def setup_user_track
+    track = create :track
+    user_track = create :user_track, track: track
+
+    [track, user_track]
+  end
```
iHiD: I think for this method, I'd just have `create :user_track` and then use `user_track.track` or `track = user_track.track`. I'm not sure this one adding much.

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,177 @@
...
+
+  def setup_from_factory(track, type, slugs)
+    slugs.map { |slug| create type, slug: slug, track: track }
+  end
```
iHiD: As this method is used exclusively by `setup_concepts` and `setup_concept_exercises`, it's good practice to put it below those methods. It's good to always be able to read a file from top to bottom in terms of execution where possible (it's often impossible to get it exactly right, but working to the spirit of it keeps things clean).

test/commands/user_track/generate_concepts_status_test.rb
```diff
@@ -0,0 +1,177 @@
...
+
+  def setup_concepts(track, *slugs)
```
iHiD: Nice use of the splat.

test/commands/user_track/generate_concepts_status_test.rb
iHiD: Meh - actually. I've changed my mind on this reading it below. It's good and makes the tests really clear. Ignore my comment 🙂

test/commands/user_track/generate_concepts_status_test.rb
iHiD: The only place it should be needed is if there is a potential conflict between two namespaces. I'm not sure why it's in `app/commands/user_track/create.rb` - feel free to remove it from there too :)

The example in `config/environments/production.rb` is a good one. Rails defines it's own Logger class, so here we're saying that we want to use the Ruby one, not the Rails one (ie, not the one in our current namespace).

iHiD:
@neenjaw Do we want to get this being used in the JS in the same PR?

neenjaw:
We can commit it, and open another to get it in the codebase. That's okay by me.  Easier to review. :)

app/commands/user_track/generate_concepts_status.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    initialize_with :user_track
+
+    def call
+      concepts_status
+    end
+
+    private
+    memoize
+    def concepts_status
```
iHiD:
```suggestion
```

app/commands/user_track/generate_concepts_status.rb
```diff
@@ -0,0 +1,27 @@
+class UserTrack
+  class GenerateConceptsStatus
```
iHiD: - I prefer pluralization as `ConceptStatuses` here as the pluralized form.
- However, as we're not generating statuses, we should add something on the end and build it around that. Probably `Mapping`.

```suggestion
  class GenerateConceptStatusMapping
```

**Updated**

iHiD: I edited this ^^

iHiD:
OK, cool. Couple of final comments then let's merge :)

app/commands/user_track/generate_concepts_status.rb
neenjaw: I think you are suggesting to merge the private method into the call method, but line 8 is missing from your suggestion

neenjaw:
Comments addressed, naming made consistent across files.

app/commands/user_track/generate_concepts_status.rb
iHiD: I was. Sorry :)

iHiD:
Thank you!! 💙

PR #197, opened 2020-10-19
Add a CI check for zeitwerk

iHiD:
This checks that there aren't any name mismatches between filenames and their contents.


iHiD:
@kntsoriano This replaces the capistrano-checks gem effectively.

neenjaw (approved):

PR #205, opened 2020-10-21
<ConceptMap /> SVG, Highlighting, CSS, oh my!

### LINKED ISSUE #202: Hovering node on the concept tree should highlight ancestors path
@neenjaw We've [discussed this before](https://exercism-team.slack.com/archives/DHGLRMJ1J/p1601402302003800), but I think it lost before we merged. Rather than just showing parents, the path should highlight ancestors and their lines.

neenjaw:
Fixes https://github.com/exercism/v3-website/issues/202

This is a bigger PR comprising a new feature for the concept map and a significant refactor to incorporate transition fading.

1) New feature: Highlight all ancestors

When hovering over a concept, it now highlights all ancestors required to unlock that concept and the paths taken to get there.

2) With this new feature in place it became apparent that the previous strategy of a single canvas to render all of the paths was becoming too difficult to extend and maintain.  **Why?** There was a split-second flash when paths overlapped concepts and caused a rendering gaffe. The gaffe is because the fading in of the paths was instantaneous and the concept boxes faded. To manage independent fade animations would be reinventing the wheel.  So rather than having a single canvas, now a multi-canvas approach is taken so that the fading can all be managed with CSS.  From my research (googling SO) this shouldn't be a performance issue.

The only downside is that it doesn't have fine-grain control over the layering, so at this point the paths may overlay each other unpredictably.

---

There is also a minor refactor to align ConceptState to ConceptStatus in the react component


app/javascript/components/concept-map/ConceptConnections.tsx
```diff
@@ -1,92 +1,96 @@
...
+        Number(canvas.style.borderBottomWidth)
+
+      canvas.width = width * devicePixelRatio
+      canvas.height = height * devicePixelRatio
+
+      canvas.style.width = `${width}px`
+      canvas.style.height = `${height}px`
+
+      const paths = mapToPaths(connections)
+
```
neenjaw:
```suggestion
```

app/javascript/components/concept-map/Concept.tsx
```diff
@@ -11,26 +11,21 @@ export const Concept = ({
...
-  isDimmed,
-  adjacentConcepts,
+  isInactive,
 }: IConcept & {
-  status: ConceptState
+  status: ConceptStatus
   isActive: boolean
-  isDimmed: boolean
-  adjacentConcepts: string[]
+  isInactive: boolean
```
ErikSchierboom: Can a concept be both active _and_ inactive?

ErikSchierboom (commented):
I was trying to test this locally, but I keep getting the following error (on all tracks I tested: Elixir, F#, Ruby):

```
Uncaught TypeError: Cannot read property 'start' of undefined
    at scalePath (draw-helpers.ts:175)
    at drawPath (draw-helpers.ts:25)
    at ConceptConnections.tsx:60
    at Array.forEach (<anonymous>)
    at ConceptConnections.tsx:28
    at commitHookEffectListMount (react-dom.development.js:19662)
    at commitPassiveHookEffects (react-dom.development.js:19699)
    at HTMLUnknownElement.callCallback (react-dom.development.js:191)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:240)
    at invokeGuardedCallback (react-dom.development.js:293)
```

iHiD:
> The only downside is that it doesn't have fine-grain control over the layering, so at this point the paths may overlay each other unpredictably.

Use `z-index` to fix this?

iHiD (changes requested):
I get the same issue as Erik.

neenjaw:
Interesting.... Will have to figure that out, because it worked for me....

app/javascript/components/concept-map/Concept.tsx
neenjaw: No, I'm using these flags to determine highlighting and local opacity, I can change the naming so it is more clear because there are 3 states. Neutral (when nothing highlighted), active (hovering) and inactive (dimmed)

neenjaw:
Okay!

> I was trying to test this locally, but I keep getting the following error (on all tracks I tested: Elixir, F#, Ruby):

I think I know what is happening, there is a race condition! On the first render of the components the paths and the concept bubbles are being drawn.  This doesn't work because the paths depend on the bubbles already having been drawn to calculate their position. So I need to just re-order it or fail more gracefully or pass a reference to the DOM object or some combination of those.

> Use z-index to fix this?

Yes, that is the likely solution

app/javascript/components/concept-map/Concept.tsx
ErikSchierboom: That would be great. Maybe introduce an enum for that?

ErikSchierboom (approved):
It now works on my machine!

neenjaw:
To reiterate the commit message:

Several changes have been made to the component for how it handles the
path generation and management.

Before the canvas was sized to the webpage, it is no longer. It is now
sized to just the size needed to hold the path shape.  then a CSS
transform is applied to translate the canvas to the correct location.

Second performance was the elimination of extra redrawing.  The crux of
the problem is that the concepts don't know about the paths and the
paths don't know about the concepts, but depend on the concepts being
mounted to the DOM.  So a "channel" for communicating updates is now
used to allow concepts to self-register their location and for paths to
declare their need for them without causing extra renders.

So now, on load there is one DOM paint for the elements, and one canvas
paint per canvas to show the paths.

Some concerns exist whether this approach will work in the longer run,
so an SVG approach should probably be explored.

Infinite gratitude to @iHiD and @SleeplessByte for their help

> That would be great. Maybe introduce an enum for that?

@ErikSchierboom I removed the third state, now just one boolean

app/javascript/components/concept-map/Concept.tsx
SleeplessByte: `'neutral' | 'active' | 'dimmed'`

Fine to use the booleans for now. (Still wanted to say this, even if it's outdated)

ErikSchierboom (approved):
Great! It now renders really fast on my machine.

neenjaw:
First time coding SVG in to place, so if there is anything, please comment

neenjaw:
@ ihid, when testing this locally, these [lines 80-81 of `application.tsx`](https://github.com/exercism/v3-website/blob/3571c5ffc2091a9d977b840188ec3970af223f2e/app/javascript/packs/application.tsx#L80-81) cause wepbacker to crash and not compile the javascript. 

It loads successfully when commented out

app/controllers/tracks/concepts_controller.rb
```diff
@@ -9,7 +9,9 @@ def authenticated_index
     @concept_map_data = Track::DetermineConceptMapLayout.(@track)
 
     if current_user.joined_track?(@track)
-      @concept_map_data[:status] = UserTrack::GenerateConceptStatusMapping.(@user_track)
+      @concept_map_data = Track::DetermineConceptMapLayout.(@track).merge(
+        status: UserTrack::GenerateConceptStatusMapping.(@user_track)
+      )
```
neenjaw:
```suggestion
      @concept_map_data[:status] = UserTrack::GenerateConceptStatusMapping.(@user_track)
```
I think there was a merge conflict and I did something weird here, this is to fix.

neenjaw:
@iHiD I think this is ready for you to look at when you get a chance re our discussion about moving styling to CSS rather than CSS-in-JS

app/javascript/components/concept-map/Concept.tsx
```diff
@@ -10,30 +17,37 @@ export const Concept = ({
...
+      removeVisibilityListener(setVisibility)
+    }
+  }, [slug, conceptRef])
+
   // Build the class list
   let classes = ['card']
   classes.push(`card-${status}`)
-  if (isActive) {
-    classes.push('card-active')
+  if (isInactive) {
```
iHiD: I think this should still be `isActive`. Having negatively-named variables is generally a bit of a smell. Even though you're only checking for negation, it's better to have `if(!active)` than `if(inactive)`.

app/javascript/components/concept-map/helpers/style-helpers.ts
```diff
@@ -0,0 +1,40 @@
...
+    | 'line-available'
+    | 'line-locked'
+    | 'line-dasharray-dashed'
+    | 'line-dasharray-solid'
+    | 'text-unlocked'
+    | 'text-locked'
+    | 'line-width'
+    | 'circle-radius'
+    | 'hover-opacity'
+) => computedRootStyle.getPropertyValue(`--c-concept-map-${type}`)
```
iHiD: So probably all of these can just be removed now we're just using the stylesheet.

app/javascript/components/concept-map/Concept.tsx
neenjaw: Fixed in update

app/javascript/components/concept-map/helpers/style-helpers.ts
neenjaw: Yes, those have been removed.

app/javascript/components/concept-map/PathLineSVG.tsx
```diff
@@ -1,10 +1,21 @@
...
+    const dx1 = start.x
+    const dy1 = start.y + halfDeltaY
+    const dx2 = end.x
+    const dy2 = end.y - halfDeltaY
+
+    return `M ${start.x}, ${start.y} C ${dx1}, ${dy1} ${dx2}, ${dy2} ${end.x}, ${end.y}`
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [path.start.x, path.start.y, path.end.x, path.end.y])
+
+  return <path d={computedBezier} className={`line-width-2 ${path.status}`} />
```
iHiD: I'd rather the `line-width-2` can be added via the stylesheet, rather than inlined.

I'm also not sure the `path.status` needs adding here. We can instead use the status reference that appears on the SVG itself for styling changes.

app/javascript/components/concept-map/ConnectionPathSVG.tsx
```diff
@@ -0,0 +1,109 @@
...
+      data-from={connection.from}
+      data-to={connection.to}
+    >
+      <g>
+        <PurePathLineSVG path={normalizedPath} />
+        <PurePathLineEndSVG
+          cx={normalizedPath.start.x}
+          cy={normalizedPath.start.y}
+          radius={radius}
+          status={normalizedPath.status}
```
iHiD: I don't think this needs the status passing in. We can infer styling from the top-level SVG.

app/javascript/components/concept-map/ConnectionPathSVG.tsx
```diff
@@ -0,0 +1,109 @@
...
+        />
+        <PurePathLineEndSVG
+          cx={normalizedPath.end.x}
+          cy={normalizedPath.end.y}
+          radius={radius}
+          status={normalizedPath.status}
```
iHiD: I don't think this needs the status passing in. We can infer styling from the top-level SVG.

app/javascript/components/concept-map/ConnectionPathSVG.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const normalizedPath = normalizePathToCanvasSize(path, width, height)
+
+  // Compute ClassNames
+  const existsActivePaths = activeConcepts.size > 0
+  const isActive =
+    !existsActivePaths ||
+    (activeConcepts.has(connection.from) && activeConcepts.has(connection.to))
+
+  const classNames = ['canvas', normalizedPath.status]
```
iHiD: We should probably rename this from `canvas` to `connection` to be semantically descriptive.

app/javascript/components/concept-map/ConceptMap.css
```diff
@@ -59,7 +68,54 @@
...
+
+        & .complete {
+            stroke: var(--c-concept-map-line-complete);
+            stroke-dasharray: var(--c-concept-map-line-dasharray-solid);
+        }
+
+        & .end-cap {
+            fill: var(--c-concept-map-card-background);
+            stroke-dasharray: var(--c-concept-map-line-dasharray-solid);
+        }
```
iHiD, line 106:
```suggestion
          stroke-width: 2px;
        }
```

app/javascript/components/concept-map/ConceptMap.css
```diff
@@ -59,7 +68,54 @@
...
+
+        & .line-width-2 {
```
iHiD:
```suggestion
        & path {
```

app/javascript/components/concept-map/PathLineEndSVG.tsx
```diff
@@ -0,0 +1,25 @@
...
+  cy: number
+  radius: number
+  status: ConceptPathStatus
+}): JSX.Element => {
+  return (
+    <circle
+      cx={cx}
+      cy={cy}
+      r={radius}
+      className={`line-width-2 end-cap ${status}`}
```
iHiD:
```suggestion
```

app/javascript/components/concept-map/ConceptMap.css
```diff
@@ -59,7 +68,54 @@
...
+            stroke: var(--c-concept-map-line-available);
+            stroke-dasharray: var(--c-concept-map-line-dasharray-dashed);
+        }
+
+        & .complete {
+            stroke: var(--c-concept-map-line-complete);
+            stroke-dasharray: var(--c-concept-map-line-dasharray-solid);
+        }
+
+        & .end-cap {
```
iHiD:
```suggestion
        & circle {
```

app/javascript/components/concept-map/PathLineSVG.tsx
iHiD:
```suggestion
  return <path d={computedBezier} />
```

app/javascript/components/concept-map/helpers/path-helpers.ts
```diff
@@ -1,82 +1,46 @@
...
+  return {
+    start: {
+      x: leftToRight ? radius + lineWidth : width - radius - lineWidth,
+      y: radius + lineWidth,
+    },
+    end: {
+      x: leftToRight ? width - radius - lineWidth : radius + lineWidth,
+      y: height - radius - lineWidth,
+    },
+    status: path.status,
```
iHiD:
```suggestion
```

app/javascript/components/concept-map/ConnectionPathSVG.tsx
iHiD:
```suggestion
  const classNames = ['canvas', path.status]
```

app/javascript/components/concept-map/ConnectionPathSVG.tsx
iHiD:
```suggestion
```

app/javascript/components/concept-map/ConnectionPathSVG.tsx
iHiD:
```suggestion
```

app/controllers/tracks/concepts_controller.rb
iHiD: It should be this:
```suggestion
      @concept_map_data[:status] = UserTrack::GenerateConceptStatusMapping.(@user_track)
```

iHiD: (I've merged your suggestion before, but then it got overriden, so I'm leaving it so you can fix it 🙂)

neenjaw: I feel like I've made this same change several times, I don't know where the commit keeps getting dropped :grinning:

app/javascript/components/concept-map/ConnectionPathSVG.tsx
neenjaw: True, Just need to go the next step of refactoring the stylesheet, thanks for the direction!

app/javascript/components/concept-map/ConceptMap.css
neenjaw: I didn't need it here, it's being inherited from elsewhere

app/javascript/components/concept-map/ConceptMap.css
neenjaw: Rather than styling the element, I've used another named class to prevent potential cascade rule problems

app/javascript/components/concept-map/helpers/path-helpers.ts
neenjaw: This change requires the type to change which cascades to a bunch of other changes.

Perhaps the better solution is to refactor so that the generated original path **is** the normalized path.

iHiD:
Great work. This is a big step forward! Thank you 💙 (And props to @SleeplessByte again!)

PR #207, opened 2020-10-21
User cancels submission before API accepts it

kntsoriano:
https://github.com/exercism/v3-project-management/issues/53

## Description
This PR adds a "Cancel" button to allow the user to cancel their submission while a response from the submissions API isn't received.

## Questions
1. @iHiD How does this look like from the backend? Is it safe to assume that we can just cancel the request if the API hasn't responded in time? I'm probably thinking that it isn't, so we might need to change the API call to not take too long, and just acknowledge that it has received the request and respond right away? The current API is already queueing jobs and other things before it responds. I decided to push through with this, though, as this solves the instance where the user might have internet problems.

### DIFF
--- a/app/javascript/components/student/Editor.tsx
+++ b/app/javascript/components/student/Editor.tsx
@@ -1,6 +1,7 @@
-import React, { useState, useRef, useEffect } from 'react'
+import React, { useReducer, useRef, useEffect, useCallback } from 'react'
 import { CodeEditor } from './editor/CodeEditor'
 import { TestRunSummary } from './editor/TestRunSummary'
+import { Submitting } from './editor/Submitting'
 import { fetchJSON } from '../../utils/fetch-json'
 import { typecheck } from '../../utils/typecheck'
 
@@ -17,6 +18,41 @@ export enum TestRunStatus {
   OPS_ERROR = 'ops_error',
 }
 
+type State = {
+  submission?: Submission
+  isSubmitting: boolean
+  code: string
+}
+
+export type Action =
+  | { type: 'submitting'; payload: { code: string } }
+  | { type: 'submitted'; payload: { submission: Submission } }
+  | { type: 'cancelled' }
+
+function reducer(state: State, action: Action): State {
+  switch (action.type) {
+    case 'submitting':
+      return {
+        submission: undefined,
+        code: action.payload.code,
+        isSubmitting: true,
+      }
+    case 'submitted':
+      return {
+        ...state,
+        submission: action.payload.submission,
+        isSubmitting: false,
+      }
+    case 'cancelled':
+      return {
+        ...state,
+        isSubmitting: false,
+      }
+    default:
+      return state
+  }
+}
+
 export function Editor({
   endpoint,
   timeout = 60000,
@@ -24,32 +60,62 @@ export function Editor({
   endpoint: string
   timeout?: number
 }) {
-  const [submission, setSubmission] = useState<Submission>()
+  const [{ submission, isSubmitting, code }, dispatch] = useReducer(reducer, {
+    submission: undefined,
+    isSubmitting: false,
+    code: '',
+  })
   const controllerRef = useRef(new AbortController())
+  const abortFetch = useCallback(() => {
+    controllerRef.current.abort()
+    controllerRef.current = new AbortController()
+  }, [controllerRef.current])
 
   useEffect(() => {
     return () => {
-      controllerRef.current.abort()
+      abortFetch()
     }
   }, [])
 
-  function submit(code: string) {
-    controllerRef.current.abort()
-    controllerRef.current = new AbortController()
-    setSubmission(undefined)
+  useEffect(() => {
+    if (isSubmitting) {
+      return
+    }
+
+    abortFetch()
+  }, [isSubmitting])
+
+  useEffect(() => {
+    if (code === '') {
+      return
+    }
+
+    if (isSubmitting) {
+      abortFetch()
+    }
 
     fetchJSON(endpoint, {
       method: 'POST',
       signal: controllerRef.current.signal,
       body: JSON.stringify({ files: { file: code } }),
-    }).then((json: any) => {
-      setSubmission(typecheck<Submission>(json, 'submission'))
     })
-  }
+      .then((json: any) => {
+        dispatch({
+          type: 'submitted',
+          payload: { submission: typecheck<Submission>(json, 'submission') },
+        })
+      })
+      .catch((responseOrError) => {
+        if (responseOrError.name === 'AbortError') {
+          console.log('Fetch cancelled')
+        }
+      })
+  }, [code])
 
   return (
     <div>
-      <CodeEditor onSubmit={submit} />
+      <CodeEditor dispatch={dispatch} />
+      {isSubmitting && <Submitting dispatch={dispatch} />}
       {submission && (
         <TestRunSummary submission={submission} timeout={timeout} />
       )}
--- a/app/javascript/components/student/editor/TestRunSummary.tsx
+++ b/app/javascript/components/student/editor/TestRunSummary.tsx
@@ -1,4 +1,4 @@
-import React, { useReducer, useEffect } from 'react'
+import React, { useReducer, useEffect, useRef } from 'react'
 import { Submission, TestRunStatus } from '../Editor'
 import { TestRunChannel } from '../../../channels/testRunChannel'
 
@@ -70,6 +70,7 @@ export function TestRunSummary({
     tests: [],
   })
   const haveTestsResolved = RESOLVED_TEST_STATUSES.includes(testRun.status)
+  const timer = useRef<number>()
 
   useEffect(() => {
     const channel = new TestRunChannel(submission, (testRun: TestRun) => {
@@ -86,8 +87,17 @@ export function TestRunSummary({
       return
     }
 
-    setTimeout(() => dispatch({ type: 'testRun.timeout' }), timeout)
+    timer.current = window.setTimeout(
+      () => dispatch({ type: 'testRun.timeout' }),
+      timeout
+    )
   }, [testRun.status])
+
+  useEffect(() => {
+    return () => {
+      clearTimeout(timer.current)
+    }
+  }, [])
 
   return (
     <div>
--- a/app/javascript/utils/fetch-json.ts
+++ b/app/javascript/utils/fetch-json.ts
@@ -5,23 +5,28 @@ export async function fetchJSON(endpoint: string, options: any) {
     'content-type': 'application/json',
     accept: 'application/json',
   }
-  const response = await fetch(
-    endpoint,
-    Object.assign(options, { headers: headers })
-  )
 
-  if (!response.ok) {
-    throw response
+  try {
+    const response = await fetch(
+      endpoint,
+      Object.assign(options, { headers: headers })
+    )
+
+    if (!response.ok) {
+      throw response
+    }
+
+    const contentType = response.headers.get('Content-Type')
+    if (
+      !contentType ||
+      (!contentType.includes('+json') &&
+        !contentType.includes('application/json'))
+    ) {
+      throw response
+    }
+
+    return camelizeKeys(await response.json())
+  } catch (responseOrError) {
+    return Promise.reject(responseOrError)
   }
-
-  const contentType = response.headers.get('Content-Type')
-  if (
-    !contentType ||
-    (!contentType.includes('+json') &&
-      !contentType.includes('application/json'))
-  ) {
-    throw response
-  }
-
-  return camelizeKeys(await response.json())
 }
--- a/app/javascript/components/student/editor/CodeEditor.tsx
+++ b/app/javascript/components/student/editor/CodeEditor.tsx
@@ -1,6 +1,11 @@
 import React, { useRef } from 'react'
+import { Action } from '../Editor'
 
-export function CodeEditor({ onSubmit }: { onSubmit: (code: string) => void }) {
+export function CodeEditor({
+  dispatch,
+}: {
+  dispatch: (action: Action) => void
+}) {
   const editor = useRef<HTMLTextAreaElement>(null)
 
   function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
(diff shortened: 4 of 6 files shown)


iHiD:
Could we have it so that once the submission comes back, it immediately gets cancelled. (We could keep this as an async function while continuing the "normal" codepath)?

kntsoriano:
Could you explain further?

iHiD:
So:
1. We create a request
2. The user wants to cancel the request.

We now want to tell the server the submission is cancelled. However, we don't have a submissionId so we can't cancel it (unless we add an extra corrolation token, which I'm a little meh about). So the easiest option would be to wait for the request to return, then when we get the response with the token, immediately send another request to cancel it.

So rather than just throwing away the request and imagining it never happened, we wait for it to come back and send another request to cancel it. This could just be a background thread - the UI doesn't have to wait for this to happen.

There's still a small chance that some requests get through but we drop the response from the server, but I'm ok with that tradeoff.

iHiD:
It's also worth noting that the actual API request is really fast. The only reason it's slow atm is that it does a Git Fetch from GitHub in development mode, which is very slow.

iHiD:
You know what, ignore me. I'm ok with this. The chance of someone cancelling within the 1s of starting is small, so this is fine.

iHiD:
@SleeplessByte @ErikSchierboom Could you both take a look at this one pls.

SleeplessByte (commented):
I need to respond to this. Please ping me if I haven't when y'all have discussed it and want to merge it in.

Few consistency (react related) issues I see -- I'll point them out then.

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
+  const [{ submission, isSubmitting, code }, dispatch] = useReducer(reducer, {
+    submission: undefined,
+    isSubmitting: false,
+    code: '',
+  })
   const controllerRef = useRef(new AbortController())
+  const abortFetch = useCallback(() => {
+    controllerRef.current.abort()
+    controllerRef.current = new AbortController()
+  }, [controllerRef.current])
```
SleeplessByte, line 72:
```suggestion
  }, [controllerRef])
```

This callback reaches _into_ the `controllerRef`. Its dependency should be the ref, not the current value. _Otherwise_ you will get aborts when for ANY other reason this component re-renders.

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
 
   useEffect(() => {
     return () => {
-      controllerRef.current.abort()
+      abortFetch()
     }
   }, [])
```
SleeplessByte, line 78:
```suggestion
  }, [abortFetch])
```

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
 
-  function submit(code: string) {
-    controllerRef.current.abort()
-    controllerRef.current = new AbortController()
-    setSubmission(undefined)
+  useEffect(() => {
```
SleeplessByte, line 80: I find it hard to follow why we need all the different effects. So apparently `isSubmitting` can change on its own. And then you want to abort? Why? It feels like you only want that because you want a `new AbortController`.

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
     fetchJSON(endpoint, {
       method: 'POST',
       signal: controllerRef.current.signal,
       body: JSON.stringify({ files: { file: code } }),
-    }).then((json: any) => {
-      setSubmission(typecheck<Submission>(json, 'submission'))
     })
-  }
+      .then((json: any) => {
+        dispatch({
```
SleeplessByte, line 103: Should check if still mounted. The abort controller takes care of this, but we'd really want to guard for it if the abortcontroller fails (because a browser is stupid). It's also good practice to think about it)

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
   endpoint: string
   timeout?: number
 }) {
-  const [submission, setSubmission] = useState<Submission>()
+  const [{ submission, isSubmitting, code }, dispatch] = useReducer(reducer, {
+    submission: undefined,
+    isSubmitting: false,
+    code: '',
+  })
   const controllerRef = useRef(new AbortController())
```
SleeplessByte, line 68:
```suggestion
  const controllerRef = useRef<undefined | AbortController>()
```

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
+  const abortFetch = useCallback(() => {
+    controllerRef.current.abort()
+    controllerRef.current = new AbortController()
```
SleeplessByte, line 71:
```suggestion
```

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
+  }, [isSubmitting])
+
+  useEffect(() => {
+    if (code === '') {
+      return
+    }
+
+    if (isSubmitting) {
+      abortFetch()
+    }
```
SleeplessByte, line 95:
```suggestion
    if (abortController.current) {
      abortController.current.abort()
      abortController.current = new AbortController()
    }
```

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
+        dispatch({
+          type: 'submitted',
+          payload: { submission: typecheck<Submission>(json, 'submission') },
+        })
+      })
+      .catch((responseOrError) => {
+        if (responseOrError.name === 'AbortError') {
+          console.log('Fetch cancelled')
+        }
+      })
```
SleeplessByte, line 112:
```suggestion
      }).finally(() => { abortController.current = undefined }))
```
You probably need to polyfill this -- webpack should take care of this.

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
+      .then((json: any) => {
+        dispatch({
+          type: 'submitted',
+          payload: { submission: typecheck<Submission>(json, 'submission') },
+        })
+      })
+      .catch((responseOrError) => {
+        if (responseOrError.name === 'AbortError') {
+          console.log('Fetch cancelled')
+        }
```
SleeplessByte, line 111:
```suggestion
        }
        
        throw responseOrError
```

Rethrow, or this error is lost forever.

app/javascript/components/student/Editor.tsx
```diff
@@ -17,39 +18,104 @@ export enum TestRunStatus {
...
-      setSubmission(typecheck<Submission>(json, 'submission'))
     })
-  }
+      .then((json: any) => {
+        dispatch({
+          type: 'submitted',
+          payload: { submission: typecheck<Submission>(json, 'submission') },
+        })
+      })
+      .catch((responseOrError) => {
```
SleeplessByte, line 108: Can this really be a response or error?

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -70,6 +70,7 @@ export function TestRunSummary({
     tests: [],
   })
   const haveTestsResolved = RESOLVED_TEST_STATUSES.includes(testRun.status)
+  const timer = useRef<number>()
```
SleeplessByte, line 73:
```suggestion
  const timer = useRef<number | undefined>()
```

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -86,9 +87,18 @@ export function TestRunSummary({
...
+      () => dispatch({ type: 'testRun.timeout' }),
+      timeout
+    )
   }, [testRun.status])
 
+  useEffect(() => {
+    return () => {
+      clearTimeout(timer.current)
+    }
+  }, [])
```
SleeplessByte, line 100:
```suggestion
  }, [timer])
```

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -86,9 +87,18 @@ export function TestRunSummary({
       return
     }
 
-    setTimeout(() => dispatch({ type: 'testRun.timeout' }), timeout)
+    timer.current = window.setTimeout(
+      () => dispatch({ type: 'testRun.timeout' }),
+      timeout
+    )
   }, [testRun.status])
```
SleeplessByte, line 94:
```suggestion
  }, [timer, testRun.status])
```

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -86,9 +87,18 @@ export function TestRunSummary({
       return
     }
 
-    setTimeout(() => dispatch({ type: 'testRun.timeout' }), timeout)
+    timer.current = window.setTimeout(
```
SleeplessByte, line 90:
```suggestion
    timer.current = setTimeout(
```

app/javascript/utils/fetch-json.ts
```diff
@@ -5,23 +5,28 @@ export async function fetchJSON(endpoint: string, options: any) {
...
+      !contentType ||
+      (!contentType.includes('+json') &&
+        !contentType.includes('application/json'))
+    ) {
+      throw response
+    }
+
+    return camelizeKeys(await response.json())
+  } catch (responseOrError) {
+    return Promise.reject(responseOrError)
```
SleeplessByte, line 30: Errors thrown in `async` functions are automatically turned into promise. Remove the entire `try` `catch`:

```typescript
async function hello(boom = false) {
  if (boom) {
    throw new Error('boom')
  }

  return 42
}

hello().then(console.log)
// => 42

hello().catch(console.error)
// => Error('boom')
```

SleeplessByte:
So. @kntsoriano , see the review above, but I actually have a different way to solve this.

```tsx
export function Editor({
  endpoint,
  timeout = 60000,
}: {
  endpoint: string;
  timeout?: number;
}) {
  const [submission, setSubmission] = useState<Submission>();
  const [{ submission, isSubmitting, code }, dispatch] = useReducer(reducer, {
    submission: undefined,
    isSubmitting: false,
    code: '',
  });

  const abortControllerRef = useRef<AbortController | undefined>();
  const abort = useCallback(() => {
    abortControllerRef.current?.abort();
    abortControllerRef.current = undefined;
  }, [abortControllerRef]);

  const submit = useCallback(
    (nextCode: string) => {
      if (nextCode.trim().length === 0) {
        return;
      }

      abort();
      abortControllerRef.current = new AbortController();

      dispatch({ type: 'submitting' });

      fetch(/* ... */)
        .then(() => {
          /* ... */
          dispatch({ type: 'submitted' });
        })
        .catch(() => {
          /* ... */
          dispatch({ type: 'cancelled' }); // it actually is! you're missing this
        })
        .finally(() => {
          abortControllerRef.current = undefined;
        });
    },
    [abort, abortControllerRef]
  );

  const cancel = useCallback(() => {
    abort();
    dispatch({ type: 'cancelled' });
  }, [dispatch, abort]);

  // Abort ongoing requests when this component unmounts
  useEffect(() => {
    return abort;
  }, [abort])
  
  // pass down cancel and submit
  return (
    <div>
      <CodeEditor onSubmit={submit} onCancel={cancel} />
      {isSubmitting && <Submitting onCancel={cancel}/>}
    </div>
  )
}
```

What this does is create callbacks for the actual actions. It allows you to have side-effects _before_ you dispatch. Makes it much easier to follow the code and you don't need all the useEffects. 

[`.finally`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally) might not exist; you can add it via `tsconfig.json`, or add the line `abortControllerRef.current = undefined;` twice (once in the `then`, once in the `catch`)

iHiD:
@kntsoriano I much prefer the callbacks syntax too. I've read the React docs and they recommend the dispatch thing (so your version is "correct"), but I think it's really hard to get your head around, compared to the callbacks. And it feels really leaky to be passing the dispatch around.

So I'd rather we use the callbacks approach if you're happy with that? (cc @ErikSchierboom for discussion too!)

---

(Thank you @SleeplessByte!)

SleeplessByte:
For the timer, I would extract a custom hook:

```typescript
function useTimeout(timeout: number, stableCallback: () => void) {
  const timerRef = useRef<number | undefined>()
  const onTimeout = useCallback(() => {
    stableCallback()
    timerRef.current = undefined
  }, [stableCallback, timerRef])

  useEffect(() => {
    timerRef.current = setTimeout(stableCallback, onTimeout)
  }, [onTimeout, stableCallback])

  useEffect(() => {
    // Cancel when hook unmounts
    return () => {
      clearTimeout(timerRef.current)
    }
  }, [timerRef])
}
```

However, this hook starts "running" as soon as you mount it. You can add another variable to the hook to make it "conditional", you could add a `start` and `cancel` callback that is returned from the hook, so you can make it imperative (in your `useEffect`), but far far far easier, is to be declarative:

```tsx
function TimeoutTestRun({ timeout: number, onTimeout }: { timeout: number, onTimeout: () => void }) {
  onTimeout(timeout, onTimout)
  return null 
}

// ...
const cancel = useCallback(() => dispatch({ type: 'testRun.timeout' }), [dispatch])

return (
    <div>
      {hasTestBeenResolved ? null : <TimeoutTestRun timeout={timeout} onTimout={cancel} />}
      <p>Status: {testRun.status}</p>
      {/*...*/}
```

kntsoriano:
@iHiD @SleeplessByte Cool. I will rewrite this to use callbacks.

app/javascript/components/student/editor/TestRunSummary.tsx
kntsoriano: I get this error: Type 'Timeout' is not assignable to type 'number'.

https://stackoverflow.com/questions/45802988/typescript-use-correct-version-of-settimeout-node-vs-window/50428377

What should I do?

PR #208, opened 2020-10-21
More design improvements for tracks

iHiD:
This PR has highlighted a couple of interesting things:

1. We're using Streamline icons. These are SVG. Setting the colours of SVG icons (either at all, or changing on hover/etc) you have to render them inline (not via the  tag). This leaves us with some options: 
-  Create multiple versions and sub them in/out. I dislike this because you get an icon flash, and because it just feels daft when we're using SVG to download two identical things with one property (color) that's different.
-  Render them inline each time. I'm pretty reluctant to render the same icon inline in dozens of places.
- Render them inline once, and use the `<use>` tag: This is what I'm doing in this PR. It's fine as a temporary fix, but I don't like it, as it means rendering **all** icons on each page load (even via turbolinks as it can't be persisted in the HEAD)
- Use an SVG font. This will be the solution we end up with. Either automatically take the icons and turn them into a font (preferable as long as we work out some nice caching strategy in dev so it's not slow to reload pages), or just manually build a font seperately and use it within the website.

2. The font-weight issue that I've mentioned in the comments is something that we'll need to fix in a few places later. The problem is that if a box is bound by the size of its text, and the font-weight changes, the boxes change size. Classic problem. Setting fixed widths on boxes solves it, but if anyone has any thoughts on a more elegant approach I'd be happy to hear it :)

---

Screenshot 2020-10-21 at 17 54 50
Screenshot 2020-10-21 at 17 54 58
Screenshot 2020-10-22 at 00 11 17


app/controllers/tracks/concepts_controller.rb
```diff
@@ -6,12 +6,13 @@ class Tracks::ConceptsController < ApplicationController
...
+    @concept_map_data = Track::DetermineConceptMapLayout.(@track)
+
     if current_user.joined_track?(@track)
-      @concept_map_data = Track::DetermineConceptMapLayout.(@track).merge(
-        status: UserTrack::GenerateConceptStatusMapping.(@user_track)
-      )
+      @concept_map_data[:status] = UserTrack::GenerateConceptStatusMapping.(@user_track)
       render action: "index/joined"
     else
+      @concept_map_data[:status] = {}
```
iHiD, line 15: @neenjaw Just an FYI to this change in this PR.

app/helpers/view_components/view_component.rb
```diff
@@ -1,6 +1,14 @@
...
     include ActionView::Helpers::TagHelper
+    include ActionView::Helpers::UrlHelper
+    include IconsHelper
+
+    # TODO: This can probably be removed
+    include ActionView::Helpers::AssetTagHelper
+
+    include ActionView::Context
+
     extend Mandate::Memoize
```
iHiD, line 12: I think it would be worth splitting this into two classes `ReactComponent` and `RenderedComponent`, as they require different `includes` etc. I'll do this in another PR.

app/helpers/tmp/icons_helper.rb
```diff
@@ -3,18 +3,6 @@ module Tmp::IconsHelper
   # rubocop:disable Rails/OutputSafety
   def reputation_icon(color:)
     '
-    <svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
```
kntsoriano, line 6: Can this whole file be deleted now? Or is it a placeholder for something?

app/helpers/view_components/student/track_nav.rb
```diff
@@ -0,0 +1,61 @@
+module ViewComponents
```
kntsoriano, line 1: What is the advantage of using something like this compared to a view partial?

kntsoriano (approved):

app/javascript/css/components/track-nav.css
```diff
@@ -0,0 +1,62 @@
...
+
+    & .container {
+        @apply flex items-center;
+    }
+
+    & .back {
+        @apply block mr-24;
+        line-height: var(--nav-height);
+        & svg {
+            height: 21px;
```
taiyab, line 16: So I've tidied up this header to remove all the abnormal sizes. It'll effect the margins between various objects and how they're displayed.

We also have something called an iconButton (rename it as you wish), which is an icon just wrapped with a clickable padded area around it. I did this so we can add mouseover effects to those things in the future (sort of material design-esque).

app/helpers/tmp/icons_helper.rb
iHiD: I failed to push my latest version where I had done this! Thanks.

ErikSchierboom (approved):
> Use an SVG font. This will be the solution we end up with. Either automatically take the icons and turn them into a font (preferable as long as we work out some nice caching strategy in dev so it's not slow to reload pages), or just manually build a font seperately and use it within the website.

This definitely seems like the best option. One thing to consider is that this (having a separate) also allows others to use the icons we have. This could for example be very useful when we'll be working on e.g. a VS Code extension.

app/helpers/view_components/student/track_nav.rb
iHiD: So:
- I think there is value in having a component that determines logical parts of things (for example here it works out the tab classes, but will also add more details to the tabs (e.g. "concepts" will change to "10/100 concepts"). It also means that can be easily tested.
- However, the `to_s` method itself could be rendering a partial instead, with data, or locals, passed into it. A little like we do for Kaido Reports. I actually started by doing that then rolled back to this, just for the sake of simplicity. But I imagine as I do a few more they can turn into partials.

taiyab (approved):

app/javascript/css/components/track-nav.css
iHiD: Cool. So that's basically what this is (where the `.back` is the button). So we just need to understand whether these IconButtons have a consistent style I guess?

PR #209, opened 2020-10-22
Cancel after submission

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/53.

## Description
This allows the user to cancel a submission after it has been queued. To do this, we send an API request to cancel the submission. When the API request responds, we unsubscribe from the test run channel.


app/serializers/serialize_submission.rb
```diff
@@ -6,7 +6,15 @@ class SerializeSubmission
   def call
     {
       uuid: submission.uuid,
-      tests_status: submission.tests_status
+      tests_status: submission.tests_status,
+      links: {
```
kntsoriano, line 10: @iHiD want to point you towards this. I've added the link to cancel in the submission serializer, rather than constructing the URL in the client side to make the front end and back end less coupled. Let me know what you think.

app/javascript/components/student/Editor.tsx
```diff
@@ -78,44 +93,40 @@ export function Editor({
...
+            dispatch({
+              type: EditorStatus.SUBMITTED,
+              payload: {
+                submission: typecheck<Submission>(json, 'submission'),
+              },
+            })
+          })
+          .catch((responseOrError) => {
+            if (responseOrError.name === 'AbortError') {
+              console.log('Fetch cancelled')
```
ErikSchierboom, line 119: Is this temporary or do you expect this to be in the actual product?

app/javascript/components/student/Editor.tsx
```diff
@@ -1,52 +1,67 @@
...
+  status?: EditorStatus
 }
 
 export type Action =
-  | { type: 'submitting'; payload: { code: string } }
-  | { type: 'submitted'; payload: { submission: Submission } }
-  | { type: 'cancelled' }
+  | { type: EditorStatus.SUBMITTING; payload: { code: string } }
+  | { type: EditorStatus.SUBMITTED; payload: { submission: Submission } }
+  | { type: EditorStatus.SUBMISSION_CANCELLED }
```
ErikSchierboom, line 44: I like this definition! Really clear

app/javascript/components/student/Editor.tsx
```diff
@@ -1,52 +1,67 @@
...
   OPS_ERROR = 'ops_error',
+  CANCELLING = 'cancelling',
+  CANCELLED = 'cancelled',
+  TIMEOUT = 'timeout',
+}
+
+export enum EditorStatus {
+  SUBMITTING = 'submitting',
+  SUBMITTED = 'submitted',
+  SUBMISSION_CANCELLED = 'submissionCancelled',
```
ErikSchierboom, line 32: Good idea to introduce an enum.

app/javascript/components/student/editor/EditorStatusSummary.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { Action, EditorStatus } from '../Editor'
+
+export function EditorStatusSummary({
+  status,
+  dispatch,
+}: {
+  status?: EditorStatus
+  dispatch: (action: Action) => void
+}) {
+  switch (status) {
```
ErikSchierboom, line 11: I assume you're expecting more statuses to handle besided the one that is now being used? If not, this could also be an if-statement.

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -57,40 +49,51 @@ export function TestRunSummary({
...
+          setTestRun({ ...testRun, status: TestRunStatus.CANCELLED })
+        })
+
+        break
+      case TestRunStatus.CANCELLED:
+        if (channel.current) {
+          channel.current.disconnect()
+        }
+
+        break
```
ErikSchierboom, line 95: Just a small suggestion: it might help readability to introduce some helper functions here. That way you could do something like:

```
switch (testRun.status) {
  case TestRunStatus.QUEUED:
    handleQueued()
    break
  case TestRunStatus.CANCELLING:
    handleCancelling()
    break
  case TestRunStatus.CANCELLED:
    handleCancelled()
    break
}
```

But feel free to ignore this if you think this does not help.

app/serializers/serialize_submission.rb
ErikSchierboom: I like that, as it will be helpful when we'll be building external integrations (like IDE plugins).

ErikSchierboom (approved):

app/javascript/components/student/editor/TestRunSummary.tsx
iHiD: I like this suggestion a lot.

app/serializers/serialize_submission.rb
iHiD: Yep - I'm happy with this too.

kntsoriano:
I will close this because we will be using callbacks now.

PR #210, opened 2020-10-23
[WIP] Move CSS outside of JS tree for my sanity

iHiD:
It's driving me mad having:
- css nested inside app/javascript (three directories to be in the `root` is too many for me)
- Having CSS files deep inside JS files. I can get like 7 `/` deep to get to a CSS file now, and it's really slowing me down.

So css is going to live in `app/css` and we're not going to nest css within the JS components. Because I can't cope this way.

---

This PR also splits up the TrackNav thing I started yesterday. I wasn't intending to commit this yet, so I'll finish it before merging, but wanted to alert y'all to the CSS thing now 🙂 

@neenjaw I've left the concept map where it is to avoid merge conflicts, but when you're ready, could you PR that to `app/css/components/concept-map.css` please 🙂


ErikSchierboom:
> css nested inside app/javascript (three directories to be in the root is too many for me)
> Having CSS files deep inside JS files. I can get like 7 / deep to get to a CSS file now, and it's really slowing me down.

What I personally liked about the current setup is that the CSS is intrinsically linked to the JavaScript. That means that when I'm working on the JavaScript component, it is very easy to locate its CSS in my editor. This may not hold for Vi but it was always really helpful to me.

That said though, as you'll be the one doing the bulk of the CSS, do whatever you prefer :)

iHiD:
> What I personally liked about the current setup is that the CSS is intrinsically linked to the JavaScript. That means that when I'm working on the JavaScript component, it is very easy to locate its CSS in my editor. 

Yeah - so I liked this in theory too. But as 9x% of our JS is going to be not component-related, it means most of our JS is in one place, then these exceptions for components are in another. Which just feels confusing to me.

> That said though, as you'll be the one doing the bulk of the CSS, do whatever you prefer :)

Thanks :)

app/helpers/images_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def rounded_bg_link(src, href, alt, css_class)
+    link_to(
+      "",
+      href,
+      alt: alt,
+      class: ['c-rounded-bg-img', css_class].compact.join(" "),
+      style: "background-image:url(#{src})"
+    )
+  end
```
iHiD:
```suggestion
  def rounded_bg_link(src, href, alt, css_class)
    link_to(
      href,
      class: ['c-rounded-bg-img', css_class].compact.join(" "),
      style: "background-image:url(#{src})"
    ) do
      image_tag(src, alt: alt, class: "sr-only")
    end
  end
```

app/helpers/images_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+    link_to(
+      href,
+      class: ['c-rounded-bg-img', css_class].compact.join(" "),
+      style: "background-image:url(#{src})",
+      "aria-title": title
+    ) do
+      image_tag(src, alt: alt, class: "tw-sr-only")
+    end
+  end
+end
```
iHiD, line 25: @SleeplessByte Please give this a check.

app/helpers/images_helper.rb
```diff
@@ -0,0 +1,26 @@
+module ImagesHelper
+  def bg_img(src, alt, css_class: 'c-bg-img', extra_css_class: nil)
+    tag.div(
+      role: "img",
```
SleeplessByte: You don't need this role, because you embed the image inside.

app/helpers/images_helper.rb
SleeplessByte: Looks good. One comment, see inline.

You might want to make them keyword args so you don't need to lookup the order all the time. But that's just me :)

app/helpers/images_helper.rb
SleeplessByte: still this :P

app/helpers/icons_helper.rb
```diff
@@ -1,15 +1,20 @@
 module IconsHelper
   def graphical_icon(icon)
-    icon(icon, nil)
+    icon(icon, nil, role: :presentation)
```
SleeplessByte, line 3: You could consider forcing the role if no alt if given. (no alt isn't the same as an empty alt).

SleeplessByte (approved):

app/helpers/images_helper.rb
iHiD: Yeah - good suggestion. I'll see how much it pisses me off looking it up :)

PR #214, opened 2020-10-25
Cancel before submission via callbacks

kntsoriano:
https://github.com/exercism/v3-project-management/issues/53


app/javascript/components/student/Editor.tsx
```diff
@@ -24,32 +66,61 @@ export function Editor({
...
+        body: JSON.stringify({ files: { file: code } }),
+      })
+        .then((json: any) => {
+          dispatch({
+            type: EditorStatus.SUBMITTED,
+            payload: { submission: typecheck<Submission>(json, 'submission') },
+          })
+        })
+        .catch((err) => {
+          if (err.name === 'AbortError' && controllerRef.current) {
```
SleeplessByte, line 99: Why must the controllerRef be checked? What's the thinking here? 

If the idea is "well on unmount we abort, and then therefore we unset it", it's probably nicer for readers to have a _dedicated_ ref that tracks this. If we now refactor `abort`, this will break.

app/javascript/components/student/editor/Submitting.tsx
```diff
@@ -0,0 +1,18 @@
...
+      e.preventDefault()
+      onCancel()
+    },
+    [onCancel]
+  )
+
+  return (
+    <div>
+      <p>Submitting...</p>
+      <button onClick={handleCancel}>Cancel</button>
```
SleeplessByte: If you add `type="button"`, you don't need to "preventDefault".

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -57,37 +48,59 @@ export function TestRunSummary({
...
+        handleQueued()
+        break
+      case TestRunStatus.TIMEOUT:
+        handleTimeout()
+        break
+      default:
+        clearTimeout(timer.current)
+        break
+    }
+  }, [testRun.status])
```
SleeplessByte:
```suggestion
  }, [testRun.status, handleQueued, handleTimeout, timer])
```

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -57,37 +48,59 @@ export function TestRunSummary({
...
+    }, timeout)
+  }, [timer])
+  const handleTimeout = useCallback(() => {
+    channel.current?.disconnect()
+  }, [channel])
+
+  useEffect(() => {
+    switch (testRun.status) {
+      case TestRunStatus.QUEUED:
+        handleQueued()
```
SleeplessByte, line 72: Is there a reason these are not just inlined here?

app/javascript/components/student/Editor.tsx
kntsoriano: You shouldn't dispatch the `cancelled` event if there is an existing fetch going on. This happens when submit is clicked twice in a row.

app/javascript/components/student/editor/TestRunSummary.tsx
kntsoriano: @iHiD and @ErikSchierboom suggested this. I don't have any preference. I'd rather you talk it out then I will apply the changes.

app/javascript/components/student/editor/Submitting.tsx
kntsoriano: This is cool!

iHiD:
@kntsoriano Awesome. Thank you :) Going to merge. If there are follow up things to work on, they can be done seperately. Thanks to DJ and Erik for the reviews that led us to here too :)

app/javascript/components/student/editor/TestRunSummary.tsx
iHiD: The level of complexity here before made this method just pretty hard to read. Might be easier for more experienced React devs, but both Erik and I found it too deep to be able to easily follow.

PR #218, opened 2020-10-27
Add monaco editor

kntsoriano:
## Description
This adds the [Monaco editor](https://microsoft.github.io/monaco-editor/) to our in browser editor component.

It took a bit long to get this working, and I needed to do the following in order to make it work:
1. Remove our react fast refresh plugin. I spent time fixing it, but decided to just move on as I couldn't really understand what the error is. We can add it back if needed in the future.
2. Create a Jest mock for the Monaco editor. It's been well documented that Monaco editor doesn't work well with Jest. I just created a mock which approximates what https://github.com/react-monaco-editor/react-monaco-editor provides. (Sidenote: I decided to use this react plugin instead of the original package as I've inspected the plugin's code and it seems to be what we will write in the future as well). We can fix this in the future, but I'd like to continue on work as we have a deadline for Halloween.


test/system/components/student/editor_test.rb
```diff
@@ -95,7 +91,6 @@ class EditorTest < ApplicationSystemTestCase
         solution = create :concept_solution, user: user
 
         visit test_components_student_editor_path(solution_id: solution.id)
-        fill_in "Code", with: "Test"
```
iHiD, line 98: Why do all these get removed?

iHiD (approved):

app/javascript/components/student/editor/CodeEditor.tsx
```diff
@@ -1,23 +1,35 @@
...
+  const editorRef = useRef<monacoEditor.editor.IStandaloneCodeEditor>()
 
-  function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+  function handleSubmit(e: React.MouseEvent<HTMLButtonElement>) {
     e.preventDefault()
 
-    if (editor.current === null) {
-      return
-    }
+    onSubmit(editorRef.current?.getValue() || '')
```
ErikSchierboom, line 11: Does this mean we're allowing submitting empty code?

app/javascript/components/student/editor/CodeEditor.tsx
```diff
@@ -1,23 +1,35 @@
...
   return (
-    <form onSubmit={handleSubmit}>
-      <label htmlFor="student-code-editor-code">Code</label>
-      <textarea ref={editor} id="student-code-editor-code"></textarea>
-      <button>Submit</button>
-    </form>
+    <div>
+      <MonacoEditor
+        width="800"
+        height="600"
```
ErikSchierboom, line 25: Is it required to specify the width and height here, or can the Monaco editor also dynamically resize based on its parent's size?

test/system/components/student/editor_test.rb
ErikSchierboom: Yes, that would also have been my question :)

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "declaration": false,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
-    "lib": ["es6", "dom"],
+    "lib": ["es6", "dom", "es2018.promise"],
```
ErikSchierboom, line 7: Do we need `es2018.promise` for the Monaco Editor?

ErikSchierboom (approved):

SleeplessByte:
Haha. I thought @iHiD was against this?

iHiD:
Monaco? I changed my mind because it supports accessiblity better, has better highlighting, and has much more active development. But the lack of mobile touch support, and the decision to not add it in the medium-term, make me happy to shift to a better alternative if there is one. But I don't think there is yet.

SleeplessByte:
@kntsoriano You'll want to look at [my implementation](https://github.com/SleeplessByte/javascript-preview/blob/master/src/editor/Editor.tsx). It's "outdated" but should still work. It's instantiated [here](https://github.com/SleeplessByte/javascript-preview/blob/master/src/play/PlayExercise.tsx#L133-L146). Check it out / Demo it [here](https://exercism-playground.derk-jan.com/javascript/play/concept/basics).

SleeplessByte:
(**please** copy my support for `global.d.ts` so I can make the lives of JS/TS users a gazillion times better. talk to me when you want to do this).

app/javascript/components/student/editor/CodeEditor.tsx
SleeplessByte: (I think we should)

test/system/components/student/editor_test.rb
kntsoriano: I didn't want to try and write JS for capybara to write into Monaco editor. Our tests right now work without this, so I'm leaving it out for now.

app/javascript/components/student/editor/CodeEditor.tsx
kntsoriano: There's an empty string check before we submit so this is fine. I wrote it like this as the `onSubmit` function expects a string.

I agree with @SleeplessByte, though. We should allow empty code :)

app/javascript/components/student/editor/CodeEditor.tsx
kntsoriano: That I don't know yet. I will have to check.

tsconfig.json
kntsoriano: For some reason, when we remove the `react-refresh` library, we need this now. It allows us to use `finally`.

kntsoriano:
@SleeplessByte Thanks! Your implementation is really helpful :) I'll look into that whenever I'm stuck. I'll make sure to talk to you about `global.d.ts` too.

app/javascript/components/student/editor/CodeEditor.tsx
SleeplessByte: In my code!

tsconfig.json
SleeplessByte: Yes. Finally is not part of es2015 promises

app/javascript/components/student/editor/CodeEditor.tsx
iHiD: > We should allow empty code :)
> (I think we should)

Why?

SleeplessByte: Because there is no difference between `" "` or `""` from a invalid code standpoint. Requiring a character to be able to see the first test case is meh. 

Alternative: show the test file.

PR #223, opened 2020-10-28
Add support for tooltips

ErikSchierboom:
This PR adds supports for using components. I originally considered using a higher-order component, but for now have tried to keep things as simple as possible. I still have to add tests, but I wanted to have the opportunity for early feedback.


app/controllers/test/components/mentoring/queue_controller.rb
```diff
@@ -15,7 +15,8 @@ def solutions
         have_mentored_previously: true,
         status: "First timer",
         updated_at: 1.year.ago.iso8601,
-        url: "https://exercism.io/solutions/1"
+        url: "https://exercism.io/solutions/1",
+        tooltip_url: student_test_components_tooltips_tooltip_path(1)
```
ErikSchierboom: As we are trying to keep our component dumb, I'm passing the tooltip URL

app/controllers/test/components/mentoring/queue_controller.rb
```diff
@@ -41,12 +43,13 @@ def solutions
...
         updated_at: 1.week.ago.iso8601,
-        url: "https://exercism.io/solutions/3"
+        url: "https://exercism.io/solutions/3",
+        tooltip_url: Exercism::Routes.student_test_components_tooltips_tooltip_path(3)
       }
     ]
 
     page = params.fetch(:page, 1).to_i
-    per = params.fetch(:per, 1).to_i
+    per = params.fetch(:per, 2).to_i
```
ErikSchierboom: This makes testing the relative positioning of the tooltips easier. I'll revert later.

app/javascript/components/mentoring/queue/Solution.jsx
```diff
@@ -14,46 +13,28 @@ export function Solution({
...
-        <div
-          ref={setPopperElement}
-          style={styles.popper}
-          {...attributes.popper}
-        >
-          Lorem ipsum
-        </div>
-      )}
-    </>
+    <tr onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
```
ErikSchierboom: We need to expose the HTML reference used in this component as that is required for the propper popper (pun intended) positioning in the parent component.

app/javascript/components/tooltips/Student.tsx
```diff
@@ -0,0 +1,63 @@
...
+export function Student({
+  endpoint,
+  styles,
+}: {
+  endpoint: string
+  styles?: React.CSSProperties
+}) {
+  const request = {
+    endpoint: endpoint,
+    options: { staleTime: 1000 * 60 },
```
ErikSchierboom: This property ensure that for 1000ms * 60 (hence: 60 seconds) there won't be no new retrieval of data for the request key. As the request key is unique per tooltip (two students have two different URLs), this means that the data is cached per student for 60 seconds.

app/javascript/packs/application.tsx
```diff
@@ -85,6 +86,9 @@ initReact({
     />
   ),
   'student-editor': (data: any) => <Student.Editor endpoint={data.endpoint} />,
+  'student-tooltip': (data: any) => (
+    <Tooltips.Student endpoint={data.endpoint} />
```
ErikSchierboom: I've tried to keep the arguments as simple as possible, just like the editor.

app/javascript/packs/react-bootloader.jsx
```diff
@@ -20,5 +21,33 @@ export const initReact = (mappings) => {
...
+        elem.insertAdjacentElement('afterend', tooltipElem)
+
+        // Link the tooltip element with the reference element
+        const popperOptions = {
+          placement: elem.dataset['placement'] || 'auto',
+        }
+        createPopper(elem, tooltipElem, popperOptions)
+
+        elem.addEventListener('mouseenter', () =>
+          render(tooltipElem, component)
```
ErikSchierboom: This renders a new React component each time the `mouseenter` event is triggered. I don't know if this is horribly inefficient though. We create the accompanying HTML element once per element. We could consider having just one tooltip element available, as I don't think we'll have overlapping tooltips.

app/javascript/components/mentoring/queue/SolutionList.jsx
```diff
@@ -12,32 +14,63 @@ export function SolutionList({ request, setPage }) {
     resolvedData,
     latestData,
   } = usePaginatedRequestQuery('mentor-solutions-list', request)
+  const [tooltipTriggerElement, setTooltipTriggerElement] = useState(null)
+  const [tooltipElement, setTooltipElement] = useState(null)
+  const [tooltipEndpoint, setTooltipEndpoint] = useState(null)
+  const { styles } = usePopper(tooltipTriggerElement, tooltipElement)
+
+  const showTooltip = (referenceElement, tooltipUrl) => {
+    setTooltipTriggerElement(referenceElement)
```
SleeplessByte: These are always set together, so they should be one state.

app/javascript/components/mentoring/queue/SolutionList.jsx
```diff
@@ -12,32 +14,63 @@ export function SolutionList({ request, setPage }) {
...
+                <th>Updated at</th>
+                <th>URL</th>
+              </tr>
+            </thead>
+            <tbody>
+              {resolvedData.results.map((solution, key) => (
+                <Solution
+                  key={key}
+                  {...solution}
+                  onMouseEnter={(e) =>
```
SleeplessByte: The shouldn't be named "onMouseEnter". It should be a callback that should display or hide more information. `Mouse` indicates "keyboard" won't work (which isn't accessible) and `Enter` indicates _when_ it should happen. It's not the `SolutionList` responsibility when these should happen.

app/javascript/components/mentoring/queue/SolutionList.jsx
```diff
@@ -12,32 +14,63 @@ export function SolutionList({ request, setPage }) {
...
+              ))}
+            </tbody>
+          </table>
+          <div ref={setTooltipElement}>
+            {tooltipEndpoint && (
+              <Tooltips.Student
+                endpoint={tooltipEndpoint}
+                styles={styles.popper}
+              />
+            )}
```
SleeplessByte: This will break when it's falsy but not `null` -- `false` works on regular React but not all other renderers.

```suggestion
            {tooltipEndpoint ? (
              <Tooltips.Student
                endpoint={tooltipEndpoint}
                styles={styles.popper}
              />
            ) : null}
```

app/javascript/components/tooltips/Student.tsx
SleeplessByte: If you need to comment it here, you probably want to comment it _there_. This is good explanation to have _in_ the source code! 

That said, I do _not_ like this local state caching. `react-query` should optimistically provide the data, and `fetch` should cache for us. Not an _extra_ cache layer right here.

app/javascript/components/tooltips/Student.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }
+  const { isLoading, isError, isSuccess, data } = useRequestQuery(
+    'student-tooltip',
+    request
+  ) as QueryResult<any>
```
SleeplessByte: The data here is `any`, but then it's destructured on line 27. That's probably a bad idea.

app/javascript/components/tooltips/index.js
```diff
@@ -0,0 +1 @@
+export { Student } from './Student'
```
SleeplessByte: I would encourage more descriptive names. What is a "student" tooltip?

app/javascript/packs/application.tsx
SleeplessByte: Passing in urls === perfect.

app/javascript/packs/react-bootloader.jsx
SleeplessByte: It's probably fine. It should not only trigger on `mousenter`. You probably want to trigger it too when the element focusses via keyboard, or just focusses in general. You might want to show it on click too. (and hide it if it's clicked again).

config/routes.rb
```diff
@@ -107,6 +107,11 @@
             end
           end
         end
+        namespace :tooltips do
+          resource :tooltip, controller: "tooltip", only: [:show] do
+            get 'student/:id', to: 'tooltip#student', as: 'student'
```
SleeplessByte: I'd personally do

```ruby
resources :students, only: %i[] do
  resource :tooltip, only: %i[show], controller: 'tooltip', action: 'student'
end
```

or


```ruby
resource :tooltip, only: %i[show], controller: 'tooltip' do
  resources :students, only: %i[show], controller: 'tooltip/students'
end
```

But I know Jeremy likes to be non-rails about all this.

iHiD: @ErikSchierboom You should entirely ignore DJ's comment here 😉

app/javascript/packs/react-bootloader.jsx
iHiD: Yeah, let's have it trigger on focus events too please, Erik.

A lot of the time these will live on links, so let's not have it on click as a general rule, but we might definitely want to add the ability for that further down the line.

config/routes.rb
SleeplessByte: I'd like to preserve my objection for the record 💯

app/javascript/packs/react-bootloader.jsx
SleeplessByte: A good way to _conditionally_ enable it on clicks would be to pass a data-attribute if it should. We can also do, semantically: `click` only if _not_ `a[href]:not([href=""])` and _not_ `button:not([action="tooltip"])`.

Would allow tooltips to trigger on links without `href` (anchors without destination), and on buttons with explicit action set.

app/javascript/components/tooltips/Student.tsx
ErikSchierboom: > `react-query` should optimistically provide the data, and fetch should cache for us. Not an extra cache layer right here.

How would we do that? By sending the appropriate cache headers from the backend?

ErikSchierboom: Scratch that. I've looked at the Rails caching options and have found this: https://github.com/exercism/v3-website/pull/223/commits/ef6ae183cf54491fd26742003b6fc1a9bdd4bb03

Seems to work well:

image

app/javascript/packs/react-bootloader.jsx
ErikSchierboom: > but we might definitely want to add the ability for that further down the line.

Let's defer this to when we actually need it then.

app/javascript/components/tooltips/index.js
ErikSchierboom: The context here is the tooltip shown for a student on the mentor's inbox/queue page.

app/javascript/components/tooltips/Student.tsx
ErikSchierboom: It is. This was also temporary, as I had a typing issue that I failed to solve. I've now pushed a new version that introduces some nice typing for the `useRequestQuery` function.

app/javascript/components/tooltips/index.js
ErikSchierboom: Some options:

- `MentoredStudent`
- `Mentoring.Student` (namespacing)

ErikSchierboom: I've gone with `MentoredStudent` for now.

SleeplessByte: Ah. Makes more sense now. 

Student details in mentoring context, but technically could be used for anything if the URL was different?

app/javascript/components/tooltips/Student.tsx
SleeplessByte: Sorry. I should have elaborated!

Rails caching instructs HTTP Cache. You can check this by doing a hard reload (cleared cache) and inspect the response. You'll see `200 ok`, and as you see in this response, `max-age=60, private`. This means that the next 60 seconds, it won't even _query_ the server.

PR #226, opened 2020-10-29
Render concept document information

ErikSchierboom:


app/models/track/concept.rb
```diff
@@ -8,4 +9,14 @@ class Track::Concept < ApplicationRecord
   scope :not_taught, lambda {
     where.not(id: Exercise::TaughtConcept.select(:track_concept_id))
   }
+
+  delegate :about, to: :data
+
+  delegate :links, to: :data
```
iHiD:
```suggestion
  delegate :about, :links, to: :data
```

app/views/tracks/concepts/show.html.haml
```diff
@@ -17,22 +17,18 @@
     %div
       .about
         %h2 About #{@concept.name}
-        %p C# structs are closely related to classes. They have state and behavior. They have constructors that take arguments, instances can be assigned, tested for equality and stored in collections.
+        %p
+          = @concept.about
```
iHiD: We want to render the mark
```suggestion
        = raw ParseMarkdown.(@concept.about)
```

app/views/tracks/concepts/show.html.haml
```diff
@@ -17,22 +17,18 @@
...
-            = link_to "#" do
-              = bg_img("https://c.s-microsoft.com/favicon.ico?v2", "Favicon of linked website")
-              Using Methods in C#
-              = icon(:external_link, "(external link)")
+          - @concept.links.each do |link|
+            %li
+              = link_to link['url'] do
+                = bg_img("https://c.s-microsoft.com/favicon.ico?v2", "Favicon of linked website")
+                = link['description']
+                = icon(:external_link, "(external link)")
```
iHiD: It would be nice if link wasn't a hash, but instead was an object. You could achieve this by mapping each links to be an openstruct.

test/test_helper.rb
```diff
@@ -58,7 +58,8 @@ def ==(other)
       # uses lots of ports on localhost for thesystem tests
       "127.0.0.1",
       "localhost:3040", "dynamodb",
-      "localhost:3041", "s3"
+      "localhost:3041", "s3",
+      "localhost:3022", "git-server"
```
iHiD: I don't think we want to require someone to have the git-server running to run the tests.

ErikSchierboom: I didn't actually need this. I initially had a bug in the stubbing of the Git Concept code. Removed.

app/views/tracks/concepts/show.html.haml
ErikSchierboom: Done

app/models/git/concept.rb
```diff
@@ -0,0 +1,36 @@
...
+    extend Mandate::Memoize
+    extend Mandate::InitializerInjector
+
+    initialize_with :track_slug, :concept_slug, :git_sha
+
+    memoize
+    def data
+      resp = RestClient.get(url_for(:data))
+      data = JSON.parse(resp.body)
+      OpenStruct.new(data['concept']).tap do |concept|
```
SleeplessByte, line 12: Jeremy might not agree (I object). But making these named ValueObject classes would have my preference.

iHiD: Yeah - I'd prefer that too normally. 

However, as I don't know if I'm going to roll all this back into the website rather than having a seperate service at all, quick and dirty works for me for now.

PR #233, opened 2020-10-30
Add Concepts Page header

iHiD:
Screenshot 2020-10-30 at 15 29 08

The progress bar needs rewriting to support two colours (learnt/mastered) and we should get the data from the existing data, not do more lookups. But it looks pretty for now.


app/views/tracks/concepts/index.html.haml
```diff
@@ -1,6 +1,16 @@
-#page-track-concepts
+#page-concepts-index
   = ViewComponents::Track::TopLevelNav.new(@track, :concepts)
 
-  .md-container
-    %h2 Concepts for #{@track.title}
+  .page-header
```
SleeplessByte, line 4: (Well aware you're just quickly running through these and later fixing it, but for the logging, `header`)

app/views/tracks/concepts/index.html.haml
```diff
@@ -1,6 +1,16 @@
...
+    .lg-container.container
+      = graphical_icon("graphic-forest", css_class: 'graphic')
+      .info
+        %h1
+          Your journey through
+          = track_icon(@track)
+        %p Learn and master concepts to forge to become fluent in #{@track.title}.
+        %progress.progress-bar{ value: @num_completed, max: @num_concepts }
+        .progress-text #{pluralize(@num_completed, 'concepts')} learnt
```
SleeplessByte, line 13: Nice. Needs to be attached to the progressbar for aria (e.g. give this an id, and progress-bar labelledby)

app/views/tracks/concepts/index.html.haml
SleeplessByte: (I'd also expect everything but the nav to be inside `main` -- probably will come)

app/views/tracks/concepts/index.html.haml
iHiD: Annoying, I can't actually use progress here in the long-term because we want multiple values on this progress bar, but noted for the one on exercises where I will do this.

SleeplessByte: You can keep using a progressbar. If you want multiple values, accessible, overlay them. Bonus, you can use `div` with `role="progressbar"` (times the amount of values you want), and then using `aria-valuenow`, `aria-valuemin`, and `aria-valuemax`.

> Note: Assistive technologies generally will render the value of `aria-valuenow` as a percent of the range between the value of `aria-valuemin` and `aria-valuemax`, unless `aria-valuetext` is specified. It is best to set the values for `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` in a manner that is appropriate for this calculation.

PR #234, opened 2020-10-30
Add site header

iHiD:
Screenshot 2020-10-30 at 16 15 44

This doesn't quite feel right, and needs some work around positioning, but it'll do for Halloween :)


app/views/layouts/_header.html.haml
```diff
@@ -0,0 +1,27 @@
+%header#site-header
+  .lg-container.container
+    = link_to root_path, class: "exercism-link" do
+      = icon "exercism-with-logo-black", "Exercism"
+
+    %ul.primary-links
```
taiyab: Is there a reason not to use a `<nav>` tag here?

taiyab (approved):

app/views/layouts/_header.html.haml
iHiD: Nope. Added it 🙂

app/views/layouts/_header.html.haml
```diff
@@ -0,0 +1,28 @@
...
+      = icon "exercism-with-logo-black", "Exercism"
+
+    %nav
+      %ul
+        %li
+          = link_to "#" do
+            = graphical_icon(:dashboard)
+            Dashboard
+
+        %li.selected
```
SleeplessByte: You'll want to add: `aria-current="page"`.

iHiD: Done. Thanks :)

PR #235, opened 2020-10-30
Redirect to iteration after submitting code

kntsoriano:


app/javascript/components/student/Editor.tsx
```diff
@@ -217,11 +210,13 @@ export function Editor({
...
-        (json: any) => {
-          dispatch({ type: ActionType.ITERATION_CREATED })
-        }
-      )
+      fetchJSON(submission.links.submit, {
+        method: 'POST',
+        body: JSON.stringify({}),
+      }).then((json: any) => {
+        const iteration = typecheck<Iteration>(json, 'iteration')
+        location.replace(iteration.links.self)
```
SleeplessByte, line 218: Thank you for using links ❤️ .

1 question: is it intended to break the back button? If no, you'd probably want to use `pushState`.

iHiD: Added as an issue here: https://github.com/exercism/v3-website/issues/238

PR #239, opened 2020-10-30
A few styling tweaks to improve the graph

iHiD:
A few small tweaks that just bring this slightly more in line with the actual designs.

Screenshot 2020-10-30 at 22 39 54

### DIFF
--- a/app/javascript/components/concept-map/CompleteIcon.tsx
+++ b/app/javascript/components/concept-map/CompleteIcon.tsx
@@ -18,6 +18,7 @@ export const CompleteIcon = ({
       xmlns="http://www.w3.org/2000/svg"
       className="complete-icon"
     >
+      <title>You have mastered this concept</title>
       <circle cx="8" cy="8" r="8" fill={fillColor} />
       <path
         d="M11.1545 5.13611L7.31292 10.6239C7.21009 10.7716 7.04251 10.8609 6.86253 10.8639C6.68254 10.8669 6.5121 10.7831 6.40445 10.6389L5.42725 9.33611"
--- a/app/javascript/components/concept-map/CompleteIcon.tsx
+++ b/app/javascript/components/concept-map/CompleteIcon.tsx
@@ -9,26 +9,22 @@ export const CompleteIcon = ({
 
   const rootStyle = getComputedStyle(document.documentElement)
   const fillColor = rootStyle.getPropertyValue('--c-concept-map-check-green')
-  const checkColor = '#000000'
-  const height = '1rem'
-  const width = '1rem'
+  const checkColor = '#26282D'
 
   return (
     <svg
-      viewBox="0 0 512 512"
-      height={height}
-      width={width}
+      viewBox="0 0 16 16"
+      fill="none"
       xmlns="http://www.w3.org/2000/svg"
       className="complete-icon"
     >
-      <title>completed</title>
+      <circle cx="8" cy="8" r="8" fill={fillColor} />
       <path
-        d="m256 0c-141.164062 0-256 114.835938-256 256s114.835938 256 256 256 256-114.835938 256-256-114.835938-256-256-256zm0 0"
-        fill={fillColor}
-      />
-      <path
-        d="m385.75 201.75-138.667969 138.664062c-4.160156 4.160157-9.621093 6.253907-15.082031 6.253907s-10.921875-2.09375-15.082031-6.253907l-69.332031-69.332031c-8.34375-8.339843-8.34375-21.824219 0-30.164062 8.339843-8.34375 21.820312-8.34375 30.164062 0l54.25 54.25 123.585938-123.582031c8.339843-8.34375 21.820312-8.34375 30.164062 0 8.339844 8.339843 8.339844 21.820312 0 30.164062zm0 0"
-        fill={checkColor}
+        d="M11.1545 5.13611L7.31292 10.6239C7.21009 10.7716 7.04251 10.8609 6.86253 10.8639C6.68254 10.8669 6.5121 10.7831 6.40445 10.6389L5.42725 9.33611"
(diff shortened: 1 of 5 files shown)


app/javascript/components/concept-map/CompleteIcon.tsx
```diff
@@ -9,26 +9,22 @@ export const CompleteIcon = ({
 
   const rootStyle = getComputedStyle(document.documentElement)
   const fillColor = rootStyle.getPropertyValue('--c-concept-map-check-green')
-  const checkColor = '#000000'
-  const height = '1rem'
-  const width = '1rem'
```
iHiD, line 14: Better to use CSS to set heights/widths

app/javascript/components/concept-map/CompleteIcon.tsx
```diff
@@ -9,26 +9,22 @@ export const CompleteIcon = ({
...
-        fill={fillColor}
-      />
-      <path
-        d="m385.75 201.75-138.667969 138.664062c-4.160156 4.160157-9.621093 6.253907-15.082031 6.253907s-10.921875-2.09375-15.082031-6.253907l-69.332031-69.332031c-8.34375-8.339843-8.34375-21.824219 0-30.164062 8.339843-8.34375 21.820312-8.34375 30.164062 0l54.25 54.25 123.585938-123.582031c8.339843-8.34375 21.820312-8.34375 30.164062 0 8.339844 8.339843 8.339844 21.820312 0 30.164062zm0 0"
-        fill={checkColor}
+        d="M11.1545 5.13611L7.31292 10.6239C7.21009 10.7716 7.04251 10.8609 6.86253 10.8639C6.68254 10.8669 6.5121 10.7831 6.40445 10.6389L5.42725 9.33611"
+        stroke={checkColor}
+        strokeWidth="1.5"
+        strokeLinecap="round"
+        strokeLinejoin="round"
```
iHiD, line 28: This is just the actual SVG Tai has used in the designs.

app/javascript/components/concept-map/ConceptMap.css
```diff
@@ -35,15 +35,11 @@
 
             & .card {
                 @apply rounded-t-8;
-                @apply box-border pointer-events-auto;
-                @apply flex flex-col;
-                justify-content: space-between;
-                margin: 42px;
+                @apply box-border pointer-events-auto shadow-base;
+                @apply flex flex-col justify-between m-40;
```
iHiD, line 39: Just using Tailwind helpers.

app/javascript/components/concept-map/ConceptMap.css
```diff
@@ -69,32 +69,25 @@
                 }
 
                 & .display {
-                    @apply pt-12 pb-8 px-24 flex items-center justify-between;
+                    @apply py-12 px-24 flex items-center justify-between;
```
iHiD, line 72: Keep the same vertical padding top and bottom and use `display:none` instead of `visiblity:hidden` further down.

app/javascript/components/concept-map/ConceptMap.css
```diff
@@ -69,32 +69,25 @@
...
-                    }
-
-                    & .status {
-                    }
-
                     & .complete-icon {
-                        @apply ml-8;
+                        @apply ml-12;
+                        width: 16px;
+                        height: 16px;
```
iHiD, line 81: This is where we add in the width/height that we removed from React.

app/javascript/components/concept-map/ConceptMap.css
```diff
@@ -69,32 +69,25 @@
...
-                    -webkit-appearance: none;
-                    appearance: none;
-
-                    width: 100%;
+                    @apply appearance-none w-100;
                     height: 4px;
 
                     &.hidden {
-                        visibility: hidden;
+                        @apply hidden;
```
iHiD, line 90: This is `display:none` (confusingly named by tailwind!)

app/javascript/components/concept-map/ConceptMap.css
neenjaw: This will throw off the alignment of the flexbox. also will make it appear as though the text is floating higher

app/javascript/components/concept-map/ConceptMap.css
neenjaw: I deviated here on purpose to maintain flex-box alignment

app/javascript/components/concept-map/ConceptMap.css
iHiD: Nice catch. I've fixed it by adding `align-items:center` in the latest commit.

Screenshot 2020-10-30 at 22 52 33

app/javascript/components/concept-map/ConceptMap.css
iHiD: Cool. Also fixed by the latest commit 🙂

app/javascript/components/concept-map/ConceptMap.css
```diff
@@ -31,19 +31,15 @@
...
 
         & .layer {
-            @apply flex justify-evenly flex-wrap;
+            @apply flex justify-evenly items-center flex-wrap;
 
             & .card {
                 @apply rounded-t-8;
-                @apply box-border pointer-events-auto;
-                @apply flex flex-col;
-                justify-content: space-between;
```
taiyab, line 40: Tailwind way of doing this is: `@apply justify-between`

app/javascript/components/concept-map/ConceptMap.css
SleeplessByte: I think it's because they use `invisible`. Good thing that `[hidden]` should be `@apply hidden`, and thus the naming is again okay 👍

PR #240, opened 2020-10-31
Use location.assign to preserve back button

### LINKED ISSUE #238: Use pushstate when redirecting from the editor
https://github.com/exercism/v3-website/pull/235#discussion_r515293510

kntsoriano:
Closes https://github.com/exercism/v3-website/issues/238.

I was a bit confused about using `pushState` as it does not create a request to the server but we want to do that. I think what was wrong was that I did `location.replace` instead of `location.assign`.


SleeplessByte:
Yes. Sorry. That's what I meant!

ErikSchierboom (approved):

PR #245, opened 2020-11-02
Pull test run data from endpoint

kntsoriano:
## Description
Since the test run data can be really big, we need to pull it from a separate endpoint. This PR does that.

### DIFF
--- a/app/controllers/api/submissions/test_runs_controller.rb
+++ b/app/controllers/api/submissions/test_runs_controller.rb
@@ -1,0 +1,23 @@
+module API
+  module Submissions
+    class TestRunsController < BaseController
+      def show
+        submission = current_user.submissions.find_by!(uuid: params[:submission_id])
+
+        test_run = if submission.tests_queued?
+                     {
+                       id: nil,
+                       submission_uuid: submission.uuid,
+                       status: :queued,
+                       tests: [],
+                       message: ""
+                     }
+                   else
+                     SerializeSubmissionTestRun.(submission.test_runs.last)
+                   end
+
+        render json: { test_run: test_run }
+      end
+    end
+  end
+end
--- a/app/controllers/api/submissions/test_runs_controller.rb
+++ b/app/controllers/api/submissions/test_runs_controller.rb
@@ -4,17 +4,7 @@ module API
       def show
         submission = current_user.submissions.find_by!(uuid: params[:submission_id])
 
-        test_run = if submission.tests_queued?
-                     {
-                       id: nil,
-                       submission_uuid: submission.uuid,
-                       status: :queued,
-                       tests: [],
-                       message: ""
-                     }
-                   else
-                     SerializeSubmissionTestRun.(submission.test_runs.last)
-                   end
+        test_run = SerializeSubmissionTestRun.(submission.test_runs.last)
 
         render json: { test_run: test_run }
       end
--- a/app/javascript/components/student/Editor.tsx
+++ b/app/javascript/components/student/Editor.tsx
@@ -17,7 +17,7 @@ export type Submission = {
   testsStatus: SubmissionTestsStatus
   uuid: string
   links: SubmissionLinks
-  testRun: TestRun
+  testRun?: TestRun
 }
 
 type APIError = {
@@ -59,6 +59,7 @@ export enum TestStatus {
 type SubmissionLinks = {
   cancel: string
   submit: string
+  testRun: string
 }
 
 export enum TestRunStatus {
@@ -118,17 +119,8 @@ function reducer(state: State, action: Action): State {
     case ActionType.SUBMISSION_CREATED:
       return {
         ...state,
+        submission: action.payload.submission,
         status: EditorStatus.SUBMISSION_CREATED,
-        submission: {
-          ...action.payload.submission,
-          testRun: {
-            id: null,
-            submissionUuid: action.payload.submission.uuid,
-            status: TestRunStatus.QUEUED,
-            tests: [],
-            message: '',
-          },
-        },
       }
     case ActionType.SUBMISSION_CANCELLED:
       return {
@@ -175,6 +167,7 @@ export function Editor({
   initialSubmission?: Submission
   files: File[]
 }) {
+  const isMountedRef = useIsMounted()
   const [{ submission, status, apiError }, dispatch] = useReducer(reducer, {
     status: undefined,
     submission: initialSubmission,
@@ -276,12 +269,16 @@ export function Editor({
   }, [dispatch, abort])
   const updateSubmission = useCallback(
     (testRun: TestRun) => {
+      if (!isMountedRef.current) {
+        return
+      }
+
       dispatch({
         type: ActionType.SUBMISSION_CHANGED,
         payload: { testRun: testRun },
       })
     },
-    [dispatch]
+    [dispatch, isMountedRef]
   )
 
   useEffect(() => {
@@ -312,10 +309,9 @@ export function Editor({
       )}
       {status === EditorStatus.CREATING_ITERATION && <p>Submitting...</p>}
       {apiError && <p>{apiError.message}</p>}
-      {submission && submission.testRun && (
+      {submission && (
         <TestRunSummary
-          cancelLink={submission.links.cancel}
-          testRun={submission.testRun}
+          submission={submission}
           timeout={timeout}
           onUpdate={updateSubmission}
         />
--- a/app/javascript/components/student/Editor.tsx
+++ b/app/javascript/components/student/Editor.tsx
@@ -182,7 +182,6 @@ export function Editor({
     status: undefined,
     submission: initialSubmission,
   })
-  const isMountedRef = useIsMounted()
   const controllerRef = useRef<AbortController | undefined>(
     new AbortController()
   )
@@ -262,14 +261,29 @@ export function Editor({
         method: 'POST',
         body: JSON.stringify({}),
         signal: controllerRef.current.signal,
-      }).then((json: any) => {
-        if (!isMountedRef.current) {
-          return
-        }
-
-        const iteration = typecheck<Iteration>(json, 'iteration')
-        location.assign(iteration.links.self)
       })
+        .then((json: any) => {
+          if (!isMountedRef.current) {
+            return
+          }
+
+          const iteration = typecheck<Iteration>(json, 'iteration')
+          location.assign(iteration.links.self)
+        })
+        .catch((err) => {
+          if (!isMountedRef.current) {
+            return
+          }
+
+          if (err instanceof Error) {
+            if (err.name === 'AbortError' && controllerRef.current) {
+              return
+            }
+          }
+        })
+        .finally(() => {
+          controllerRef.current = undefined
+        })
     },
     [controllerRef, isMountedRef, submission]
   )
@@ -300,11 +314,32 @@ export function Editor({
       return
     }
 
+    controllerRef.current = new AbortController()
+
     fetchJSON(submission.links.testRun, {
       method: 'GET',
-    }).then((json: any) => {
-      updateSubmission(typecheck<TestRun>(camelizeKeys(json), 'testRun'))
+      signal: controllerRef.current.signal,
     })
+      .then((json: any) => {
+        if (!isMountedRef.current) {
+          return
+        }
+        updateSubmission(typecheck<TestRun>(camelizeKeys(json), 'testRun'))
+      })
+      .catch((err) => {
+        if (!isMountedRef.current) {
+          return
+        }
+
+        if (err instanceof Error) {
+          if (err.name === 'AbortError' && controllerRef.current) {
+            return
+          }
+        }
+      })
+      .finally(() => {
+        controllerRef.current = undefined
+      })
   }, [])
 
   return (
--- a/app/javascript/components/student/Editor.tsx
+++ b/app/javascript/components/student/Editor.tsx
@@ -12,6 +12,7 @@ import { fetchJSON } from '../../utils/fetch-json'
 import { typecheck } from '../../utils/typecheck'
 import { Iteration } from '../track/IterationSummary'
 import { useIsMounted } from 'use-is-mounted'
+import { camelizeKeys } from 'humps'
 
 export type Submission = {
   testsStatus: SubmissionTestsStatus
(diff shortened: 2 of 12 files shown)


app/controllers/api/submissions/test_runs_controller.rb
```diff
@@ -0,0 +1,23 @@
+module API
+  module Submissions
+    class TestRunsController < BaseController
+      def show
+        submission = current_user.submissions.find_by!(uuid: params[:submission_id])
+
+        test_run = if submission.tests_queued?
```
kntsoriano: @iHiD Is it fine to place this here? Or should we add it to the serializer?

iHiD: I don't think we should mock out a fake test-run. I think if the test run doesn't exist, we should be returning a `404` with `json: {test_run: nil}` and then the React should deal with that.

I presume the React can already cope with that anyway, as it's the same as the initial state of the exercise? And the React should also be rendering something for a missing test-run than an empty test-run I'd have thought?

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -3,16 +3,16 @@ import { Submission, TestRun, TestRunStatus } from '../Editor'
 import { TestRunChannel } from '../../../channels/testRunChannel'
 import { TestRunSummaryContent } from './TestRunSummaryContent'
 import { fetchJSON } from '../../../utils/fetch-json'
+import { typecheck } from '../../../utils/typecheck'
+import { camelizeKeys } from 'humps'
 
 export function TestRunSummary({
-  testRun,
+  submission,
```
iHiD: I'm not clear on why you've changed this in this file. It feels like we've more tightly coupled things, rather than losening the coupling. Can you explain?

test/system/components/student/editor_test.rb
```diff
@@ -1,106 +1,119 @@
...
+          wait_for_submission
+          click_on "Cancel"
 
-        assert_text "Status: cancelled"
+          assert_text "Status: cancelled"
+        end
       end
 
       private
       def wait_for_submission
```
ErikSchierboom, line 115: This method name implies that it is waiting for something. With the removal of the `wait` it no longer waits. Maybe inline this method, or else add a `wait_for_cancellation` method to make the 

```
click_on "Run tests"
wait_for_submission

click_on "Cancel"
assert_text "Status: cancelled"
```

be slightly more consistent.

app/javascript/components/student/editor/TestRunSummary.tsx
ErikSchierboom: It looks like the reason for this change is the use in several places of `submission.links` and `submission.uuid`. So instead of passing in the test run and the submission properties separately, it provides just the submission. Is that correct?

ErikSchierboom (approved):

app/javascript/components/student/editor/TestRunSummary.tsx
kntsoriano: Yes. It depends on a lot of things on submission that's why I just passed submission down.

app/controllers/api/submissions/test_runs_controller.rb
kntsoriano: Cool. I will do this in React then. I was just thinking about having some kind of consistency with a "null object" which the server supplies, instead of the client just guessing what it is. It's fine though.

test/system/components/student/editor_test.rb
kntsoriano: Will do :)

app/controllers/api/submissions/test_runs_controller.rb
SleeplessByte: I _agree_ with `404`, but i'd return:

```ruby
render json: { errors: [ { message: 'Test run doesn\'t exist.' } ] }, 
           status: :not_found, 
           content_type: 'application/vnd.exercism.errors.v1+json'
```

On the _react_ side, this is a regular 404, with a *structured* error message. You can "just" use `response.json()`, like regular, but you know for a fact that it will be `errors: [ array of errors ]`.

It's probably a _good_ idea, to use `rescue_from ActiveRecord::RecordNotFound` in the `api` base controller, and rescue from with that response. You *can* use a simple gem [`erratum_responsum`](https://github.com/XPBytes/erratum_responsum), which adds all the regular api error codes as messages. This is what we use:

```ruby
  # initializers/<...>.rb  
  self.error_media_type = 'application/vnd.xpbytes.errors.v1+json'

  # api_controller.rb
  include ErratumResponsum

  rescue_from CanCan::AccessDenied, AuthorizedTransaction::TransactionUnauthorized, with: :forbidden
  rescue_from OptimisticallyStale::MissingLockVersion, ActionController::BadRequest,
              RequestMissingParam, with: :bad_request
  rescue_from ActiveRecord::RecordNotFound, with: :not_found
  rescue_from ActionController::UnknownFormat, NoAcceptSerializer, with: :not_acceptable
  rescue_from ResourceGone, with: :gone
  rescue_from ActiveRecord::StatementInvalid, ActiveRecord::RecordInvalid,
              ContentDoesNotMatchContentType, with: :unprocessable_entity
  rescue_from ActiveRecord::RecordNotUnique, ActiveRecord::StaleObjectError, with: :conflict
  rescue_from ContentTypeNotRecognised, with: :unsupported_media_type
```

Additionally, it gives you [_stable_ error codes](https://github.com/XPBytes/erratum_responsum/blob/de6590825bb74625058460819c3944dc4f2cb212/lib/erratum_responsum.rb#L82-L91) you can track:
```ruby
  # inside the gem
  
  def serialize_errors(errors)
    Array(errors).map do |error|
      { message: error.respond_to?(:message) ? error.message : error, code: error_code(error) }
    end
  end

  def error_code(error)
    return 'Ex' + String(error.send(:error_code)) if error.respond_to?(:error_code)
    'Gx' + Digest::MD5.hexdigest(error.class.name)
  end
```

app/javascript/components/student/Editor.tsx
```diff
@@ -58,6 +59,7 @@ export enum TestStatus {
 type SubmissionLinks = {
```
SleeplessByte, line 60: At some point I recommend to switch to

```typescript
{
  cancel: { href: string },
  submit: { href: string },
  testRun: { href: string }
}
```

It allows for metadata and parameters to be part of the URL value, and doesn't require adding weird types later to do this.

app/javascript/components/student/Editor.tsx
```diff
@@ -11,12 +11,13 @@ import { FileEditor, FileEditorHandle } from './editor/FileEditor'
...
 
 export type Submission = {
   testsStatus: SubmissionTestsStatus
   uuid: string
   links: SubmissionLinks
-  testRun: TestRun
+  testRun?: TestRun
 }
 
 type APIError = {
```
SleeplessByte, line 24: I recommend using the error type given above. There will be times where compound errors will be a thing:
```json
{
  "errors": [
    { "message": "Something went wrong" },
    { "message": "Second error", "param": "value", "code": "whatever" }
  ]
}
```

Nesting inside `errors` allows _request_ metadata to be present at the top-level.

app/javascript/components/student/Editor.tsx
```diff
@@ -11,12 +11,13 @@ import { FileEditor, FileEditorHandle } from './editor/FileEditor'
 import { fetchJSON } from '../../utils/fetch-json'
 import { typecheck } from '../../utils/typecheck'
 import { Iteration } from '../track/IterationSummary'
+import { useIsMounted } from 'use-is-mounted'
```
SleeplessByte, line 14: ✔️

app/javascript/components/student/Editor.tsx
```diff
@@ -11,12 +11,13 @@ import { FileEditor, FileEditorHandle } from './editor/FileEditor'
...
 
 export type Submission = {
   testsStatus: SubmissionTestsStatus
   uuid: string
   links: SubmissionLinks
-  testRun: TestRun
+  testRun?: TestRun
```
SleeplessByte, line 21: FYI, this does *not* means `testRun` can be `TestRun` or `undefined`. It means: `testRun` may not be present. Doesn't matter here, but might at some point in the future.

app/javascript/components/student/Editor.tsx
```diff
@@ -11,12 +11,13 @@ import { FileEditor, FileEditorHandle } from './editor/FileEditor'
 import { fetchJSON } from '../../utils/fetch-json'
 import { typecheck } from '../../utils/typecheck'
 import { Iteration } from '../track/IterationSummary'
+import { useIsMounted } from 'use-is-mounted'
 
 export type Submission = {
   testsStatus: SubmissionTestsStatus
   uuid: string
```
SleeplessByte, line 19: Instead of passing back the `uuid`, add ` link` to the `links` object that refers to this, e.g.
```typescript
{ self: { href: 'string' } }
```

Reasons:

- discourages building URLs.
- allows testsubmissions to live on different domains (can track this later).
- don't need an extra URL later if you want to re-request it.

You might ask: "But we already know the test submission URL, right?"

Yes, but it might redirect. This would be the canonical URL :)

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -78,14 +82,22 @@ export function TestRunSummary({
...
+  useEffect(() => {
+    channel.current = new TestRunChannel(submission, (testRun: TestRun) => {
       setTestRun(testRun)
     })
 
     return () => {
       channel.current?.disconnect()
     }
-  }, [testRun.submissionUuid])
+  }, [submission.uuid])
```
SleeplessByte: Here you would use the link.self.href instead of the uuid.

app/serializers/serialize_submission.rb
```diff
@@ -15,9 +15,12 @@ def call
           submission.solution.uuid,
           submission_id: submission.uuid,
           auth_token: user.auth_tokens.first.to_s
+        ),
+        test_run: Exercism::Routes.api_submission_test_run_url(
+          submission.uuid,
```
SleeplessByte, line 20: This is fine, because this is at serialization. Here you can build the URL and then never again.

SleeplessByte (commented):
A few pointers that will take *less* time to implement now, and make your life easier in the future. I *strongly* recommend doing these now.

app/controllers/api/submissions/test_runs_controller.rb
kntsoriano: This is nice! I'll put this in a separate issue.

test/system/components/student/editor_test.rb
kntsoriano: I thought about this again, and I think it's fine. When capybara asserts things, it waits for the result implicitly. I wouldn't also want to inline this as it takes the form of `assert_text` which sort of means like it's an assertion. But this is the only way for us to ensure that we wait for the API request to get back.

app/javascript/components/student/Editor.tsx
kntsoriano: In our websockets connection, we would need the UUID. How do you think should we deal with this?

app/javascript/components/student/Editor.tsx
kntsoriano: Cool. I'll put this in a separate issue.

app/javascript/components/student/editor/TestRunSummary.tsx
kntsoriano: I've updated the PR and I've moved things back to how they were before. Agree with making the coupling less loose.

iHiD:
(@kntsoriano I'm on Slack btw if you're around :))

app/javascript/components/student/Editor.tsx
```diff
@@ -269,6 +279,18 @@ export function Editor({
     return abort
   }, [abort])
 
+  useEffect(() => {
+    if (!submission) {
+      return
+    }
+
+    fetchJSON(submission.links.testRun, {
```
ErikSchierboom, line 319: Do we need to add the abort code used in https://github.com/exercism/v3-website/pull/249#pullrequestreview-524090289 here too?

app/javascript/components/student/editor/TestRunSummary.tsx
```diff
@@ -82,22 +78,14 @@ export function TestRunSummary({
...
-    fetchJSON(submission.links.testRun, {
-      method: 'GET',
-    }).then((json: any) => {
-      setTestRun(typecheck<TestRun>(camelizeKeys(json), 'testRun'))
-    })
-  }, [submission.uuid])
-
-  useEffect(() => {
-    channel.current = new TestRunChannel(submission, (testRun: TestRun) => {
+    channel.current = new TestRunChannel(testRun, (testRun: TestRun) => {
```
ErikSchierboom: So much cleaner!

ErikSchierboom: I do find the re-using of the `testRun` name for both the argument and the lambda parameter slightly confusing. Maybe you could use a different name for the lambda parameter?

ErikSchierboom (approved):

app/javascript/components/student/Editor.tsx
kntsoriano: Yes. I've added this now :)

app/javascript/components/student/editor/TestRunSummary.tsx
kntsoriano: Agreed. I renamed it to `updatedTestRun`.

iHiD (approved):
I'm happy to merge this from the Ruby side of things btw :)

app/javascript/components/student/Editor.tsx
kntsoriano: @SleeplessByte Merging this one now, but feel free to reply to this one.

SleeplessByte: Do you need the uuid, or the _channel_?

It might be that the channel is the uuid, but we don't want to call it uuid then. Instead, subscription channel or whatever.

test/system/components/student/editor_test.rb
SleeplessByte: That's a perfectly fine reason. Should probably document it inside at `wait_for_submission`, to explain WHY.

iHiD:
Thanks @kntsoriano and @SleeplessByte 💙

PR #246, opened 2020-11-04
Add icon and graphical-icon components

ErikSchierboom:


app/javascript/components/common/GraphicalIcon.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+export function GraphicalIcon({ icon }: { icon: string }) {
+  return (
+    <svg className="c-graphical-icon" role="img">
```
iHiD:
```suggestion
    <svg className="c-graphical-icon" role="presentation">
```

app/javascript/components/common/index.ts
```diff
@@ -1 +1,5 @@
 export { CopyToClipboardButton } from './CopyToClipboardButton'
+export { GraphicalIcon } from './GraphicalIcon'
+export { Icon } from './Icon'
+export { Loading } from './Loading'
```
iHiD, line 4: What does this LOC do?

kntsoriano (commented):
Looks good! I think we'd want the test where the react component and the view helper render the same thing.

iHiD:
@kntsoriano Yes!! Thanks for remembering. I want that too.

So @ErikSchierboom , so the output of (haml):
```haml
=icon(:foobar)
```

Should be the same as the equiv React (once React has rendered):

```
<Icon icon="foobar">
```

iHiD:
@ErikSchierboom Ignore the test for now, because it won't work for reasons I will explain to you, which https://github.com/exercism/v3-website/issues/251 will solve. So feel free to go ahead without the test, but create an issue for you to add it later pls?

app/javascript/components/common/index.ts
ErikSchierboom: It is a React convention for the `index.ts` to export all the components in the directory. You can then do the following to emulate some form of namespacing:

```jsx
import * as Common from '../components/common'

<Common.CopyToClipboardButton textToCopy={data.text_to_copy} />
```

ErikSchierboom:
@iHiD I've created an issue: https://github.com/exercism/v3-website/issues/253

app/javascript/components/common/index.ts
iHiD: Ah, cool.

iHiD:
This is red, but good to merge when green.

iHiD (approved):

PR #249, opened 2020-11-05
Clean up API requests

kntsoriano:
## Description
When we have API requests, we must do the ff:
1. Abort it when component unmounts.
2. Make sure that the component is still mounted when running side effects.

### DIFF
--- a/app/javascript/components/student/Editor.tsx
+++ b/app/javascript/components/student/Editor.tsx
@@ -11,6 +11,7 @@ import { FileEditor, FileEditorHandle } from './editor/FileEditor'
 import { fetchJSON } from '../../utils/fetch-json'
 import { typecheck } from '../../utils/typecheck'
 import { Iteration } from '../track/IterationSummary'
+import { useIsMounted } from 'use-is-mounted'
 
 export type Submission = {
   testsStatus: SubmissionTestsStatus
@@ -178,6 +179,7 @@ export function Editor({
     status: undefined,
     submission: initialSubmission,
   })
+  const isMountedRef = useIsMounted()
   const controllerRef = useRef<AbortController | undefined>(
     new AbortController()
   )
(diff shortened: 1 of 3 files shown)


app/javascript/components/student/Editor.tsx
```diff
@@ -233,25 +243,32 @@ export function Editor({
...
+        if (!isMountedRef.current) {
+          return
+        }
+
         const iteration = typecheck<Iteration>(json, 'iteration')
         location.assign(iteration.links.self)
       })
     },
-    [submission]
+    [controllerRef, isMountedRef, submission]
```
ErikSchierboom, line 271: Do we also need to add the following code here:

```
.finally(() => {
          controllerRef.current = undefined
        })
```

This is used in the above `fetchJSON` call.

app/javascript/components/student/Editor.tsx
```diff
@@ -207,12 +209,20 @@ export function Editor({
         body: JSON.stringify({ files: files }),
       })
         .then((json: any) => {
+          if (!isMountedRef.current) {
```
ErikSchierboom, line 212: If we have this pattern at multiple places (where there is a `isMountedRef ` and `AbortController`), we could consider encoding this in some helper function at some point.

ErikSchierboom (approved):

app/javascript/components/student/Editor.tsx
kntsoriano: Yes. We need to :) Thanks.

app/javascript/components/student/Editor.tsx
kntsoriano: I agree! I created https://github.com/exercism/v3-website/issues/255 for this one.

kntsoriano: @SleeplessByte Have you got this as a custom hook?

PR #250, opened 2020-11-05
Make <IterationSummary /> more accessible

kntsoriano:


test/javascript/components/track/IterationSummary.test.ts
```diff
@@ -19,7 +19,7 @@ test('shows details', async () => {
 
   expect(getByText('Iteration 2')).toBeInTheDocument()
   expect(getByTitle('Submitted via CLI')).toBeInTheDocument()
-  expect(getByRole('details')).toHaveTextContent(
+  expect(getByTestId('details')).toHaveTextContent(
```
iHiD, line 22: Can you get by classname rather than adding the extra attribute for this?

SleeplessByte: (it's a trade-off. you'll break the test by changing the classnames, or you have an extra attribute).

iHiD: (I'd rather the class name, because the class name is semantic, so if the class name changes, this bit of the test probably should too)

ErikSchierboom (approved):

test/javascript/components/track/IterationSummary.test.ts
kntsoriano: (I'm going to send my response in parentheses too)

(I'll use the class name.)

iHiD: (Great choice! Your wages will arrive as normal this month 😉)

kntsoriano: Oh. It doesn't seem we can use the class name as a selector for our library. Could we probably assign a role to this block? I'm unsure what role to use for this one.

SleeplessByte: Oh. Right. 

The reason the library doesn't allow you to do that is because a _user_ can't do that either. 

> React Testing Library aims to test the component, like how users would, users see the button, heading, form and other elements by their role, not by their id or class, or element tag name. When you use React Testing Library, you should not try to access the DOM like how you would do by document.querySelector API (which you can incidentally use in your tests, but it's not recommended for the reasons stated in this paragraph).

> **Manual Queries**
On top of the queries provided by the testing library, you can use the regular querySelector DOM API to query elements. Note that using this as an escape hatch to query by class or id is a bad practice because users can't see or identify these attributes. Use a testid if you have to.

```javascript
// @testing-library/react
const { container } = render(<MyComponent />)
const foo = container.querySelector('[data-foo="bar"]')
```

iHiD: Urgh.

I really hate it when production code gets populated with noise just for the sake of tests reading things.

Can we do this instead?

```suggestion
  expect(getByText('Submitted via CLI, a few seconds ago').toBeInTheDocument)
```

SleeplessByte: It's not just for the sake of tests reading things -- basically this means that we're testing too low-level, or lacking a good "role" (semantic) structure.

To take your comment and up it one: it's perfectly fine to check if the text is present if we don't care that much where it shows up. ANd in this case, I think you're right.

iHiD (approved):

test/javascript/components/track/IterationSummary.test.ts
kntsoriano: We can't do this as our text is broken up into different HTML elements. We can write a custom matcher, but it's arguably more complex vs just using the testid. I think the solution here is to use a correct role, but I don't know which to use. I'll merge this PR in for now.

PR #254, opened 2020-11-05
Readd Git into repo

iHiD:
This adds Git back into this repo, and removes the need for the `git-server`.


iHiD:
@kntsoriano @ErikSchierboom I don't think there's huge point in you reviewing this. Fundamentally it moves all of the `git-server` stuff into the website, continues to use Rugged for reading, but uses native `git` for cloning and pulling as Rugged can't cope with EFS.

test/system/components/tooltips/tooltip_test.rb
```diff
@@ -23,8 +23,8 @@ class TooltipTest < ApplicationSystemTestCase
         user_summaries[0].hover
         within(".c-user-summary-tooltip") { assert_text "Erik Schierboom" }
 
-        user_summaries[1].hover
-        within(".c-user-summary-tooltip") { assert_text "Rob Keim" }
+        # user_summaries[1].hover
+        # within(".c-user-summary-tooltip") { assert_text "Rob Keim" }
```
iHiD, line 27: @ErikSchierboom This test fails for me. I think the first one isn't unhovering, so the wrong element is coming up.

iHiD:
@ErikSchierboom @kntsoriano @neenjaw One important thing as a result of this. 

The `concept_exercise` and `practice_exercise` factories contain the correct setup for the git reading. So if you want to rely on the correct data being in place (e.g. for system tests that have to render things), make sure you start your test setup by defining those then getting the `track` afterwards. 

e.g.
```ruby
# DO THIS
exercise = create :concept_exercise
track = exercise.track

# NOT THIS
track = create :track
exercise = create :concept_exercise, track: track
```

The second of those would create a track that mismatches to the exercise, so the git lookup will fail.

neenjaw:
How to I refactor tests that have multiple exercises for the same track?

iHiD:
> How to I refactor tests that have multiple exercises for the same track?

With great pain and suffering.

PR #256, opened 2020-11-06
Components styling

iHiD:
This does a few things:
1. Introduces the `--` syntax for sub-classes of components. This gives us nice protection against accidentally applying unexpected styles.
2. Renames the `c-graphical-icon` to be `c-icon` as they're considered the same thing at the CSS level.
3. Inlines the `c-track` into the `c-tracks-list` as it's not used outside it.

It's mainly housekeeping and keeping things neat :)


SleeplessByte:
(There is a trade-off here. You've got very high nesting, without `>` , and this breaks composability, ánd css now determines what's a component. 🤷 I like the general idea of using `--something` to do variants tho! forces it to be on a `c-something` 👍 )

PR #257, opened 2020-11-08
Language Server Protocol POC

kntsoriano:
*Opening this PR to demonstrate a simple implementation of connecting to a Language Server via Monaco*

## Screenshot
Nov-08-2020 11-20-03

## How it works
In this example, we use the `solargraph` language server for Ruby. We'd need to have this running separately, behind this websocket proxy (https://github.com/wylieconlon/jsonrpc-ws-proxy). Browsers don't support direct TCP connections that's why we need to hide it behind a websocket proxy. I imagine that we'd do the same for the other language servers.

We then use https://github.com/TypeFox/monaco-languageclient as the language client to interpret the data from the language server.

### DIFF
--- a/app/javascript/components/student/editor/FileEditor.tsx
+++ b/app/javascript/components/student/editor/FileEditor.tsx
@@ -7,19 +7,9 @@ import React, {
   useState,
 } from 'react'
 import { File } from '../Editor'
-import MonacoEditor from 'react-monaco-editor'
+import { ExercismMonacoEditor } from './ExercismMonacoEditor'
 import * as monacoEditor from 'monaco-editor/esm/vs/editor/editor.api'
 import { useStorage } from '../../../utils/use-storage'
-import { listen, MessageConnection } from 'vscode-ws-jsonrpc'
-import {
-  MonacoLanguageClient,
-  CloseAction,
-  ErrorAction,
-  MonacoServices,
-  createConnection,
-} from 'monaco-languageclient'
-import normalizeUrl from 'normalize-url'
-import ReconnectingWebsocket from 'reconnecting-websocket'
 
 export type FileEditorHandle = {
   getFile: () => File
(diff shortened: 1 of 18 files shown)


app/javascript/components/student/editor/FileEditor.tsx
```diff
@@ -45,6 +57,41 @@ export const FileEditor = forwardRef<FileEditorHandle, FileEditorProps>(
           keybindings: [monacoEditor.KeyCode.F2],
           run: onRunTests,
         })
+
+        MonacoServices.install(editor)
+        const url = normalizeUrl('ws://localhost:7658/ruby')
+        const webSocket = new (ReconnectingWebSocket as any)(url, [], {
```
kntsoriano: The `ReconnectingWebsocket` constructor doesn't have any type definitions. Thus I just typed it to `any`.

babel.config.js
```diff
@@ -16,6 +16,7 @@ module.exports = function (api) {
   }
 
   return {
+    sourceType: 'unambiguous',
```
kntsoriano, line 19: The language client is exported as CommonJS and I needed to do this to make it work. I'm not sure if this is the right solution though.

ErikSchierboom: @SleeplessByte do you know if this a good way to solve @kntsoriano's issue?

app/javascript/utils/use-storage.ts
```diff
@@ -1,8 +1,8 @@
 import React, { useCallback } from 'react'
 import { StoredMemoryValue, useMutableMemoryValue } from 'use-memory-value'
 
-export function useStorage(key, initialValue) {
-  const memoryValue = new StoredMemoryValue(key, true, initialValue)
+export function useStorage<T>(key: string, initialValue: T) {
+  const memoryValue = new StoredMemoryValue<T>(key, true, initialValue)
```
ErikSchierboom, line 5: Nice!

app/javascript/components/student/editor/FileEditor.tsx
ErikSchierboom: I think you can use the `@types/reconnectingwebsocket` [NPM package](`https://www.npmjs.com/package/@types/reconnectingwebsocket`) for that.

ErikSchierboom (commented):
This is really exciting! The LSP is implemented in many, many languages: https://microsoft.github.io/language-server-protocol/implementors/servers/

babel.config.js
SleeplessByte: If someone can detail the issue, maybe 🗡️ 🥇

app/javascript/components/student/editor/FileEditor.tsx
SleeplessByte: In general `@types/xxx` will be available. However, when you download it, manually go into `node_modules/@types/xxx` and check the index file. It will have the _version_ it's for. If this is close to the version you use: its fine. 

ALternatively, add `declarations.d.ts` and :

```typescript
declare module 'reconnecting-websocket' {
  type CreateWebSocket = {
    (url: string, args: unknown[], options: Record<string, any>) => WebSocket
  }

  export = CreateWebSocket
}
```

(Above won't work, it's merely to demonstrate how to supply your own definitions, often WORTH it! You can grab the actual options from the documentation and build your own quick type.

kntsoriano: Thanks :) I've done this now.

babel.config.js
kntsoriano: We use this package https://github.com/TypeFox/vscode-ws-jsonrpc. But it's exported as CommonJS and we need this line in order to be able to use it. I'm actually not sure if I'm correct in this. This is how I just understood things.

PR #258, opened 2020-11-10
Add user activities

iHiD:
This adds user activities - a log of what a user has done, which can be filtered by a track.

It adds two activities to start with:
- Start exercise
- Submit iteration.

It follows the same sort of pattern to notifications.

The main issue right now is a lot of N+1 queries. I've got a couple of different choices on how to solve them and am waiting to decide that before working on it.

It renders like this:
Screenshot 2020-11-10 at 01 42 54


app/views/tracks/show/joined.html.haml
```diff
@@ -15,6 +15,35 @@
...
+                    .exercise
+                      = exercise_icon(activity.exercise)
+                    .exercise_name= activity.exercise.title
+
+                  .iteration
+                    = ViewComponents::Track::IterationSummary.new(activity.iteration)
+                    = graphical_icon :"chevron-right", css_class: "action-icon"
+
+                  %time{ datetime: activity.occurred_at } #{time_ago_in_words(activity.occurred_at)} ago
+
```
iHiD: This will all get split out into view components, but I've not yet decided whether they should be haml-side or React, so for now, I've just inlined.

app/commands/user/activity/create.rb
```diff
@@ -0,0 +1,20 @@
+class User::Activity
+  class Create
+    include Mandate
+
+    initialize_with :type, :user, :params
```
ErikSchierboom, line 5: Really nice how `:params` can automatically capture remaining arguments.

ErikSchierboom (approved):

app/models/user/activity.rb
```diff
@@ -0,0 +1,59 @@
...
+  #
+  # Any non-object params are left as the were passed in.
+  def params=(hash)
+    self[:params] = hash.transform_values do |v|
+      v.respond_to?(:to_global_id) ? v.to_global_id.to_s : v
+    end
+  end
+
+  # This reverses params= to explode back out
+  # {discussion: "gid://exercism/Solution::MentorDiscussion/186"}
```
ErikSchierboom, line 37: What is the use of `gid://`, as in: where it that used?

iHiD: https://github.com/rails/globalid#global-id---reference-models-by-uri

It's used by Rails to serlalized and reload objects based on their ids.

app/commands/user/activity/create.rb
iHiD: Yeah :)

PR #261, opened 2020-11-10
Allow passing class name to Icon and GraphicalIcon components

ErikSchierboom:
@iHiD Note that I've used `className`, not `classNames` as the former is the default react naming convention.


iHiD:
This is red.

iHiD (approved):

iHiD:
> @iHiD Note that I've used className, not classNames as the former is the default react naming convention.

Cool. Thanks! :)

PR #262, opened 2020-11-10
Remove react layer

iHiD:
(This is @ErikSchierboom's work - I've just opened it so I can look/comment on the code more easily. Erik - feel free to close/reopen a different PR, or use this one)

### DIFF
--- a/app/javascript/packs/application.tsx
+++ b/app/javascript/packs/application.tsx
@@ -62,6 +62,7 @@ import { camelizeKeys } from 'humps'
 import { Iteration } from '../components/track/IterationSummary'
 import { Submission, File } from '../components/student/Editor'
 import * as Tooltips from '../components/tooltips'
+import { Bootloader } from '../utils/bootloader'
 
 // Add all react components here.
 // Each should map 1-1 to a component in app/helpers/components
@@ -123,7 +124,7 @@ initReact({
   ),
   'common-icon': (data: any) => <Common.Icon icon={data.icon} alt={data.alt} />,
   'common-graphical-icon': (data: any) => (
-    <Common.GraphicalIcon icon={data.icon} />
+    <Bootloader.GraphicalIcon icon={data.icon} />
   ),
 })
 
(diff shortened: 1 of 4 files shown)


app/javascript/components/common/GraphicalIcon.tsx
```diff
@@ -1,9 +1,22 @@
...
+import { wrap } from '../../utils/wrapper'
 
-export function GraphicalIcon({ icon }: { icon: string }) {
-  return (
-    <svg className="c-graphical-icon" role="presentation">
-      <use xlinkHref={`#${icon}`} />
-    </svg>
-  )
+function GraphicalIconContents({ icon }: { icon: string }) {
+  return <use xlinkHref={`#${icon}`} />
```
iHiD: Maybe:
```suggestion
  return <><use xlinkHref={`#${icon}`} /></>
```

iHiD: @ErikSchierboom As a side note, this is one of the components we said we **don't** need to do this for, because we don't call these icons via the bootloader 🙂

ErikSchierboom: Yeah I know, it was just a very easy example for me :D

SleeplessByte: Wrapping it in a fragment is likely not going to fix it. My guess is that react now checks the validity of nesting. So the result still needs to be valid HTML. That means this will likely work for _everything_ *except* web-components and svgs.

ErikSchierboom: No, this does indeed not fix it:

```
react-dom.development.js:91 Warning: The tag <use> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.
    in use (at GraphicalIcon.tsx:5)
    in GraphicalIconContents (at application.tsx:128)
```

@iHiD I know the icons are not actually the components where this work should be done for. What are the right components?

iHiD: The `IterationSummary` is the main one atm that's called from both the Bootloader and within React.

PR #269, opened 2020-11-11
Add API endpoint for concept

iHiD:
@neenjaw You can cherry-pick this commit into your work, build on it, and this can be closed 🙂

### DIFF
--- a/app/controllers/api/concepts_controller.rb
+++ b/app/controllers/api/concepts_controller.rb
@@ -1,0 +1,18 @@
+module API
+  class ConceptsController < BaseController
+    skip_before_action :authenticate_user!
+    before_action :authenticate_user
+
+    def show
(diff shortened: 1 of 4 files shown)


app/controllers/api/concepts_controller.rb
```diff
@@ -0,0 +1,18 @@
+module API
+  class ConceptsController < BaseController
+    skip_before_action :authenticate_user!
+    before_action :authenticate_user
+
+    def show
+      track = Track.find_by!(slug: params[:track_id])
+      concept = track.concepts.find_by!(slug: params[:id])
+
+      render json: {
```
SleeplessByte, line 10: If you're just going to prerender the blurb, I recommend using content negotiation here to just return the HTML. removes a _lot_ of logic required.

PR #271, opened 2020-11-11
Add concept tooltip rendered server-side

iHiD:
Adds a ~tooltip~ (thing that's a bit like a tooltip but this is the wrong thing to call it) component for concepts.

Accessible via `/tracks/ruby/concepts/booleans/tooltip` or `tooltip_track_concept_path(:ruby, :booleans)`

Screenshot 2020-11-11 at 21 55 49


iHiD:
Now with completed state too.

Screenshot 2020-11-11 at 22 13 11

PR #272, opened 2020-11-11
Implement progress bar on concept page

iHiD:
The first commit implements the progress bar at the RHS.

Screenshot 2020-11-11 at 22 38 12

---

The second commit is a more comprehensive implementation of the user_track methods onto the new summary class, which builds on my work in https://github.com/exercism/v3-website/pull/270

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -76,6 +76,9 @@ Style/Documentation:
 Style/DocumentationMethod:
   Enabled: false
 
+Style/DoubleNegation:
+  Enabled: false
+
 Style/RedundantSelf:
   Enabled: false
 
--- a/app/commands/submission/create.rb
+++ b/app/commands/submission/create.rb
@@ -29,7 +29,6 @@ class Submission
 
       create_submission!
       create_files!
-      update_solution!
       schedule_jobs!
       submission.broadcast!
 
@@ -100,11 +99,6 @@ class Submission
           file.slice(:uuid, :filename, :digest, :content)
         )
       end
-    end
-
-    def update_solution!
-      solution.status = :submitted if solution.pending?
-      solution.save!
     end
 
     def schedule_jobs!
--- a/app/commands/user_track/generate_summary.rb
+++ b/app/commands/user_track/generate_summary.rb
@@ -4,24 +4,7 @@ class UserTrack
 
     initialize_with :track, :user_track
 
-    # We want to ensure we only do this once, but we
-    # provide two access vectors, this and the
-    # user_track.summary. This slightly hacky method
-    # checks whether we've already done this work and
-    # if so returns what we already have. Otherwise
-    # it does the work and then sets it.
-    #
-    # If we don't have a user_track, then we just generate
-    # the data and don't worry about this bit.
     def call
-      return generate! unless user_track
-
-      return user_track.summary if user_track.summary_generated?
-
-      user_track.summary = generate!
-    end
-
-    def generate!
       t = Time.now.to_f
       Rails.logger.info "[BM] Starting generating user summary"
 
@@ -174,6 +157,10 @@ class UserTrack
 
       def num_completed_exercises
         num_completed_concept_exercises + num_completed_practice_exercises
+      end
+
+      def available?
+        available
       end
 
       def mastered?
--- a/app/commands/user_track/generate_summary.rb
+++ b/app/commands/user_track/generate_summary.rb
@@ -2,19 +2,46 @@ class UserTrack
   class GenerateSummary
     include Mandate
 
-    initialize_with :user_track
-    delegate :track, to: :user_track
+    initialize_with :track, :user_track
 
+    # We want to ensure we only do this once, but we
+    # provide two access vectors, this and the
+    # user_track.summary. This slightly hacky method
+    # checks whether we've already done this work and
+    # if so returns what we already have. Otherwise
+    # it does the work and then sets it.
+    #
+    # If we don't have a user_track, then we just generate
+    # the data and don't worry about this bit.
     def call
-      #####################
-      # Populate concepts #
-      #####################
-      concepts = {}
-      track.concepts.each do |concept|
-        concept_exercises = concept_exercise_data.values.
+      return generate! unless user_track
+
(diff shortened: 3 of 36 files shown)


app/commands/user_track/generate_summary.rb
```diff
@@ -2,75 +2,189 @@ class UserTrack
...
+      Rails.logger.info "[BM] Starting generating user summary"
+
+      UserTrack::Summary.new(
+        concepts.map { |k, v| [k, ConceptSummary.new(v)] }.to_h,
+        exercises.map { |k, v| [k, ExerciseSummary.new(v)] }.to_h
+      ).tap do
+        Rails.logger.info "[BM] Finished generating user summary"
+        Rails.logger.info "[BM] Generating User Summary: #{Time.now.to_f - t}"
+      end
+    end
```
iHiD, line 18: The logging in this is temporary. But useful for now!

It takes about 30ms on a track with only one solution in development. I imagine that means with lots of solutions it hits a worse case of 50ms, and probably faster in production. Which is fine as this is getting cached as the normal situation anyway.

app/commands/user_track/generate_summary.rb
```diff
@@ -2,75 +2,189 @@ class UserTrack
...
+
+    def generate!
+      t = Time.now.to_f
+      Rails.logger.info "[BM] Starting generating user summary"
+
+      UserTrack::Summary.new(
+        concepts.map { |k, v| [k, ConceptSummary.new(v)] }.to_h,
+        exercises.map { |k, v| [k, ExerciseSummary.new(v)] }.to_h
+      ).tap do
```
iHiD, line 14: We're doing this mapping here (rather than lower down) because we'll cache JSON, not the serialized objects, so this method will eventually have a "cache or generate stuff - then map to the structs"

.rubocop.yml
```diff
@@ -76,6 +76,9 @@ Style/Documentation:
 Style/DocumentationMethod:
   Enabled: false
 
+Style/DoubleNegation:
+  Enabled: false
+
```
iHiD, line 81: This can fuck right off. What a ridiculous default!

app/commands/submission/create.rb
```diff
@@ -102,11 +101,6 @@ def create_files!
       end
     end
 
-    def update_solution!
-      solution.status = :submitted if solution.pending?
-      solution.save!
-    end
```
iHiD, line 108: This never actually got used, and I don't have plans for it now, so I'm removing the field altogether.

app/commands/user_track/generate_summary.rb
```diff
@@ -2,75 +2,189 @@ class UserTrack
   class GenerateSummary
     include Mandate
 
-    initialize_with :user_track
-    delegate :track, to: :user_track
+    initialize_with :track, :user_track
```
iHiD, line 5: `user_track` can be `nil`, so we need to specify the track explicitly. I might end up with two versions of this generator, one specifically for the nil scenario, but this is fine for now

app/models/track/concept.rb
```diff
@@ -17,9 +17,20 @@ def blurb
...
+  memoize
+  def practice_exercises
+    PracticeExercise.that_practice(self)
+  end
+
+  # TODO: Cache this?
+  memoize
   def num_exercises
-    3
+    concept_exercises.size + practice_exercises.size
```
iHiD: We'll almost certainly end up removing all references to this and using the summary instead.

app/helpers/view_components/widgets/exercise.rb
```diff
@@ -28,33 +29,15 @@ def to_s
...
-      # pre-fetches all the relevant solutions for a
-      # user, which we then pass into this. This whole
-      # method should then be replaces and the solution
-      # should be passed in directly.
-      memoize
-      def solution
-        # TODO: This will break once devise is added
-        Solution.for(User.first, exercise)
-      end
-
```
iHiD, line 41: This is an example of what all this new code is fixing. It removes this `n+1`, which is deeply nested in a view component, and ensures we only calculate this data once across a whole request (at most!)

app/models/user_track.rb
```diff
@@ -50,65 +57,13 @@ def available_practice_exercises
...
-            GROUP BY prereqs.exercise_id
-          ) a
-        INNER JOIN
-          (
-            SELECT exercise_id, COUNT(*) AS num_concepts
-            FROM exercise_prerequisites
-            GROUP BY exercise_id
-          ) b ON a.exercise_id = b.exercise_id AND a.num_concepts = b.num_concepts
-      })
-    end
```
iHiD, line 111: All this SQL has now been replaced by working it out in Ruby.

.rubocop.yml
SleeplessByte: Turn this off, always.

iHiD:
@ErikSchierboom @kntsoriano This is probably worth a read for you both. I've commented bits (both in the PR and the code) which I hope are useful. Very happy to answer any questions etc. It's still a little bit of a work in progress. Eventually I think we'll probably end up hiding this "summary" class altogether, and just delegating from user_track, then having a `NullUserTrack` for the external view (or some such thing). But it's already a good step forward from where we were.

app/models/user_track.rb
```diff
@@ -13,40 +13,38 @@ def self.for!(user_param, track_param)
...
+    user.solutions.joins(:exercise).where("exercises.track_id": track)
   end
 
-  delegate :exercise_available?, :concept_available?, to: :summary
+  delegate :exercise_available?, :exercise_completed?,
+    :num_completed_exercises,
+    :num_concepts, :num_concepts_mastered,
+    :num_exercises_for_concept, :num_completed_exercises_for_concept,
+    :concept_available?, :concept_mastered?,
+    to: :summary
```
iHiD, line 41: This whole commit is basically this.

ErikSchierboom: I love how there is stuff like this built-in :D

iHiD: Yeah. It's good huh. This is Rails, not Ruby btw.

PR #273, opened 2020-11-12
<Tooltips.Concept /> New feature for <ConceptMap />

neenjaw:
This PR adds tooltip/popover/dialog-type things to the <ConceptMap /> React component

This is based off of the work done for the other tooltips, but it has diverged a little for a couple reasons:
- they don't directly overlay
- they have some mild interactivity (links)
- they pull HTML content directly from the API using `react-query`
- the tooltip is self-controlled component, rather than a controlled component from the parent

---

## Goals of the feature

- Accessibility, this was designed first as a tab-navigable feature, then mouse interactivity added after.
- Consistent behavior
  - Tab-navigation takes precedence.  When the concept/tooltip has focus, the tooltip is open, mouse cannot open more tooltips u ntil focus is lost (blur)
  - Mouse navigation must allow for the tooltip to be detached
- Must allow for direct HTML injection from the API
- tooltips must adjust their placement based on the window size
  - This is done using popper, like the other tooltips

## Screenshots of features

> popper below (default)

tooltip-show-popper-bottom

> popper above (when window position doesn't allow)

tooltip-show-popper-top

> Showing focus preference to mouse

tooltip-show-focus-preference

> Showing tab-navigation (forward only, but works in reverse too)

tooltip-show-tab-nav

> Showing hover behavior

tooltip-show-hover

---

New files introduced:

- `app/javascript/components/tooltips/Concept.tsx`
  - This is the main new component, arguably I think there is room here for some codesplitting
    - Perhaps the state reducer and open-tooltip management functions could be moved into another file

Refactoring done:

- `app/javascript/components/tooltips/index.js` :arrow_right:  `app/javascript/components/tooltips/index.ts`
  - prevent loss of type information
- `app/commands/track/determine_concept_map_layout.rb`
  - added tooltip url to concept map generation
  - tests fixed to support change
- `app/javascript/components/tooltips/MentoredStudent.tsx`
  - added exported function return type
- `<ConceptMap />` CSS moved to the `css` dir :arrow_right: `app/css/components/track/concept-map.css`
  - imports updated


neenjaw:
test failing is because now the tooltip_url is included to the concept hash which is missing from the test

app/css/components/tooltips/concept.css
```diff
@@ -5,6 +5,8 @@
 
     @apply shadow-lg rounded-8 py-20 px-24 bg-white;
 
+    transition: opacity 120ms cubic-bezier(0.4, 0, 0.2, 1);
```
iHiD: Could we comment this out (but leave it there) and put a `/* TODO: Add fade in */` here pls. I'm trying to keep all design stuff to only things that that Tai has approved for now.

neenjaw: Sure

app/javascript/components/concept-map/Concept.tsx
```diff
@@ -50,6 +57,9 @@ export const Concept = ({
   if (isActive) {
     classes.push('active')
   }
+  if (isActiveHover) {
+    classes.push('hover')
```
iHiD: What does this class do?

app/javascript/components/concept-map/Concept.tsx
```diff
@@ -60,44 +70,44 @@ export const Concept = ({
...
       />
-    </a>
+    </div>
   )
 }
 
+// focus on concept / focus-blur
+// focus on tooltip / focus-blur
+// focus inside tooltip
+// no focus
```
iHiD: Can these be removed now?

neenjaw: Ya, I'll clean them up when this is ready for final review. It's only half working so far

app/javascript/components/concept-map/Concept.tsx
neenjaw: Nothing anymore, my plan was to remove it

app/javascript/components/concept-map/Concept.tsx
iHiD: OK :)

app/css/components/tooltips/concept.css
SleeplessByte: Thank you for using the "correct" transition tho.

neenjaw:
[Description revised and updated](https://github.com/exercism/v3-website/pull/273#issue-519669847)

I think that one test that isn't passing is a git repo test?

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+    requestHover: false,
+    requestFocus: false,
+    hover: false,
+    focus: false,
+    show: false,
+  })
+
+  const popper = usePopper(parent, tooltipRef.current, {
+    placement: 'bottom',
+    modifiers: [{ name: 'offset', options: { offset: [0, 8] } }],
```
neenjaw, line 267: This line changes the offset of the tooltip from the concept.

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  useEffect(() => {
+    if (hoverRequestToShow && popper?.update) {
+      popper.update()
+      dispatch(tooltipId, showDispatcher, 'request-hover', 'show')
+      return
+    }
+
+    setTimeout(
+      () => dispatch(tooltipId, showDispatcher, 'request-hover', 'hide'),
+      150
```
neenjaw: this is the delay once the mouse leaves the concept before tool-tip disappears (unless it has another reason to stay open). It facilitates the ability to move the mouse across the gap.  I just picked a duration that felt good. Up for change.

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+}
+
+function dispatch(
+  id = '',
+  dispatcher: React.Dispatch<any>, // eslint-disable-line @typescript-eslint/no-explicit-any
+  source: ShowActionSource,
+  action?: ShowActionValue
+): void {
+  dispatcher([id, dispatcher, source, action])
+}
```
neenjaw: This could probably be moved to another file

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  focusRequestToShow: boolean
+  parentSlug?: string
+}): JSX.Element => {
+  const tooltipId = `concept-tooltip${
+    parentSlug ? `-${parentSlug}` : parentSlug
+  }`
+  // Retrieve the HTML contents from the endpoint
+  const { isLoading, isError, data: html_content } = useQuery(tooltipId, () =>
+    fetch(endpoint).then((res) => res.text())
+  ) as { isLoading: boolean; isError: boolean; data: string }
```
neenjaw: If there is merit, this could be abstracted to a custom hook, like in the other tooltips

app/javascript/components/concept-map/Concept.tsx
```diff
@@ -1,5 +1,7 @@
 import React, { useEffect, useRef, useState, MouseEventHandler } from 'react'
 import { CompleteIcon } from './CompleteIcon'
+import * as Tooltips from '../tooltips'
```
iHiD: Should this just import `Tooltips.Concept` rather than everything?

neenjaw: It could be, I just followed the already set pattern.

app/javascript/components/tooltips/Concept.tsx
SleeplessByte: This entire function should probably go (it doesn't do anything except for forwarding the arguments given).

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+      break
+    case 'force-close':
+      nextState.show = false
+      break
+    default:
+      throw new Error('unhandled tooltip show action')
+  }
+
+  if (nextState.show) {
+    addOpenTooltip(id, dispatch)
```
SleeplessByte: This action is a side effect. There shouldn't be side-effects inside the reducer. If you want to _act_ on a _state change_, instead, use a `useEffect` or `useLayoutEffect` outside of the reducer.

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  } else {
+    removeOpenTooltip(id)
```
SleeplessByte: Same

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  // request-focus and focus actions take precedence over mouse events
+  // so it can close any other for accessibility purposes
+  if (
```
SleeplessByte: Same

app/javascript/components/tooltips/Concept.tsx
SleeplessByte: Doesn't `useQuery` return this as a type?? I don't think you need to cast it.

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  dispatcher([id, dispatcher, source, action])
+}
+
+export const Concept = ({
+  endpoint,
+  parent,
+  hoverRequestToShow,
+  focusRequestToShow,
+  parentSlug = '',
+}: {
```
SleeplessByte: Extract to `ConceptProps`

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const tooltipId = `concept-tooltip${
+    parentSlug ? `-${parentSlug}` : parentSlug
+  }`
+  // Retrieve the HTML contents from the endpoint
+  const { isLoading, isError, data: html_content } = useQuery(tooltipId, () =>
+    fetch(endpoint).then((res) => res.text())
+  ) as { isLoading: boolean; isError: boolean; data: string }
+
+  const tooltipRef = useRef(null)
+  const [{ show }, showDispatcher] = useReducer(showReducer, {
```
SleeplessByte: consider extracting the initial state to a constant or a function that returns this (reducers can take functions for intiial state)

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  return (
+    <div
+      ref={tooltipRef}
+      className={classNames.join(' ')}
+      style={styles}
+      {...popper.attributes.popper}
+      role="tooltip"
+      tabIndex={show ? undefined : -1}
+      onFocus={() => {
+        dispatch(tooltipId, showDispatcher, 'focus', 'show')
```
SleeplessByte: Consider instead renaming `showDispatcher` to `dispatch` and changing the array to an object (think positional arguments to keyword arguments):

```javascript
dispatch({ type: 'show', cause: 'focus', target: tooltipId })
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+      }}
```
iHiD:
```suggestion
      onFocus={() => { showDispatcher(tooltipId, 'focus', 'show') }}
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+  useEffect(() => {
+    if (hoverRequestToShow && popper?.update) {
+      popper.update()
+      dispatch(tooltipId, showDispatcher, 'request-hover', 'show')
+      return
+    }
+
+    setTimeout(
+      () => dispatch(tooltipId, showDispatcher, 'request-hover', 'hide'),
```
SleeplessByte: Should guard against unmount before the timer ends.

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  useEffect(() => {
+    if (focusRequestToShow && popper?.update) {
+      popper.update()
+      dispatch(tooltipId, showDispatcher, 'request-focus', 'show')
+      return
+    }
+
+    dispatch(tooltipId, showDispatcher, 'request-focus', 'hide')
+  }, [focusRequestToShow]) // eslint-disable-line react-hooks/exhaustive-deps
+
```
SleeplessByte, line 383: Should hide itself on unmount

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+      role="tooltip"
+      tabIndex={show ? undefined : -1}
+      onFocus={() => {
+        dispatch(tooltipId, showDispatcher, 'focus', 'show')
+      }}
+      onBlur={() => {
+        dispatch(tooltipId, showDispatcher, 'focus', 'hide')
+      }}
+      onMouseEnter={() => dispatch(tooltipId, showDispatcher, 'hover', 'show')}
+      onMouseLeave={() => dispatch(tooltipId, showDispatcher, 'hover', 'hide')}
```
iHiD: As far as I can tell, `dispatch` is unneccessary and can be replaced as follows

```suggestion
      onFocus={() => { showDispatcher(tooltipId, 'focus', 'show') }}
      onBlur={() => { showDispatcher(tooltipId, 'focus', 'hide') }}
      onMouseEnter={() => showDispatcher(tooltipId, 'hover', 'show')}
      onMouseLeave={() => showDispatcher(tooltipId, 'hover', 'hide')}
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+    setTimeout(
+      () => dispatch(tooltipId, showDispatcher, 'request-hover', 'hide'),
+      150
+    )
+  }, [hoverRequestToShow]) // eslint-disable-line react-hooks/exhaustive-deps
+
+  useEffect(() => {
+    if (focusRequestToShow && popper?.update) {
+      popper.update()
+      dispatch(tooltipId, showDispatcher, 'request-focus', 'show')
```
iHiD:
```suggestion
      showDispatcher(tooltipId, 'request-focus', 'show')
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+      return
+    }
+
+    dispatch(tooltipId, showDispatcher, 'request-focus', 'hide')
```
iHiD:
```suggestion
    showDispatcher(tooltipId, 'request-focus', 'hide')
```

app/javascript/components/tooltips/Concept.tsx
iHiD:
```suggestion
      () => showDispatcher(tooltipId, 'request-hover', 'hide'),
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+  // For the next two useEffects:
+  // If I ignore the exhaustive deps warning and only have requestToShow, it
+  // works.  If I have the exhaustive deps added (popper) it causes inf loop
+  // due to popper reference changing on every render.
+
+  useEffect(() => {
+    if (hoverRequestToShow && popper?.update) {
+      popper.update()
+      dispatch(tooltipId, showDispatcher, 'request-hover', 'show')
```
iHiD:
```suggestion
      showDispatcher(tooltipId, 'request-hover', 'show')
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const index = indexOfOpenTooltip(id)
+  if (index !== -1) {
+    openTooltips.splice(index, 1)
+  }
+}
+
+const closeOtherOpenTooltips = (id: string) => {
+  openTooltips.forEach(([openId, openDispatcher]) => {
+    if (openId !== id) {
+      dispatch(openId, openDispatcher, 'force-close')
```
iHiD:
```suggestion
      openDispatcher(openId, 'force-close')
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * showReducer
+ * This serves as a state reducer for the Concept (tooltip) Component for use with
+ * the useReducer hook.
+ */
+function showReducer(
+  state: ShowState,
+  [id = '', dispatch, source, action]: [
+    string,
+    React.Dispatch<unknown>,
```
iHiD:
```suggestion
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/**
+ * showReducer
+ * This serves as a state reducer for the Concept (tooltip) Component for use with
+ * the useReducer hook.
+ */
+function showReducer(
+  state: ShowState,
+  [id = '', dispatch, source, action]: [
```
iHiD: This `dispatch` object isn't used in this method.

```suggestion
  [id = '', source, action]: [
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+    string,
+    React.Dispatch<unknown>,
+    ShowActionSource,
+    ShowActionValue
+  ]
+) {
+  const nextState: ShowState = { ...state }
+  const parsedAction = action === 'show'
```
iHiD: I found this variable name confusing.

```suggestion
  const shouldShow = action === 'show'
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const nextShow = (state: ShowState): boolean =>
+    state.requestHover || state.requestFocus || state.hover || state.focus
+
+  switch (source) {
+    case 'request-hover':
+      nextState.requestHover = parsedAction && !hasOpenTooltip(id)
```
iHiD:
```suggestion
      nextState.requestHover = shouldShow && !hasOpenTooltip(id)
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+      nextState.show = nextShow(nextState)
+      break
+    case 'request-focus':
+      nextState.requestFocus = parsedAction
```
iHiD:
```suggestion
      nextState.requestFocus = shouldShow
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+      nextState.show = nextShow(nextState)
+      break
+    case 'hover':
+      nextState.hover = parsedAction && !hasOpenTooltip(id)
```
iHiD:
```suggestion
      nextState.hover = shouldShow && !hasOpenTooltip(id)
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+      nextState.show = nextShow(nextState)
+      break
+    case 'focus':
+      nextState.focus = parsedAction
```
iHiD:
```suggestion
      nextState.focus = shouldShow
```

app/javascript/components/tooltips/Concept.tsx
iHiD: I can't see the purpose of this method. It seems to just indirectly call a method on one of its paramaters.

```suggestion
```

app/javascript/components/tooltips/Concept.tsx
iHiD: Does the `id` need a default value? What is the situation where that happens?

app/javascript/components/tooltips/Concept.tsx
iHiD: Regardless of this change, here we have `action?` but in the dispatcher, we're missing the `?`. I would expect them both to be the same, but I'm quite possibly wrong :)

iHiD (commented):
OK, first pass review. This removes a bit of indirection that confused me. I've started here because the rest didn't confuse me in the same way! :)

I'll look at more of it next.

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+): void {
+  dispatcher([id, dispatcher, source, action])
+}
+
+export const Concept = ({
+  endpoint,
+  parent,
+  hoverRequestToShow,
+  focusRequestToShow,
+  parentSlug = '',
```
iHiD: The "parent" and the "slug" are only related here via some coincidence. This tooltip might reasonable appear on things that are not concepts. So this would be better loosely coupled from that, and named `slug` (or possible `conceptSlug`)

```suggestion
  slug = '',
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+  dispatcher: React.Dispatch<any>, // eslint-disable-line @typescript-eslint/no-explicit-any
+  source: ShowActionSource,
+  action?: ShowActionValue
+): void {
+  dispatcher([id, dispatcher, source, action])
+}
+
+export const Concept = ({
+  endpoint,
+  parent,
```
iHiD: I don't think this is actually a `parent`. It's not hierachical (the element actually sits next to it in the tree) and parent is quite non-descript as to the purpose it serves in this class. I think we can name it relevantly to *this* class. So  It's the `triggeringElement` (in the context of a tooltip) or `referenceElement` (popper's name for it) etc

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+  // Get styles and classes to apply
+  const styles = { ...popper.styles.popper, zIndex: 10 }
+  const classNames = ['c-tooltip', 'c-concept-tooltip']
+  if (!show) {
+    classNames.push('tw-invisible')
+  }
+  if (show) {
+    classNames.push('tw-pointer-events-auto')
+  }
```
iHiD: ~Two~ Three comments here.

1. If/else

```suggestion
  if (show) {
    classNames.push('tw-pointer-events-auto')
  }
  else {
    classNames.push('tw-invisible')
  }
```

2. Do we need to turn `tw-pointer-events-auto` off? If the element has `display:none`, then I don't think it matter if these are on. So I think this can just be added to the default set on L200, and you can *only* have the `if(!show)..` bit

3. Is there a reason you're using `tw-invisible` and not `tw-hidden`? I think we should use `display:none` here unless there's a good reason not to.

4. I think this should have a default state and a pivot. I'm presuming we consider tooltips to be shown by default, so we can just put `tw-pointer-events-auto` in the stylesheet?

iHiD: Also, if we have (2) on always (as opposed to requiring conditional logic), that should be added to the CSS class for `c-concept-tooltip` not added here..

app/javascript/components/tooltips/Concept.tsx
iHiD: (For clarity, this is the same as my comment below but with the addition of using the object)

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,244 @@
...
+        role="tooltip"
+      >
+        {isLoading && <Loading />}
+        {isError && <p>Something went wrong</p>}
+      </div>
+    )
+  }
+
+  return (
+    <div
```
iHiD, line 395: Should we use `<dialog>` for these?

app/javascript/components/tooltips/Concept.tsx
neenjaw: It was erroring without it I thought. :shrug: I can try again. Maybe I just need to restart `TSServer` :smile:

app/javascript/components/tooltips/Concept.tsx
neenjaw: it's just how prettier formatted it for me when I hit `ctrl+s`, I'll see if I can get that to stick

app/javascript/components/tooltips/Concept.tsx
neenjaw: Thanks! This was like the 5th name and I just couldn't find one I liked.  Hindsight 20-20 :eyeglasses:

test/commands/track/determine_concepts_layout_test.rb
```diff
@@ -28,7 +28,8 @@ def test_layout_with_one_level
             {
               slug: 'basics',
               name: 'Basics',
-              web_url: 'https://test.exercism.io/tracks/ruby/concepts/basics'
+              web_url: 'https://test.exercism.io/tracks/ruby/concepts/basics',
+              tooltip_url: '/tracks/ruby/concepts/basics/tooltip'
```
iHiD: Can we make these URLs, not paths pls? (You'll obviously need to change the code to match -`_path` to `_url`.)

app/javascript/components/tooltips/Concept.tsx
neenjaw: Ya, it started as something else then mutated over time, you and @SleeplessByte are right.  It just needs to go.

app/javascript/components/tooltips/Concept.tsx
iHiD: This is nice. Can you add this as a comment to the code please. The first two sentences of your comment here would be perfect 🙂 

Always good to add a comment for any "magic numbers" that aren't worth turning into a variable name (and I agree that here, inlining is best).

app/javascript/components/tooltips/Concept.tsx
neenjaw: ya, `referenceElement` or `triggeringElement` is probably the way to go

app/javascript/components/tooltips/Concept.tsx
iHiD: Apparently, we shouldn't because they don't cause a change in focus.

app/javascript/components/tooltips/Concept.tsx
iHiD: That's cool. Refactoring to delete code once everything works is the perfect way to solve things. Red/green/refactor FTW! :)

app/javascript/components/tooltips/Concept.tsx
iHiD: Yeah, totally. (To continue to metaphor) a second pair of eyes is invaluable for naming!

app/javascript/components/tooltips/Concept.tsx
neenjaw: 1) Sure, I think they were separated by something else which I removed
2) So, `pointer-events: none` is set on the figure, so it is inherited by the tooltip, and strangely, without it set to none seems to trigger it to show even when hidden.
3) Using **invisible** so that popper can calculate the position of the tooltip even when not seen and to make it invisible to screen-readers (rather than opacity)
4) I hear what you're saying, but I think what I'm saying is that without manipulating the pointer events (as in `(2)`) it causes some triggering problems.

app/javascript/components/tooltips/Concept.tsx
neenjaw: I don't know. I thought dialog has a different meaning with semantic HTML tho?

test/commands/track/determine_concepts_layout_test.rb
neenjaw: Oh, I think I just copied the ruby helper code from your example.  I didn't know that the ending changed things! :laughing:

app/javascript/components/tooltips/Concept.tsx
iHiD: > without it set to none seems to trigger it to show even when hidden.

My guess would be because it's `invisible` and not `hidden`

> so that popper can calculate the position of the tooltip even when not seen

Why is this desirable?

app/javascript/components/tooltips/Concept.tsx
SleeplessByte: Don't. It's prettier. You don't need to always agree with it :)

app/javascript/components/tooltips/Concept.tsx
SleeplessByte: You need:

```typescript
useQuery<ReturnType>()
```

or even

```typescript
useQuery<ReturnType, ErrorType>()
```

app/javascript/components/tooltips/Concept.tsx
neenjaw: :relieved:

app/javascript/components/tooltips/Concept.tsx
neenjaw: removed

app/javascript/components/concept-map/Concept.tsx
iHiD: Let's import the single thing. Could you change this when you get a min pls? :)

test/commands/track/determine_concepts_layout_test.rb
```diff
@@ -57,12 +58,14 @@ def test_layout_with_two_levels
...
```
iHiD:
```suggestion
              tooltip_url: 'https://test.exercism.io/tracks/ruby/concepts/basics/tooltip'
```

etc

neenjaw: Ya, haven't finished all the changes just yet. Ran out of time last night. Will be working on this more today

iHiD: No probs! I figured you knew :)

app/javascript/components/tooltips/Concept.tsx
neenjaw: This system has been changed a bit, I can't really comment on this implmentation now.  Will explain when the PR is ready again for re-review

neenjaw: So now, if the tooltip is not hovered, the div is not mounted to the DOM, when it is, it is initially invisible so it can be positioned by popper (visibility: hidden) then made visible.

app/javascript/components/tooltips/Concept.tsx
ErikSchierboom: Aha, that must be what's causing this:

tooltip

app/javascript/components/concept-map/Concept.tsx
```diff
@@ -10,26 +12,31 @@ import {
   Visibility,
 } from './helpers/concept-visibility-handler'
 import { wrapAnimationFrame } from './helpers/animation-helpers'
-import { PureExerciseProgressBar } from './ExerciseProgressBar'
 
 export const Concept = ({
   slug,
   name,
   web_url,
+  tooltip_url,
```
ErikSchierboom: We're using camelCase for the other properties. Can we use them here too?

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,423 @@
...
+  focusRequestToShow,
+  referenceElement,
+  referenceConceptSlug,
+}: ConceptProps): JSX.Element | null => {
+  const tooltipId = `concept-tooltip${
+    referenceConceptSlug ? `-${referenceConceptSlug}` : referenceConceptSlug
+  }`
+
+  // Retrieve the HTML contents from the contentEndpoint
+  const { isLoading, isError, data: html_content } = useQuery<string>(
```
ErikSchierboom: Could `html_content` use camelCase?

```suggestion
  const { isLoading, isError, data: htmlContent } = useQuery<string>(
```

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,423 @@
...
+      nextState.showState = 'error'
+      break
+    default:
+      throw new Error('unhandled tooltip reducer action')
+  }
+
+  return nextState
+}
+
+export const Concept = ({
```
ErikSchierboom, line 227: Just to check: most of the code is not specific to the Concept tooltip, but could be applied to other tooltips too, right? Were you planning to extract this to a separate component, for it to be re-used?

app/javascript/components/tooltips/Concept.tsx
```diff
@@ -0,0 +1,423 @@
...
+  referenceConceptSlug,
+}: ConceptProps): JSX.Element | null => {
+  const tooltipId = `concept-tooltip${
+    referenceConceptSlug ? `-${referenceConceptSlug}` : referenceConceptSlug
+  }`
+
+  // Retrieve the HTML contents from the contentEndpoint
+  const { isLoading, isError, data: html_content } = useQuery<string>(
+    tooltipId,
+    () => fetch(contentEndpoint).then((res) => res.text())
```
ErikSchierboom: The editor uses the `AbortController` class to allow cancellation of pending `fetch` requests. It might be nice to add that here too. See https://github.com/exercism/v3-website/blob/84863725436106c30e5e04bc41ca43031d1acd0d/app/javascript/components/student/Editor.tsx#L187

app/javascript/components/concept-map/Concept.tsx
```diff
@@ -60,41 +67,40 @@ export const Concept = ({
...
+          exercises={exercises}
+          hidden={!hasExercises || !isStarted}
+        />
+      </a>
+      <ConceptTooltip
+        contentEndpoint={tooltip_url}
+        hoverRequestToShow={isActiveHover}
+        focusRequestToShow={hasFocus}
+        referenceElement={conceptRef.current}
+        referenceConceptSlug={slug}
```
ErikSchierboom: Nice and simple way to include the tooltip! 👍

app/javascript/components/concept-map/ConceptMap.tsx
```diff
@@ -68,6 +68,7 @@ export const ConceptMap = ({
                     slug={slug}
                     name={concept.name}
                     web_url={concept.web_url}
+                    tooltip_url={concept.tooltip_url}
```
ErikSchierboom: Could we use camelCase here?

```suggestion
                    tooltipUrl={concept.tooltip_url}
```

ErikSchierboom (approved):
Love it! Left some minor comments.

ErikSchierboom:
I feel that this "let's have the tooltips just render HTML" idea was excellent.

app/javascript/components/concept-map/Concept.tsx
neenjaw: I think these are like this because these properties are coming from the rails api. So either I need to use camelcase in rails, or snake case in typescript.

app/javascript/components/tooltips/Concept.tsx
neenjaw: We had discussed this in slack, and the one case that @iHiD had been thinking of we weren't sure if it just deserved just another type of tooltip. So I haven't made an active effort to ensure generalization

app/javascript/components/tooltips/Concept.tsx
neenjaw: Yes, I think I just had this separate because of its "unsafe" usage in my head.

app/javascript/components/tooltips/Concept.tsx
neenjaw: Ya, should just tweak the timing there a bit

neenjaw: Thanks for the gif, makes it easier to see what's happening!

app/javascript/components/concept-map/Concept.tsx
SleeplessByte: @kntsoriano has a helper for this!

app/javascript/components/tooltips/Concept.tsx
SleeplessByte: (it's not unsafe tho :P)

app/javascript/components/tooltips/Concept.tsx
ErikSchierboom: First time making a GIF!

Zvxx53q

app/javascript/components/tooltips/Concept.tsx
ErikSchierboom: Makes sense!

app/javascript/components/tooltips/Concept.tsx
neenjaw: fixed

app/javascript/components/concept-map/Concept.tsx
neenjaw: help me @kntsoriano, you're my only hope!

app/javascript/components/tooltips/Concept.tsx
neenjaw: This should now be fixed in https://github.com/exercism/v3-website/pull/273/commits/d803d7780035a47d9f9e87f7cd019873e43b8fd6

app/css/components/tooltips/concept.css
```diff
@@ -91,4 +93,27 @@
...
         }
     }
+
+    &::before {
+        content: "";
+        width: 100%;
+        height: 8px;
+        left: 0;
+        top: 0;
+        display: inline-block;
```
iHiD: Is this needed?

neenjaw: Yes so that it can be used to cover the gap between the `<Concept>` and `<ConceptTooltip>`. Otherwise the tooltip closes before you have a chance to hover over it when there is a gap

neenjaw: I think `display: block` might work too, but untested

iHiD: Yeah. I think you can remove this line altogether and it will default to `display: block`.

iHiD:
100th comment 🦜 

Screenshot 2020-11-17 at 21 55 50

neenjaw:
Shoulda used :100: :P

app/javascript/components/concept-map/Concept.tsx
ErikSchierboom: @neenjaw See https://github.com/exercism/v3-website/blob/784415bd5b845e978d8ea3949fa3465600c3e9de/app/javascript/hooks/request-query.ts#L37

kntsoriano: @neenjaw I use this to import the helper: `import { camelizeKeys } from 'humps'`

We usually apply the helper when we get the response from the API. But in your case, there is no API request, so I would probably camelize the keys inside `application.tsx`. Hope that helps!

neenjaw: Yes! Thanks!

SleeplessByte (commented):
Tell me when I should look at this.

neenjaw:
Review when you get a chance :)

iHiD:
@SleeplessByte I think we're just waiting on you for this one. No urgency but just in case it fell past your radar :)

app/css/components/tooltips/concept.css
```diff
@@ -91,4 +93,25 @@
             @apply text-darkGray leading-paragraph font-medium;
         }
     }
+
+    &::before {
```
SleeplessByte, line 99: I tend to comment these so the next reader understands _why_ this before and after pseudo element is necessary. In this case it's not because of visuals, so it's important!

app/css/components/tooltips/concept.css
```diff
@@ -91,4 +93,25 @@
...
+        content: "";
+        width: 100%;
+        height: 8px;
+        left: 0;
+        top: 0;
+        transform: translateY(-100%);
+        background: rgba(0, 0, 0, 0);
```
SleeplessByte:
```suggestion
        background: transparent;
```

Showing the colour picker for the transparent keyword colour. It matches #00000000, which is rbga(0, 0, 0, 0)

It's usually easier to parse with intent.

app/javascript/components/concept-map/Concept.tsx
SleeplessByte: (This).

app/javascript/components/concept-map/Concept.tsx
```diff
@@ -10,26 +12,31 @@ import {
...
 }: IConcept & {
   handleEnter: MouseEventHandler
   handleLeave: MouseEventHandler
   status: ConceptStatus
   isActive: boolean
+  isActiveHover: boolean
 }): JSX.Element => {
   const [visibility, setVisibility] = useState<Visibility>('hidden')
   const conceptRef = useRef(null)
+  const [hasFocus, setHasFocus] = useState<boolean>(
```
SleeplessByte, line 44: This is ambiguous. Usually when an element is focusable and it contains focusable elements, we'd say "focus within", but in this case I think focus is meant to be "focus is on this concept (or within)". Its initial value is "focus is on this concept (but NOT within)".

So:
- either the logic isn't completely sound
- or the naming isn't perfect

It's fine to keep, but add a comment!

app/javascript/components/concept-map/Concept.tsx
```diff
@@ -10,26 +12,31 @@ import {
...
   web_url,
+  tooltip_url,
   handleEnter,
   handleLeave,
   status,
   isActive,
+  isActiveHover,
   exercises = 0,
   exercisesCompleted = 0,
 }: IConcept & {
```
SleeplessByte: Perhaps `ConceptProps` is a good idea now. This list has become long.

SleeplessByte (approved):

app/javascript/components/concept-map/Concept.tsx
neenjaw: done

neenjaw: Thanks for the tips!

iHiD:
Great work. Super excited for this!! Thanks for all your hard work @neenjaw, and for the reviews @SleeplessByte, @ErikSchierboom and @kntsoriano  💙 

(@ErikSchierboom It'll be a good place to test your concept "blurb" syncing code manually too 🙂)

PR #276, opened 2020-11-12
Convert components to TypeScript

ErikSchierboom:
This PR converts some of the existing JavaScript code to TypeScript. The goal here is to catch errors at an earlier stage, while still coding. I've not converted all components as that was quite a bit of work. I'd be happy to do so if everyone feels that is a useful way to spend my time. But we can just do this incrementally.


iHiD (approved):
Approving as Karlo has no power in the Philipines (big storms) so might be a while before he can reply. But I don't know the code well enough to be useful!

ErikSchierboom:
@iHiD It would be good for @SleeplessByte to take a quick look then.

app/javascript/components/common/TrackIcon.tsx
```diff
@@ -1,6 +1,11 @@
 import React from 'react'
 
-export function TrackIcon({ track }) {
+type TrackIconProps = {
+  iconUrl: string
+  title: string
+}
+
+export function TrackIcon({ track }: { track: TrackIconProps }) {
```
SleeplessByte: I would have expected less nesting (because of the naming convention ComponentProps):

```typescript
export function TrackIcon({ iconUrl, title }: TrackIconProps) {
```

app/javascript/components/maintaining/SubmissionsSummaryTable.tsx
```diff
@@ -23,11 +36,15 @@ function SubmissionsSummaryTableRow({
...
+export function SubmissionsSummaryTable({
+  submissions,
+}: {
+  submissions: Submission[]
+}) {
   const [stateSubmissions, setSubmissions] = useState(submissions)
 
   useEffect(() => {
-    const received = (data) => {
+    const received = (data: SubmissionFromChannel) => {
```
SleeplessByte: I haven't looked into actioncable's typing yet, but assuming this `data` param is _actual data_, and not an `event` param with the data merged in? If yes, this is fine. If no, I recommend using the `XXXEvent` convention for naming.

app/javascript/components/maintaining/SubmissionsSummaryTable.tsx
```diff
@@ -23,11 +36,15 @@ function SubmissionsSummaryTableRow({
   )
 }
 
-export function SubmissionsSummaryTable({ submissions }) {
+export function SubmissionsSummaryTable({
+  submissions,
+}: {
+  submissions: Submission[]
```
SleeplessByte: If this is intended to be readonly (and I think it is. you don't want this table component to modify the prop's array which would cause potentially other higher up components that use the same value to break), mark it as such:
```suggestion
  submissions: readonly Submission[]
```

app/javascript/components/notifications/Icon.tsx
```diff
@@ -1,19 +1,30 @@
 import React, { useReducer, useEffect } from 'react'
 import consumer from '../../utils/action-cable-consumer'
 
-function reducer(state, action) {
+type IconProps = {
+  count: number
+}
+
+type IconActionType = 'notifications.changed'
```
SleeplessByte: This type isn't used, and can be in-lined on line 11.

app/javascript/hooks/request-query.ts
```diff
@@ -10,17 +9,17 @@ import { camelizeKeys } from 'humps'
 
 type Request = {
   endpoint: string
-  query?: string
+  query?: object
```
SleeplessByte: Probably not accepting _any_ object, linter/ts should have complained about this.

```suggestion
  query?: Record<string, string>
```

Or since you use it as such:

```suggestion
  query?: Parameters<UrlParams>[0]
```

app/javascript/utils/time.ts
```diff
@@ -2,6 +2,6 @@ import dayjs from 'dayjs'
 import RelativeTime from 'dayjs/plugin/relativeTime'
 dayjs.extend(RelativeTime)
 
-export function fromNow(dateTime) {
+export function fromNow(dateTime: dayjs.ConfigType) {
```
SleeplessByte: Very good. Should be able to import this type too:

```typescript
import type { ConfigType } from 'dayjs'
```

In this case not that much difference, but sometimes you want to do this explicitly, because 

1. it's easier to strip `import type` 
2. doesn't convolute types on namespaces with properties on namespaces/objects/default export bags.

app/javascript/utils/url-params.ts
```diff
@@ -1,12 +1,18 @@
+type Params = {
+  [key: string]: any
+}
+
```
SleeplessByte: It's easier to write

```suggestion
type Params = Record<string, any>

```

Are you certain it can be _any_ tho? From line 18-25, the reality is it can only be an `Array` of `string` (`string[]`), or whatever `encodeURIComponent` takes (which is probably only `string`).

```suggestion
type Params = Record<string, string | string[]>
```

SleeplessByte (approved):

app/javascript/utils/url-params.ts
ErikSchierboom: I've refined this is bit

app/javascript/utils/time.ts
ErikSchierboom: Great!

app/javascript/hooks/request-query.ts
ErikSchierboom: Never new about `Parameters`! Awesome. I found that there is even a `ConstructorParameters` option.

ErikSchierboom:
Thanks @SleeplessByte!

PR #282, opened 2020-11-13
Refactor tooling jobs

iHiD:
Goes with:
- https://github.com/exercism/tooling-orchestrator/pull/18
- https://github.com/exercism/tooling-invoker/pull/23

Refactors the tooling jobs across the pathway to simplify things:
- Removes S3 for storing results (this will come back, but as as secondary action via the website, not via the orchestrator, which will speed things up)
- Uses one key for all the job metadata
- Adds a ToolingJob model wrapping the DynamoDB table
- Renders errors out on the maintaining dashboard (this was the cause of all this).

There's some small work to replicate the changes for representers and analyzers, but I'll do that once this is all good.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -41,6 +41,36 @@ class Submission < ApplicationRecord
   end
 
   def serialized
+    # tests_data = tests_status
+    # if tests_exceptioned?
+    #   client = ExercismConfig::SetupDynamoDBClient.()
+    #   tests_data += client.query(
+    #     table_name: Exercism.config.dynamodb_tooling_jobs_table,
+    #     index_name: "submission_type",
+    #     expression_attribute_values: {
+    #       ":SU" => submission_uuid,
+    #       ":TP" => type,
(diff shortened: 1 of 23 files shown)


app/models/submission.rb
```diff
@@ -41,11 +41,17 @@ def exercise_version
   end
 
   def serialized
+    tests_data = tests_status
+    if tests_exceptioned?
+      job = ToolingJob.find(test_runs.last.tooling_job_id, full: true)
+      tests_data += "\n\n#{JSON.pretty_generate(job.execution_metadata)}"
+    end
```
iHiD, line 48: This is temporary so I didn't have to fiddle with the React, but it's fine for now.

app/models/tooling_job.rb
```diff
@@ -0,0 +1,89 @@
+class ToolingJob
```
iHiD, line 1: All the DynamoDB stuff lives in this one class now.

PR #285, opened 2020-11-16
Small language server addons

kntsoriano:


iHiD (approved):
Looks good to me!

app/javascript/components/student/editor/ExercismMonacoEditor.tsx
```diff
@@ -54,7 +55,9 @@ export function ExercismMonacoEditor({
     })
 
     MonacoServices.install(editor)
-    const url = normalizeUrl(`ws://localhost:3000/${language}`)
+    const url = normalizeUrl(
+      `ws://${process.env.LANGUAGE_SERVER_HOST}/${language}/${uuidv4()}`
```
SleeplessByte: Probably want to also pass in the _protocol. I recommend the env has both `protocol` + `hostname` + `port` (so the entire origin).

Why?
- `wss`: secure web socket
- `ws`: regular (insecure) web socket

SleeplessByte (approved):

ErikSchierboom (approved):

app/javascript/components/student/editor/ExercismMonacoEditor.tsx
kntsoriano: Sounds good! I've created https://github.com/exercism/language-servers/issues/2 to remind us to migrate to wss.

app/javascript/components/student/editor/ExercismMonacoEditor.tsx
```diff
@@ -54,7 +55,7 @@ export function ExercismMonacoEditor({
     })
 
     MonacoServices.install(editor)
-    const url = normalizeUrl(`ws://localhost:3000/${language}`)
+    const url = normalizeUrl(`${process.env.LANGUAGE_SERVER_HOST}/${language}/${uuidv4()}`)
```
iHiD, line 58: How would this work if we wanted to set this via some rails config rather than an ENV var? Is that doable? We don't use ENV vars anywhere in the whole stack for different environments - everything goes through Exercism.config - I want to keep that pattern here so need to understand options for how we could do this.

Setting ENV vars via Exercism Config is an option, but I'd rather a different one if possible 🙂 

What is `process` here?

SleeplessByte: You could either inject the variables you need into the layout in a meta tag in the head (probably what I would do), or create an `env.ts.erb` and do the same.

`process` is node. `process.env` is the node environment.

iHiD: > You could either inject the variables you need into the layout in a meta tag in the head 

@kntsoriano Could you do this pls?

iHiD: So this broke exercism.lol. It broke because the environment variable is not set. However, I don't get how that variable can actually get set in production anyway. We don't run this in node in production, so there isn't a "process" (other than maybe the browser). Does webpacker interpolate this into the code when it builds maybe?

Regardless, I'll revert this for now to fix the staging site, and let's use the meta-tag moving forward

SleeplessByte: Webpack has several plugins, such as [`EnvironmentPlugin`](https://webpack.js.org/plugins/environment-plugin/) which _regex search and replace_ `process.env.XXX` with the actual value. 

> Does webpacker interpolate this into the code when it builds maybe?

So yes.

iHiD: @kntsoriano DJ tells me that it interpolates it when it compiles.

So let's do the following:
- You add the meta tag and read from that.
- Once LSPs are deployed, I'll update the metatag to populate from Exercism.config.

I'll either revert this commit, or leave the staging site broken for today.

SleeplessByte: You can also add the env variable to the staging site and rebuild and deploy?

SleeplessByte: (I like the meta tag because it's not more/less safe, but it is less magic ánd can be used from other non-webpack places)

iHiD: I could, but I don't want to add env vars because it means we have to keep them in sync with the server endpoints in the staging site, which change independently with terraform changes. (This is what Exercism.config) insulates against)

PR #290, opened 2020-11-18
Add temporary rootUri

kntsoriano:


ErikSchierboom (approved):
What does this do?

kntsoriano:
@ErikSchierboom The C# language server needs this for some reason. I don't know why. I'm pretty sure that I'm not supplying the correct value for the `rootUri`, but Omnisharp needs some value in order for it not to crash.

PR #291, opened 2020-11-18
Sync tracks and exercises from git

ErikSchierboom:


app/models/track.rb
```diff
@@ -16,7 +16,7 @@ class Track < ApplicationRecord
     :ignore_regexp,
     to: :git
 
-  delegate :head_sha, to: :git, prefix: :git
+  delegate :head_sha, to: :git, prefix: true
```
iHiD: This is implicit, not explicit now 🙂 

I preferred explicit

iHiD: My logic is that if we rename the `git` method in `Track` to `repo` (for example), we don't want all the things that use `git_head_sha` to fail because this method gets renamed to `repo_head_sha`.

test/factories/concept_exercises.rb
```diff
@@ -4,6 +4,7 @@
     uuid { SecureRandom.uuid }
     slug { 'datetime' }
     title { slug.titleize }
+    git_sha { "HEAD" }
```
iHiD, line 9: Once this gets close to finished, it might be worth updating these to be whatever the *actual* head commit in the fixture is.

ErikSchierboom: Indeed

db/seeds.rb
```diff
@@ -128,7 +129,7 @@ def rugged_repo
   rescue => e
     #puts e.message
     #puts e.backtrace
-    puts "Error creating concept exercises for Track #{track_slug}: #{e}"
+    puts "Error seeding Track #{track_slug}: #{e}"
```
ErikSchierboom, line 106: This made it easier to find any issues with tracks as well as skipping over any issues without crashing

app/views/tracks/show/unjoined.html.haml
```diff
@@ -7,8 +7,7 @@
       .info
         %h1= @track.title
         %p
-          / TODO: Add real data
-          C# is a modern, object-oriented language with lots of great features, such as type-inference and async/await. The tooling is excellent, and there is extensive, well-written documentation.
+          = @track.blurb
```
iHiD: You can merge this with the LOC above:
```%p= @track.blurb```

File: db/migrate/20200504163928_create_tracks.rb [hunk 5]
```diff
@@ -3,6 +3,7 @@ def change
     create_table :tracks do |t|
       t.string :slug, null: false, unique: true
       t.string :title, null: false
+      t.string :blurb, null: false, :limit => 400
```
iHiD: Use colon syntax: `limit: 400`

ErikSchierboom: Ah, I "borrowed" this from another file. Will update

app/views/tracks/show/unjoined.html.haml
ErikSchierboom: Doh, of course

app/models/git/track.rb
```diff
@@ -22,7 +22,11 @@ def ignore_regexp
     end
 
     def config(commit: repo.head_commit)
-      repo.read_json_blob(commit, "languages/#{slug}/config.json")
+      repo.read_json_blob(commit, config_file)
+    end
+
+    def config_file
```
iHiD:
```suggestion
    def config_filepath
```

app/commands/git/sync_track.rb
```diff
@@ -0,0 +1,46 @@
...
+      head_config = track.git.config(commit: head_commit)
+
+      track.blurb = head_config[:blurb]
+      track.active = head_config[:active]
+      track.title = head_config[:language]
+
+      # TODO: validate track to prevent invalid track data
+
+      track.git_sha = head_commit.oid if track.changed?
+      track.save
```
iHiD: A couple of thoughts here.

1. You probably want to use `save!` not `save` to raise an exception if it fails.

2. You might want to use an update block:

```suggestion
      track.update!(
        blurb: head_config[:blurb]
        active: head_config[:active]
        title: head_config[:language]
        git_sha: head_commit.oid
      )
```

app/models/exercise.rb
```diff
@@ -47,7 +47,6 @@ def prerequisite_exercises
 
   memoize
   def git
-    # TODO: Change to sha, not HEAD
-    Git::Exercise.new(track.slug, slug, "HEAD", git_type)
+    Git::Exercise.new(track.slug, slug, git_sha, git_type)
```
iHiD: FYI, the reason this is currently like this is so that we always show the HEAD to maintainers, as I've not added in the ability for them to update their solutions yet. We might want to keep it that way for now?

test/commands/git/sync_track_test.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    refute track.changed?
+  end
+
+  test "git SHA changes to HEAD SHA when there are changes" do
+    track = create :track, slug: 'fsharp', active: false, git_sha: "f290a29144b93b21e2399cd532b22562d83b6a52"
+
+    Git::SyncTrack.(track)
+
+    assert_equal "70a8473aa074c1ee9c843665347481fe5c8ee95f", track.git_sha
```
iHiD: I wonder if it's worth on this line reading the `"70a..."` from git itself. Otherwise whenever the repo changes in the tests, this test is going to break.

app/commands/git/sync_track.rb
```diff
@@ -0,0 +1,46 @@
...
+    def call
+      lookup_head_and_current_commit
+      update_track! if track_config_modified?
+    end
+
+    private
+    attr_reader :current_commit, :head_commit
+
+    def lookup_head_and_current_commit
+      track.git.send(:repo).update!
```
iHiD: We shouldn't use `send()` in production code. So it's probably best to just make `update!` public.

iHiD: > We shouldn't use send() in production code. 

To clarify, we shouldn't use it to access private methods. There are other good reasons to actually use it :)

app/models/exercise.rb
ErikSchierboom: Sure!

app/commands/git/sync_exercise.rb
```diff
@@ -0,0 +1,92 @@
...
+      @head_commit = exercise.git.head_commit
+    end
+
+    def exercise_synced_to_head?
+      current_commit.oid == head_commit.oid
+    end
+
+    def update_exercise!
+      update_concept_exercise! if exercise.concept_exercise?
+      update_practice_exercise! if exercise.practice_exercise?
```
iHiD: This is a good example of where you can use `send`:
```suggestion
      send("update_#{exercise.git_type}_exercise!")
```

app/commands/git/sync_exercise.rb
```diff
@@ -0,0 +1,92 @@
...
+        exercise.update!(
+          slug: config_exercise[:slug],
+          title: config_exercise[:name],
+          git_sha: head_commit.oid,
+          synced_to_git_sha: head_commit.oid
+        )
+      else
+        exercise.update!(
+          synced_to_git_sha: head_commit.oid
+        )
```
iHiD: I'd probably treat this more like a guard:

```ruby
def update_concept_exercise!
  return exercise.update!(synced_to_git_sha: head_commit.oid) unless exercise.modifed?

  ...
```

app/commands/git/sync_exercise.rb
```diff
@@ -0,0 +1,92 @@
...
+        config_exercise[:name] != exercise.title ||
+        config_exercise[:deprecated] != exercise.deprecated
+    end
+
+    def exercise_files_modified?
+      return false if current_commit.oid == head_commit.oid
+
+      diff.each_delta.any? do |delta|
+        delta.old_file[:path] == exercise.track.git.config_filepath ||
+          delta.new_file[:path] == exercise.track.git.config_filepath
```
iHiD: Is this the wrong filepath?

app/models/exercise.rb
```diff
@@ -47,7 +47,6 @@ def prerequisite_exercises
 
   memoize
   def git
```
iHiD, line 53: I think we should keep this private and have a method for `update_git!` that calls `git.update!`. I feel like other things shouldn't need to access this thing's git info.

iHiD:
Looking good so far! :)

app/commands/git/sync_exercise.rb
ErikSchierboom: No, because some of the track metadata is defined in the track's config file

ErikSchierboom: It is. Work in progress :)

app/models/exercise.rb
ErikSchierboom: I understand. There are quite some things that I will need to access though:

- Updating git
- Getting the HEAD commit
- Getting the current commit
- Getting the non_ignored_filepaths

Does it make sense to still do this via the `Exercise` class?

app/commands/git/sync_concept.rb
```diff
@@ -0,0 +1,62 @@
...
+    private
+    attr_reader :synced_git_concept, :head_git_concept, :head_git_track
+
+    def fetch_git_data!
+      git_repo = Git::Repository.new(concept.track.slug, repo_url: concept.track.repo_url)
+      git_repo.update!
+
+      @synced_git_concept = Git::Concept.new(concept.track.slug, concept.slug, concept.synced_to_git_sha, repo: git_repo)
+      @head_git_concept = Git::Concept.new(concept.track.slug, concept.slug, git_repo.head_sha, repo: git_repo)
+      @head_git_track = Git::Track.new(concept.track.slug, git_repo.head_sha, repo: git_repo)
```
iHiD: I'd suggest having these as memoized methods, rather than instance variables with attr readers.

app/commands/git/sync_concept.rb
```diff
@@ -0,0 +1,62 @@
...
+    def concept_modified?
+      return false unless track_config_modified?
+
+      config_concept[:slug] != concept.slug ||
+        config_concept[:name] != concept.name ||
+        config_concept[:blurb] != concept.blurb
+    end
+
+    def track_config_modified?
+      return false if synced_git_concept.commit.oid == head_git_concept.commit.oid
```
iHiD:
```suggestion
      return false if synced_to_head? 
```

app/commands/git/sync_concept.rb
```diff
@@ -0,0 +1,43 @@
+module Git
+  class SyncConcept < Sync
+    include Mandate
+
+    def initialize(concept)
+      super(concept.track, concept.synced_to_git_sha)
```
iHiD, line 6:
```suggestion
      super(concept.track, concept.synced_to_git_sha)
      
```

app/commands/git/sync.rb
```diff
@@ -0,0 +1,54 @@
...
+    def head_git_track
+      Git::Track.new(track.slug, git_repo.head_sha, repo: git_repo)
+    end
+
+    memoize
+    def synced_git_track
+      Git::Track.new(track.slug, synced_to_git_sha, repo: git_repo)
+    end
+
+    def synced_to_head?
```
iHiD, line 28:
```suggestion
    memoize
    def synced_to_head?
```

app/commands/git/sync.rb
```diff
@@ -0,0 +1,54 @@
...
+    memoize
+    def track_config_modified?
+      filepath_in_diff?(head_git_track.config_filepath)
+    end
+
+    def filepath_in_diff?(filepath)
+      diff.each_delta.any? do |delta|
+        [delta.old_file[:path], delta.new_file[:path]].include?(filepath)
+      end
+    end
```
iHiD: I'd probably move this above the memoized methods.

app/commands/git/sync.rb
```diff
@@ -0,0 +1,54 @@
...
+      sync! unless synced_to_head?
+    end
+
+    def sync!
+      raise NotImplementedError
+    end
+
+    def update_git_repo!
+      git_repo.update!
+    end
```
iHiD: You could probably inline this now. Katrina definitely would. I don't mind either way :)

ErikSchierboom: I prefer not to to keep everything at the same "level":

```ruby
def call
  update_git_repo!
  sync! unless synced_to_head?
end
```

to me reads better than:

```ruby
def call
  git_repo.update!
  sync! unless synced_to_head?
end
```

The dotting into the `git_repo` field to me seems to suggest more of an implementation detail than the first version.

app/models/git/exercise.rb
```diff
@@ -83,7 +94,11 @@ def file_entries
     def tree
       # TODO: When things are exploded back into repos, do this
       # repo.fetch_tree(commit, "exercises/#{exercise_type}/#{slug}")
-      repo.fetch_tree(commit, "languages/#{track_slug}/exercises/#{exercise_type}/#{exercise_slug}")
+      repo.fetch_tree(commit, dir)
+    end
+
+    def dir
+      "languages/#{track_slug}/exercises/#{exercise_type}/#{exercise_slug}"
```
iHiD, line 109:
```suggestion
      # TODO: Needs changing once we're out of the monorepo
      "languages/#{track_slug}/exercises/#{exercise_type}/#{exercise_slug}"
```

test/factories/track/concepts.rb
```diff
@@ -3,8 +3,10 @@
     track
     uuid { SecureRandom.uuid }
     slug { "concept_#{SecureRandom.hex(4)}" }
+    blurb { "Description of #{slug.titleize}" }
     name { slug.titleize }
     blurb { "A cracking concept" }
```
iHiD:
```suggestion
```

db/migrate/20200505152413_create_exercises.rb
```diff
@@ -3,7 +3,7 @@ def change
     create_table :exercises do |t|
       t.belongs_to :track, foreign_key: true, null: false
 
-      t.string :uuid, null: false
+      t.string :uuid, null: false, index: { unique: true }
```
iHiD, line 6: Be aware that this currently will block ETL, as exercises in different tracks share the same UUIDs.

It's the right thing to do for sure, but just will break for practice exercises.

ErikSchierboom: Want me to temporary make it such that it is UUID/track that is unique?

app/commands/concept_exercise/create.rb
```diff
@@ -0,0 +1,19 @@
+class ConceptExercise
+  class Create
+    include Mandate
+
+    initialize_with :uuid, :slug, :title, :taught_concepts, :prerequisites, :deprecated, :git_sha, :synced_to_git_sha, :track
```
iHiD: I'd probably pass in the `uuid`, `track`, and then an `attributes` hash for the rest. You can then do `e.attributes = attributes` in the block on L9.

iHiD: (and the same comment on other places that do similar)

app/commands/track/create.rb
```diff
@@ -0,0 +1,17 @@
+class Track
+  class Create
+    include Mandate
+
+    initialize_with :slug, :title, :blurb, :repo_url, :synced_to_git_sha, :tags
+
+    def call
+      ::Track.create_or_find_by!(slug: slug) do |t|
```
iHiD: You don't need the `::` now as you're not in the Git scope.

app/commands/git/sync_track.rb
```diff
@@ -0,0 +1,86 @@
...
+          exercise[:uuid],
+          exercise[:slug],
+          exercise[:name], # TODO: the DB used title, config.json used name. Consider if we want this
+          find_concepts(exercise[:concepts]),
+          find_concepts(exercise[:prerequisites]),
+          exercise[:deprecated] || false,
+          head_git_track.commit.oid,
+          head_git_track.commit.oid,
+          track
+        )
```
iHiD: So with my other comment, this would be:

```suggestion
        ::ConceptExercise::Create.(
          track,
          exercise[:uuid],
          slug: exercise[:slug],
          name: exercise[:name], # TODO: the DB used title, config.json used name. Consider if we want this
          concepts: find_concepts(exercise[:concepts]),
          prerequisites: find_concepts(exercise[:prerequisites]),
          deprecated: exercise[:deprecated] || false,
          synced_to_git_sha: head_git_track.commit.oid
        )
```

app/commands/concept_exercise/create.rb
iHiD:
```suggestion
    initialize_with :track, :uuid, :attributes
```

app/commands/concept_exercise/create.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def call
+      ::ConceptExercise.create_or_find_by!(uuid: uuid, track: track) do |e|
+        e.slug = slug
+        e.title = title
+        e.taught_concepts = taught_concepts
+        e.prerequisites = prerequisites
+        e.deprecated = deprecated
+        e.git_sha = git_sha
+        e.synced_to_git_sha = synced_to_git_sha
```
iHiD:
```suggestion
        e.attributes = attributes
        e.git_sha = e.synced_to_git_sha
```

iHiD: You might also want to consider moving that last line to into the `before_create` of the `Track` itself. We can discuss pros/cons of that together tomorrow :)

app/commands/git/sync_track.rb
```diff
@@ -46,11 +46,15 @@ def concepts
 
     def concept_exercises
       config_concept_exercises.map do |exercise|
+        next unless exercise[:uuid] # TODO: decide if we want to allow null as the uuid
```
iHiD:
```suggestion
        next unless exercise[:uuid].present? # TODO: decide if we want to allow null as the uuid
```

Guard against empty strings with `present?`

app/commands/git/sync.rb
```diff
@@ -0,0 +1,72 @@
+module Git
+  class Sync
+    include Mandate
+    initialize_with :track, :synced_to_git_sha
```
iHiD, line 5:
```suggestion

    initialize_with :track, :synced_to_git_sha
```

app/commands/git/sync.rb
```diff
@@ -0,0 +1,72 @@
...
+      Git::Repository.new(track.slug, repo_url: track.repo_url)
+    end
+
+    memoize
+    def head_git_track
+      Git::Track.new(track.slug, git_repo.head_sha, repo: git_repo)
+    end
+
+    memoize
+    def synced_git_track
```
iHiD: I'm finding this name hard to reason about. Maybe:

```suggestion
    def current_git_track
```

app/commands/git/sync.rb
```diff
@@ -0,0 +1,72 @@
...
+      filepath_in_diff?(head_git_track.config_filepath)
+    end
+
+    memoize
+    def diff
+      head_git_track.commit.diff(synced_git_track.commit)
+    end
+
+    memoize
+    def config_concept_exercises
```
iHiD: How about we switch the `config` on these to be a suffix?
```suggestion
    def concept_exercises_config
```

app/commands/git/sync.rb
```diff
@@ -0,0 +1,72 @@
+module Git
+  class Sync
+    include Mandate
+    initialize_with :track, :synced_to_git_sha
+
+    def call
+      sync! unless synced_to_head?
```
iHiD: I'm trying to work out the logic of this a bit. 

[Further down](https://github.com/exercism/v3-website/pull/291/files#diff-571fbde4167b7daec296052a69f079a05369faa3147dc34bdedce25eb6e60062R15) we do this: ```return concept.update!(synced_to_git_sha: head_git_concept.commit.oid) unless concept_needs_updating?```

But that will never get called if `synced_to_head?` here is true.

I can't quite work out if that's correct or not etc.

I'm wondering if we actually want this guard in this parent class, or if it should be down to each child class to do it' own guard.

What do you think?

app/commands/git/sync_concept_exercise.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    def find_concepts(slugs)
+      slugs.map do |slug|
+        config_concept = config_concepts.find { |e| e[:slug] == slug }
+        ::Track::Concept.find_by!(uuid: config_concept[:uuid])
+      end
+    end
+
+    memoize
+    def config_exercise
```
iHiD: Definitely feel like this would be better named `exercise_config`
```suggestion
    def exercise_config
```

app/commands/git/sync_concept.rb
```diff
@@ -0,0 +1,44 @@
...
+    def concept_needs_updating?
+      return false unless track_config_modified?
+
+      config_concept[:slug] != concept.slug ||
+        config_concept[:name] != concept.name ||
+        config_concept[:blurb] != concept.blurb
+    end
+
+    memoize
+    def config_concept
```
iHiD:
```suggestion
    def concept_config
```

app/commands/git/sync_practice_exercise.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def find_concepts(slugs)
+      slugs.map do |slug|
+        config_concept = config_concepts.find { |e| e[:slug] == slug }
+        ::Track::Concept.find_by!(uuid: config_concept[:uuid])
+      end
+    end
+
+    memoize
+    def config_exercise
```
iHiD:
```suggestion
    def exercise_config
```

app/commands/git/sync_track.rb
```diff
@@ -0,0 +1,103 @@
...
+  class SyncTrack < Sync
+    include Mandate
+
+    def initialize(track)
+      super(track, track.synced_to_git_sha)
+      @track = track
+    end
+
+    def call
+      update_git_repo!
```
iHiD: I think we should rename this `update` to `fetch` all the way through things for clarity.

db/migrate/20200505152413_create_exercises.rb
iHiD: No, I think it'll be a good way to see failures when we pull the practice exercises in.

app/commands/git/sync_track.rb
```diff
@@ -0,0 +1,103 @@
...
+
+        ::PracticeExercise::Create.(
+          exercise[:uuid],
+          track,
+          slug: exercise[:slug],
+          title: exercise[:slug].titleize, # TODO: what to do with practice exercise names?
+          prerequisites: find_concepts(exercise[:prerequisites]),
+          deprecated: exercise[:deprecated] || false,
+          git_sha: head_git_track.commit.oid,
+          synced_to_git_sha: head_git_track.commit.oid
```
iHiD: I think this line should move into a `before_create` block in Exercise.

```ruby
self.synced_to_git_sha = git_sha unless self.synced_to_git_sha
```

iHiD:
I few things as I read it:

- [x] I've made a couple of wide-reaching naming comments:
  - [x] Changing "update" to "fetch" (as update signfies some DB update to me)
  - [x] Changing from prefix to suffix for `config`
- [x] I think as well in `Sync` there are certain methods that should be `private`, so ensuring those are at the bottom and adding the `private` keyword. Check the other classes to make sure this is correct too pls.
- [x] I think it would be worth adding a few lines of comment to the top of the `Sync` class that explain how everything pieces together. As there's a lot of moving parts here :)
- [x] It's red.

But fundamentally I think it's basically there and that you've done a really great job 🦜

iHiD:
Brilliant work. Let's see what happens!!

PR #297, opened 2020-11-20
Revert to simpler localstorage hook

### LINKED ISSUE #289: Empty files cause infinite loop
@kntsoriano We have an issue where if the initial stub is empty, or you create an empty submission then refresh, stuff goes wrong.

See this video:
https://files.internetwolke.ovh/2020/11/ydncVtcAPD.mp4

I think there’s an issue with the local storage stuff. Maybe because `""` in JS is `false`, not `true` and some something is happening that you’re not expecting?

The Julia track has lots of empty stubs to experiment with (until @SaschaMann implements them :)). I've updated Julia/Lasagna as part of diagnosing this though.

kntsoriano:
Closes https://github.com/exercism/v3-website/issues/289.

When using the`useMemory` hook, it somehow changes `''` to `null`, which causes the revert code to not run properly. I'm unsure what causes this, but I decided to use the `localStorage` for now in order for us to not have this problem.

### DIFF
--- a/app/javascript/components/student/editor/FileEditor.tsx
+++ b/app/javascript/components/student/editor/FileEditor.tsx
@@ -9,7 +9,7 @@ import React, {
 import { File } from '../Editor'
 import { ExercismMonacoEditor } from './ExercismMonacoEditor'
 import * as monacoEditor from 'monaco-editor/esm/vs/editor/editor.api'
-import { useStorage } from '../../../utils/use-storage'
+import { useLocalStorage } from '../../../utils/use-storage'
 
 export type FileEditorHandle = {
   getFile: () => File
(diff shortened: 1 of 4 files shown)


app/javascript/components/student/editor/FileEditor.tsx
```diff
@@ -59,7 +59,7 @@ export const FileEditor = forwardRef<FileEditorHandle, FileEditorProps>(
     )
     const revertContent = useCallback(
       (e) => {
-        setContent(file.content)
+        editorRef.current?.setValue(file.content)
```
kntsoriano, line 62: This was also a crucial line as Monaco editor doesn't behave like a controlled component.

ErikSchierboom (approved):
I'll leave this to @SleeplessByte who has much more knowledge about this :)

iHiD:
This is red, @kntsoriano

SleeplessByte:
I'll check out the hook and get back to you. Can merge this now, and I'll ping ya later 👍

PR #301, opened 2020-11-23
Allow user to toggle between default / vim / emacs keybindings

kntsoriano:
## Description
This PR allows the user to toggle between keybindings.

### DIFF
--- a/app/javascript/components/student/editor/ExercismMonacoEditor.tsx
+++ b/app/javascript/components/student/editor/ExercismMonacoEditor.tsx
@@ -12,7 +12,7 @@ import {
 import normalizeUrl from 'normalize-url'
 import ReconnectingWebsocket from 'reconnecting-websocket'
 import { v4 as uuidv4 } from 'uuid'
-import { initVimMode } from 'monaco-vim'
+import { initVimMode, VimMode } from 'monaco-vim'
 
 export type FileEditorHandle = {
   getFile: () => File
@@ -22,6 +22,11 @@ type FileEditorProps = {
   file: File
   language: string
   onRunTests: () => void
+}
+
+export enum Keybindings {
+  DEFAULT = 'default',
+  VIM = 'vim',
 }
 
 const SAVE_INTERVAL = 500
@@ -35,6 +40,7 @@ export function ExercismMonacoEditor({
   options,
   value,
   theme,
+  keybindings,
 }: {
   width: string
   height: string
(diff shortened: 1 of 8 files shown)


app/javascript/components/student/editor/ExercismMonacoEditor.tsx
```diff
@@ -107,18 +123,47 @@ export function ExercismMonacoEditor({
...
+      <MonacoEditor
+        width="800"
+        height="600"
+        language={language}
+        editorDidMount={handleEditorDidMount}
+        options={options}
+        value={value}
+        theme={theme}
+      />
+      <div ref={statusBarRef}></div>
```
kntsoriano, line 166: This is needed as the vim and emacs editors have a status bar underneath.

app/javascript/declarations.d.ts
```diff
@@ -28,3 +28,18 @@ declare module 'reconnecting-websocket' {
   const ReconnectingWebsocket: ReconnectingWebsocket
   export = ReconnectingWebsocket
 }
+
```
kntsoriano, line 31: I needed to declare my own type declarations for the vim plugin. There are no defaults.

app/javascript/components/student/editor/__mocks__/ExercismMonacoEditor.tsx
```diff
@@ -1,5 +1,11 @@
 import React, { useEffect, useRef } from 'react'
 
+export enum Keybindings {
```
kntsoriano, line 3: Is there any other way to do this? I've mocked this module since jest doesn't work well with Monaco, but I need this enum defined as well.

Is it a good idea instead to declare a global enums file or something like that?

iHiD: (@SleeplessByte)

ErikSchierboom (approved):
LGTM!

app/javascript/components/student/editor/__mocks__/ExercismMonacoEditor.tsx
SleeplessByte: What is "this"?

I can't quite figure out what you need, or what doesn't work.

kntsoriano: I've mocked this component out in our tests, as how Monaco imports modules doesn't work well with Jest yet. The only problem is that within this component, I define this `Keybindings` enum. I end up with two versions of the enum which is in the actual component, and within the mocked component.

Do we have a way of defining a global enum or something like that? Or is it okay to do it this way still?

SleeplessByte: I'd probably declare it in a different file (module) and import it in both places, but I might be missing why that doesn't work.

PR #305, opened 2020-11-25
Rename Exercise::Complete to Solution::Complete

iHiD:


app/commands/solution/complete.rb
```diff
@@ -37,17 +37,14 @@ def record_activity!
       Rails.logger.error e.message
     end
 
-    def guard!
-      raise SolutionNotFoundError unless solution
-      raise UserTrackNotFoundError unless user_track
-    end
```
ErikSchierboom, line 43: The fact that these are now gone is a strong indicator that is a good refactoring! :)

ErikSchierboom (approved):

app/commands/solution/complete.rb
iHiD: Yes! Good point :D

PR #306, opened 2020-11-25
Style notification widget

iHiD:
This adds the two notifications (top-right) and also the ":+1: Joined" faux-button.

Screenshot 2020-11-25 at 21 40 58

Screenshot 2020-11-25 at 21 34 50

@kntsoriano or @ErikSchierboom: I changed `Notification::Icon` to `Notification::Widget`. I'd actually like it to be `NotificationIcon` (without the namespacing). Could one of you change that pls and push a commit to this with it working pls?


app/css/layout.css
```diff
@@ -16,3 +16,8 @@
     max-width: 1440px;
     padding: 0 48px;
 }
+
+.c-react-component {
+    width: fit-content;
+    height: fit-content;
+}
```
iHiD: @SleeplessByte @ErikSchierboom This might go some way to fixing the need for the awkward-as-fuck React vs non-react component problem thing.

kntsoriano (approved):

app/css/layout.css
ErikSchierboom: Ooh, nice!

ErikSchierboom:
The styling looks a bit off for the tracks page (all pulled to the left):

image

Is that intentional?

iHiD:
> Is that intentional?

No. That's because the `fit-content` doesn't work. Sadly.

ErikSchierboom:
@kntsoriano I've worked on the renaming of the Notification::Icon component.

app/css/layout.css
SleeplessByte: I think [this will be fine](https://caniuse.com/intrinsic-width).

ErikSchierboom:
@iHiD There is a merge conflict.

PR #311, opened 2020-11-26
Assign correct CSS class to notification widget

kntsoriano:
Closes https://github.com/exercism/v3-website/issues/307.

### DIFF
--- a/app/javascript/components/notifications/Widget.tsx
+++ b/app/javascript/components/notifications/Widget.tsx
@@ -1,4 +1,4 @@
-import React, { useReducer, useEffect } from 'react'
+import React, { useReducer, useEffect, useMemo } from 'react'
 import consumer from '../../utils/action-cable-consumer'
 import { Icon } from '../common/Icon'
 
@@ -20,7 +20,18 @@ function reducer(state: WidgetProps, action: WidgetAction) {
 
 export function Widget({ count }: WidgetProps) {
   const [state, dispatch] = useReducer(reducer, { count: count })
-  const isUnread = state.count > 0
+  const variantClass = useMemo(() => {
+    switch (true) {
+      case state.count === 0:
+        return '--none'
+      case state.count >= 1 && state.count <= 9:
+        return '--digital'
+      case state.count >= 10 && state.count <= 99:
+        return '--double-digit'
+      case state.count >= 100:
+        return '--triple-digit'
+    }
+  }, [state.count])
 
   useEffect(() => {
     const received = (data: WidgetAction) => {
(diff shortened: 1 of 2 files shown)


app/javascript/components/notifications/Widget.tsx
```diff
@@ -20,7 +20,18 @@ function reducer(state: WidgetProps, action: WidgetAction) {
...
+    switch (true) {
+      case state.count === 0:
+        return '--none'
+      case state.count >= 1 && state.count <= 9:
+        return '--digital'
+      case state.count >= 10 && state.count <= 99:
+        return '--double-digit'
+      case state.count >= 100:
+        return '--triple-digit'
+    }
```
ErikSchierboom, line 33: I've never seen this pattern before! Very interesting. What made you use this over regular if/else if statements?

test/javascript/components/notifications/Widget.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+  component = render(<Widget count={10} />)
+  expect(component.container.firstChild).toHaveClass('--double-digit')
+
+  component = render(<Widget count={99} />)
+  expect(component.container.firstChild).toHaveClass('--double-digit')
+
+  component = render(<Widget count={100} />)
+  expect(component.container.firstChild).toHaveClass('--triple-digit')
+})
```
ErikSchierboom, line 26: Maybe it's me, but I think I would prefer to have separate tests for the different cases.

ErikSchierboom (approved):

iHiD:
Gonna merge this into my PR. Thanks @kntsoriano! I'll leave @ErikSchierboom's comments to be discussed separately :)

app/javascript/components/notifications/Widget.tsx
SleeplessByte: You can use Math too if you'd like:

```typescript
const CLASSES = ['--none', '--digital', '--double-digit', '--triple-digit']

...

const digits = Math.log(state.count) * Math.LOG10E + 1 | 0;
return CLASSES[digits]  || CLASSES[CLASSES.length - 1]
```

test/javascript/components/notifications/Widget.test.js
SleeplessByte: It should have separate tests for separate cases.

app/javascript/components/notifications/Widget.tsx
iHiD: Please add a comment if you do this!

SleeplessByte: Hah yes!

kntsoriano: @ErikSchierboom I just like how every case is laid out nicely. But I don't have any preference if we go back to use if / else statements.

test/javascript/components/notifications/Widget.test.js
kntsoriano: I will separate this to different test cases.

PR #313, opened 2020-11-26
Fix concept map test

### LINKED ISSUE #312: Console error in ConceptMap
@neenjaw I get this when compiling the assets:

```
js-tests |     console.error
js-tests |       Warning: An update to Concept inside a test was not wrapped in act(...).
js-tests |       
js-tests |       When testing, code that causes React state updates should be wrapped into act(...):
js-tests |       
js-tests |       act(() => {
js-tests |         /* fire events that update state */
js-tests |       });
js-tests |       /* assert on the output */
js-tests |       
js-tests |       This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
js-tests |           in Concept (at Concept.tsx:103)
js-tests |           in div (at Concept.tsx:78)
js-tests |           in Concept (at ConceptMap.tsx:66)
js-tests |           in div (at ConceptMap.tsx:57)
js-tests |           in div (at ConceptMap.tsx:55)
js-tests |           in figure (at ConceptMap.tsx:54)
js-tests |           in ConceptMap (at ConceptMap.test.tsx:93)
js-tests | 
js-tests |       at printWarning (node_modules/react-dom/cjs/react-dom.development.js:88:30)
js-tests |       at error (node_modules/react-dom/cjs/react-dom.development.js:60:5)
js-tests |       at warnIfNotCurrentlyActingUpdatesInDEV (node_modules/react-dom/cjs/react-dom.development.js:23284:7)
js-tests |       at setState (node_modules/react-dom/cjs/react-dom.development.js:15656:9)
js-tests |       at node_modules/react-query/src/react/utils.ts:46:11
```

neenjaw:
The concept  map component starts hidden, then unhides itself when it
detects that all styles have been loaded. This wasn't being accounted
for in the tests. So now it is using a `waitFor` promise to allow the
component to reach a steady state before testing the assertion.

close #312

### DIFF
--- a/test/javascript/components/concept-map/ConceptMap.test.tsx
+++ b/test/javascript/components/concept-map/ConceptMap.test.tsx
@@ -3,108 +3,90 @@ import React from 'react'
 import '@testing-library/jest-dom'
 
 // Test deps
-import {
-  getByText,
-  getByTitle,
-  queryByTitle,
-  render,
-} from '@testing-library/react'
+import { screen, render, waitFor } from '@testing-library/react'
 import '@testing-library/jest-dom/extend-expect'
 
 // Component
 import { ConceptMap } from '../../../../app/javascript/components/concept-map/ConceptMap'
-import {
-  ConceptStatus,
-  IConcept as Concept,
-  ConceptConnection,
-  ConceptLayer,
-} from '../../../../app/javascript/components/concept-map/concept-map-types'
 
 describe('<ConceptMap />', () => {
   test('renders empty component', () => {
-    const { container } = renderMap([], [], [], {}, {})
+    const { container } = render(
+      <ConceptMap
+        concepts={[]}
+        levels={[[]]}
+        connections={[]}
+        status={{}}
+        exercise_counts={{}}
+      />
+    )
     const map = container.querySelector('.c-concepts-map')
     expect(map).not.toBeNull()
   })
 
-  test('renders single incomplete concept', () => {
+  test('renders single incomplete concept', async () => {
     const testConcept = concept('test')
-    const { container } = renderMap(
-      [testConcept],
-      [[testConcept.slug]],
-      [],
-      {
-        test: 'unavailable',
-      },
-      {
-        test: {
-          exercises: 1,
-          exercises_completed: 0,
-        },
-      }
+
+    render(
+      <ConceptMap
+        concepts={[testConcept]}
+        levels={[[testConcept.slug]]}
+        connections={[]}
+        status={{
+          test: 'unavailable',
+        }}
+        exercise_counts={{
+          test: {
+            exercises: 1,
+            exercises_completed: 0,
+          },
+        }}
+      />
     )
-    const conceptEl = getByText(container, 'Test')
-    expect(conceptEl).toBeInTheDocument()
-    const completeIconEl = queryByTitle(
-      container,
-      'You have mastered this concept'
-    )
-    expect(completeIconEl).not.toBeInTheDocument()
+
+    await waitFor(() => {
+      expect(screen.getByText('Test')).toBeInTheDocument()
+      expect(
+        screen.queryByTitle('You have mastered this concept')
+      ).not.toBeInTheDocument()
+    })
   })
 
-  test('renders single completed concept', () => {
+  test('renders single completed concept', async () => {
     const testConcept = concept('test')
-    const { container } = renderMap(
-      [testConcept],
-      [[testConcept.slug]],
-      [],
-      {
-        test: 'completed',
-      },
-      {
-        test: {
-          exercises: 1,
-          exercises_completed: 1,
-        },
-      }
+    render(
+      <ConceptMap
+        concepts={[testConcept]}
+        levels={[[testConcept.slug]]}
+        connections={[]}
+        status={{
+          test: 'completed',
+        }}
+        exercise_counts={{
+          test: {
+            exercises: 1,
+            exercises_completed: 1,
+          },
+        }}
+      />
     )
-    const conceptEl = getByText(container, 'Test')
-    expect(conceptEl).toBeInTheDocument()
-    const completeIconEl = getByTitle(
-      container,
-      'You have mastered this concept'
-    )
-    expect(completeIconEl).toBeInTheDocument()
+
+    await waitFor(() => {
+      expect(screen.getByText('Test')).toBeInTheDocument()
+      expect(
+        screen.getByTitle('You have mastered this concept')
+      ).toBeInTheDocument()
+    })
   })
 })
 
-const concept = (conceptName: string): Concept => {
+const concept = (conceptName: string) => {
   return {
     slug: conceptName,
     name: slugToTitlecase(conceptName),
     webUrl: `link-for-${conceptName}`,
     tooltipUrl: `tooltop-link-for${conceptName}`,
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -13,7 +13,7 @@
     "strict": true,
     "noEmit": true,
     "allowSyntheticDefaultImports": true,
-    "types": ["webpack-env", "app/javascript/declarations"]
+    "types": ["webpack-env", "app/javascript/declarations", "jest"]
```
iHiD: Does this mean we "leak" Jest into our production compile?

SleeplessByte: You don't leak `types` :)

SleeplessByte: @neenjaw instead you want to create a `tsconfig.json` inside the tests folder that **extends** the one in the root, that adds these types, otherwise `jest` types are available in non-test files.

Same for `includes`/`excludes`, you want test files to be able to see (import) app files, but not the other way around.

neenjaw: :+1:

neenjaw: I think this is what you meant? (see commit changes)

test/javascript/components/concept-map/ConceptMap.test.tsx
```diff
@@ -1,105 +1,94 @@
 // Deps
 import React from 'react'
+import '@testing-library/jest-dom'
```
SleeplessByte, line 3: It's fine to have this in every test file, but you probably want it in a [`setupFilesAfterEnv`](https://jestjs.io/docs/en/configuration#setupfilesafterenv-array).

test/javascript/components/concept-map/ConceptMap.test.tsx
```diff
@@ -1,105 +1,94 @@
...
+          test: {
+            exercises: 1,
+            exercises_completed: 0,
+          },
+        }}
+      />
     )
-    expect(completeIconEl).toBeNull()
+
+    await waitFor(() => {
```
SleeplessByte: Does the bottom one (`"You have mastered this concept"`) appear on initial mount but disappear after? Otherwise it shouldn't be in this wait for. You might want to add a comment _why_ this waitFor is necessary for future readers.

test/javascript/components/concept-map/ConceptMap.test.tsx
```diff
@@ -1,105 +1,94 @@
...
+        exercise_counts={{
+          test: {
+            exercises: 1,
+            exercises_completed: 1,
+          },
+        }}
+      />
     )
+
+    await waitFor(() => {
```
SleeplessByte: Same

test/javascript/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
```
SleeplessByte, line 1: Great

tsconfig.json
```diff
@@ -15,6 +15,12 @@
     "allowSyntheticDefaultImports": true,
     "types": ["webpack-env", "app/javascript/declarations"]
   },
-  "exclude": ["**/*.spec.ts", "node_modules", "vendor", "public"],
+  "exclude": [
+    "**/*.spec.ts",
```
SleeplessByte, line 19: If you want the _exact_ same, you probably want `test/javascript/**/*.spec.ts`, but I am cool with this.

SleeplessByte (commented):
I think this will work, but seomoen should check in their IDE and running jest 💯

test/javascript/components/concept-map/ConceptMap.test.tsx
neenjaw: The title comes from the "completed" svg checkmark, so it shouldn't appear on a concept that isn't complete (the second test) and should appear when it is complete (the third test)

neenjaw: So they all appear when the concept is rendered, but doesn't appear visible at first

SleeplessByte: If it's not visible, do you need to wait for it to not be visible?

neenjaw: So the only test indicator that a concept is complete or not is the svg checkmark and I wanted to write a test to make sure that incomplete concepts don't have an svg checkmark.  Should I not do that?

SleeplessByte: Okay. I get it now:

1. the `waitFor` you need because it might still be triggering re-renders. Only put inside the `waitFor` what will be the _end_ state. In this case, the thing is now visible.
2. then you want a "invariant" check. It will be a regular assertion that checks that the icon is NOT visible.
3. now you do the interaction
4. now you _wait for_ the interaction to complete
5. if the _wait for_ isn't checking that the icon becomes visible, you should assert here, otherwise you're done.

So basically what I'm saying is: separate the code that waits for the component to _come to rest_ from the code that is testing an invariant. Your idea is solid.

iHiD:
I'm not going to keep up with your cool kids and your JS, so I'll unsubscribe and y'all just tag me when this is good to merge 💙

test/javascript/components/concept-map/ConceptMap.test.tsx
```diff
@@ -8,78 +8,82 @@ import '@testing-library/jest-dom/extend-expect'
...
+      concepts: [testConcept],
+      levels: [[testConcept.slug]],
+      connections: [],
+      status: { test: 'unavailable' },
+      exercise_counts: { test: { exercises: 1, exercises_completed: 0 } },
+    }
+
+    await waitForConceptMapReady(config)
+
+    expect(
```
SleeplessByte, line 43: You probably want a positive expectation here too. (e.g. expect that the concept is visible) -- even if `waitForConceptMapReady` already checks that. That will allow you to _change_ the control logic, but keep the test coherent/the same.

test/javascript/components/concept-map/ConceptMap.test.tsx
```diff
@@ -8,78 +8,82 @@ import '@testing-library/jest-dom/extend-expect'
...
+      screen.queryByTitle('You have mastered this concept')
+    ).toBeInTheDocument()
   })
 })
 
+const waitForConceptMapReady = async (config: IConceptMap) => {
+  const renderResult = render(
```
SleeplessByte: Is this renderResult necessary? I'd either expect this to be called `renderConceptMap` or take in a rendered concept map and replace `screen.getByText` to use the `render.getByText`

neenjaw: I had just named it after the return type from `render`. I can split them up for versatility, sure.

SleeplessByte: Splitting up might make it easier for others to read; it's weird that the result of this function is never used. You often want to scope your `getByXXX` to the render result, so it can't find something somewhere else on the page! (Not really applicable to your minimal example, but will be in compound/larger tests).

ErikSchierboom (approved):
LGTM! Running the tests does indeed fix the issue!

SleeplessByte (approved):

neenjaw:
@iHiD, come baaaaack!

iHiD:
Thanks! :D

PR #314, opened 2020-11-27
User onboarding

kntsoriano:
## Description
Before the user proceeds to use our site, they must first accept the terms of service and the privacy policy.


Gemfile
```diff
@@ -48,6 +48,9 @@ gem 'devise'
 # Omniauth
 gem 'omniauth-github'
 
+# Turbolinks form errors compatibility
+gem 'turbolinks_render'
```
kntsoriano, line 52: This is needed to let turbolinks support this piece of code:

```
  def create
    @onboarding = UserOnboardingForm.new(
      user_onboarding_params.merge(user: current_user)
    )

    if @onboarding.save
      redirect_to after_sign_in_path_for(current_user)
    else
      render :show
    end
  end
```

Without the gem, the `render` method won't work.

app/controllers/user_onboarding_controller.rb
```diff
@@ -0,0 +1,26 @@
+class UserOnboardingController < ApplicationController
+  skip_before_action :ensure_onboarded!
+
+  before_action :authenticate_user!
```
iHiD:
```suggestion
```

This is in ApplicationController already.

app/forms/user_onboarding_form.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  validates :terms_of_service, acceptance: true
+  validates :privacy_policy, acceptance: true
+
+  def save
+    return unless valid?
+
+    user.update!(accepted_privacy_policy_at: Time.current, accepted_terms_at: Time.current)
+  end
+end
```
iHiD, line 14: What's the advantage of using this form to updating the user directly? I can't see where it's adding extra value.

db/migrate/20201127032403_add_accept_fields_to_user.rb
```diff
@@ -0,0 +1,6 @@
+class AddAcceptFieldsToUser < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :accepted_privacy_policy_at, :datetime
+    add_column :users, :accepted_terms_at, :datetime
+  end
+end
```
iHiD: Add these to the original create_user migration pls.

test/application_system_test_case.rb
```diff
@@ -3,21 +3,20 @@
...
-  # driven_by :selenium, using: :chrome, screen_size: [1400, 1400]
-  driven_by :selenium, using: :headless_chrome do |driver_option|
-    # Without this argument, Chrome cannot be started in Docker
-    driver_option.add_argument('no-sandbox')
-  end
+  driven_by :selenium, using: :chrome, screen_size: [1400, 1400]
+  # driven_by :selenium, using: :headless_chrome do |driver_option|
+  #   # Without this argument, Chrome cannot be started in Docker
+  #   driver_option.add_argument('no-sandbox')
+  # end
```
iHiD: Why this change? This always causes me pain as I end up with Chrome doing stuff visibly, not invisibly. If there is a specific test that needs this, could we just do it for the one test maybe?

test/controllers/tracks/exercises_controller_test.rb
```diff
@@ -12,7 +12,7 @@ class Tracks::ExercisesControllerTest < ActionDispatch::IntegrationTest
   end
 
   test "index: renders correctly for joined" do
-    user = create :user
+    user = create :user, :onboarded
```
iHiD: I think we should presume that all users are onboarded and set these to be default in the factory. Then if we **don't** want someone to be onboarded (the edge-case) have a trait for `:not_onboarded`. 

It's worth having done it this way to check everything is right first, but if you could invert it now, I'd appreciate it 💙

iHiD:
Looks good mate 🙂  I've made a couple of small comments.

app/forms/user_onboarding_form.rb
kntsoriano: There's a validation to make sure that the users checked the checkbox. We can remove that if you don't want that. Also, it allows us to return an error message related to the validation just as how you save a normal AR record.

test/application_system_test_case.rb
kntsoriano: This wasn't intended. I will revert this.

db/migrate/20201127032403_add_accept_fields_to_user.rb
iHiD: @kntsoriano I think you missed this one.

iHiD:
Sweet. Merging this into the main devise branch. Thanks dude :)

SleeplessByte:
For future updates: it's not a bad idea to store the "policy version" that was accepted. Yes, it can likely be derived from the accepted_at date, given no rolling-updates, but it's still good practice. I usually hardcode a date I wrote the file as a version identifier and store that.

iHiD:
Yeah - we discussed this at the launch of v2 actually. I think we'd store the git_sha of the repo if we were to do it now. Thanks for the reminder.

PR #319, opened 2020-11-28
[WIP] Editor

iHiD:
**This is WIP. Please don't review yet.**

---

Todo list of things I find, @kntsoriano.

- [x] Move code into React
- [x] Implement tabs
- [x] Make monaco full width
- [x] Make monaco full height and only scroll if there's something to see
- [x] Move the Editor out of the Student namespace
- [x] Fix bug where everything blows up when pressing “Run tests” three times in a run (like 0.5s apart)
- [x] Implements all TODOs added by PR.
- [x] Group tests (see below)
- [x] Don't show test groups if they're empty (ie, if there are no failures, don't show the failures bit)
- [x] When opening the editor, if the tests have previously run, start with that tab opened. If not, start with the instructions (as currently)
- [x] The settings dialog (that opens when you click the button) should be 16px to right of the button and 2px from the bottom of the button. (See below)
- [x] The settings on the setting modal should be radio buttons with labels, not selects.
- [x] If no files have changed and the message returns saying that, then the test-run pane should not clear. The easiest way to do this would probably to be to disable the "Run Tests" button until the person changes something, enable it when they do, and disable it again if they go back to the previous state the code was in.

### Settings modal positioning

Screenshot 2020-12-07 at 17 42 39

### Group Tests
The tests should be grouped like this:

Screenshot 2020-11-28 at 21 33 14

Basically another details/summary/tests block that surrounds each group of tests. It would be better to do this grouping at the JS level I think.


kntsoriano:
Do we move the `<Editor />` component to the common namespace?

iHiD:
> Do we move the <Editor /> component to the common namespace?

Or just the top level namespace? What do you think?

kntsoriano:
@iHiD We can move it to the top level namespace, but should that mean that there isn't a common namespace anymore? What is the difference between the two?

iHiD:
I was thinking the common namespace was for things that get embedded into other components. Like shared components (e.g. Icons). Whereas the Editor is always going to be rendered via the Bootloader.

@ErikSchierboom Any thoughts on this?

ErikSchierboom:
I think that's a nice distinction. Common components would be components that would exclusively be used from other components then, never stand-alone?

iHiD:
That works well for me!

kntsoriano:
Agree! I'll add it to the top level namespace.

---

A few questions:
1. Are we adding hints now? If so, where do I pull them?
2. Are we adding report bugs (found inside the ... section)? Do we have a mockup for it?
3. Are we still adding the keyboard shortcuts modal now? I think we only have shortcuts to submit and run tests at the moment.

SleeplessByte:
IMO namespaces should _only_ be used when components are so related they should be namespaced, but not for *namespacing* as you would in Ruby. That's what modules are for.

Having re-exports from index files also disables tree-shaking _accross_ bundles (and you'll end up with larger JavaScript files than necessary when multiple entry points / packs / bundles use the same "namespace").

**Example**: [Card (and Card.Title etc)](https://callstack.github.io/react-native-paper/card.html), vs [typography](https://callstack.github.io/react-native-paper/subheading.html). It's intended to use `Card.Title` inside a `Card`, but the typography ones are related but not coupled.

iHiD:
Yeah, that is where I am too, @SleeplessByte. So things like the Editor should sit as a top-level object (and namespace). But shouldn't be inside something like "Student" which is more of a name wrapper.

iHiD:
@kntsoriano For the test groups, don't show test groups if they're empty (ie, if there are no failures, don't show the failures bit)

iHiD:
@kntsoriano When opening the editor, if the tests have previously run, start with that tab opened. If not, start with the instructions (as currently)

iHiD:
@kntsoriano Where do we actually define the monaco theme? [This bit](https://microsoft.github.io/monaco-editor/playground.html#customizing-the-appearence-exposed-colors):

```
monaco.editor.defineTheme('myTheme', {
    base: 'vs',
    inherit: true,
    rules: [{ background: 'EDF9FA' }],
    colors: {
        'editor.foreground': '#000000',
        'editor.background': '#EDF9FA',
        'editorCursor.foreground': '#8B0000',
        'editor.lineHighlightBackground': '#0000FF20',
        'editorLineNumber.foreground': '#008800',
        'editor.selectionBackground': '#88000030',
        'editor.inactiveSelectionBackground': '#88000015'
    }

});
monaco.editor.setTheme('myTheme');
```

That bit should also get the following scrollbar setup pls:
```
	scrollbar: {
		useShadows: true,
		verticalHasArrows: false,
		horizontalHasArrows: false,
		vertical: 'auto',
		horizontal: 'auto',
		verticalScrollbarSize: 18,
		horizontalScrollbarSize: 18,
		arrowSize: 30
	}
```

iHiD:
@kntsoriano The editor currently doesn't fill the vertical space correctly.

For example, I can scroll down where are there no lines:
Screenshot 2020-12-07 at 16 57 27

kntsoriano:
@iHiD In https://microsoft.github.io/monaco-editor/playground.html, it seems that the editor behaves that way with regards to the scroll bar. I've checked VSCode as well.

iHiD:
@kntsoriano Fixed in https://github.com/exercism/v3-website/pull/371

I've also noticed that if I use the vim bindings and type `:` (to get into command mode) it adds a div at the bottom which breaks the layout of the page. Can you see how we can control where that lives please? At the least we'll need to add a css class to it so I can style it :)

iHiD:
@kntsoriano One more thing. I'm not sure whether for this PR or if you want to do it afterwards is being able to resize the LHS and RHS. It might be easier for you if I change the layout to be flexbox rather than grid - if so just let me know and I'll do it and then we can put that in a subsequent PR. We should probably get this merged before adding even more to it!

app/css/components/file-editor.css
```diff
@@ -0,0 +1,5 @@
+@import "../styles";
+
```
iHiD:
```suggestion
```

app/css/components/file-editor.css
```diff
@@ -0,0 +1,5 @@
...
+.c-file-editor {
+    height: 100%;
+}
```
iHiD: I don't think we necessarily need a whole component of this unless you think it's going to be reusable. I'd instead suggesting renaming the class to `--file-editor` and moving this CSS into the the general editor component.

kntsoriano:
@iHiD I'll create a new issue to make the panels resizable.

kntsoriano (approved):

kntsoriano:
I'm merging this in as it's too big now!

iHiD:
You have to remove the `[WIP]` from the title before you merge! :P 

**Update:** I force pushed with it fixed :)

iHiD:
Nice work dude!! :)

PR #320, opened 2020-11-30
Add settings panel

kntsoriano:
This adds the settings panel to the editor.

## Screenshot
Screen Shot 2020-11-30 at 7 47 28 PM

(Closed and opened when clicking on the gear icon)

## Questions
1. Any suggestions for how to split the `<Editor />` component? It's getting pretty hard to read right now as the file is too long. I've been covering it with tests to make sure that we can easily refactor.

### DIFF
--- a/app/javascript/components/student/Editor.tsx
+++ b/app/javascript/components/student/Editor.tsx
@@ -14,7 +14,13 @@ import { Settings } from './editor/Settings'
 import { FileEditor, FileEditorHandle } from './editor/FileEditor'
 import { fetchJSON } from '../../utils/fetch-json'
 import { typecheck } from '../../utils/typecheck'
-import { Submission, TestRun, TestRunStatus, File } from './editor/types'
+import {
+  Submission,
+  TestRun,
+  TestRunStatus,
+  File,
+  Keybindings,
+} from './editor/types'
 import { Iteration } from '../track/IterationSummary'
 import { GraphicalIcon } from '../common/GraphicalIcon'
 import { Icon } from '../common/Icon'
@@ -145,6 +151,9 @@ export function Editor({
   exampleSolution: string
 }) {
   const [theme, setTheme] = useState('vs')
+  const [keybindings, setKeybindings] = useState<Keybindings>(
+    Keybindings.DEFAULT
+  )
   const [tab, setTab] = useState<TabIndex>(TabIndex.INSTRUCTIONS)
   const isMountedRef = useIsMounted()
   const [{ submission, status, apiError }, dispatch] = useReducer(reducer, {
@@ -336,7 +345,7 @@ export function Editor({
           <Icon icon="keyboard" alt="Keyboard Shortcuts" />
         </button>
 
-        <Settings setTheme={setTheme} />
+        <Settings setTheme={setTheme} setKeybindings={setKeybindings} />
 
         <button className="more-btn">
           <Icon icon="more-horizontal" alt="Open more options" />
@@ -351,6 +360,7 @@ export function Editor({
             ref={editor.ref}
             language={language}
             theme={theme}
+            keybindings={keybindings}
             onRunTests={runTests}
           />
         ))}
--- a/app/javascript/components/student/Editor.tsx
+++ b/app/javascript/components/student/Editor.tsx
@@ -10,6 +10,7 @@ import { TestRunSummary } from './editor/TestRunSummary'
 import { Submitting } from './editor/Submitting'
 import { Tab } from './editor/Tab'
 import { TabPanel } from './editor/TabPanel'
+import { Settings } from './editor/Settings'
 import { FileEditor, FileEditorHandle } from './editor/FileEditor'
 import { fetchJSON } from '../../utils/fetch-json'
 import { typecheck } from '../../utils/typecheck'
@@ -143,6 +144,7 @@ export function Editor({
   instructions: string
   exampleSolution: string
 }) {
+  const [theme, setTheme] = useState('vs')
   const [tab, setTab] = useState<TabIndex>(TabIndex.INSTRUCTIONS)
   const isMountedRef = useIsMounted()
   const [{ submission, status, apiError }, dispatch] = useReducer(reducer, {
@@ -334,9 +336,7 @@ export function Editor({
           <Icon icon="keyboard" alt="Keyboard Shortcuts" />
         </button>
 
-        <button className="settings-btn">
-          <Icon icon="settings" alt="Settings" />
-        </button>
+        <Settings setTheme={setTheme} />
 
         <button className="more-btn">
           <Icon icon="more-horizontal" alt="Open more options" />
@@ -350,6 +350,7 @@ export function Editor({
             file={editor.file}
             ref={editor.ref}
             language={language}
+            theme={theme}
             onRunTests={runTests}
           />
         ))}
--- a/app/javascript/components/student/Editor.tsx
+++ b/app/javascript/components/student/Editor.tsx
@@ -154,6 +154,9 @@ export function Editor({
   const [keybindings, setKeybindings] = useState<Keybindings>(
     Keybindings.DEFAULT
   )
+  const [wrap, setWrap] = useState<'off' | 'on' | 'wordWrapColumn' | 'bounded'>(
+    'on'
+  )
   const [tab, setTab] = useState<TabIndex>(TabIndex.INSTRUCTIONS)
   const isMountedRef = useIsMounted()
   const [{ submission, status, apiError }, dispatch] = useReducer(reducer, {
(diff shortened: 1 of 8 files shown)


app/javascript/components/student/Editor.tsx
```diff
@@ -143,6 +150,13 @@ export function Editor({
   instructions: string
   exampleSolution: string
 }) {
+  const [theme, setTheme] = useState('vs')
+  const [keybindings, setKeybindings] = useState<Keybindings>(
+    Keybindings.DEFAULT
+  )
+  const [wrap, setWrap] = useState<'off' | 'on' | 'wordWrapColumn' | 'bounded'>(
```
kntsoriano: Is there a way to write this better? The `off' | 'on' | 'wordWrapColumn' | 'bounded'` is expected by Monaco.

app/javascript/components/student/editor/Settings.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const handleClickOutside = (e: MouseEvent) => {
+    if (componentRef.current?.contains(e.target as Node)) {
+      return
+    }
+
+    setOpen(false)
+  }
+
+  useEffect(() => {
```
kntsoriano: This is to detect whether the user has clicked outside the panel, and if they do, we close the panel.

app/javascript/components/student/editor/FileEditor.tsx
```diff
@@ -53,9 +56,9 @@ export const FileEditor = forwardRef<FileEditorHandle, FileEditorProps>(
           editorDidMount={editorDidMount}
           onRunTests={onRunTests}
           options={options}
-          value={file.content}
-          theme={'vs'}
-          keybindings={Keybindings.DEFAULT}
+          defaultValue={file.content}
```
kntsoriano, line 59: I changed this to `defaultValue` as the value gets reset when we change the theme.

app/javascript/components/student/Editor.tsx
iHiD: @SleeplessByte?

SleeplessByte: Extract a type

SleeplessByte: In particular: in `Settings.tsx`, I would start by writing:

```typescript
export type SettingsProps = {
  ...
}

export function Settings({ ... }: SettingsProps) {
 ...
}
```

You could then use:

```typescript
... useState<SettingsProps['wrap']>('...')
```

Need it in more places?

```typescript
export type WrapSetting = 'off' | 'on' | 'wordWrapColumn' | 'bounded'
export type SettingsProps = {
  ...
  wrap: WrapSetting 
}

...

... useState<WrapSetting>('...')
```

app/javascript/components/student/editor/Settings.tsx
SleeplessByte: If you need a comment here in the review, you probably want to place the comment on line 40 (It's a good comment to have)

app/javascript/components/student/editor/Settings.tsx
```diff
@@ -0,0 +1,109 @@
...
+    document.addEventListener('click', handleClickOutside)
+
+    return () => {
+      document.removeEventListener('click', handleClickOutside)
+    }
+  }, [])
```
SleeplessByte, line 49: This will work, but it will stop working once you add _enclosed_ values to `handleClickOutside`, because the bound click handler is the one on initial mount, not the one on the current/last render. It's fine for now, but something to look out for. In this case, you can _define_ `handleClickOutside` *inside* the `useEffect` to stop worrying about this:

```typescript
  useEffect(() => {
    const handleClickOutside = (e: MouseEvent) => {
      if (componentRef.current?.contains(e.target as Node)) {
        return
      }

      setOpen(false)
    }
  
    document.addEventListener('click', handleClickOutside)
    return () => {
      document.removeEventListener('click', handleClickOutside)
    }
  }, [componentRef, setOpen])
```

app/javascript/components/student/editor/Settings.tsx
```diff
@@ -0,0 +1,109 @@
...
+              <option value={Keybindings.DEFAULT}>Default</option>
+              <option value={Keybindings.VIM}>Vim</option>
+              <option value={Keybindings.EMACS}>Emacs</option>
+            </select>
+            <label htmlFor="wrap">Wrap</label>
+            <select
+              id="wrap"
+              onChange={(e) =>
+                setWrap(
+                  e.target.value as 'off' | 'on' | 'wordWrapColumn' | 'bounded'
```
SleeplessByte: This would then become 
```suggestion
                  e.target.value as WrapSetting
```

or 

```suggestion
                  e.target.value as SettingsProps['wrap']
```

SleeplessByte:
> Any suggestions for how to split the `<Editor />` component? It's getting pretty hard to read right now as the file is too long. I've been covering it with tests to make sure that we can easily refactor.

1. extract `useRunTests` https://github.com/exercism/v3-website/blob/2fb1dfb0222ea1c6e4782f56b6baf211762edfab/app/javascript/components/student/Editor.tsx#L178-L235
2. extract `useCreateIteration` https://github.com/exercism/v3-website/blob/2fb1dfb0222ea1c6e4782f56b6baf211762edfab/app/javascript/components/student/Editor.tsx#L236-L275

If `cancel` cancels both, I'd probably have _both_ return cancel. If cancel only cancels create iteration, id' have `useCreateIteration` return both `create` and `cancel`.

However, I see they're all tied into the dispatcher / reducer, so you can also choose to extract _all_ of the actions + state changes related to the reducer into a single function, basically extracting the *implementation*. This would allow you to later replace the reducer with something completely different, but keeping the interface the same.

I would likely componentize more of the resulting tree https://github.com/exercism/v3-website/blob/2fb1dfb0222ea1c6e4782f56b6baf211762edfab/app/javascript/components/student/Editor.tsx#L332-L456

e.g.

```tsx
<section id="page-editor">
  <Header 
       exercisePath={exercisePath}
       title={trackTitle}
       exerciseTitle={exerciseTitle}

       ...
   />
  
  <FileEditors />
  <Tabs />
</section>
```

then you'd say "but now I need to drill those props down". You don't, you can still compose everything:

```tsx
<section id="page-editor">
  <Header>
    <Header.Back href={exercisePath} />
    <Header.Title track={trackTitle} exercise={exerciseTitle} />
    
    <Header.ActionHints />
    <Header.ActionKeyboardShortcuts />
     
    <Settings
          theme={theme}
          keybindings={keybindings}
          wrap={wrap}
          setTheme={setTheme}
          setKeybindings={setKeybindings}
          setWrap={setWrap}
        />
   
     <Header.ActionMore />
  </Header>

   ...
</section>
```

As you can see, this actually becomes _more_ readible, because all the noise (css classes, which html elements to use etc) is all abstracted away. You just have the composition left. You can still combine "layout" html with components:

```tsx
<TabsProvider initial={TabIndex.INSTRUCTIONS}>
  <section id="page-editor">
    <Header>
      ...
    </Header>

    <section className="main-lhs">
      {editorsRef.current.map((editor) => (
        <FileEditor
            key={editor.file.filename}
            file={editor.file}
            ref={editor.ref}
            language={language}
            theme={theme}
            keybindings={keybindings}
            wrap={wrap}
            onRunTests={runTests}
          />
      ))}
    </section>
    <section className="main-rhs">
      <Tab.Panel which={TabIndex.INSTRUCTIONS}>
        <Instructions
            introduction={introduction}
            instructions ={instructions}
            exampleSolution={exampleSolution}
            />
      </Tab.Panel>
      <Tab.Panel which={TabIndex.TESTS}>
        ...
       </Tab.Panel>
       <Tab.Panel which={TabIndex.RESULTS}>
         <TestResults submission={submission} />
       </Tab.Panel>
      </Tabs>
    </section>
   ...
  </section >
</TabsProvider>
```

**note**: REFRAIN from using abbrevations such as `lhs` and `rhs`, and don't use `left` or `right`. Use `start` and `end` if you _must_ but probably better to describe what they show instead of where they show.

You may have seen that I wrapped the entire thing in `TabsProvider`, which will allow you to get rid of `tab`, `setTab` and syncing this state inside this component. What you can do is:

```tsx
const TabsContext = createContext({ tab: '', switchToTab: () => {} })

export function TabsProvider({ children, initial }) {
  const tabs = useProvideTabs(initial)
  return <TabsContext.Provider value={tabs}>{children}</TabsContextProvider>
}

export function useTab() {
  return useContext(TabsContext)
}

export function useProvideTabs(initial) {
  const [tab, setTab] = useState(initial)
  return { tab, switchToTab: setTab }
}
```

(All written without checking if it works, but the gist is this).

You can then use `useTab`  inside the TabPanel and compare `tab` with `which` to determine if its active. Same for the tab actions in the footer, and use `switchToTab(nextTab)` to switch.

iHiD:
@SleeplessByte You're wonderful. Thank you.
@kntsoriano There's some commonality to consider with the track selector on the mentor screen that I wrote up here, if you want to consider where the extraction might be. Or you might want to care about this later. https://github.com/exercism/v3-project-management/issues/88

app/javascript/components/student/editor/Settings.tsx
kntsoriano: Good catch!

kntsoriano:
@SleeplessByte Thank you so much!

I'm going to merge this now and do the extractions in a separate PR.

PR #327, opened 2020-12-02
Support unique slug in tracks

iHiD:
I want to refactor this further down the line, but it works well for now.

### DIFF
--- a/app/commands/practice_exercise/create.rb
+++ b/app/commands/practice_exercise/create.rb
@@ -5,8 +5,8 @@ class PracticeExercise
     initialize_with :uuid, :track, :attributes
 
     def call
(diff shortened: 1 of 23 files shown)


app/commands/practice_exercise/create.rb
```diff
@@ -5,8 +5,8 @@ class Create
     initialize_with :uuid, :track, :attributes
 
     def call
-      PracticeExercise.create_or_find_by!(uuid: uuid, track: track) do |pe|
-        pe.attributes = attributes
+      ::PracticeExercise.create_or_find_by!(uuid: uuid, track: track) do |ce|
+        ce.attributes = attributes
```
iHiD, line 9: You probably want to revert this.

ErikSchierboom: I do :D

PR #339, opened 2020-12-08
Respond to push webhook event

ErikSchierboom:


app/controllers/webhooks/base_controller.rb
```diff
@@ -0,0 +1,49 @@
+# This is the base controller of Exercism's webhooks.
+#
+# It is intended to only be called from GitHub webhooks
+# and should not be called by any other source
+
+module Webhooks
+  class BaseController < ApplicationController
+    skip_before_action :verify_authenticity_token
+    skip_before_action :authenticate_user!
```
ErikSchierboom, line 9: Is this okay? We don't have need of a user or or CSRF tokens, as we do the signature validation. This is also how the SPI controller does it.

app/controllers/webhooks/base_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    private
+    def verify_github_webhook
+      return if signature_valid?
+
+      render json: {
+        error: {
+          type: :invalid_webhook_signature,
+          message: "The auth token provided is invalid. Delivery: #{github_delivery}"
+        }
+      }, status: :internal_server_error
```
ErikSchierboom: The API controllers use I8N for the error messages. I could do that here, but it didn't know how to add in the delivery argument (which I've copied from the v2 code).

app/controllers/webhooks/base_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      Exercism.secrets.github_webhooks_secret
+    end
+
+    def github_signature
+      request.headers['HTTP_X_HUB_SIGNATURE_256']
+    end
+
+    def github_delivery
+      request.headers['HTTP_X_GITHUB_DELIVERY']
+    end
```
ErikSchierboom, line 47: These headers (and the signature calculating code) have been updated to the code specified by GitHub: https://developer.github.com/webhooks/securing/#validating-payloads-from-github

app/controllers/webhooks/base_controller.rb
iHiD: Yep. This seems correct! :)

app/controllers/webhooks/base_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+#
+# It is intended to only be called from GitHub webhooks
+# and should not be called by any other source
+
+module Webhooks
+  class BaseController < ApplicationController
+    skip_before_action :verify_authenticity_token
+    skip_before_action :authenticate_user!
+
+    before_action :verify_github_webhook
```
iHiD: Throw a `!` on the end of this as it's an explosive action.

ErikSchierboom: Bang!

app/controllers/webhooks/base_controller.rb
iHiD: I don't think we need i18n here as it's not public facing.

app/controllers/webhooks/repo_updates_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    def track_slug
+      params[:repo][:name]
+    end
+
+    def pushed_to_master?
+      params[:ref] == MASTER_REF
+    end
+
+    MASTER_REF = "refs/heads/master".freeze
+    private_constant :MASTER_REF
```
iHiD: So none of this should get leaked into the controller itself. The controller is responsible for taking the request and handing it to to something to process, and then rendering the response. It shouldn't have any logic about the handling itself. 

Defining these methods inside the controller also leak them into every other controller action that might live in this at some time. As a general (pretty-hard) rule, controllers should only contain methods for actions, and methods for shared code between those actions (e.g. auth or retrieval of record across things).

So this should be moved either inline in the controller method, or should be moved into `app/commands/webhooks/process_repo_update`. The second would be my preference in this scenario.

test/controllers/webhooks/base_test_case.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Webhooks
+  class BaseTestCase < ActionDispatch::IntegrationTest
+    def headers(payload)
+      delivery = SecureRandom.uuid
+      signature = 'sha256=' + OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha256'),
+        Exercism.secrets.github_webhooks_secret, payload.to_json)
+
+      @headers = {
+        'HTTP_X_GITHUB_DELIVERY' => delivery,
```
iHiD: You can inline the `SecureRandom.uuid` here as it's only used here.

test/controllers/webhooks/repo_updates_controller_test.rb
```diff
@@ -0,0 +1,49 @@
...
+      ref: 'refs/heads/master',
+      repo: { name: 'csharp' }
+    }
+
+    invalid_headers = headers(payload)
+    invalid_headers['HTTP_X_HUB_SIGNATURE_256'] = "invalid_signature"
+
+    post webhooks_repo_updates_path, headers: invalid_headers, as: :json, params: payload
+    assert_response 500
+  end
```
iHiD, line 15: This should probably be 403 not a 500. Reserve 500s for actual unexpected errors. That allows us to then capture those via Bugsnag and fix them all.

app/controllers/webhooks/repo_updates_controller.rb
iHiD: (The command should probably take the `params[:repo]` and `params[:ref]` as params)

app/controllers/webhooks/base_controller.rb
iHiD: Nice.

PR #347, opened 2020-12-09
Fix repo updates webhook

ErikSchierboom:
This PR contains three things:

1. It fixes a bug where the repository name was fetched from the wrong object
1. It adds support for the GH ping event, which GH sends whenever the webhook is added
1. It uses a background job to do the actual track syncing, as otherwise the GH webhook UI would show a timeout for the webhook

image

Let me know if I've correctly handled the background job. Looking at the logs, I can see that a background task is queued for the track syncing, but I don't see any output of the background job. Is that normal?

### DIFF
--- a/app/controllers/webhooks/base_controller.rb
+++ b/app/controllers/webhooks/base_controller.rb
@@ -26,12 +26,12 @@ module Webhooks
     end
 
     def handle_ping_event!
-      return unless is_ping_event?
+      return unless ping_event?
 
-      render json: {}, status: :ok
+      head :no_content
     end
 
-    def is_ping_event?
+    def ping_event?
       github_event == 'ping'
     end
 
--- a/app/controllers/webhooks/base_controller.rb
+++ b/app/controllers/webhooks/base_controller.rb
@@ -9,6 +9,7 @@ module Webhooks
     skip_before_action :authenticate_user!
 
     before_action :verify_github_webhook!
+    before_action :handle_ping_event!
 
     layout false
 
@@ -22,6 +23,16 @@ module Webhooks
           message: "The auth token provided is invalid. Delivery: #{github_delivery}"
         }
       }, status: :forbidden
+    end
+
+    def handle_ping_event!
+      return unless is_ping_event?
+
+      render json: {}, status: :ok
+    end
+
+    def is_ping_event?
+      github_event == 'ping'
     end
 
     def signature_valid?
@@ -44,6 +55,10 @@ module Webhooks
 
     def github_delivery
       request.headers['HTTP_X_GITHUB_DELIVERY']
+    end
+
+    def github_event
+      request.headers['HTTP_X_GITHUB_EVENT']
     end
   end
 end
--- a/test/jobs/sync_track_job_test.rb
+++ b/test/jobs/sync_track_job_test.rb
@@ -1,0 +1,10 @@
+require "test_helper"
+
+class SyncTrackJobTest < ActiveJob::TestCase
+  test "track is synced" do
+    track = create :track, slug: 'csharp'
+
+    Git::SyncTrack.expects(:call).with(track)
+    SyncTrackJob.perform_now(track)
+  end
+end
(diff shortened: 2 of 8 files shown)


iHiD:
This looks really solid. The downside is that we don't have background jobs in production yet...

ErikSchierboom:
Oh, oops. That's unfortunate. Should I rewrite?

iHiD:
Nope. It's the right way to do it. I'll just make a decision about background jobs and get it deployed!

app/controllers/webhooks/base_controller.rb
```diff
@@ -24,6 +25,16 @@ def verify_github_webhook!
       }, status: :forbidden
     end
 
+    def handle_ping_event!
+      return unless is_ping_event?
```
SleeplessByte:
```suggestion
      return unless ping_event?
```

app/controllers/webhooks/base_controller.rb
```diff
@@ -24,6 +25,16 @@ def verify_github_webhook!
...
+
+      render json: {}, status: :ok
+    end
+
+    def is_ping_event?
```
SleeplessByte:
```suggestion
    def ping_event?
```

app/controllers/webhooks/base_controller.rb
```diff
@@ -24,6 +25,16 @@ def verify_github_webhook!
       }, status: :forbidden
     end
 
+    def handle_ping_event!
+      return unless is_ping_event?
+
+      render json: {}, status: :ok
```
SleeplessByte: Is it supposed to render an empty JSON? I don't know what GH expects, but if it accepts _any_ ok status , perhaps:

```suggestion
      head :no_content
```

SleeplessByte (approved):

app/controllers/webhooks/base_controller.rb
iHiD: (@ErikSchierboom The `?` implies `is_` in Ruby, so the `is_` isn't needed)

app/controllers/webhooks/base_controller.rb
ErikSchierboom: Yes, that works:

image

test/jobs/sync_track_job_test.rb
```diff
@@ -0,0 +1,10 @@
+require "test_helper"
+
+class SyncTrackJobTest < ActiveJob::TestCase
+  test "track is synced" do
+    track = create :track, slug: 'csharp'
+
+    Git::SyncTrack.expects(:call).with(track)
+    SyncTrackJob.perform_now(track)
+  end
```
SleeplessByte, line 9: Is it synced now?

SleeplessByte (approved):

test/jobs/sync_track_job_test.rb
ErikSchierboom: I don't understand the question :)

SleeplessByte: - The test is "test that track is sycned".
- The test expects a command to be called.

Questions:
- Are spies automatically verified at the end of the test?
- Is testing if the command is called the same as "the track is synced"? If yes, do we test the command? If no, is there a better name? _Gewetensvraag_: is it worth it to test an implementation detail like this? Or is this test too low-level to be meaningful?

ErikSchierboom: > Are spies automatically verified at the end of the test?

They are.

> If yes, do we test the command

We test that separately, yes.

> is it worth it to test an implementation detail like this? Or is this test too low-level to be meaningful?

🤷‍♂️ I've used the same attempt that was already taken in https://github.com/exercism/v3-website/blob/master/test/jobs/award_badge_job_test.rb

I do agree that the test name is slightly misleading.

SleeplessByte: Yah. YMMV, I know that Jeremy also sometimes has tests like this because "things sometimes break", but I don't. Just a thought exercise, nothing blocking.

I'd personally test:

- the ref of the track is not up to date (invariant)
- perform the sync job
- the ref of the track is up to date (assert)

I really don't care if it's calling a certain command, but I do care that the job syncs the track.

iHiD:
Gonna merge and try this.

PR #349, opened 2020-12-09
Sync authorship

ErikSchierboom:


app/commands/git/sync_concept_exercise.rb
```diff
@@ -17,13 +17,29 @@ def call
...
+
+      update_reputation!
     end
 
     private
     attr_reader :exercise
 
+    def update_reputation!
+      exercise.authorships.each do |authorship|
+        User::ReputationAcquisition.find_or_create_by!(
```
ErikSchierboom: I used `find_or_create_by!` instead of `create_or_find_by!` as there is no unique key on any of the keys, nor do I think there should be a unique key as we might have objects on which we want to bestow multiple reputation acquisitions I think.

app/commands/git/sync_concept_exercise.rb
```diff
@@ -17,13 +17,29 @@ def call
...
+          user: authorship.author,
+          reason_object: authorship,
+          reason: "exercise_authorship",
+          category: "exercise_authorship"
+        ) do |ra|
+          ra.amount = 10
```
ErikSchierboom: We probably need to find a good location to define these reputation amounts. Would making them constants in `User::ReputationAcquisition` make sense as a location?

app/models/exercise.rb
```diff
@@ -14,6 +14,15 @@ class Exercise < ApplicationRecord
...
 
+  has_many :authorships,
+    class_name: "Exercise::Authorship",
+    inverse_of: :exercise,
+    dependent: :destroy
+
+  has_many :authors,
+    through: :authorships,
+    source: :author
+
```
ErikSchierboom: I modelled this after the existing `prerequisites` properties

app/models/exercise/authorship.rb
```diff
@@ -0,0 +1,6 @@
+class Exercise::Authorship < ApplicationRecord
+  belongs_to :exercise
+  belongs_to :author, # rubocop:disable Rails/InverseOf
```
iHiD: Rather than disabling the InverseOf here, you should probably add it. So that `User` has:
```ruby
has_many :authorships, source: :author
has_many :authored_exercises, through: :authorships
```

(I think that's right)

db/migrate/20201208144440_create_exercise_authorships.rb
```diff
@@ -0,0 +1,10 @@
+class CreateExerciseAuthorships < ActiveRecord::Migration[6.1]
+  def change
+    create_table :exercise_authorships do |t|
+      t.belongs_to :exercise, foreign_key: true
+      t.belongs_to :user, foreign_key: true
```
iHiD:
```suggestion
      t.belongs_to :exercise, foreign_key: true, null: false
      t.belongs_to :user, foreign_key: true, null: false
```

app/commands/git/sync_concept_exercise.rb
ErikSchierboom: One thing we also need to decide is whether we want to allow _removal_ of reputation acquisitions, if someone is removed as an author (not very likely).

app/commands/git/sync_concept_exercise.rb
```diff
@@ -51,6 +86,26 @@ def find_concepts(slugs)
       end
     end
 
+    def authors
+      head_git_exercise.authors.map do |author|
+        ::User.find_by!(handle: author["exercism_username"])
```
iHiD: Rather than having the n+1, you can do
```
::User.where(handle: head_git_exercise.authors.map{|a|a["exercism_username"]})
```

You could then compare the original list of usernames to the results of that query to get any that are missing for outputting.

app/models/exercise/authorship.rb
ErikSchierboom: I copied it from `Exercise::Prerequisites`:

```ruby
class Exercise::Prerequisite < ApplicationRecord
  belongs_to :exercise
  belongs_to :concept, # rubocop:disable Rails/InverseOf
    class_name: "Track::Concept",
    foreign_key: :track_concept_id
end
```

Does that also need updating?

app/commands/git/sync_concept_exercise.rb
iHiD: I was thinking this when I wrote my thoughts in Slack to you. Let's presume that you can't remove authors for now. And then we'll worry about this a bit further down the line.

app/commands/git/sync_concept_exercise.rb
iHiD: Yeah. I think we want a constant hash of `{reason: amount}` and then in the `before_create` we can set `amount`. So the external API (ie, what you're calling here) only has to set the correct `reason`. We can also then check that the `reason` is valid based on whether it's a key in that hash.

app/commands/exercise/authorship/create.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      def call
+        authorship = exercise.authorships.create!(author: author)
+        User::ReputationAcquisition.find_or_create_by!(
+          user: author,
+          reason_object: authorship,
+          reason: "exercise_authorship",
+          category: "exercise_authorship"
+        )
+      rescue ActiveRecord::RecordNotUnique
```
iHiD: I reckon it's safer (and clearer) to wrap L9 in this rescue, and return in it. As we could end up with a situation where something inside `User::ReputationAcquisition. find_or_create_by` causes the exception, and then we'd be in a muddled state but not know.

app/commands/exercise/authorship/create.rb
```diff
@@ -0,0 +1,21 @@
...
+      initialize_with :exercise, :author
+
+      def call
+        authorship = exercise.authorships.create!(author: author)
+        User::ReputationAcquisition.find_or_create_by!(
+          user: author,
+          reason_object: authorship,
+          reason: "exercise_authorship",
+          category: "exercise_authorship"
+        )
```
ErikSchierboom: I was wondering if it would make sense to have strongly-typed commands here too. Something like:

```suggestion
        User::ReputationAcquisition::ExerciseAuthorship::Create.(author, authorship)
```

Or would that be overkill?

app/commands/git/sync_concept_exercise.rb
```diff
@@ -19,11 +19,40 @@ def call
...
+      author_usernames_config.each do |username|
+        author = authors.find { |a| a.handle == username }
+
+        if author
+          ::Exercise::Authorship::Create.(exercise, author)
+        else
+          Rails.logger.error "Missing author: #{username}"
+        end
+      end
+    end
```
iHiD: How about something like:
```ruby
authors.each { ::Exercise::Authorship::Create.(exercise, author) }

missing_authors = (author_usernames_config - authors.map(&:handle)
Rails.logger.error "Missing authors: #{missing_authors.join(', ')}" if missing_authors.present?
```

app/models/user/reputation_acquisition.rb
```diff
@@ -2,7 +2,19 @@ class User::ReputationAcquisition < ApplicationRecord
...
+    self.amount = REASON_AMOUNTS[self.reason] unless self.amount
+  end
+
   def reason
     super.to_sym
   end
+
+  REASON_AMOUNTS = {
+    'exercise_authorship': 10,
+    'exercise_contributorship': 5
```
iHiD:
```suggestion
    exercise_authorship: 10,
    exercise_contributorship: 5
```

app/models/user/reputation_acquisition.rb
```diff
@@ -2,7 +2,19 @@ class User::ReputationAcquisition < ApplicationRecord
   belongs_to :user
   belongs_to :reason_object, polymorphic: true, optional: true
 
+  before_create do
+    raise ReputationAcquisitionReasonMissingAmount unless self.amount || REASON_AMOUNTS[self.reason]
```
iHiD: Rather than raising in `before_create` use validators.

iHiD: Also, I'd probably go for `REASON_AMOUNTS.has_key?(self.reason)`

iHiD: And I think that we should always check the reason is valid, not only if the `amount` is missing

app/models/user/reputation_acquisition.rb
```diff
@@ -2,7 +2,19 @@ class User::ReputationAcquisition < ApplicationRecord
...
+
+    self.amount = REASON_AMOUNTS[self.reason] unless self.amount
```
iHiD: I don't think amount should be overridable, so I think you can remove the `unless self.amount`

config/initializers/exceptions.rb
```diff
@@ -9,3 +9,4 @@ class TrackHasCyclicPrerequisiteError < RuntimeError; end
 class TrackSearchStatusWithoutUserError < RuntimeError; end
 class TrackSearchInvalidStatusError < RuntimeError; end
 class UserTrackNotFoundError < RuntimeError; end
+class ReputationAcquisitionReasonMissingAmount < RuntimeError; end
```
iHiD:
```suggestion
class ReputationAcquisitionReasonInvalid < RuntimeError; end
```

app/models/user/reputation_acquisition.rb
ErikSchierboom: I've converted to `validates`, which fixes the above comments, right?

app/commands/exercise/authorship/create.rb
ErikSchierboom: Agreed

app/commands/git/sync_concept_exercise.rb
ErikSchierboom: LGTM!

app/commands/git/sync_concept_exercise.rb
```diff
@@ -19,11 +19,30 @@ def call
...
+      update_authors!
+      update_contributors!
     end
 
     private
     attr_reader :exercise
 
+    def update_authors!
+      authors = ::User.where(handle: author_usernames_config)
+      authors.each { |author| ::Exercise::Authorship::Create.(exercise, author) }
```
SleeplessByte:
```suggestion
      authors.find_each { |author| ::Exercise::Authorship::Create.(exercise, author) }
```

Ok yes, it's _unlikely_ that we'll hit > 500 records here, but good practice to use `find_each` over `each when you can.

- `each` loads everything into memory
- `find_each` batches

app/commands/git/sync_concept_exercise.rb
```diff
@@ -19,11 +19,30 @@ def call
...
+
+      missing_authors = author_usernames_config - authors.map(&:handle)
+      Rails.logger.error "Missing authors: #{missing_authors.join(', ')}" if missing_authors.present?
+    end
+
+    def update_contributors!
+      contributors = ::User.where(handle: contributor_usernames_config)
+      contributors.each { |contributor| ::Exercise::Contributorship::Create.(exercise, contributor) }
```
SleeplessByte:
```suggestion
      contributors.find_each { |contributor| ::Exercise::Contributorship::Create.(exercise, contributor) }
```

same as above

app/commands/git/sync_concept_exercise.rb
```diff
@@ -19,11 +19,30 @@ def call
...
 
     private
     attr_reader :exercise
 
+    def update_authors!
+      authors = ::User.where(handle: author_usernames_config)
+      authors.each { |author| ::Exercise::Authorship::Create.(exercise, author) }
+
+      missing_authors = author_usernames_config - authors.map(&:handle)
+      Rails.logger.error "Missing authors: #{missing_authors.join(', ')}" if missing_authors.present?
```
SleeplessByte, line 36: Is this completely hidden? Wouldn't it be a better idea to either throw, or create a model/db entry/ send an e-mail/give us something _actionable_?

If no action should be taken, then it's not an error.

app/commands/git/sync_concept_exercise.rb
```diff
@@ -19,11 +19,30 @@ def call
...
+    end
+
+    def update_contributors!
+      contributors = ::User.where(handle: contributor_usernames_config)
+      contributors.each { |contributor| ::Exercise::Contributorship::Create.(exercise, contributor) }
+
+      missing_contributors = contributor_usernames_config - contributors.map(&:handle)
+      Rails.logger.error "Missing contributors: #{missing_contributors.join(', ')}" if missing_contributors.present?
```
SleeplessByte, line 45: Same as above

app/commands/exercise/authorship/create.rb
iHiD: "strongly-typed commands" - I like that phrase :)

And yes, I'd be pro-that.

app/models/user/reputation_acquisition.rb
```diff
@@ -2,7 +2,23 @@ class User::ReputationAcquisition < ApplicationRecord
...
 
+  before_create do
+    self.amount = REASON_AMOUNTS[self.reason]
+  end
+
+  validates :reason, inclusion: {
+    in: %i[exercise_authorship exercise_contributorship],
+    message: "%<value>s is not a valid reason",
+    strict: ReputationAcquisitionReasonInvalid
+  }
```
iHiD, line 9: Nice! 

Put this before the `before_create`.

I try and keep the order as associations, validations, scopes, callbacks, methods

app/commands/git/sync_concept_exercise.rb
```diff
@@ -51,6 +70,20 @@ def find_concepts(slugs)
       end
     end
 
+    memoize
+    def author_usernames_config
+      return [] unless head_git_exercise.authors
+
+      head_git_exercise.authors.map { |a| a["exercism_username"] }
```
iHiD: You could also merge these three lines using `to_a`
```
head_git_exercise.authors.to_a.map { |a| a["exercism_username"] }
```

I see pros and cons for either! :)

iHiD:
This is coming along really nicely 🙂

app/commands/git/sync_concept_exercise.rb
ErikSchierboom: I like the suggestion!

SleeplessByte (commented):
This review I did hours and hours ago, but it wouldn't let me submit (500)

app/commands/git/sync_concept_exercise.rb
ErikSchierboom: Excellent.

app/commands/git/sync_concept_exercise.rb
ErikSchierboom: This is a temporary implementation. I forgot to add a TODO, which I've just done.

app/commands/exercise/authorship/create.rb
ErikSchierboom: I've pushed a commit that implements the above. Let me know what you think

PR #351, opened 2020-12-10
Don't show empty test groups

kntsoriano:
In the screenshot, the "Passed" test group is hidden because no tests have passed.

## Screenshot
Screen Shot 2020-12-10 at 11 38 17 AM


app/javascript/components/editor/TestsGroupList.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import { TestStatus, Test } from './types'
+import { GraphicalIcon } from '../common/GraphicalIcon'
+import { TestsGroup } from './TestsGroup'
+
+export function TestsGroupList({ tests }: { tests: Test[] }): JSX.Element {
+  const testsWithIndex = tests.map((test, i) => ({ index: i + 1, ...test }))
+  const firstFailedTestIdx = testsWithIndex.findIndex(
```
ErikSchierboom: I think you can use [`some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some), which returns a `boolean`:

```suggestion
  const firstFailedTestIdx = testsWithIndex.some(
```

ErikSchierboom (approved):

app/javascript/components/editor/TestRunSummary.tsx
```diff
@@ -1,30 +1,10 @@
...
-          All tests passed
-        </div>
-      )
-    default:
-      return null
-  }
-}
-
-export function TestRunSummary({
+export const TestRunSummary = ({
```
SleeplessByte, line 7: What did converting this to `const` get you? Was that necessary for TypeScript to allow the properties?

app/javascript/components/editor/TestRunSummary.tsx
```diff
@@ -121,8 +101,78 @@ export function TestRunSummary({
...
+          <span className="--dot" />
+          All tests passed
+        </div>
+      )
+    default:
+      return null
+  }
+}
+
+TestRunSummary.Content = ({
```
SleeplessByte, line 130: This is going to "live" update without user interaction, right?

Consider wrapping the status message in a [status role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_status_role). What gives? It will allow screenreaders to _read out_ the change automatically when it happens, without you needing to do anything for that.

app/javascript/components/editor/TestsGroupList.tsx
SleeplessByte: True, but I think further down firstFailedTestIdx is used as _id_, not as _state_.

app/javascript/components/editor/TestsGroupList.tsx
```diff
@@ -0,0 +1,49 @@
...
+    (test) =>
+      test.status === TestStatus.FAIL || test.status === TestStatus.ERROR
+  )
+  const passed = testsWithIndex.filter(
+    (test) => test.status === TestStatus.PASS
+  )
+  const failed =
```
SleeplessByte: You could consider building `passed` and `failed` the same way, and then using `failed[0]` to get the first failed test, instead.

SleeplessByte (approved):
Looking good

iHiD (approved):
Approving from a Codeowner perspective. Not read the code :)

app/javascript/components/editor/TestRunSummary.tsx
kntsoriano: I wasn't able to assign `TestRunSummary.Header` and `TestRunSummary.Content` if this wasn't a `const`.

app/javascript/components/editor/TestRunSummary.tsx
kntsoriano: Thanks. I've added this :)

app/javascript/components/editor/TestsGroupList.tsx
kntsoriano: Yes, we use an ID here. I've rewritten the code for it to be clearer now.

app/javascript/components/editor/TestsGroupList.tsx
kntsoriano: Thanks. I did this :)

PR #352, opened 2020-12-10
<Tooltip> make react-query lazy-fire

neenjaw:
By default, react-query fires its query when the Component is mounted. This is a problem as it may trigger unnecessary requests. react-query has an option to disable the query setting the `enabled` option to false. So now, while the tool-tip is in a `hidden` or `error` state, the query is disabled.

--- 

I've been able to get it going locally now, see image for gif of current lazy loading

lazy_load

### DIFF
--- a/app/javascript/components/tooltips/Tooltip.tsx
+++ b/app/javascript/components/tooltips/Tooltip.tsx
@@ -1,7 +1,21 @@
 import React, { useEffect, useState } from 'react'
 import { usePopper } from 'react-popper'
 import { useQuery } from 'react-query'
-import { useStatefulTooltip } from '../../hooks/use-stateful-tooltip'
+import {
+  useStatefulTooltip,
+  dispatchError,
+  dispatchLoaded,
+  dispatchShow,
+  // dispatchHide,
+  dispatchRequestHideFromRefHover,
+  dispatchRequestHideFromRefFocus,
+  dispatchRequestShowFromRefHover,
+  dispatchRequestShowFromRefFocus,
+  dispatchRequestHideFromHover,
+  dispatchRequestHideFromFocus,
+  dispatchRequestShowFromHover,
+  dispatchRequestShowFromFocus,
+} from '../../hooks/use-stateful-tooltip'
 
 interface TooltipProps {
   id: string
@@ -63,22 +77,12 @@ export const Tooltip = ({
   // useEffect for responding to the reference element's request to show when hovered
   useEffect(() => {
     if (hoverRequestToShow) {
-      return dispatch({
-        id: id,
-        action: {
-          type: 'request-show-from-ref-hover',
-        },
-      })
+      return dispatchRequestShowFromRefHover(dispatch, id)
     }
 
     // setTimeout used to fire this dispatch AFTER the tooltip self-hover has time to fire
     const timeoutRef = setTimeout(() => {
-      dispatch({
-        id: id,
-        action: {
-          type: 'request-hide-from-ref-hover',
-        },
-      })
+      dispatchRequestHideFromRefHover(dispatch, id)
     }, 0)
 
     return () => {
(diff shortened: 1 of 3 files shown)


neenjaw:
@ErikSchierboom - does this interfere with how you extracted the tooltip?
@iHiD - sorry about not being able to test it myself, advise?

app/javascript/components/tooltips/Tooltip.tsx
```diff
@@ -262,6 +247,28 @@ export const Tooltip = ({
...
+        fetch(contentEndpoint, {
+          method: 'get',
+          signal,
+        }).then((res) => res.text()),
+        { cancel: () => controller.abort() }
+      )
+    },
+    {
+      // Enable the query to fetch only if it isn't in a hidden or error state
+      enabled: !(showState == 'hidden' || showState == 'error'),
```
SleeplessByte: This is something different than: don't fire on mount.

Your implementation sounds like the right approach for the problem, but double confirming that you're not actually looking to make it "not fire on mount" but rather "lazily fire".

Not fire on mount has a dedicated option in react query

neenjaw: yes, I'm looking to lazily fire.  I didn't see that option.  Will look again

neenjaw: I don't see that api call, or another option.  Do you have a reference I can review?

SleeplessByte: What you're doing right now is lazily fire. This PR is called "shouldn't fire on mount".

Your code does what you want it to do, but the PR is named incorrectly. My comment was trying to figure out if the code was incorrect or the PR was incorrect. The "shouldn't fire on mount" option is called `refetchOnMount`.

ErikSchierboom: Personally, I find the following version slightly easier to read:

```suggestion
      enabled: showState != 'hidden' && showState != 'error',
```

But feel free to ignore.

ErikSchierboom (approved):

app/javascript/components/tooltips/Tooltip.tsx
ErikSchierboom: This PR does not break the popups @neenjaw

iHiD:
The tests are green but the codeclimate stuff apparently breaks on branches now. @neenjaw Tell me when to merge :)

app/javascript/components/tooltips/Tooltip.tsx
neenjaw: DeMorgan's law hard at work! I'll change it, it doesn't matter to me

neenjaw:
I've extracted the query to `useTooltipContentQuery` and the reducer to `useStatefulTooltip` custom hooks.  I can't seem to reduce the **cognitive complexity** lower than 7 (to the required 5).

These probably should have been extracted before, so I don't think it's been too much hassle.

One line to note is in the content query. See following comment

app/javascript/hooks/use-tooltip-content-query.ts
```diff
@@ -0,0 +1,46 @@
+import { useQuery } from 'react-query'
+
+function fetchContent(endpoint: string, signal: AbortSignal) {
+  return fetch(endpoint, {
+    method: 'get',
+    signal,
+  }).then((res) => {
+    if (res.status != 200 && res.status != 304) {
```
neenjaw: Is this good by others to have here? So that it only tries to display to tool tip on 200 (success) or 304 (cached)?

iHiD: My Ruby brain would rather group the statuses:

```
if(![200,304].include?(res.status))
```

Maybe even pull them into "validStatuses" and add the comment about `200 = success and 304 = cached=` in the code for future reference.

But that might be less good JS 🤷

neenjaw:
```suggestion
    if (![200, 304].includes(res.status)) {
```

SleeplessByte: They're equally fine!

PR #353, opened 2020-12-10
Focus results tab if previous test results are present

kntsoriano:
If we get a previous test results from the API request, we switch to the results tab. Otherwise, the instructions tab is selected by default.


ErikSchierboom (approved):

SleeplessByte (approved):
Seems fine 🤷🏽 😄

iHiD:
Nice. Thanks @kntsoriano :)

iHiD (approved):

PR #354, opened 2020-12-10
Change settings to radio buttons for styling

kntsoriano:


ErikSchierboom (approved):

iHiD:
@kntsoriano A few things.

----

I've added a commit with styling

---

Problem: Tabbing between the keybinding options means that the Monaco Editor takes focus away from the dialog and into monaco (e.g. when you tab to VIM and press tab again, the cursor jumps into Monaco)
Solution: Only apply the keybindings change once the dialog closes.

I think this only needs to happen for the keybindings one. The other two work fine as is and the theme switching being "live" is nice.

----

Problem: The settings button and dialog needs aria attributes for hasPopup and expanded.
Fix: Duplicate the aria + role + tabindex functionality from app/javascript/components/student/tracks-list/TagsFilter.jsx

---

Have ESC close the dialog, and refocs on the settings cog.

Procfile.dev
```diff
@@ -1,4 +1,4 @@
 server: bin/rails server -p 3020
 assets: bin/webpack-dev-server
 anycable: RAILS_ENV=development bundle exec anycable --server-command "anycable-go --host=localhost --port 3334"
-js-tests: yarn test-watch
+#js-tests: yarn test-watch
```
ErikSchierboom: Was this intentional?

iHiD: Nope. It was in my procfile for some reason. So. Many. Branches.

PR #356, opened 2020-12-10
Don't clear submission until a valid submission comes in

kntsoriano:
## Description
This fixes the bug where the previous test results were cleared even if the newest submission is invalid. We shouldn't clear the submission until we get a new one.

### DIFF
--- a/test/javascript/components/Editor.test.js
+++ b/test/javascript/components/Editor.test.js
@@ -7,67 +7,6 @@ import { rest } from 'msw'
 import { setupServer } from 'msw/node'
 import { Editor } from '../../../app/javascript/components/Editor'
 
-test('clears current submission when resubmitting', async () => {
-  const server = setupServer(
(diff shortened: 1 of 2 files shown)


test/javascript/components/Editor.test.js
```diff
@@ -7,67 +7,6 @@ import { rest } from 'msw'
 import { setupServer } from 'msw/node'
 import { Editor } from '../../../app/javascript/components/Editor'
 
-test('clears current submission when resubmitting', async () => {
```
kntsoriano, line 10: This test shouldn't apply anymore as we already cancel requests just in case a user submits two iterations at the same time.

ErikSchierboom (approved):

iHiD:
Screenshot 2020-12-10 at 12 43 29
 🦜

iHiD (approved):

PR #357, opened 2020-12-10
Add reputation from commits

ErikSchierboom:


app/models/user/reputation_acquisition.rb
```diff
@@ -5,4 +5,9 @@ class User::ReputationAcquisition < ApplicationRecord
   def reason
     super.to_sym
   end
+
+  after_save do
+    total_amount = Arel.sql("(#{user.reputation_acquisitions.select('SUM(amount)').to_sql})")
```
iHiD: Can you add a comment above this saying "We want to update this in the DB rather than through two queries to avoid race-conditions" or similar pls.

test/models/user_test.rb
```diff
@@ -35,6 +35,7 @@ class UserTest < ActiveSupport::TestCase
     create :user_reputation_acquisition, user: user, category: "track_javascript", amount: 4
     create :user_reputation_acquisition, user: user, category: "docs", amount: 8
 
+    user.reload
     assert_equal 15, user.reputation
```
iHiD:
```suggestion
    assert_equal 15, user.reload.reputation
```

app/commands/git/sync.rb
```diff
@@ -14,13 +14,6 @@ def filepath_in_diff?(filepath)
       end
     end
 
-    def commits_from_current_to_head
-      walker = Rugged::Walker.new(git_repo.send(:rugged_repo))
```
iHiD: Removing my name from the codebase! 🙂

app/commands/webhooks/process_push_update.rb
```diff
@@ -1,13 +1,13 @@
 module Webhooks
-  class ProcessRepoUpdate
+  class ProcessPushUpdate
```
iHiD, line 2: This is good.

app/controllers/webhooks/pull_request_updates_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Webhooks
+  class PullRequestUpdatesController < BaseController
+    def create
+      ::Webhooks::ProcessPullRequestUpdate.(
+        # params[:action] does not work as it is populated by Rails with the action method name
```
iHiD, line 5: Thanks for leaving the comment.

app/jobs/process_pull_request_update_job.rb
```diff
@@ -0,0 +1,7 @@
+class ProcessPullRequestUpdateJob < ApplicationJob
+  queue_as :default
+
+  def perform(action, author, url, html_url)
```
iHiD:
```suggestion
  def perform(action, github_username, url, html_url)
```

I think we should name this consistently inside for us. It's unclear to me whether "author" is a string, object, DB record, etc.

iHiD:
```suggestion
  def perform(action, author, pr_url, html_url)
```

(I'm guessing that's what this is?)

app/commands/user/reputation_token/code_contribution/create.rb
```diff
@@ -0,0 +1,20 @@
...
+      class Create
+        include Mandate
+
+        initialize_with :user, :external_link
+
+        def call
+          User::ReputationToken.find_or_create_by!(
+            user: user,
+            external_link: external_link,
+            reason: 'committed_code',
```
iHiD: Maybe "contributed_code" to go with the command name. Or change the command name to `CodeCommitted`. I'm not sure which is the better term?

app/commands/git/sync.rb
ErikSchierboom: Makes sense. Should I make an effort of removing any actual usernames from the codebase?

app/jobs/process_pull_request_update_job.rb
ErikSchierboom: It is the PR's API URL. So maybe either:

`def perform(action, author, api_url, html_url)`

or else

`def perform(action, author, pr_url, pr_html_url)`

?

app/commands/user/reputation_token/code_contribution/create.rb
ErikSchierboom: I like contributed code more, as Exercism cares more about contributions than about commits. I'll update.

app/commands/user/reputation_token/code_contribution/create.rb
```diff
@@ -0,0 +1,32 @@
...
+
+        private
+        def reason
+          case size
+          when :minor
+            'contributed_code/minor'
+          when :major
+            'contributed_code/major'
+          else
+            'contributed_code'
```
iHiD: We should probably do `/regular` on ones that also have minor/major.

db/migrate/20201215122646_add_reputation_token_context_key.rb
```diff
@@ -0,0 +1,6 @@
+class AddReputationTokenContextKey < ActiveRecord::Migration[6.1]
+  def change
+    add_column :user_reputation_tokens, :context_key, :string, null: false
+    add_index :user_reputation_tokens, %i[context_key user_id], unique: true
+  end
+end
```
iHiD: You can add this to the main migration as that's in this PR too.

Gemfile
```diff
@@ -20,6 +20,7 @@ gem 'bootsnap', '>= 1.4.2', require: false
 # Interaction gems
 gem 'rugged' # Git
 gem 'rest-client' # http
+gem 'octokit' # GitHub
```
iHiD, line 22: 💙 for putting this in the right place in the Gemfile!

app/commands/user/reputation_token/award_for_pull_request.rb
```diff
@@ -22,6 +23,26 @@ def award_reputation_to_author
         User::ReputationToken::CodeContribution::Create.(user, external_link, repo, number, reason)
       end
 
+      def award_reputation_to_reviewers
+        return unless closed?
```
iHiD: I think we only award reputation once the PR is merged. So should we guard on line 15 too?

iHiD: Actually, should we just guard the whole thing on L11?

app/commands/user/reputation_token/code_review/create.rb
```diff
@@ -4,12 +4,12 @@ module CodeReview
       class Create
         include Mandate
 
-        initialize_with :user, :external_link, :context_key, :reason
+        initialize_with :user, :external_link, :repo, :number, :reason
```
iHiD:
```suggestion
        initialize_with :user, :external_link, :repo, :pr_id, :reason
```

app/controllers/webhooks/pull_request_updates_controller.rb
```diff
@@ -0,0 +1,49 @@
+module Webhooks
+  class PullRequestUpdatesController < BaseController
+    def create
+      ::Webhooks::ProcessPullRequestUpdate.(action, github_username,
+        url: url,
+        html_url: html_url,
+        labels: labels,
+        state: state,
+        number: number,
+        repo: repo)
```
iHiD: This should probably be a hash.

app/commands/user/reputation_token/award_for_pull_request.rb
ErikSchierboom: Yes. I hadn't yet gotten around the merged bit. I've updated

db/migrate/20201215122646_add_reputation_token_context_key.rb
ErikSchierboom: Doh! Will update

app/commands/git/sync_concept.rb
```diff
@@ -23,7 +23,6 @@ def call
     attr_reader :concept
 
     def concept_needs_updating?
-      return false if synced_to_head?
```
ErikSchierboom, line 26: This was not needed as the synced_to_head check was already done at the level of the track syncing (which calls this file's class)

app/commands/git/sync_concept_exercise.rb
```diff
@@ -47,8 +47,6 @@ def update_contributors!
     end
 
     def exercise_needs_updating?
-      return false if synced_to_head?
```
ErikSchierboom, line 50: This was not needed as the synced_to_head check was already done at the level of the track syncing (which calls this file's class)

app/commands/git/sync_practice_exercise.rb
```diff
@@ -24,8 +24,6 @@ def call
     attr_reader :exercise
 
     def exercise_needs_updating?
-      return false if synced_to_head?
```
ErikSchierboom, line 27: This was not needed as the synced_to_head check was already done at the level of the track syncing (which calls this file's class)

app/commands/user/reputation_token/award_for_pull_request.rb
```diff
@@ -0,0 +1,77 @@
...
+          User::ReputationToken::CodeReview::Create.(reviewer, external_link, repo, pr_id, 'reviewed_code')
+        end
+
+        # TODO: consider what to do with missing reviewers
+        missing_reviewers = reviewer_usernames - reviewers.map(&:handle)
+        Rails.logger.error "Missing reviewers: #{missing_reviewers.join(', ')}" if missing_reviewers.present?
+      end
+
+      def merged?
+        params[:merged].present? && params[:merged]
```
ErikSchierboom: Is this the Ruby way of doing this? I saw somewhere that some people use a `!!` to do something like this.

db/migrate/20201210173500_create_user_reputation_tokens.rb
```diff
@@ -0,0 +1,24 @@
+class CreateUserReputationTokens < ActiveRecord::Migration[6.1]
+  def change
+    # TODO: Remove
+    drop_table :user_reputation_acquisitions, if_exists: true
```
ErikSchierboom, line 4: @iHiD Does this need to be removed? In other words, does the exercism.lol environment have this table?

app/jobs/process_pull_request_update_job.rb
iHiD: Yeah, `api_url` then I think.

app/commands/user/authenticate_from_omniauth.rb
```diff
@@ -18,6 +19,7 @@ def find_by_uid
         user.save
       end
 
+      user.update(github_username: auth.info.nickname)
```
iHiD: So, fun one for you here. This will fail if for some reason the user is not "valid" for some unrelated reason. It'll fail silently. We could use `update!` which would raise an exception, but we don't want that. So instead we can use `update_column(:github_username, auth.info.nickname)` which bypasses all the Rails checks and writes it even if the model is invalid.

app/commands/user/reputation_token/award_for_pull_request.rb
```diff
@@ -0,0 +1,77 @@
+require 'octokit'
+
+class User
+  class ReputationToken
+    class AwardForPullRequest
+      include Mandate
+
+      initialize_with :action, :github_username, :params
+
+      def call
```
iHiD, line 10: This was more what I meant with my comment.

I think having the `merged?` guard inside `award_reputation_to_author` was correct.

```suggestion
      def call
        return unless closed?
        
```

app/commands/user/reputation_token/award_for_pull_request.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      private
+      def award_reputation_to_author
+        user = User.find_by(github_username: github_username)
+
+        # TODO: decide what to do with user that cannot be found
+        Rails.logger.error "Missing author: #{github_username}" unless user
+        return unless user
+
+        User::ReputationToken::CodeContribution::Create.(user, external_link, repo, pr_id, reason)
```
iHiD: Let's rename `reason` to be specific to the thing it's a reason about:

```suggestion
        User::ReputationToken::CodeContribution::Create.(user, external_link, repo, pr_id, author_reputation_reason)
```

app/commands/user/reputation_token/award_for_pull_request.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+
+      def award_reputation_to_reviewers
+        reviews = octokit_client.pull_request_reviews(repo, pr_id)
+        reviewer_usernames = reviews.map { |reviewer| reviewer[:user][:login] }
```
iHiD, line 32:
```suggestion
        reviewer_usernames = reviews.map { |reviewer| reviewer.dig(:user, :login) }
```

app/commands/user/reputation_token/award_for_pull_request.rb
iHiD: `params[:merged].present?` is fine without the `&& ...`

`nil.present? == false`
`false.present? == false`
`true.present? == true`

app/commands/user/reputation_token/award_for_pull_request.rb
```diff
@@ -0,0 +1,77 @@
...
+
+        reviewers = ::User.where(handle: reviewer_usernames)
+        reviewers.find_each do |reviewer|
+          User::ReputationToken::CodeReview::Create.(reviewer, external_link, repo, pr_id, 'reviewed_code')
```
iHiD:
```suggestion
          User::ReputationToken::CodeReview::Create.(reviewer, external_link, repo, pr_id, :reviewed_code)
```

app/commands/user/reputation_token/award_for_pull_request.rb
```diff
@@ -0,0 +1,77 @@
...
+      def repo
+        params[:repo]
+      end
+
+      def pr_id
+        params[:pr_id]
+      end
+
+      def reason
+        return 'contributed_code/major' if params[:labels].include?('reputation/contributed_code/major')
```
iHiD: I'd probably have these as symbols.

```suggestion
        return :'contributed_code/major' if params[:labels].include?('reputation/contributed_code/major')
```

app/commands/user/reputation_token/code_contribution/create.rb
```diff
@@ -0,0 +1,24 @@
+class User
+  class ReputationToken
+    module CodeContribution
+      class Create
+        include Mandate
+
+        initialize_with :user, :external_link, :repo, :number, :reason
```
iHiD:
```suggestion
        initialize_with :user, :external_link, :repo, :pr_id, :reason
```

iHiD: Tiny one, but I think I'd reorder the params so it's `:user, :reason, :repo, :pr_id, :external_link`

My logic is that these reputation token classes will always have a `user` and `reason` so having those first will give us some nice consistency. Then having the things that make up the `context_key` (repo and pr_id) feel good next, then finally an extra things.

app/commands/user/reputation_token/award_for_pull_request.rb
iHiD: If `'reviewed_code'` is a hard-coded string, we could remove it from here and put it into the `Create` command

app/commands/user/reputation_token/code_contribution/create.rb
```diff
@@ -0,0 +1,24 @@
...
+          reputation_token = User::ReputationToken.create_or_find_by!(
+            user: user,
+            context_key: "contributed_code/#{repo}/pulls/#{number}"
+          ) do |rt|
+            rt.external_link = external_link
+            rt.reason = reason
+            rt.category = :building
+          end
+
+          reputation_token.update!(reason: reason)
```
iHiD, line 21:
```suggestion
          # Reasons are mutable and this class may be called multiple times
          # if the reason changes (via labels being changed on PRs)
          reputation_token.update!(reason: reason)
```

app/commands/user/reputation_token/exercise_authorship/create.rb
```diff
@@ -0,0 +1,22 @@
...
+
+        initialize_with :authorship
+
+        def call
+          User::ReputationToken.create_or_find_by!(
+            user: authorship.author,
+            context_key: "authored_exercise/#{authorship.exercise.uuid}"
+          ) do |rt|
+            rt.context = authorship
+            rt.reason = 'authored_exercise'
```
iHiD:
```suggestion
            rt.reason = :authored_exercise
```

app/commands/user/reputation_token/exercise_contributorship/create.rb
```diff
@@ -0,0 +1,22 @@
...
+
+        initialize_with :contributorship
+
+        def call
+          User::ReputationToken.create_or_find_by!(
+            user: contributorship.contributor,
+            context_key: "contributed_to_exercise/#{contributorship.exercise.uuid}"
+          ) do |rt|
+            rt.context = contributorship
+            rt.reason = 'contributed_to_exercise'
```
iHiD:
```suggestion
            rt.reason = :contributed_to_exercise
```

app/commands/webhooks/process_pull_request_update.rb
```diff
@@ -0,0 +1,23 @@
+module Webhooks
+  class ProcessPullRequestUpdate
+    include Mandate
+
+    initialize_with :action, :github_username, :params
+
+    def call
+      return unless closed?
+      return unless valid_action?
```
iHiD, line 9: Nice.

app/controllers/webhooks/pull_request_updates_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  class PullRequestUpdatesController < BaseController
+    def create
+      ::Webhooks::ProcessPullRequestUpdate.(action, github_username,
+        { url: url,
+          html_url: html_url,
+          labels: labels,
+          state: state,
+          pr_id: pr_id,
+          repo: repo,
+          merged: merged? })
```
iHiD: Hashes are implicit if they're the last param.

```suggestion
      ::Webhooks::ProcessPullRequestUpdate.(
          action,
          github_username,
          url: url,
          html_url: html_url,
          labels: labels,
          state: state,
          pr_id: pr_id,
          repo: repo,
          merged: merged? 
        )
```

app/controllers/webhooks/pull_request_updates_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def repo
+      params[:repository][:full_name]
+    end
+
+    def merged?
+      params[:pull_request][:merged]
+    end
+  end
```
iHiD: I don't love these being methods on a controller. Controller methods should only be `actions`, or things that guard or set up actions. We're leaking one action's submethods into the whole controller by doing this.

I'd instead either:
- inline them all into the `create` action.
- pass in `params[:pull_request]` as the object itself into `Webhooks::ProcessPullRequestUpdate` and do this work there.

app/controllers/webhooks/pull_request_updates_controller.rb
ErikSchierboom: I thought that was what I had before, but I guess I did not add the field: prefixes. Will fix!

app/models/user.rb
```diff
@@ -53,13 +53,15 @@ def admin?
   end
 
   def reputation(track_slug: nil, category: nil)
+    return super() if track_slug.nil? && category.nil?
```
iHiD: If you're explicitely checking for just `nil` and don't mind also checking for `false` you can do:
```
return super() unless track_slug || category
```

(I think my bool logic is right there!)

You might instead be checking to see if they're nil or blank/empty, in which case you'd be better using `.present?` or `.blank?` rather than `.nil?`

If you *really* care just about `nil` then what you have is fine, but that's unusual

db/migrate/20201210173500_create_user_reputation_tokens.rb
iHiD: Yes, it does. So this line can't be removed. But once this is merged and deployed and run on the server, we can then remove it.

app/models/user.rb
iHiD: (In this context, I think you actually care about empty strings too, so present or blank)

test/factories/users.rb
```diff
@@ -4,6 +4,7 @@
     name { "User" }
     password { "password" }
     handle { "handle-#{SecureRandom.hex(4)}" }
+    github_username { nil }
```
iHiD: You don't need this line. It's implicit.
```suggestion
```

app/models/user.rb
ErikSchierboom: My original reasoning was that `nil` is the default value of the parameters, but I do indeed not really care about `nil` in particular.

app/commands/user/authenticate_from_omniauth.rb
ErikSchierboom: Interesting!

app/commands/user/reputation_token/award_for_pull_request.rb
ErikSchierboom: This does not seem to work. It returns `nil`

iHiD:
Please rebase.

app/commands/user/reputation_token/award_for_pull_request.rb
iHiD: Very weird.

PR #360, opened 2020-12-11
Fix sync track tests

ErikSchierboom:


test/commands/git/sync_track_test.rb
```diff
@@ -79,17 +79,41 @@ class Git::SyncTrackTest < ActiveSupport::TestCase
...
   end
 
   test "adds new concept exercises defined in config.json" do
     track = create :track, slug: 'fsharp', synced_to_git_sha: 'ab0b9be3162f6ec4ed6d7c46b55a8bf2bd117ffb'
 
     Git::SyncTrack.(track)
 
-    cars_assemble = ConceptExercise.find_by(uuid: '6ea2765e-5885-11ea-82b4-0242ac130003')
-    assert_includes track.concept_exercises, cars_assemble
+    assert track.concept_exercises.find_by(uuid: '6ea2765e-5885-11ea-82b4-0242ac130003').present?
```
iHiD:
```suggestion
    assert track.concept_exercises.where(uuid: '6ea2765e-5885-11ea-82b4-0242ac130003').exists?
```

iHiD: Exists simply checks for the presence in the db of something. `find_by` will init the object so is a lot heavier. It doesn't really matter for tests, but it's a good habit to use `exists?` if you're not then going to use the returned object.

PR #365, opened 2020-12-11
Switch to Rails 6.1

iHiD:


test/commands/mentor/reply_to_discussion_test.rb
```diff
@@ -31,7 +31,7 @@ class Mentor::ReplyToDiscussionTest < ActiveSupport::TestCase
       "foobar"
     )
     assert_equal 1, user.notifications.size
-    notification = user.notifications.first
+    notification = Notification.where(user: user).first
```
iHiD, line 34: These are from me working on getting strict_loading working. It's problematic with FactoryBot so I've stopped for now. But these are still fine to merge.

app/commands/iteration/create.rb
```diff
@@ -5,9 +5,10 @@ class Create
     initialize_with :solution, :submission
 
     def call
+      time = Time.now.utc.to_s(:db)
```
kntsoriano, line 8: Nice.

test/commands/mentor/reply_to_discussion_test.rb
kntsoriano: What's strict loading?

kntsoriano (approved):

test/commands/mentor/reply_to_discussion_test.rb
iHiD: It basically causes an exception if an `n+1` might happen.

https://github.com/rails/rails/pull/37400
https://github.com/rails/rails/pull/39491

app/commands/iteration/create.rb
iHiD: `NOW()` doesn't work with the latest version of MySQL on mac for some reason.

PR #366, opened 2020-12-11
change react-queryt id to endpoint

neenjaw:
Currently, there is a collision occuring between tracks because how turbolinks works.
Turbolinks prevents react-query from being unloaded, so the cache is persisted, so using `basics` can cause a collision between different tracks with same-named concepts.


iHiD:
Can we get rid of `id` altogether now?

neenjaw:
it's used for other things for the tooltip managing its state

PR #367, opened 2020-12-13
Revert to exercise start

kntsoriano:
## Description
This reverts the files in the editor to the original files of the exercise. In order to do this, we send `GET /api/solutions/:solution_uuid/original_solution_files` to retrieve the files.


app/controllers/test/base_controller.rb
```diff
@@ -1,4 +1,6 @@
 class Test::BaseController < ApplicationController
   skip_before_action :verify_authenticity_token
   skip_before_action :authenticate_user!
+
+  disable_site_header!
```
kntsoriano, line 5: There was a problem with the "Run tests" button not being  seen in the test layout because the header was present. I disabled it for now, and will fix the CSS in another PR.

app/javascript/components/Editor.tsx
```diff
@@ -154,6 +88,11 @@ export function Editor({
 }) {
   const [tab, switchToTab] = useState(TabIndex.INSTRUCTIONS)
   const [theme, setTheme] = useState(Themes.LIGHT)
+  const [
```
kntsoriano, line 91: I actually think that the reducer for the revert and submission status shouldn't be here and should be abstracted behind a custom hook. I've held off on doing that for now, and will do it in a separate PR.

kntsoriano: I also decided to break the state up into two reducers: one that deals with the submission lifecycle, another that deals with the revert lifecycle.

app/javascript/components/Editor.tsx
```diff
@@ -309,14 +251,96 @@ export function Editor({
...
+            revertDispatch({
+              type: RevertStatus.FAILED,
+              payload: { apiError: res.error },
+            })
+          })
+        }
+      })
+  }, [revertDispatch, sendRequest, submission])
+
+  useEffect(() => {
```
kntsoriano, line 317: When the submission status updates, we set the editor status.

app/javascript/components/Editor.tsx
```diff
@@ -309,14 +251,96 @@ export function Editor({
...
+      case SubmissionStatus.CREATING:
+        setStatus(EditorStatus.CREATING_SUBMISSION)
+        break
+      case SubmissionStatus.CREATING_ITERATION:
+        setStatus(EditorStatus.CREATING_ITERATION)
+        break
+    }
+  }, [submissionStatus])
+
+  useEffect(() => {
```
kntsoriano, line 334: When the revert status is set, we set the editor status.

app/javascript/components/Editor.tsx
```diff
@@ -309,14 +251,96 @@ export function Editor({
...
+      case RevertStatus.SUCCEEDED:
+        setStatus(EditorStatus.REVERTED)
+        break
+      case RevertStatus.FAILED:
+        setStatus(EditorStatus.REVERT_FAILED)
+        break
+    }
+  }, [revertStatus])
+
+  useEffect(() => {
```
kntsoriano, line 348: When an error comes in from either reverting and submitting, we set the API error.

Currently, we only have one place in the editor to show API errors. We can decide to have a separate place for each kind of error, but I decided to just put them in one place for now.

test/javascript/components/Editor/features/keyboard-shortcuts.test.js
```diff
@@ -0,0 +1,19 @@
+jest.mock('../../../../../app/javascript/components/editor/FileEditor')
```
kntsoriano, line 1: I decided to break up the `Editor.test.js` into smaller parts. It was getting too hard just to keep on adding tests to it.

app/serializers/serialize_submission.rb
```diff
@@ -19,6 +19,10 @@ def call
         test_run: Exercism::Routes.api_submission_test_run_url(
           submission.uuid,
           auth_token: user.auth_tokens.first.to_s
+        ),
+        files: Exercism::Routes.api_solution_original_solution_files_url(
```
iHiD:
```suggestion
        files: Exercism::Routes.api_solution_initial_files_url(
```

app/controllers/api/solutions/original_solution_files_controller.rb
```diff
@@ -0,0 +1,13 @@
+module API
+  module Solutions
+    class OriginalSolutionFilesController < BaseController
```
iHiD:
```suggestion
    class InitialFilesController < BaseController
```

app/controllers/api/solutions/original_solution_files_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+      def index
+        solution = current_user.solutions.find_by!(uuid: params[:solution_id])
```
iHiD: See top of api/files_controller for how we do this line elsewhere.

```suggestion
      begin
        solution = Solution.find_by!(uuid: params[:solution_id])
      rescue ActiveRecord::RecordNotFound
        return render_solution_not_found
      end
      
      return render_403(:solution_not_accessible) unless current_user.may_view_solution?(solution)
```

app/controllers/test/base_controller.rb
iHiD: I think we should just leave the header off anyway. It clashed on the notifications one too. So I'd just leave this as you've done it - not fix the CSS.

iHiD (approved):
Approving from a Rails POV pending my suggestions :)

No idea on the JS ;)

app/controllers/api/solutions/original_solution_files_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+
+        files = SerializeFiles.(solution.original_solution_files)
+
+        render json: { files: files }
```
iHiD: Please add a controller test that covers the various cases in this. See the other API controller tests for bits you can probably copy/paste in.

app/models/solution.rb
```diff
@@ -59,8 +59,12 @@ def editor_language
     track.slug
   end
 
+  def original_solution_files
```
iHiD:
```suggestion
  def initial_solution_files
```

app/models/solution.rb
```diff
@@ -59,8 +59,12 @@ def editor_language
...
+    git_exercise.editor_solution_files
+  end
+
   def editor_solution_files
-    files = git_exercise.editor_solution_files
+    files = original_solution_files
```
iHiD:
```suggestion
    files = initial_solution_files
```

app/models/solution.rb
```diff
@@ -59,8 +59,12 @@ def editor_language
     track.slug
   end
 
+  def original_solution_files
+    git_exercise.editor_solution_files
+  end
```
iHiD, line 64: This method needs a test adding.

app/serializers/serialize_files.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def call
+    files.map do |filename, content|
+      {
+        filename: filename,
+        content: content
+      }
+    end
+  end
+end
```
iHiD, line 14: This needs a test adding.

app/serializers/serialize_submission.rb
iHiD:
```suggestion
        initial_files: Exercism::Routes.api_solution_original_solution_files_url(
```

iHiD: Should this not be on the solution, not the submission? This doesn't feel like a property of a submission to me, it feels like a property of the whole solution.

app/javascript/components/Editor.tsx
ErikSchierboom: Extracting to a custom hook sounds good!

ErikSchierboom (approved):
This really helps structure things! Separating out code to different files was a very good idea.

app/serializers/serialize_submission.rb
kntsoriano: I agree. We pass in the submission to the editor, and not the solution, though. I'll open an issue to look into passing the whole solution instead.

PR #385, opened 2020-12-18
Add <Common.MarkdownEditor /> component

kntsoriano:
## Description
This adds the markdown editor component, which will be reused throughout the site.

## Screenshot
Screen Shot 2020-12-18 at 11 59 16 AM

## Questions
1. What options do we use from [here](https://github.com/Ionaru/easy-markdown-editor#options-list)?
2. Do we have a final design for the preview tab? For now, I just used the preview button (eye icon) in the markdown editor.


app/javascript/components/common/MarkdownEditor.tsx
```diff
@@ -0,0 +1,49 @@
...
+          if (!url) {
+            return 'Preview unavailable'
+          }
+
+          sendRequest({
+            endpoint: url,
+            body: JSON.stringify({ markdown: markdown }),
+            method: 'POST',
+            isMountedRef: isMountedRef,
+          }).then((json: any) => {
```
ErikSchierboom: What happens when there is an error returned by the HTTP request? Will it keep on showing 'Loading...'?

app/javascript/components/common/MarkdownEditor.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import SimpleMDE from 'react-simplemde-editor'
+import 'easymde/dist/easymde.min.css'
+import { sendRequest } from '../../utils/send-request'
+import { useIsMounted } from 'use-is-mounted'
+
+export const MarkdownEditor = ({ uuid }: { uuid: string }): JSX.Element => {
```
ErikSchierboom: I was initially wondering why the markdown editor needs a uuid :D

ErikSchierboom (approved):
Do we need some integration tests to verify that markdown in the component is actually rendered as HTML?

app/javascript/components/common/MarkdownEditor.tsx
iHiD: Why does it? :)

app/views/test/components/common/markdown_editor/show.html.haml
```diff
@@ -0,0 +1 @@
+= ReactComponents::Common::MarkdownEditor.new(SecureRandom.uuid)
```
iHiD: What's the purpose of the `uuid` here? If it's just random each time, then this should be moved into the `Component` class. 

```suggestion
= ReactComponents::Common::MarkdownEditor.new(SecureRandom.uuid)
```

iHiD:
Re the ID: I see now that it's used for the auto-save. I think my confusion is that it's called `uuid` (which is a type) rather than `id` (which is a property). So I think we just need a more descriptive name. Maybe `context_id` would make the most sense, as we're scoping it to some context (e.g. We might use `"mentor_discussion:#{discussion.uuid}"` as the id on a mentoring discussion page, or `bug_report_new` on that page.

---

1. What options do we use from here?

- `autosave:` enabled (You have all this set up)
- `blockStyles/italic:` `_`

I think beyond those we should just see how it works then improve it form there.

Nice work :)

app/controllers/api/markdown_controller.rb
```diff
@@ -0,0 +1,9 @@
+module API
+  class MarkdownController < BaseController
+    def parse
+      html = ParseMarkdown.(params[:markdown].to_s)
+
+      render json: { html: html }
+    end
+  end
+end
```
iHiD, line 9: Pls add a controller test that checks calling it with markdown returns html and is a 200.

app/views/layouts/application.html.haml
```diff
@@ -12,6 +12,7 @@
     %meta{ name: "language-server-url", content: Exercism.config.language_server_url }
     - if user_signed_in?
       %meta{ name: "bug-reports-url", content: Exercism::Routes.api_bug_reports_url(auth_token: current_user.auth_tokens.first.to_s) }
```
iHiD: This `auth_token` thing shouldn't be needed now.

app/views/layouts/application.html.haml
```diff
@@ -12,6 +12,7 @@
...
+      %meta{ name: "parse-markdown-url", content: Exercism::Routes.api_parse_markdown_url }
```
iHiD: I don't think any of these need to rely on `user_signed_in`. Let's remove that check, and just always render them in head. Makes it easier to consistently cache <head> when I add caching.

iHiD (approved):
Approved pending my comments :)

app/views/test/components/common/markdown_editor/show.html.haml
kntsoriano: I think we should configure the id per context, so it reads better when viewing the keys on localstorage. I just generated a random id for here as this is just a test page.

kntsoriano:
@ErikSchierboom Thanks. I've added the integration test and also added a `catch` block to show an error message when the API returns an error.

ErikSchierboom:
@kntsoriano Great!

PR #392, opened 2020-12-22
Add force_sync option to git track syncer

iHiD:
This adds a `force_sync` option that I can use in the bastion to sync things (and that we can have as a nightly task too). It's currently shallow (ie, it doesn't also force sync concepts and exercises). We can add that as a second step if desired.

### DIFF
--- a/test/commands/git/sync_track_test.rb
+++ b/test/commands/git/sync_track_test.rb
@@ -4,7 +4,17 @@ class Git::SyncTrackTest < ActiveSupport::TestCase
   test "no change when git sync SHA matches HEAD SHA" do
     track = create :track, slug: 'fsharp', synced_to_git_sha: "HEAD"
 
+    Git::SyncConceptExercise.expects(:call).never
     Git::SyncTrack.(track)
+
+    refute track.changed?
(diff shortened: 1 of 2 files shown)


test/commands/git/sync_track_test.rb
```diff
@@ -4,11 +4,21 @@ class Git::SyncTrackTest < ActiveSupport::TestCase
...
+    Git::SyncConceptExercise.expects(:call).never
     Git::SyncTrack.(track)
 
     refute track.changed?
   end
 
+  test "resyncs when force_sync is passed" do
+    track = create :track, slug: 'fsharp', synced_to_git_sha: "HEAD"
+
+    Git::SyncConceptExercise.expects(:call).at_least_once
```
ErikSchierboom, line 19: Perhaps also verify that the concepts are synced?

```suggestion
    Git::SyncConceptExercise.expects(:call).at_least_once
    Git::SyncConcept.expects(:call).at_least_once
```

And perhaps add a commented line for the practice exercise syncing?

ErikSchierboom (approved):

PR #396, opened 2021-01-04
Guidance panel

kntsoriano:
## Description
This adds the Guidance panel to the RHS of the mentoring page. As we don't have any endpoints yet to pull the contents of each, this PR only contains the accordion logic. With this as the base, it should be easy to slot in the contents of each when we have the endpoints ready.

### DIFF
--- a/app/javascript/components/common/Accordion.tsx
+++ b/app/javascript/components/common/Accordion.tsx
@@ -1,30 +1,30 @@
 import React from 'react'
 
 export type AccordionContext = {
-  index: string
+  id: string
   open: boolean
-  onClick: (index: string) => void
+  onClick: (id: string) => void
 }
 
 const Context = React.createContext<AccordionContext>({
-  index: '',
+  id: '',
   open: false,
-  onClick: (index: string) => {},
+  onClick: (id: string) => {},
 })
 
 export const Accordion = ({
-  index,
+  id,
   open,
   onClick,
   children,
 }: {
-  index: string
+  id: string
   open: boolean
-  onClick: (index: string) => void
+  onClick: (id: string) => void
   children?: React.ReactNode
 }): JSX.Element => {
   return (
-    <Context.Provider value={{ open, index, onClick }}>
+    <Context.Provider value={{ open, id, onClick }}>
       {children}
     </Context.Provider>
   )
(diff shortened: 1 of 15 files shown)


app/javascript/components/mentoring/Guidance.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import { Accordion } from '../common/Accordion'
+
+export const Guidance = (): JSX.Element => {
+  return (
+    <div>
+      <Accordion index="solution" open={true}>
+        <Accordion.Header>How you solved the exercise</Accordion.Header>
+        <Accordion.Panel>
+          <p>Solution here</p>
```
kntsoriano, line 56: This is where we will put the content of each.

iHiD (approved):
Thanks dude!

app/javascript/components/common/Accordion.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+
+export type AccordionContext = {
+  index: string
```
ErikSchierboom: I was slightly confused about `index` being of type `string`. With `index` I immediately assumed it was a number. Maybe a different name can be used, like `id` or `identifier` or something like that?

app/javascript/components/common/Accordion.tsx
```diff
@@ -0,0 +1,70 @@
...
+  open: boolean
```
ErikSchierboom: One small suggestion: maybe `isOpen` would be a more descriptive name? `open` sounds like it could also be a method. Just a suggestion though.

test/javascript/components/common/Accordion.test.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import { render, screen } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import { Accordion } from '../../../../app/javascript/components/common/Accordion'
+
+test('collapsed accordion', () => {
+  render(
+    <Accordion index="accordion" open={false}>
```
ErikSchierboom: The tests in this file refer to an accordion being collapsed or expanded, but the property of the accordion is named `open`. It might be good to also use the collapsed/expanded terminology for the property of the accordion.

ErikSchierboom (approved):
Great!

app/javascript/components/common/Accordion.tsx
kntsoriano: Cool. I've used `id` for this.

app/javascript/components/common/Accordion.tsx
kntsoriano: Cool. I've used `isOpen`.

test/javascript/components/common/Accordion.test.js
kntsoriano: I will change the test descriptions to use `open`.

PR #401, opened 2021-01-06
Add mentor discussion top bar

kntsoriano:
## Description
This adds the top bar to the `<MentorDiscussion />` component.

For reference, this is the [mock up](https://projects.invisionapp.com/share/RFWR89879K3#/screens/428025057).

In order for me to have done that, I needed to wrap our previous component (the RHS of this page) within this. With that, I also took the time to put the subcomponents within a separate `mentor-discussion/` directory so as not to clutter the main `mentoring/` directory.

The top bar includes the ff features:
1. Add a back button to go the exercise.
2. Show the solution info (track, student, exercise)
3. Add buttons of all iterations which includes the number of comments on each. There is also an unread state which colors the number of comments a bit different. When the iteration button is clicked, the posts for that specific iteration is shown. As an edge case, I've also hidden the number of comments when it is 0.
 
## Questions
1. I've called the component `<MentorDiscussion />`. Are we okay with this? I just inferred this since the class name is called set to this name.


app/helpers/react_components/mentoring/mentor_discussion.rb
```diff
@@ -0,0 +1,47 @@
...
+              icon_url: track.icon_url
+            },
+            exercise: {
+              title: exercise.title
+            },
+            iterations: discussion.solution.iterations.map do |iteration|
+              {
+                idx: iteration.idx,
+                num_comments: Solution::MentorDiscussionPost.where(discussion: discussion, iteration: iteration).count,
+                # TODO: @iHiD, I don't know how to get the correct unread state
```
kntsoriano: @iHiD pinging you here just in case you don't see this.

app/javascript/components/mentoring/mentor-discussion/SolutionInfo.tsx
```diff
@@ -0,0 +1,29 @@
...
+  track,
+  exercise,
+}: {
+  student: Student
+  track: Track
+  exercise: Exercise
+}): JSX.Element => {
+  return (
+    <>
+      <TrackIcon title={track.title} iconUrl={track.iconUrl} />
```
kntsoriano: @iHiD There is a similar block of code in `<Mentoring.Queue />`. Should we extract to one component or are they different?

app/javascript/components/mentoring/MentorDiscussion.tsx
```diff
@@ -0,0 +1,74 @@
...
+}: {
+  student: Student
+  track: Track
+  exercise: Exercise
+  links: Links
+  discussionId: number
+  iterations: Iteration[]
+}): JSX.Element => {
+  const [currentIteration, setCurrentIteration] = useState(
+    iterations[iterations.length - 1]
```
ErikSchierboom: Is there any scenario where `iterations` can be empty? Put another way: do we need to guard against the iterations possibly being empty?

app/javascript/components/mentoring/mentor-discussion/IterationButton.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+import { Iteration } from '../MentorDiscussion'
+
+const formatCommentCount = (count: number) => {
+  if (count >= 0 && count <= 9) {
+    return count
+  } else {
+    return '9+'
+  }
```
ErikSchierboom: Is the `count >= 0` needed? Currently, if the count is somehow less than zero, it would display `9+`. Maybe invert the logic to make it clear what the exceptional case is?

```suggestion
  if (count > 9) {
    return '9+'
  } else {
    return count
  }
```

Or even more concise:

```suggestion
  return count > 9 ? '9+' : count
```

app/javascript/packs/application.tsx
```diff
@@ -112,10 +118,21 @@ initReact({
...
     <Mentoring.Queue request={data.request} sortOptions={data.sort_options} />
   ),
-  'mentoring-mentoring-panel-list': (data: any) => (
-    <Mentoring.MentoringPanelList
+  'mentoring-mentor-discussion': (data: any) => (
+    <Mentoring.MentorDiscussion
       discussionId={data.discussion_id}
-      iterationIdx={data.iteration_idx}
+      student={
+        (camelizeKeys(data.student) as unknown) as MentorDiscussionStudent
```
ErikSchierboom: We have quite some `(camelizeKeys(X) as unknown) as Y` calls in our code base now. Maybe it would be nice to add a function that would allow us to do: `camelizeKeysAs<Y>(x)`?

ErikSchierboom (approved):

app/javascript/components/mentoring/MentorDiscussion.tsx
kntsoriano: I don't think so. You can't have a mentor discussion without submitting an iteration.

app/javascript/packs/application.tsx
kntsoriano: Sounds good! I will write that function.

app/javascript/components/mentoring/mentor-discussion/SolutionInfo.tsx
iHiD: Leave them as different for now. Thanks for checking.

app/helpers/react_components/mentoring/mentor_discussion.rb
iHiD: Me neither :) 

Thanks. I'll work it out tomorrow.

app/helpers/react_components/mentoring/mentor_discussion.rb
```diff
@@ -0,0 +1,47 @@
+module ReactComponents
+  module Mentoring
+    class MentorDiscussion < ReactComponent
```
iHiD: I think this can just be `Discussion`. We don't need the `Mentor` bit as it's namespaced.

app/helpers/react_components/mentoring/mentor_discussion.rb
```diff
@@ -0,0 +1,47 @@
...
+      initialize_with :discussion
+
+      def to_s
+        scratchpad = ScratchpadPage.new(about: discussion.solution.exercise)
+        student = discussion.solution.user
+        track = discussion.solution.track
+        exercise = discussion.solution.exercise
+
+        super(
+          "mentoring-mentor-discussion",
```
iHiD: And this should just be `mentoring-discussion` too.

iHiD:
@kntsoriano I've added a commit that adds flags for seen by mentor and student, and used it in the helper.

In a next step, we'll need to add something that marks an iteration's discussion posts as read when they get rendered by React. But let's discuss that once the first pass of things is done :)

kntsoriano:
@iHiD Sounds good. I've removed the `Mentor` prefix.

iHiD:
Sweet :)

PR #420, opened 2021-01-13
Small changes after submitting a post

kntsoriano:
## Description
This PR adds the following:
1. Close add a comment form after submission
2. Clear add a comment form after submission
3. Scroll to latest comment when a websocket message comes in

### DIFF
--- a/app/javascript/components/mentoring/discussion/DiscussionPostList.tsx
+++ b/app/javascript/components/mentoring/discussion/DiscussionPostList.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo } from 'react'
+import React, { useEffect, useMemo, useRef } from 'react'
 import { useRequestQuery } from '../../../hooks/request-query'
 import { DiscussionPost, DiscussionPostProps } from './DiscussionPost'
 import { DiscussionPostChannel } from '../../../channels/discussionPostChannel'
@@ -17,10 +17,9 @@ export const DiscussionPostList = ({
   endpoint: string
   discussionId: number
 }): JSX.Element | null => {
-  const { isSuccess, isLoading, data, refetch } = useRequestQuery<{
+  const { isLoading, data, refetch } = useRequestQuery<{
     posts: DiscussionPostProps[]
   }>(endpoint, { endpoint: endpoint, options: {} })
-
   const iterations = useMemo(() => {
     if (!data) {
       return []
@@ -40,6 +39,7 @@ export const DiscussionPostList = ({
       return iterations
     }, [])
   }, [data])
+  const lastPostRef = useRef<HTMLDivElement>(null)
 
   useEffect(() => {
     const channel = new DiscussionPostChannel(
@@ -52,6 +52,14 @@ export const DiscussionPostList = ({
     }
   }, [discussionId, refetch])
 
+  useEffect(() => {
+    if (!lastPostRef.current) {
+      return
+    }
+
+    lastPostRef.current.scrollIntoView()
+  }, [data])
+
   if (isLoading) {
     return (
       <div role="status" aria-label="Discussion post list loading indicator">
@@ -60,7 +68,7 @@ export const DiscussionPostList = ({
     )
   }
 
-  if (isSuccess && data) {
+  if (data) {
     return (
       <div className="discussion">
         {iterations.map((iteration) => {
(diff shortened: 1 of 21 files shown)


ErikSchierboom (approved):

app/javascript/components/mentoring/discussion/DiscussionPostList.tsx
```diff
@@ -51,6 +51,14 @@ export const DiscussionPostList = ({
     }
   }, [discussionId, refetch])
 
+  useEffect(() => {
+    if (!lastPostRef.current) {
+      return
+    }
+
+    lastPostRef.current.scrollIntoView()
```
iHiD:
```suggestion
    // TODO: Show a label saying "New message" and when the user clicks on that the scrolling happens
    lastPostRef.current.scrollIntoView()
```

iHiD:
I think we actually want to show a label saying "New message" when the websocket arrives, and when the user clicks on that the scrolling happens, else it will be annoying for a user if they're reading and it suddenly scrolls. But for now I've added a TODO for this.

If you submit a new post, the message should scroll then. I do still think that we shouldn't rely on the websocket to get that syncronous message back. I think the POST to the server should return the message, which gets rendered and scrolled to. Because if the websocket goes down for any reason (e.g. the server reboots), we'll get very weird behaviour for the user.

kntsoriano:
@iHiD I decided to solve the problems outlined above in this PR.

Instead of adding the data from the `POST` request, I decided to just refetch from the API when we receive a 200, as that's how I saw how people usually use `react-query`. We could also opt to use optimistic updates in the future.

The "new messages" button was also easy to implement. Just need some styling left.

iHiD:
> The "new messages" button was also easy to implement. Just need some styling left.

Sweet. I'll work this out the styling for this with @taiyab. This can be merged without that :)

---

> Instead of adding the data from the POST request, I decided to just refetch from the API when we receive a 200, as that's how I saw how people usually use react-query. We could also opt to use optimistic updates in the future.

OK. So this makes sense, but then I think whatever "loading" state the button has should stay in that state until the refetch has happened. So:
- Button clicked
- Button becomes disabled and text changes to "Submitting"
- API POST
- API GET new post
- Post is rendered
- Button resets, box resets/closes, etc

And if the GET new post fails, we need to handle that by retrying and then maybe refreshing the page if it fails again?

I basically don't want it to be possible that someone could have a successful submission state, but the post not appear in the discussion, and if we're doing multiple network calls that will happen unless we coordinate similarly to how I have it above.

SleeplessByte:
> Instead of adding the data from the POST request, I decided to just refetch from the API when we receive a 200, as that's how I saw how people usually use react-query. We could also opt to use optimistic updates in the future.

Refetching is _usually_ possible, so that's fine. However, in react-query, you're not **meant** to do that. The fastest loop is to merge the post response (if possible) with what is cached.

Examples in the react-query docs.

That's indeed 50% of the optimistic update approach. The difference is that the optimistic update approach merged the data _before_ the request (and removes it if it fails). This approach merges the data _after_ the request (and does nothing if it fails).

For this endpoint, which is really fast, it really doesn't matter. Just do whatever you have right now but know of the different options.

> And if the GET new post fails, we need to handle that by retrying and then maybe refreshing the page if it fails again?

You do not need to handle this @iHiD . React-query (by default) will retry 3 times before showing the error, with exponential back-off. My recommendation would be that the *error* message as a MANUAL retry button. If you're refreshing the page on failure, it might get into a loop state, DDOSing the server if there is an actual issue.

iHiD:
>  However, in react-query, you're not meant to do that. The fastest loop is to merge the post response (if possible) with what is cached.

>  React-query (by default) will retry 3 times before showing the error, with exponential back-off. My recommendation would be that the error message as a MANUAL retry button

So my take away from these two is that it's going to be lots more work to have the extra lookup, than to merge the initial post response. I'm writing in understanding that you're saying "POST + render response" is idiomatic here, @SleeplessByte? If so, I think that's what we should do @kntsoriano, as it's going to save us lots of extra work in the manual retry stuff.

SleeplessByte:
Manual retry is:

```javascript


if (error) {
  return <Button onPress={refetch}>...
}
```

Because `refetch` is already returned from react query. It will actually reset the state and take care of all the hard things.

That said: yes, using the post data directly is idiomatic, if there is post data.

iHiD:
Ok. Good to know. There's also messaging, error states, etc, to make it clear which I'd want to work out.

> That said: yes, using the post data directly is idiomatic, if there is post data.

I definitely think this is best then :)

kntsoriano:
@SleeplessByte @iHiD Thanks :) I've updated the code to use the data from the `POST` request.

iHiD:
Thanks @kntsoriano, and @SleeplessByte for your thoughts! :)

PR #422, opened 2021-01-13
Improvements

iHiD:
Various small improvements and stylings.


app/css/components/copy-text-to-clipboard.css
```diff
@@ -1,5 +1,8 @@
 @import "../styles";
 
+.c-react-wrapper-common-copy-to-clipboard-button {
+    min-height: 58px;
+}
```
iHiD, line 5: @kntsoriano @ErikSchierboom FYI, I'm adding these so that the page doesn't jump around when the react components render. This class automatically gets added to the divs with the react data now.

app/javascript/packs/application.tsx
```diff
@@ -199,6 +200,10 @@ initReact({
   ),
 })
 
+// TODO: Make this work with turbolinks
+import * as hljs from 'highlight.js'
+hljs.initHighlightingOnLoad()
+
```
iHiD, line 212: cc @kntsoriano

iHiD: also cc @joshgoebel as an FYI.

kntsoriano: I've changed this now in my new commit.

app/css/components/copy-text-to-clipboard.css
ErikSchierboom: 👍

app/helpers/view_components/published_solution.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def to_s
+      ApplicationController.render "components/published_solution",
+        locals: {
+          exercise: solution.exercise,
+          exercise_title: solution.exercise.title,
+          track_title: solution.track.title,
+          published_at: solution.published_at,
+          stars_count: 10, # TODO
```
ErikSchierboom: We're using `num_` elsewhere as the prefix for a count. Do we want to use that here too?

app/javascript/components/mentoring/discussion/FileViewer.tsx
```diff
@@ -9,15 +9,16 @@ export const FileViewer = ({
   language: string
   file: File
 }): JSX.Element => {
-  const [content, setContent] = useState('')
-
-  useLayoutEffect(() => {
-    setContent(highlight(language, file.content))
-  }, [file, language])
+  const parentRef = useHighlighting<HTMLPreElement>()
```
ErikSchierboom, line 12: Great to make this a React hook!

ErikSchierboom (approved):

app/helpers/view_components/published_solution.rb
iHiD: Good catch. Thanks 🙂

iHiD:
@kntsoriano Can you fix this pls?

PR #433, opened 2021-01-18
Add mentoring-sessions modal

iHiD:
@kntsoriano I've moved all modals that need implementing into this `temp` workspace. (Don't ask why I didn't use `tmp` like everywhere else - long story).

There are currently three - all three should be converted to React. We've discussed the bottom two, but there's also this new one, which shows all the sessions you've had as a mentor with a particular student. This will also have pagination.

---

http://local.exercism.io:3020/temp/modals/mentoring_sessions

Screenshot 2021-01-18 at 22 55 43

---

http://local.exercism.io:3020/temp/modals/publish_exercise

Screenshot 2021-01-18 at 22 56 42

---

http://local.exercism.io:3020/temp/modals/completed_exercise

Screenshot 2021-01-18 at 22 56 49


kntsoriano (approved):

iHiD:
Yo! :)

PR #440, opened 2021-01-20
Refactor badges to move the actual badge part into the DB

iHiD:
The first implementation of badges had the actual badge part (name, rarity, description, etc) hardcoded into the code of the website. This changes that to move them into an actual DB table. There are pros and cons to both. I'm going to sleep on it before deciding if this is the pattern I want to use.


app/models/badge.rb
```diff
@@ -1,26 +1,52 @@
 class Badge < ApplicationRecord
-  RARIRIES = %i[common rare ultimate legendary].freeze
+  has_many :acquired_badges, class_name: "User::AcquiredBadge", dependent: :destroy
 
-  belongs_to :user
+  RARIRIES = %i[common rare ultimate legendary].freeze
```
ErikSchierboom, line 4:
```suggestion
  RARITIES = %i[common rare ultimate legendary].freeze
```

app/models/badge.rb
```diff
@@ -1,26 +1,52 @@
...
+      rarity: rarity,
+      icon: icon,
+      description: description
+    }
   end
 
-  %i[name rarity icon description].each do |attr|
-    define_method(attr) { self.class.instance_variable_get("@#{attr}") }
+  # Badges are created on-demand, so if a new badge is added
+  # to the code, it will get added to the database on first usage.
```
ErikSchierboom, line 18: What is the reason for doing it this way instead of adding records manually? Easier to add new badges I guess?

ErikSchierboom (approved):
This looks nice. The main disadvantage would be that it becomes slightly slower to retrieve the badges I guess, due to a JOIN being required.

app/models/badge.rb
iHiD: Yeah. Just the fact that we'd have a race condition on deploy, and we'd need to have data being added during migrations which isn't an ideal pattern.

ErikSchierboom: Makes sense.

PR #447, opened 2021-01-22
Analysis improvements

iHiD:
@kntsoriano This is a load of work getting us to a point where we can do the automated-feedback properly.

I'll list any changes here that you need to know about as I go:

1. I've unified the submission serializer everywhere. This means you could have one "global" `Submission` class in your React stuff now, which is used everywhere, rather than varying it per component. There are a couple of things to check as there were variations as I merged the different serializers:
  a) The key is `id` not `uuid`. 
  b) The `test_status` etc are underscored. Previously in some places there were camelcase.
  c) The representer and analysis statuses are gone

2. I've changed the statuses for representation and analysis. I don't think this affects you, but it might do. Check `app/models/submission.rb` for the new statuses

3. I've added `automated_feedback_status` to the iteration serializer. This is `pending`, `present`, or `none`.
4. I've added `automated_feedback` to the iteration serializer. This is:

```ruby
{:html=>"<p>TODO: this should be json in reality</p>", :author=>{:name=>"The Ruby Analysis Team", :avatar_url=>"https://avatars.githubusercontent.com/u/5624255?s=200&v=4", :profile_url=>"#"}}
```

5. I've added the various states to `app/views/iterations/_iterations.html.haml` - these should work via the webhooks using 3 + 4 above.

---

@kntsoriano Could you PR to this branch to get everything working against the new serializers please. Then we can merge it and add any new stuff via a seperate PR. Thanks! 🙂


app/views/tracks/iterations/index/_iterations.html.haml
```diff
@@ -10,15 +10,33 @@
...
+                    = rounded_bg_img(feedback[:author][:avatar_url], "Erik SchierBOOM's uploaded avatar")
+                    = link_to "by #{feedback[:author][:name]}", feedback[:author][:profile_url]
+                    - if feedback[:author][:reputation]
+                      = render ViewComponents::Reputation.new(feedback[:author][:reputation])
+            - else
+              .automated-feedback-absent
+                = graphical_icon "graphic-conversation-1"
+                %h3 No auto suggestions? Try human mentoring.
+                %p Get real 1-to-1 human mentoring on the Leap exercise and start writing better C#.
+                = link_to "Get Mentoring", track_exercise_mentoring_index_path(iteration.track, iteration.exercise), class: 'btn-secondary'
```
iHiD, line 42: @kntsoriano This whole block should become a React Component basically. And we should listen to the `IterationChannel` for the changing `automated_feedback_status`.

You can get this working by doing the following:
1. Run up the tooling orchestrator and invoker, and make sure you have the Ruby Representer available in your dev env.
2. Create an iteration in the UI.
3. Go to http://local.exercism.io/maintaining/exercise_representations and enter some feedback by the iteration you've just made.
4. You should now see the feedback in the UI on the iterations page
5. Change states by running: `Submission.last.update(representation_status: :queued)` or `:generated` to toggle `automated_feedback_status` between `:pending` and `:present`, and toggle the presence of feedback (step 3) to switch `automated_feedback_status` between `:present` and `:none`.

---

If you need to rerun the representer, you can just do this in a `rails c`:
```reload!; Submission::Representation::Init.(Iteration.last.submission)```

File: app/models/submission.rb [hunk 2]
```diff
@@ -54,19 +54,60 @@ def broadcast!
...
   end
 
   memoize
   def automated_feedback
-    feedback = []
-    feedback << exercise_representation if exercise_representation&.has_feedback?
-    feedback
+    return nil unless has_automated_feedback?
+
+    exercise_representation&.has_feedback? ? representer_feedback : analyzer_feedback
```
ErikSchierboom, line 88: This means that can have either representer feedback or analyzer feedback, but not both, right? I think this makes sense.

test/models/submission_test.rb
```diff
@@ -39,8 +39,43 @@ class SubmissionTest < ActiveSupport::TestCase
...
+    assert_equal :pending, submission.automated_feedback_status
+
+    create :submission_representation, ast_digest: "foobar", submission: submission
+    er = create :exercise_representation, ast_digest: "foobar", exercise: submission.exercise
+    assert_equal :pending, Submission.find(submission.id).automated_feedback_status
+
+    er.update!(feedback_markdown: "foobar", feedback_author: create(:user))
+    assert_equal :present, Submission.find(submission.id).automated_feedback_status
+
+    # Present only if there is acutal feedback on analysis
```
ErikSchierboom:
```suggestion
    # Present only if there is actual feedback on analysis
```

ErikSchierboom (approved):

app/models/submission.rb
iHiD: Yes, although now we're changing that!

iHiD:
@kntsoriano I'm going to merge this. We have these three weird intermitent failing test though: https://github.com/exercism/v3-website/pull/447/checks?check_run_id=1772714232

```

Failure:
Flows::UserLoginTest#test_user_logs_in,_onboards,_and_is_redirected_to_the_correct_page [/home/runner/work/v3-website/v3-website/test/system/flows/user_login_test.rb:90]:
expected to find text "Join The Ruby Track" in "Just before you get started... 📝\nWe’re so excited you’re here and hope that Exercism will be a valuable and fun experience for you. We need to get a couple of quick legal bits out of the way to ensure you get the experience you expect, and to protect our community.\nOur Terms of Service\nIn summary, our Terms of Service say that:\nYou must act in a way that is in line with Exercism's values of respect and empathy per our\nCode of conduct\n. If you do not, we have the right to block your usage of the platform.\nAnything you submit to the site (e.g. your solutions, feedback, etc) remains your property, but you grant Exercism a licence to use it as we wish in perpetuity.\nYou may use exercises (provided by Exercism) and solutions (created by users) in accordance with the\nMIT\nand\nCC-BY-SA-4.0\nlicences respectively.\nYou can read our full\nTerms of Service\n.\nI accept Exercism's Terms of Service\nPrivacy Policy\nOur Privacy Policy is nice and simple. It says that we will never share your details with anyone outside of Exercism, and that we don't use third-party cookies, so no-one will track you here. You can\nread it in full here\n.\nI accept Exercism's Privacy Policy\nSave & Get Started". (However, it was found 1 time including non-visible text.)

rails test test/system/flows/user_login_test.rb:73
```

Could you take a look pls? They all pass on my machine. The `(However, it was found 1 time including non-visible text.)` feels odd to me...!

PR #456, opened 2021-01-27
Fix flakey tests

iHiD:


SleeplessByte (commented):
Are these failing because the data doesn't come back within 5 seconds?

iHiD:
No. We basically have no understanding of this :) Three tests sporadically fail unless this timeout is long. SSHing into the GHA and retrieving the screenshot shows what is expected, but the test says the (visible in screenshot) HTML isn't visible. Reducing the timeout to `1` also breaks it locally for Karlo. Increasing it fixes it here. What's annoying is these tests have worked for weeks, and *just* started doing this. The only relevant change I can see is that GitHub changed the version of Chrome on GHA 7 days ago.

iHiD:
> Are these failing because the data doesn't come back within 5 seconds?

Well, I say no, but maybe.

iHiD:
Closing in favour of https://github.com/exercism/v3-website/pull/458

PR #457, opened 2021-01-27
Bump gems

iHiD:
@kntsoriano Putting you in charge of upgrading us to oauth 2 when it's out: https://stackoverflow.com/questions/65702896/latest-omniauth-facebook-gem-breaks-devise

https://github.com/heartcombo/devise/pull/5327

Keep an eye on it over the next week or two :)


Gemfile.lock
```diff
@@ -338,7 +338,7 @@ GEM
     reverse_markdown (2.0.0)
       nokogiri
     rexml (3.2.4)
-    rubocop (1.10.0)
+    rubocop (1.11.0)
```
ErikSchierboom, line 341: Just as a reminder, you probably also want to update the Rubocop version here: https://github.com/exercism/website/blob/main/.github/workflows/tests.yml#L28

ErikSchierboom (approved):

PR #458, opened 2021-01-27
Observe body to fix flakey tests

iHiD:
This works, for reasons that are unclear. But it works.


iHiD:
@kntsoriano This fixes it. I'm merging. I have nothing more to say than 🤷 !!

iHiD:
(cc @ErikSchierboom too - this fixes the flakey tests)

PR #468, opened 2021-02-01
Add styling to guidance tab

iHiD:
@kntsoriano This adds styling to the guidance tab. It's going to need a little bit of work from you. Take a read of the comments - I've added TODOs for the various bits. I think it should be a quick one 🙂 

There is also a bug/missing-feature. Clicking an Accordion Heading that is open should close it (so that all the sections become closed). Could you fix that on this PR too please?

### DIFF
--- a/app/helpers/react_components/mentoring/discussion.rb
+++ b/app/helpers/react_components/mentoring/discussion.rb
@@ -9,10 +9,6 @@ module ReactComponents
           {
             discussion_id: discussion.uuid,
             user_id: current_user.id,
-            mentor: {
-              handle: mentor.handle,
-              avatar_url: mentor.avatar_url
-            },
             student: {
               name: student.name,
               handle: student.handle,
@@ -88,7 +84,12 @@ module ReactComponents
           num_stars: ms.num_stars,
           num_comments: ms.num_comments,
           published_at: ms.published_at,
-          web_url: Exercism::Routes.private_solution_url(ms)
+          web_url: Exercism::Routes.private_solution_url(ms),
+          mentor: {
+            handle: mentor.handle,
+            avatar_url: mentor.avatar_url
+          },
+          language: ms.editor_language
         }
       end
 
--- a/app/helpers/react_components/mentoring/discussion.rb
+++ b/app/helpers/react_components/mentoring/discussion.rb
@@ -9,6 +9,10 @@ module ReactComponents
           {
             discussion_id: discussion.uuid,
             user_id: current_user.id,
+            mentor: {
+              handle: mentor.handle,
+              avatar_url: mentor.avatar_url
+            },
             student: {
               name: student.name,
               handle: student.handle,
@@ -31,7 +35,8 @@ module ReactComponents
               title: exercise.title
             },
             iterations: iterations,
-            links: links
+            links: links,
+            mentor_solution: mentor_solution
           }
         )
       end
@@ -72,21 +77,22 @@ module ReactComponents
         end
       end
 
-      memoize
-      delegate :student, to: :discussion
+      def mentor_solution
+        ms = Solution.for(mentor, exercise)
+        return nil unless ms
 
-      memoize
-      delegate :mentor, to: :discussion
-
-      memoize
-      def track
-        discussion.solution.track
+        {
+          snippet: ms.snippet,
+          num_loc: ms.num_loc,
+          num_stars: ms.num_stars,
+          num_comments: ms.num_comments,
+          published_at: ms.published_at,
+          web_url: Exercism::Routes.private_solution_url(ms)
+        }
       end
 
       memoize
-      def exercise
-        discussion.solution.exercise
-      end
+      delegate :student, :mentor, :exercise, :track, to: :discussion
 
       memoize
       def scratchpad
--- a/app/helpers/react_components/mentoring/discussion.rb
+++ b/app/helpers/react_components/mentoring/discussion.rb
@@ -36,7 +36,8 @@ module ReactComponents
             },
             iterations: iterations,
             links: links,
-            mentor_solution: mentor_solution
+            mentor_solution: mentor_solution,
+            notes: notes
           }
         )
       end
@@ -89,6 +90,24 @@ module ReactComponents
           published_at: ms.published_at,
           web_url: Exercism::Routes.private_solution_url(ms)
         }
+      end
+
+      # TODO
+      def notes
+        '
+<h3>Talking points</h3>
+<ul>
+  <li>
+    <code>each_cons</code> instead of an iterator
+    <code>with_index</code>: In Ruby, you rarely have to write
+    iterators that need to keep track of the index. Enumerable has
+    powerful methods that do that for us.
+  </li>
+  <li>
+    <code>chars</code>: instead of <code>split("")</code>.
+  </li>
+</ul>
+        '.strip
       end
 
       memoize
--- a/app/helpers/react_components/mentoring/discussion.rb
+++ b/app/helpers/react_components/mentoring/discussion.rb
@@ -35,9 +35,9 @@ module ReactComponents
               title: exercise.title
             },
             iterations: iterations,
-            links: links,
             mentor_solution: mentor_solution,
-            notes: notes
+            notes: notes,
+            links: links
           }
         )
       end
--- a/app/javascript/components/common/Accordion.tsx
+++ b/app/javascript/components/common/Accordion.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { GraphicalIcon } from './GraphicalIcon'
 
 export type AccordionContext = {
   id: string
@@ -25,7 +26,7 @@ export const Accordion = ({
 }): JSX.Element => {
   return (
     <Context.Provider value={{ isOpen, id, onClick }}>
-      {children}
+      <div className="c-accordion-section">{children}</div>
     </Context.Provider>
   )
 }
@@ -45,7 +46,12 @@ const AccordionHeader = ({
       aria-controls={`accordion-panel-${id}`}
       id={`accordion-header-${id}`}
     >
-      {children}
+      {isOpen ? (
+        <GraphicalIcon icon="minus-circle" />
+      ) : (
+        <GraphicalIcon icon="plus-circle" />
+      )}
+      <div className="--title">{children}</div>
     </button>
   )
 }
--- a/app/javascript/components/mentoring/discussion/Guidance.tsx
+++ b/app/javascript/components/mentoring/discussion/Guidance.tsx
@@ -1,15 +1,18 @@
 import React, { useState, useCallback } from 'react'
 import { Accordion } from '../../common/Accordion'
+import { GraphicalIcon } from '../../common/GraphicalIcon'
+import { Avatar } from '../../common/Avatar'
+import { Icon } from '../../common/Icon'
 
 export const Guidance = (): JSX.Element => {
   const [accordionState, setAccordionState] = useState([
     {
-      id: 'solution',
-      isOpen: true,
-    },
-    {
       id: 'notes',
       isOpen: false,
+    },
+    {
+      id: 'solution',
+      isOpen: true,
     },
     {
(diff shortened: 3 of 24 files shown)


app/helpers/react_components/mentoring/discussion.rb
```diff
@@ -9,6 +9,10 @@ def to_s
           {
             discussion_id: discussion.uuid,
             user_id: current_user.id,
+            mentor: {
+              handle: mentor.handle,
+              avatar_url: mentor.avatar_url
+            },
```
iHiD: I've added this new key for displaying the info of the mentor. You could add these as prefixed keys (e.g. `mentor_handle`) into the `mentor_solution` block instead if you prefered that. I don't have a huge preference. I put them here as I figured we may choose to use them elsewhere in the future too.

app/helpers/react_components/mentoring/discussion.rb
```diff
@@ -72,21 +77,22 @@ def iterations
...
-      def track
-        discussion.solution.track
+        {
+          snippet: ms.snippet,
+          num_loc: ms.num_loc,
+          num_stars: ms.num_stars,
+          num_comments: ms.num_comments,
+          published_at: ms.published_at,
+          web_url: Exercism::Routes.private_solution_url(ms)
+        }
```
iHiD, line 93: These values are all correct and don't need you to do anything.

app/javascript/components/common/Accordion.tsx
```diff
@@ -45,7 +46,12 @@ const AccordionHeader = ({
       aria-controls={`accordion-panel-${id}`}
       id={`accordion-header-${id}`}
     >
-      {children}
+      {isOpen ? (
+        <GraphicalIcon icon="minus-circle" />
+      ) : (
+        <GraphicalIcon icon="plus-circle" />
+      )}
```
iHiD, line 53: Maybe you could add a test for this please?

app/helpers/react_components/mentoring/discussion.rb
```diff
@@ -72,21 +78,40 @@ def iterations
...
+    <code>each_cons</code> instead of an iterator
+    <code>with_index</code>: In Ruby, you rarely have to write
+    iterators that need to keep track of the index. Enumerable has
+    powerful methods that do that for us.
+  </li>
+  <li>
+    <code>chars</code>: instead of <code>split("")</code>.
+  </li>
+</ul>
+        '.strip
```
iHiD, line 111: I will do this in a seperate commit/PR. It's not blocking this.

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
   )
 
   return (
-    <div>
+    <>
+      <Accordion id="notes" isOpen={isOpen('notes')} onClick={handleClick}>
+        <Accordion.Header>Mentor notes</Accordion.Header>
+        <Accordion.Panel>
+          <div className="c-textual-content --small">
```
iHiD: **TODO for you.**

If notes is `present?` then this should show. There will also be a CTA to improve the notes. 

If notes is not present, then we'll show a (different) CTA to contribute instead.

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
+            {/* TODO: Replace with `notes` */}
```
iHiD: **TODO for you.**

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
         <Accordion.Header>How you solved the exercise</Accordion.Header>
         <Accordion.Panel>
-          <p>Solution here</p>
-        </Accordion.Panel>
-      </Accordion>
-      <Accordion id="notes" isOpen={isOpen('notes')} onClick={handleClick}>
-        <Accordion.Header>Mentor notes</Accordion.Header>
-        <Accordion.Panel>
-          <p>Notes here</p>
+          {/* TOOD: Replace "#" with mentor_solution.web_url */}
```
iHiD: **TODO for you.**

- There are various things that need the proper data putting in here.
- If mentor_solution is `null` then this whole accordion section (including the AccordionHeader) should not render.

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
+          <a href="#" className="c-published-solution">
```
iHiD: Note, that there is a normal published-solution component, but has slightly different data from this so I don't think we want to use `Common` component here (at least for now)

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
+                      alt="Number of times solution has been commented on"
+                    />
+                    {/*TODO: Replace with num_comments */}
+                    <div className="--num">2</div>
+                  </div>
+                </div>
+              </div>
+            </header>
+            <pre>
+              {/* TODO: Replcae with mentor_solution.snippet */}
```
iHiD: **TODO for you.**

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
+                  <span className="hljs-keyword">return</span>{' '}
+                  <span className="hljs-literal">true</span>
+                </span>
+                ;
+              </code>
+            </pre>
+            <footer className="--footer">
+              <div className="locs">
+                <GraphicalIcon icon="loc" />
+                {/* TODO: Replace with mentor_solution.num_loc */}
```
iHiD: **TODO for you.**

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
+                    />
+                    {/*TODO: Replace with num_stars */}
+                    <div className="--num">10</div>
+                  </div>
+                  <div className="--count">
+                    <Icon
+                      icon="comment"
+                      alt="Number of times solution has been commented on"
+                    />
+                    {/*TODO: Replace with num_comments */}
```
iHiD: **TODO for you.**

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
+                </div>
+
+                {/*TODO: If mentor_solution.published_at? */}
+                <div className="--counts">
+                  <div className="--count">
+                    <Icon
+                      icon="star"
+                      alt="Number of times solution has been stared"
+                    />
+                    {/*TODO: Replace with num_stars */}
```
iHiD: **TODO for you.**

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
+                  src="https://avatars2.githubusercontent.com/u/5337876?s=460&v=4"
+                />
+
+                <div className="--info">
+                  <div className="--exercise-title">Your Solution</div>
+                  {/* TODO: Replace "Leap" and "Ruby" */}
+                  <div className="--track-title">to Leap in Ruby</div>
+                </div>
+
+                {/*TODO: If mentor_solution.published_at? */}
```
iHiD: **TODO for you.**

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
+              <div className="--exercise">
+                {/* TODO: Replace with mentor.handle and mentor.avatar_url */}
+                <Avatar
+                  handle="iHiD"
+                  src="https://avatars2.githubusercontent.com/u/5337876?s=460&v=4"
+                />
+
+                <div className="--info">
+                  <div className="--exercise-title">Your Solution</div>
+                  {/* TODO: Replace "Leap" and "Ruby" */}
```
iHiD: **TODO for you.**

app/javascript/components/mentoring/discussion/Guidance.tsx
```diff
@@ -45,21 +48,178 @@ export const Guidance = (): JSX.Element => {
...
-      </Accordion>
-      <Accordion id="notes" isOpen={isOpen('notes')} onClick={handleClick}>
-        <Accordion.Header>Mentor notes</Accordion.Header>
-        <Accordion.Panel>
-          <p>Notes here</p>
+          {/* TOOD: Replace "#" with mentor_solution.web_url */}
+          <a href="#" className="c-published-solution">
+            <header className="--header">
+              <div className="--exercise">
+                {/* TODO: Replace with mentor.handle and mentor.avatar_url */}
```
iHiD: **TODO for you.**

app/javascript/components/common/Accordion.tsx
kntsoriano: I think this is fine. We already have a test for the correct aria attributes in the JS. We can't also assert on the tests unless we change this to an `<Icon />`and add the correct alt attribute. The alt attribute might be redundant as we already have supplied the correct aria attributes.

kntsoriano:
@iHiD Added the commits.

app/helpers/react_components/mentoring/discussion.rb
```diff
@@ -72,21 +74,45 @@ def iterations
         end
       end
 
-      memoize
-      delegate :student, to: :discussion
+      def mentor_solution
+        ms = Solution.for(mentor, exercise)
+        return nil unless ms
```
ErikSchierboom, line 79: I see a lot of places where `return unless` is used, without the explicit `nil`. Is there a reason when to add it or not?

app/javascript/components/mentoring/discussion/MentorNotes.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+export const MentorNotes = ({ notes }: { notes?: string }): JSX.Element => {
+  if (notes) {
+    return (
+      <React.Fragment>
```
ErikSchierboom, line 6: I'm seeing usage of both `<>` and `React.Fragment` in various places. They are the same, right? And if so, what is the benefit of using `React.Fragment` over `<>`?

ErikSchierboom (approved):

app/helpers/react_components/mentoring/discussion.rb
iHiD: I never remember that `return unless ms` returns `nil` so I always add it. I quite like the explicitness. But the two are equivalent. I'm surprised Rubocop doesn't standardise.

app/javascript/components/mentoring/discussion/MentorNotes.tsx
iHiD: I always use `<>` if I can, but if attributes are needed (for example a `ref="..."` then you have to use the full `<React.Fragment>`. 

@kntsoriano may know of some other reasons that I'm not aware of.

iHiD:
Sweet work @kntsoriano. I'm merging this down! Thanks for the review @ErikSchierboom :)

app/javascript/components/mentoring/discussion/MentorNotes.tsx
kntsoriano: I don't know why I choose one over the other. I guess doing `<React.Fragment />`is the wiser choice as we don't need to change much if we want to add ref to it? No spectific preference, though.

PR #471, opened 2021-02-02
Generate ids for dropdown component

### LINKED ISSUE #469: Add random ID to dropdown component
From https://github.com/exercism/website/pull/405.

kntsoriano:
Closes https://github.com/exercism/website/issues/469.

### DIFF
--- a/app/javascript/components/dropdowns/Dropdown.tsx
+++ b/app/javascript/components/dropdowns/Dropdown.tsx
@@ -1,9 +1,9 @@
-import React, { useState, useEffect, useRef } from 'react'
+import React, { useState, useEffect, useRef, useMemo } from 'react'
 import { usePanel } from '../../hooks/use-panel'
+import { v4 as uuidv4 } from 'uuid'
 
 type MenuButton = {
   label: string
-  id: string
   className: string
   html: string
 }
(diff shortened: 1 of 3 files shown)


app/javascript/components/dropdowns/Dropdown.tsx
```diff
@@ -1,9 +1,9 @@
-import React, { useState, useEffect, useRef } from 'react'
+import React, { useState, useEffect, useRef, useMemo } from 'react'
 import { usePanel } from '../../hooks/use-panel'
+import { v4 as uuidv4 } from 'uuid'
```
ErikSchierboom, line 3: Good idea to rename!

test/javascript/components/dropdowns/Dropdown.test.js
```diff
@@ -7,7 +7,6 @@ import userEvent from '@testing-library/user-event'
 test('down arrow opens menu on first item', async () => {
```
ErikSchierboom, line 7: Would it make sense to add a test to verify that the ID is generated randomkly?

ErikSchierboom (approved):

iHiD (approved):
This is great. Thanks :)

app/javascript/components/dropdowns/Dropdown.tsx
SleeplessByte: image

From the docs 🖌️

test/javascript/components/dropdowns/Dropdown.test.js
SleeplessByte: Meh, it's more important that each item has a unique ID, than that they are random. I would test: each item has an ID and non of the ID is the same.

PR #473, opened 2021-02-03
Rework the track feed

iHiD:
This provides the code to power the new track activity feed.

Running `rake db:migrate:reset db:seed`, logging in, then going to `http://local.exercism.io:3020/tracks/ruby` should present you with this:

Screenshot 2021-02-03 at 22 01 18

@kntsoriano The activities section needs turning into a React Component. I've tried to make the code as friendly as possible for you in: `app/views/tracks/show/joined/_activities_section.html.haml`

For now, you can just make this a static component, but in the future we'll add some websockets and the ability to see more, both in terms of more exercises, but also more activities within an exercise. But we'll worry about that later. For now, let's just get the first pass static.


iHiD:
@kntsoriano This is ready for you to add commits to.

iHiD:
@kntsoriano I'm merging this down. I decided I agree that this isn't a priority to get into React now. If you've already started it, that's fine though - we still need it done before launch 🙂

PR #474, opened 2021-02-03
Add mentoring queue filter

kntsoriano:
## Screenshot
I've enclosed some sections in different colored rectangles. I will be using those to them in the sections below.

Screen Shot 2021-02-03 at 5 27 30 PM

### Expected API response
Currently while the API isn't implemented yet, I've just assumed that this is the API response:

```
{
  results: [],
  meta: { total: number, queryTotal: number, page: number }
}
```

#### Result attributes
Not really firm on this. I will change once we firm up the API. Currently, the [mock controller](https://github.com/exercism/website/pull/474/files#diff-49ed32c5665559e611356bfdfe72c25a99754b0725b8c86adda8d50c9a786780) that we have has the result attributes listed there.

#### Meta attributes
`total`: number of all mentor requests.
`queryTotal`: number of filtered mentor requests.
`page`: current page number

### Requests count
This section is enclosed in the yellow rectangle.

A distinction between `total` and `queryTotal` in the API response was needed in order to implement this section.

### Filtering by track
The section is enclosed in the green rectangle.

In this PR, we pass in all the track filters to the component as props. We don't get them from the API.

### Filtering by exercise
The section is enclosed in the blue rectangle.

Like the tracks, we pass in the exercise filters as props. We don't get them from the API.

### Exercise search bar
The section is enclosed in the red rectangle.

I haven't implemented this.

### Add or remove tracks modal
The section is enclosed in the orange rectangle.

I haven't implemented this. This will be implemented in another PR.

## Questions
1. Is it correct to just pass in the track filters as props? Not getting them via the API?
2. Is it correct to just pass in the exercise filters as props? Not getting them via the API?
3. How do we implement exercise search? Do we do it via API? Or just a local filtering of the passed exercises prop?


SleeplessByte:
From a software standpoint:

1. probably fine
2. probably fine

You'll update the UI when you support more tags.

app/controllers/test/components/mentoring/queue_controller.rb
```diff
@@ -56,9 +60,11 @@ def solutions
     end
     results = sort_solutions(results) if params[:order].present?
 
+    results = filter_solutions(results) if params[:filter]
+
     render json: {
       results: results[page - 1, per],
-      meta: { current: page, total: results.size }
+      meta: { current: page, queryTotal: results.size, total: requests.size }
```
ErikSchierboom: I didn't immediately understand what `queryTotal` meant. Maybe `pageTotal`? I often see `count` used instead of `total` BTW, but that could be me :)

app/javascript/components/mentoring/queue/SolutionList.jsx
```diff
@@ -31,9 +27,9 @@ export function SolutionList({ request, setPage }) {
 
   return (
     <div>
-      {isLoading && <Loading />}
-      {isError && <p>Something went wrong</p>}
-      {isSuccess && (
+      {status === 'loading' && <Loading />}
+      {status === 'error' && <p>Something went wrong</p>}
+      {status === 'success' && (
```
ErikSchierboom: What was the reason for this change? I sort of liked the previous version better (not that I mind it much)

ErikSchierboom (approved):

iHiD:
> `total` and `queryTotal`

I suggest `total` for the smaller response total, so that it's consistent with our existing pagination serializer; and using `unscopedTotal` for the wider, larger total.

---

> Filtering by track

- It's fine to pass these via props.
- I want to change these to be radio buttons. 

> Filtering by exercise

- It's fine to pass these via props too.
- I think the structure for all this should be and array of tracks, each containing their exercises with `name`, `count` and also a `mentor_completed` boolean (see below)
- The exercises list should be filtered by the selected track (so changing the track radio button changes the exercises that are selectable)
- Checking/unchecking a track or exercise changes executes a query.

> Exercise search bar

This should be local to JS. There will also be a filter for "only show exercises I've completed" (which uses the `mentor_completed` boolean), so we'll need to add that too. I'm waiting on Tai for the design. There will also be "select all" and "select none" buttons for exercises.

> Add or remove tracks modal

Changing the tracks via this modal will then need to trigger a change in the list of tracks (and list of exercises)

---

In terms of the API, we use `api/mentor_requests` endpoint already, and it already uses pagination. I think this needs `track_id` and `exercise_ids` params honouring, and an extra response field for the `unscopedTotal`?

iHiD:
@kntsoriano I get the following errors when running this PR:

Screenshot 2021-02-03 at 22 08 15

---

Update: This is because the `query_total` field was missing from the real serializer. We should update the test code to use the real serializer soon to protect us here.

iHiD:
@kntsoriano I've added a commit that ff:
- Honours `track_id` and `exercise_ids`
- Adds a test for paginated serialiser
- Adds ability to add keys to meta and adds stubbed `query_total` (which I think should be renamed to `unpaginated_total` here)

app/javascript/components/mentoring/queue/SolutionList.jsx
kntsoriano: We pass in the status into this component now. If we pass each of the booleans, it will add extra props.

ErikSchierboom: Ah okay. Makes sense

iHiD:
@kntsoriano 
Various things I've done:
- Change tracks to be radio buttons
- Added checkboxes (unimplemented) for only showing exercise that "need mentoring" are "mentor has completed".
- In the React helper initializer:
  -  Added a `selected` key to the tracks
  -  Added all the exercises of the selected track.
  - Efficiently calculated the counts.
  - The React component requires a user now.
- Implemented `count` (track + exercises), `completed_by_mentor` (exercise) and `selected` (track)
- Renamed `queryTotal` to `unscopedTotal` and implemented it
- Added an endpoint to retrieve tracks when the user changes their subscribed tracks (`tracks_api_mentor_requests_path`)
- - Added an endpoint to retrieve exercises when the user chooses a different track via the radio button (`exercises_api_mentor_requests_path` - expect a track slug as `params[:track_id]`)

Various things I've noticed:
- Changing the track on the Queue sends a request to `/api/v1/mentor_discussions?page=1`. I think this is a bug.
- Changing the track on the Queue sends a request to `/api/v1/mentor_discussions/tracks". This should be `/api/v1/mentor_requests/tracks`
- Changing the track on the Queue sends a request to `/api/v1/mentor_requests?page=1&filter%5Btrack%5D%5B%5D=csharp&filter%5Btrack%5D%5B%5D=ruby`.
  - This ideally should use `track_id` not, `filter[track]`. If it is a lot easier for you to have the `filter[]` bit, just pls change it to `filter[track_id]`, but I'd rather without to be consistent with elsewhere.
  - `track_id` should be a single thing, not an array.
  - This ideally should use `exercise_ids`, which should be an array.
  - They use slugs, which is fine
- Pagination isn't implemented/working correctly on the React side.

Extra things we need to add:
- Select all (All exercises that are visible become checked and an API request is sent. If you have a filter box ("only completed" or "require mentoring" selected, then those are still honoured and those exercises do not become checked)
- Select None (Deselects all exercises. Sends an API request with an empty array).
- Only show exercises I've completed (checking this uncheck any exercises that you've not completed, hide them, and fire an API request)
- Only show exercises that require mentoring (`count > 0`) (have this on by default. Changing it only changes the list - doesn't need an API request)

Then:
- By default all exercises that require mentoring should be checked.
- Question: Can you actually easily pass an empty array for exercises to the API (as opposed to the key missing), or should we just have special flag to indicate no exercises?

kntsoriano:
@iHiD Yes, we're able to pass an array. I did that now in the commits.

I've implemented all the changes you've outlined above. Could you please double check?

I have a few questions that follows from those changes:
1. When do we clear the textbox?

iHiD:
@kntsoriano There was a bug in Pagination where the serializers were using `total_count`, but the JS was expecting this to be `total_pages`. So if there were `n` results, the JS was showing `n` pages.

I've fixed this by renaming things for clarity.

iHiD:
@kntsoriano I think this is all good for now 🙂 Thanks :)

There will be some improvements/tweaks we need to make but I think they're probably all CSS/HTML at this stage, so let's merge this and I can do that seperately when we've got final designs on bits from @taiyab. Nice work!

kntsoriano:
@iHiD I've fixed the tests. Let's merge.

iHiD:
Nice work. Thanks @kntsoriano 💙

PR #479, opened 2021-02-09
Inbox filter

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/97.
Closes https://github.com/exercism/v3-project-management/issues/88.

## Description
This implements the inbox track filter as a menu.


app/javascript/components/dropdowns/Dropdown.tsx
```diff
@@ -20,123 +19,30 @@ export const Dropdown = ({
...
-          role="menu"
-          hidden={!open}
-        >
+      <div {...panelAttributes}>
+        <ul className={`${menuButton.className}-dropdown`} {...listAttributes}>
           {menuItems.map((item, i) => {
             return (
               <li
-                ref={(instance) => handleMenuItemMount(instance, i)}
                 key={item.html}
```
ErikSchierboom: I don't know much React, but would setting the key to the index result in less data in the HTML?

app/javascript/components/dropdowns/Dropdown.tsx
```diff
@@ -20,123 +19,30 @@ export const Dropdown = ({
...
-      return
-    }
-
-    menuItemElementsRef.current[focusIndex].focus()
-  }, [open, focusIndex, buttonRef])
+  const {
+    buttonAttributes,
+    panelAttributes,
+    listAttributes,
+    itemAttributes,
```
ErikSchierboom, line 26: So clean!

app/javascript/components/mentoring/inbox/TrackList.tsx
```diff
@@ -0,0 +1,80 @@
...
+  value: string | null
+  setTrack: (value: string | null) => void
+}): JSX.Element => {
+  const {
+    buttonAttributes,
+    panelAttributes,
+    listAttributes,
+    itemAttributes,
+  } = useDropdown(tracks.length)
+  const selected = tracks.find((track) => track.slug === value) || tracks[0]
```
ErikSchierboom, line 55: Are these the tracks that a mentor is mentoring? I guess there will always be at least one?

ErikSchierboom (approved):

app/javascript/components/mentoring/inbox/TrackList.tsx
iHiD: If there isn't then the attribute shouldn't appear at all. I'll discuss this with Tai.

iHiD:
@kntsoriano Nice work :) I've added commits for styling. 

There's one bug I found, which is you don't seem to be able to select one of the options using the keyboard. Pressing space or enter just closes the dialog.

kntsoriano:
@iHiD Fixed.

PR #482, opened 2021-02-10
Add notifications styling

iHiD:
@kntsoriano I added this so that you have the styling for the notifications.

I'll be adding more commits to this branch to improve the functionality of notifications and implement some of my TODOs for it.

You can see the styling here: http://local.exercism.io:3020/temp/modals/notifications

Screenshot 2021-02-10 at 22 15 10


iHiD:
@kntsoriano Two things

1. I've added some badges code to the top of the notifications one. The logic in the HAML should be the same when you get to React :)

2. When you've done the notifications dropdown (ie we've got it all merged and are happy with it) you should now be able to do the reputation dropdown too. That lives at http://local.exercism.io:3020/temp/modals/reputation and looks like this:
Screenshot 2021-02-12 at 00 07 38

iHiD:
(the failing test is fixed in #485)

PR #484, opened 2021-02-11
Generate README/HINTS/HELP markdown files

ErikSchierboom:
Closes https://github.com/exercism/v3-project-management/issues/126


app/models/git/exercise.rb
```diff
@@ -73,24 +93,38 @@ def solution_files
...
+
+    memoize
+    def cli_filepaths
+      special_filepaths = ['README.md', 'HELP.md']
+      special_filepaths << 'HINTS.md' if filepaths.include?('.docs/hints.md')
+
+      filtered_filepaths = filepaths.select do |filepath| # rubocop:disable Style/InverseMethods
+        !filepath.match?(track.ignore_regexp) &&
+          !filepath.start_with?('.docs/') &&
+          !filepath.start_with?('.meta/')
```
ErikSchierboom: I've hardcoded the excluding of files in the `.docs` and `.meta` directory, but let me know if that is too course. I could do individual files.

app/models/git/exercise.rb
```diff
@@ -52,12 +62,22 @@ def example
 
     memoize
     def authors
-      config[:authors]
+      config[:authors] || []
```
ErikSchierboom: This makes it slightly easier to work with.

app/controllers/api/files_controller.rb
```diff
@@ -16,7 +16,7 @@ def show
 
       unless content
         begin
-          content = Git::Exercise.for_solution(solution).read_file_blob(params[:filepath])
+          content = solution.read_file(params[:filepath])
```
ErikSchierboom, line 19: I've opted for introducing a `read_file` method on the `Solution` class, which itself can then handle the distinction between real files and our dynamically created files (like `README.md`). Let me know if this makes sense.

app/commands/markdown/preprocess.rb
```diff
@@ -1,9 +1,17 @@
 class Markdown::Preprocess
   include Mandate
 
-  initialize_with :text
+  def initialize(text, strip_h1: true) # rubocop:disable Naming/VariableNumber
```
ErikSchierboom: Rubocop doesn't like the `h1` suffix. I've disabled it via comments for now. Let me know if that makes sense.

app/commands/markdown/parse.rb
```diff
@@ -8,12 +8,13 @@ def self.call(*args, **kwargs)
     new(*args, **kwargs).()
   end
 
-  def initialize(text, nofollow_links: false)
+  def initialize(text, nofollow_links: false, strip_h1: true) # rubocop:disable Naming/VariableNumber
     # TODO: We almost certainly don't want to do this!
     # but for now let's reduce the heading level of all
     # headings, as they're too high in the actual docs atm
     @text = text.to_s.gsub(/^##/, '###')
```
ErikSchierboom: I wondered if this would make more sense in the preprocessor.

app/commands/markdown/preprocess.rb
```diff
@@ -1,9 +1,17 @@
...
+    @text = text.to_s
+    @strip_h1 = strip_h1 # rubocop:disable Naming/VariableNumber
+  end
 
   def call
-    text.gsub(/^`{3,}(.*?)`{3,}\s*$/m) { "\n#{Regexp.last_match(0)}\n" }
+    preprocessed_text = text.gsub(/^`{3,}(.*?)`{3,}\s*$/m) { "\n#{Regexp.last_match(0)}\n" }
+    preprocessed_text = preprocessed_text.gsub(/^# .+?$/, '') if strip_h1 # rubocop:disable Naming/VariableNumber
```
ErikSchierboom: I've gone with a very simple replace for h1s.

app/commands/solution/generate_readme_file.rb
```diff
@@ -0,0 +1,81 @@
+class Solution
+  class GenerateReadmeFile
+    include Mandate
+
+    initialize_with :solution
+
+    def call
+      [preamble, introduction, instructions, source].compact_blank.join("\n\n")
```
ErikSchierboom: I have no idea if think somewhat functional style makes sense, or if you'd prefer a private field that was updated by methods.

test/commands/markdown/parse_test.rb
```diff
@@ -6,8 +6,7 @@ class Markdown::ParseTest < ActiveSupport::TestCase
   end
 
   test "converts markdown to html" do
-    expected = '<h1>OHAI</h1>
-<p>So I was split between two ways of doing this.</p>
+    expected = '<p>So I was split between two ways of doing this.</p>
```
ErikSchierboom, line 9: I had to remove the h1 as that is now stripped by default. I felt that removing it here didn't invalidate the test.

test/controllers/api/files_controller_test.rb
```diff
@@ -53,6 +53,39 @@ class API::FilesControllerTest < API::BaseTestCase
...
 
+  test "show should return special README.md solution file" do
+    setup_user
+    solution = create :concept_solution, user: @current_user
+    create :submission, solution: solution
+    content = "README"
+
+    get "/api/v1/solutions/#{solution.uuid}/files/README.md", headers: @headers, as: :json
+    assert_response 200
+    assert content, response.body
```
ErikSchierboom: I copied this from the other tests, but is this not actually a bad test? It doesn't fail where I think it should be. It looks like the original intent might have been to check if the body contained a certain text.

test/controllers/api/markdown_controller_test.rb
```diff
@@ -5,12 +5,12 @@ class API::MarkdownControllerTest < API::BaseTestCase
     setup_user
 
     post api_parse_markdown_path,
-      params: { markdown: "# Hello" },
+      params: { markdown: "*Hello*" },
```
ErikSchierboom, line 10: Changed due to the h1 stripping being the default now

test/models/git/exercise_test.rb
```diff
@@ -18,27 +18,128 @@ class ExerciseTest < ActiveSupport::TestCase
...
+      expected_filepaths = [
+        "README.md",
+        "HELP.md",
+        "anagram.rb",
+        "anagram_test.rb"
+      ]
+      assert_equal expected_filepaths, exercise.cli_filepaths
+    end
+
+    test "retrieves instructions" do
```
ErikSchierboom, line 76: Some test cases were missing, so I've added them

test/models/solution_test.rb
```diff
@@ -127,6 +127,112 @@ class SolutionTest < ActiveSupport::TestCase
     assert solution.exercise_solution_files["log_line_parser.rb"].start_with?("module LogLineParser")
   end
 
+  test "read_file for concept exercise returns exercise file" do
```
ErikSchierboom, line 130: For now, I opted to add the tests to this file instead of separate test files for the generate classes. I think that if we have the files controller tests verify that the right generated file is returned, it is fine to move these tests.

app/commands/markdown/parse.rb
iHiD: > I wondered if this would make more sense in the preprocessor.

I don't really understand where these different bits fit together now, so I'll be guided by you.

---

I suspect we want another flag which is `lower_heading_levels_by: 1`, which controls this. In different places, we'll want to drop headings by different amount of levels, now we've increased them all to h2 everywhere.

ErikSchierboom: Indeed. I think that makes sense.

app/commands/markdown/parse.rb
```diff
@@ -8,12 +8,13 @@ def self.call(*args, **kwargs)
     new(*args, **kwargs).()
   end
 
-  def initialize(text, nofollow_links: false)
+  def initialize(text, nofollow_links: false, strip_h1: true) # rubocop:disable Naming/VariableNumber
```
iHiD: Can you see if rubocop allows you to specifiy exceptions. I think the key might be `AllowedIdentifier` in the rubocop.yml 

See here: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/VariableNumber

app/commands/markdown/preprocess.rb
iHiD: Replied to this above.

app/commands/markdown/preprocess.rb
iHiD: Few thoughts:
1. Could you split this (and the line above it) into methods so that we can see what they do from their names pls?
2. In terms of the regexp, it looks sane, yep.
3. However, I don't think this approach will work. This will also remove any lines containing code with comments (e.g. `# xyz` is the standard format for comments in Ruby), so I think we should actually post-process the H1 out of the resulting HTML, rather than preprocess it here. Which will also mean we'll have to not lower h1 headings in the bit that does that.

app/commands/solution/generate_help_file.rb
```diff
@@ -0,0 +1,26 @@
+class Solution
+  class GenerateHelpFile
+    include Mandate
+
+    initialize_with :solution
+
+    def call
+      "# Help\n\n#{tests}\n\n#{submitting}\n\n#{help}"
```
iHiD: Could you use [HEREDOC](https://www.rubyguides.com/2018/11/ruby-heredoc/) for this rather than embedding the newlines pls?

app/commands/markdown/preprocess.rb
iHiD: Actually, thinking about this more, I think there's an easier approach, which is to limit this to the very first line of the file. You'll also need to strip the whitespace line below it.

app/commands/solution/generate_readme_file.rb
iHiD: I'm good with this.

I'd probably format it like this though for readibility:

```suggestion
      [
        preamble, 
        introduction,
        instructions, 
        source
      ].compact_blank.join("\n\n")
```

Bonus points for `compact_blank` - that's new to me! :)

app/commands/solution/generate_readme_file.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    private
+    def preamble
+      if solution.git_exercise.hints.present?
+        hints_text = "If you get stuck on the exercise, check out `HINTS.md`, but try and solve it without using those first :)" # rubocop:disable Layout/LineLength
```
iHiD: Can you move this and any other copy into i18n pls? Probably a new file for it - probably called `exercises` (see notifications for an example of how to do this - look at `config/locales` and great for `I18n` in `app/models`

app/commands/solution/generate_readme_file.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      "# #{solution.exercise.title}
+
+Welcome to #{solution.exercise.title} on Exercism's #{solution.track.title} Track.
+If you need help running the tests or submitting your code, check out `HELP.md`.
+#{hints_text}".strip
+    end
+
+    def introduction
+      return unless solution.git_exercise.introduction
```
iHiD:
```suggestion
      return unless solution.git_exercise.introduction.present?
```

(I could see tracks having blank files)

app/commands/solution/generate_readme_file.rb
```diff
@@ -0,0 +1,81 @@
...
+        if user[:exercism_username].blank?
+          "- @#{user[:github_username]}"
+        else
+          "- #{user[:exercism_username]} (@#{user[:github_username]})"
+        end
+      end.join("\n")
+    end
+
+    def based_on
+      return if solution.git_exercise.source.blank? && solution.git_exercise.source_url.blank?
```
iHiD: I have a small preference for inverting this:
```
return unless solution.git_exercise.source.present? || solution.git_exercise.source_url.present?
```

It feels more like I'm saying "I want to do this method if there's content to do it with". The same applies for these guards in general. However, I'm happy for you to make the decision as it's your code - there's a not a standard that needs enforcing here.

app/controllers/api/files_controller.rb
iHiD: Yeah - I like this.

app/models/git/exercise.rb
iHiD:
```suggestion
      config[:authors].to_a
```

app/models/git/exercise.rb
```diff
@@ -52,12 +62,22 @@ def example
...
     end
 
     memoize
     def contributors
-      config[:contributors]
+      config[:contributors] || []
```
iHiD:
```suggestion
      config[:contributors].to_a
```

app/models/git/exercise.rb
iHiD: No, I think that's correct.

I'd probably rather guard, than daisy chain logic though:

e.g.

```ruby
filtered_filepaths = filepaths.select do |filepath|
  next if filepath.match?(ignore_regexp)
  next if filepath.start_with?('.docs/')
  next if filepath.start_with?('.meta/')

  true
end
```

---

What did we decide about the ignore regexp? Are we keeping it? I can't remember. I think we said it's only used for example files, so we don't need that if they all move to `.meta`. But I'm not sure.

app/models/git/track.rb
```diff
@@ -36,6 +36,21 @@ def snippet
...
+
+    memoize
+    def help
+      repo.read_text_blob(commit, help_filepath)
+    end
+
+    memoize
+    def tests
+      repo.read_text_blob(commit, tests_filepath)
+    end
```
iHiD: You can metaprogram all this. Something like this:

```ruby
%i[about snippet debug help tests].each do |meth|
  define_method meth do
    iv = "@__#{meth}__"
    return instance_variable_get(iv) if instance_variable_defined?(iv)
    instance_variable_set(iv, repo.read_text_blob(commit, meth)
  end
end
```

Or you could go one step funkier and DSL it:
```ruby
# Move this into a git helper module that you include at the top
def self.git_text_blob_reader(*meths)
  meths.each do |meth|
    define_method meth do
      iv = "@__#{meth}__"
      return instance_variable_get(iv) if instance_variable_defined?(iv)
      instance_variable_set(iv, repo.read_text_blob(commit, send("{meth}_filepath"))
    end
  end
end

# Then have this in track
git_text_blob_reader :about, :snippet, :debug, :help, :tests
```

app/models/git/track.rb
```diff
@@ -49,6 +64,18 @@ def snippet_filepath
...
+
+    def help_filepath
+      "exercises/shared/.docs/help.md"
+    end
+
+    def tests_filepath
+      "exercises/shared/.docs/tests.md"
+    end
+
```
iHiD: I'd probably use some metaprogramming here too.

I'd probably define a FILEPATHS constant with hash of `{help: "exercises/shared/.docs/help.md", ...}` and then generate these methods from that.

That keeps all the "data" in one structure and makes it easier/clearer to see/update

test/commands/git/sync_concept_exercise_test.rb
```diff
@@ -11,7 +11,7 @@ class Git::SyncConceptExerciseTest < ActiveSupport::TestCase
   end
 
   test "git SHA does not change when there are no changes" do
-    head_sha = "73b8e389daac4a3e08747e6c16fd420072386e2b"
+    head_sha = "7a42b345ebfa4e74899174abb4ed6b9301cf93b9"
```
iHiD: Can we look this up from git, rather than having to change it each time?

test/commands/markdown/parse_test.rb
```diff
@@ -106,4 +103,12 @@ class Markdown::ParseTest < ActiveSupport::TestCase
...
     assert_equal "<p><del>Hello</del></p>\n", Markdown::Parse.("~~Hello~~")
   end
+
+  test "removes level one headings by default" do
+    assert_equal "<p>Content</p>\n", Markdown::Parse.("# Top heading\n\nContent")
+  end
+
+  test "can keep level one headings" do
+    assert_equal "<h1>Top heading</h1>\n<p>Content</p>\n", Markdown::Parse.("# Top heading\n\nContent", strip_h1: false) # rubocop:disable Naming/VariableNumber
+  end
```
iHiD, line 118: Nice. So if you add an extra test here to guard against when `# ...` is used in code, then you can TDD the solution.

test/commands/markdown/parse_test.rb
```diff
@@ -106,4 +103,12 @@ class Markdown::ParseTest < ActiveSupport::TestCase
   test "parses double tildes as strikethrough" do
     assert_equal "<p><del>Hello</del></p>\n", Markdown::Parse.("~~Hello~~")
   end
+
+  test "removes level one headings by default" do
+    assert_equal "<p>Content</p>\n", Markdown::Parse.("# Top heading\n\nContent")
```
iHiD: I see it's already stripping the newlines. Nice :)

test/controllers/api/files_controller_test.rb
iHiD: >  It doesn't fail where I think it should be.

Where is it failing and where do you think it should?

I can't see what's wrong with it, but I may well be just missing something.

test/models/solution_test.rb
iHiD: Yeah - I'm ok with this for now. My preference would be to probably duplicate the relevant tests to each class, just so we have command-level test-coverage in case we ever access them from anywhere else, or add exceptions for something etc.

app/models/git/exercise.rb
ErikSchierboom: We decided we'd remove it. I wanted to do that later though, or would you like me to remove it in this PR?

app/commands/markdown/parse.rb
ErikSchierboom: Yes, it does: https://github.com/exercism/website/pull/484/files#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5cR97

test/commands/git/sync_concept_exercise_test.rb
ErikSchierboom: Yeah we can. Fixed

app/commands/solution/generate_help_file.rb
ErikSchierboom: Just this specific example or also the others?

test/controllers/api/files_controller_test.rb
ErikSchierboom: What does `assert content, response.body` do? If I change the content to any random string, it still passes the test. I assume that we want to do something like `assert_includes response.body, content`. I've updated the tests.

test/models/solution_test.rb
ErikSchierboom: Done

app/models/git/exercise.rb
iHiD: No, we can take that as a second step. Maybe add a `# TODO` in this Pr for it to remind us?

test/controllers/api/files_controller_test.rb
iHiD: Ah. I see. Yes. Or `assert_equal` probably/maybe.

app/commands/solution/generate_readme_file.rb
ErikSchierboom: Only the hints text, or also the other strings in this file?

test/models/solution_test.rb
iHiD: Nice. Thanks :)

app/commands/markdown/parse.rb
iHiD: Great :)

app/commands/solution/generate_help_file.rb
iHiD: The others too.

test/commands/git/sync_concept_exercise_test.rb
iHiD: Great. Thanks :)

app/commands/solution/generate_readme_file.rb
iHiD: All the strings in this file.

iHiD: Mainly from a seperating data/code POV, rather than me thinking about i18n specifically at this point.

app/models/git/exercise.rb
ErikSchierboom: Done

app/models/git/has_git_filepaths.rb
```diff
@@ -0,0 +1,24 @@
+module Git
+  module HasGitFilepaths
+    def git_filepaths(**fields_to_filepaths)
+      fields_to_filepaths = Hash(fields_to_filepaths)
```
iHiD: I don't think we don't need this line.

app/models/git/has_git_filepaths.rb
```diff
@@ -0,0 +1,24 @@
...
+      fields_to_filepaths.each do |field, filepath|
+        filepath.end_with?('.json') ? read_method = "read_json_blob" : read_method = "read_text_blob"
```
iHiD, line 9: NICE.

app/models/git/track.rb
```diff
@@ -64,13 +42,6 @@ def commit
...
 
-    FILEPATHS = {
-      about: "docs/ABOUT.md",
-      snippet: "docs/SNIPPET.txt",
-      debug: "exercises/shared/.docs/debug.md",
-      help: "exercises/shared/.docs/help.md",
-      tests: "exercises/shared/.docs/tests.md",
-      config: "config.json"
-    }.freeze
+    delegate :read_text_blob, :read_json_blob, to: :repo
```
iHiD: Can this be moved into the `Git::HasGitFilepaths` too I wonder?

app/models/git/track.rb
```diff
@@ -26,36 +34,6 @@ def key_features
...
-    memoize
-    def tests
-      repo.read_text_blob(commit, FILEPATHS[:tests])
-    end
-
-    memoize
-    def config
-      repo.read_json_blob(commit, FILEPATHS[:config])
-    end
-
```
iHiD, line 55: This makes me very happy :)

app/models/git/has_git_filepaths.rb
```diff
@@ -0,0 +1,24 @@
...
+        end
+
+        # Define a <field>_filepath method to allow easy access to the filepath
+        define_method "#{field}_filepath" do
+          filepath
+        end
+      end
+    end
+  end
+end
```
iHiD, line 27: This is basically what the insides of Rails looks like btw :)

ErikSchierboom: :D

app/models/git/track.rb
ErikSchierboom: Good!

app/models/git/track.rb
ErikSchierboom: Probably not, for reasons that will become apparent when I will be adding this to the other two Git classes :)

app/models/git/has_git_filepaths.rb
ErikSchierboom: Yes, although I find the duplicate assignment really, really weird :)

app/models/git/has_git_filepaths.rb
ErikSchierboom: We probably don't. I just copy-pasted it.

app/commands/solution/generate_readme_file.rb
SleeplessByte: `compact_blank` _is_ new.

SleeplessByte: (https://github.com/rails/rails/blob/v6.1.0/activesupport/CHANGELOG.md, since 6.1 ActiveSupport)

iHiD: Modern developers, slowly pulling me away from Rails 3.2 one step at a time!

ErikSchierboom: Lol

app/commands/solution/generate_readme_file.rb
ErikSchierboom: I've added some basic localization. Let me know what you think

app/models/git/has_git_filepaths.rb
```diff
@@ -10,7 +14,7 @@ def git_filepaths(**fields_to_filepaths)
           iv = "@__#{field}__"
           return instance_variable_get(iv) if instance_variable_defined?(iv)
 
-          instance_variable_set(iv, send(read_method, commit, filepath))
+          instance_variable_set(iv, repo.send(read_method, commit, send("full_filepath", filepath)))
```
iHiD:
```suggestion
          instance_variable_set(iv, repo.send(read_method, commit, full_filepath(filepath)))
```

app/models/git/track.rb
```diff
@@ -42,6 +42,8 @@ def commit
     private
     attr_reader :repo, :git_sha
 
-    delegate :read_text_blob, :read_json_blob, to: :repo
+    def full_filepath(filepath)
```
iHiD: I suggest this should be called `expand_filepath`

ErikSchierboom: I just renamed it to `absolute_filepath`, which was also how it was described somewhere else in the source code.

app/models/concept_exercise.rb
```diff
@@ -9,8 +9,6 @@ class ConceptExercise < Exercise
     through: :exercise_taught_concepts,
     source: :concept
 
-  delegate :instructions, to: :git
```
ErikSchierboom, line 12: We're using this for both concept and practice exercises, so I've moved it to the base Exercise class.

.github/workflows/rubocop.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       - name: Install gems
         run: |
-          gem install rubocop -v 0.91.0
+          gem install rubocop -v 1.7.0
```
ErikSchierboom, line 26: This syncs the rubocop version used in the GHA workflow with the one defined in the Gemfile.lock file

app/commands/git/sync_concept_exercise.rb
```diff
@@ -79,12 +79,12 @@ def find_concepts(slugs)
 
     memoize
     def author_usernames_config
-      head_git_exercise.authors.to_a.map { |a| a["exercism_username"] }
+      head_git_exercise.authors.to_a.map { |a| a[:exercism_username] }
```
iHiD, line 82: Rather than doing this, can we make authors a hash with indifferent access pls

app/helpers/react_components/editor.rb
```diff
@@ -34,5 +34,11 @@ def introduction
     def instructions
       Markdown::Parse.(solution.instructions)
     end
+
+    def example_files
```
iHiD, line 47: Can you add a TODO to remove this pls once we don't want this any more.

app/javascript/components/Editor.tsx
```diff
@@ -72,6 +72,7 @@ export function Editor({
   instructions,
   exampleFiles,
   storageKey,
+  debugging,
```
iHiD: This might be more clearly named as `debuggingInstructions`

test/integration/exercises/show_external_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    get "/tracks/ruby/exercises/bob/"
+    assert_select ".introduction", text: /Introduction for bob/
+  end
+
+  test "the introduction is not shown for a practice exercise without an introduction" do
+    create :practice_exercise, slug: 'allergies'
+    get "/tracks/ruby/exercises/allergies/"
+    assert_select ".introduction", false
+  end
+end
```
iHiD, line 45: Nice tests :)

test/commands/solution/generate_readme_file_test.rb
```diff
@@ -6,7 +6,8 @@ class Solution::GenerateReadmeFileTest < ActiveSupport::TestCase
...
...nt log levels:\n\n- `INFO`\n- `WARNING`\n- `ERROR`\n\nYou have three tasks, each of which will take a log line and ask you to do something with it.\n\n## 1\\. Get message from a log line\n\nImplement the `LogLineParser.message` method to return a log line's message:\n\n\n```ruby\nLogLineParser.message('[ERROR]: Invalid operation')\n// Returns: \"Invalid operation\"\n```\n\n\nAny leading or trailing white space should be removed:\n\n\n```ruby\nLogLineParser.message('[WARNING]:  Disk almost full\\r\\n')\n// Returns: \"Disk almost full\"\n```\n\n\n## 2\\. Get log level from a log line\n\nImplement the `LogLineParser.log_level` method to return a log line's log level, which should be returned in lowercase:\n\n\n```ruby\nLogLineParser.log_level('[ERROR]: Invalid operation')\n// Returns: \"error\"\n```\n\n\n## 3\\. Reformat a log line\n\nImplement the `LogLineParser.reformat` method that reformats the log line, putting the message first and the log level after it in parentheses:\n\n\n```ruby\nLogLineParser.reformat('[INFO]: Operation completed')\n// Returns: \"Operation completed (info)\"\n```\n\n## Source\n\n### Created by\n\n- pvcarrera (@pvcarrera)" # rubocop:disable Layout/LineLength
```
iHiD: Does the `1\\`. thing matter? Gonna look weird, no?

app/commands/markdown/preprocess.rb
```diff
@@ -1,9 +1,64 @@
...
+      next unless node.type == :header && node.header_level == 1
+
+      mutations << { type: :delete_header, node: node }
+    end
+  end
+
+  def lower_heading_levels!
+    # TODO: We almost certainly don't want to do this!
+    # but for now let's reduce the heading level of all
+    # headings, as they're too high in the actual docs atm
```
iHiD: I think you can safely remove this comment now :)

app/commands/markdown/preprocess.rb
```diff
@@ -1,9 +1,64 @@
...
+        lines[mutation[:node].sourcepos[:start_line] - 1].insert(mutation[:node].sourcepos[:start_column] - 1, '#')
+        mutation[:node].header_level = mutation[:node].header_level + lower_heading_levels_by
+      end
+    end
+
+    @text = lines.join("\n")
+  end
+
+  def mutations_in_reverse_line_order
+    mutations.sort_by { |mutation| -mutation[:node].sourcepos[:start_line] }
```
iHiD: I have a slight preference over adding `.reverse` to the end rather than using `-` as I feel the minus is slighty easier to miss and harder to read.

```suggestion
    mutations.sort_by { |mutation| mutation[:node].sourcepos[:start_line] }.reverse
```

app/commands/markdown/render_markdown.rb
```diff
@@ -0,0 +1,18 @@
+class Markdown::RenderMarkdown
```
iHiD: I wonder if it's worth merging this with `RenderDoc` and having the second argument passed as either `:doc` or `:text` instead? Unless you imagine them diverging later?

app/commands/markdown/preprocess.rb
```diff
@@ -1,9 +1,64 @@
...
+  def apply_mutations!
+    @lines = text.lines(chomp: true)
+
+    # As we'll possibly be deleting lines too, we have to process the
+    # mutations in reverse order
+    mutations_in_reverse_line_order.each do |mutation|
+      case mutation[:type]
+      when :delete_header
+        lines.delete_at(mutation[:node].sourcepos[:start_line] - 1)
+        mutation[:node].delete
```
iHiD, line 48: Why is this needed?

app/models/track.rb
```diff
@@ -15,6 +15,7 @@ class Track < ApplicationRecord
     to: :git
 
   delegate :head_sha, to: :git, prefix: :git
+  delegate :debug, to: :git
```
iHiD: I feel this should be called `debugging_instructions` throughout.

app/serializers/serialize_exercise_instructions.rb
```diff
@@ -13,8 +13,7 @@ def call
 
   private
   def overview
-    # TODO: remove once we figured out how to handle practice exercises
-    return '' if exercise.practice_exercise?
+    return instructions if exercise.practice_exercise?
```
iHiD, line 18: This isn't correct I don't think, because practice_exercises might have introductions too, which need to be rendered here?

app/serializers/serialize_exercise_instructions.rb
```diff
@@ -75,22 +71,24 @@ def tasks
       end
   end
 
+  def instructions
+    [
+      Markdown::RenderMarkdown.(exercise.git.instructions).strip,
+      Markdown::RenderMarkdown.(exercise.git.instructions_append).strip
+    ].join("\n").strip
```
iHiD:
```suggestion
    ].join("\n\n").strip
```

I'm not sure if this is correct, but we need to decide whether we're adding a full line break or not. I feel like it's safer that we do, in which case we need the double newline I think.

app/serializers/serialize_exercise_instructions.rb
```diff
@@ -75,22 +71,24 @@ def tasks
...
   end
 
   memoize
   def hints_doc
-    render_doc(exercise.git.hints || '')
+    Markdown::RenderDoc.(exercise.git.hints || '')
   end
 
   def parse_title(header)
     header.to_plaintext.strip.gsub(/^(^\d+)\.\s*(.*)/, '\2')
```
iHiD, line 96: This needs a comment adding.

app/commands/markdown/preprocess.rb
ErikSchierboom: Because the doc is returned later on and this is more efficient that re-parsing the doc

app/serializers/serialize_exercise_instructions.rb
ErikSchierboom: Introductions are passed to the editor via a different property

test/commands/solution/generate_readme_file_test.rb
ErikSchierboom: That was some GH caching, because I've fixed that a couple of hours ago :)

app/helpers/react_components/editor.rb
```diff
@@ -10,7 +10,7 @@ def to_s
           track_title: solution.track.title,
           exercise_title: solution.exercise.title,
           introduction: introduction,
-          instructions: SerializeExerciseInstructions.(solution.exercise),
+          assignment: SerializeExerciseAssignment.(solution.exercise),
```
iHiD, line 13: @kntsoriano Worth you seeing this specific commit in this PR. The word `instructions` ended up being too overloaded, as its a specific file to maintainers, but here it also has tests and other things. So we renamed it here to `assignment`.

app/commands/markdown/render_markdown.rb
```diff
@@ -0,0 +1,18 @@
+class Markdown::Render
```
iHiD:
```suggestion
class Markdown::RenderMarkdown
```

ErikSchierboom: This file shouldn't even be here. Whoops

iHiD (approved):

ErikSchierboom:
@kntsoriano @iHiD The tests are a bit flaky. I had to run the tests three times for a valid run. There were two system test failures and one JS failure.

PR #492, opened 2021-02-16
Fix distinction between mentors and students

iHiD:
@kntsoriano This reverts some of the work you did changing Student to Partner and instead splits them out into more separate concepts. 

As I understand it, your work was in trying to keep the distinction between Students and Mentors DRY and avoid duplication, but I think the net result for both me and Erik was that it made it a bit more confusing. The main source of the reason for this is the shared `PartnerInfo` component. I've duplicated this (and they are slightly different too) and added `Student` and `Mentor` as similar concepts. I've also added a `userIsStudent` flag that we can use rather than having to compare students and mentors to each other, which I think was one reason to unify them.

I think its really hard for you to know what is intended to be the same and what is different so these decisions can be quite hard to make. With my wider big-picture hat, I think this is a better refactor. As this is a change to your work and you may disagree, I've PR'd it separately so we can discuss if if you want to first, but feel free just to merge if you're happy! 🙂 

You may need to fix the tests up. I still can't run them locally 🙁

### DIFF
--- a/app/css/components/mentor/discussion.css
+++ b/app/css/components/mentor/discussion.css
@@ -250,7 +250,8 @@
             overflow-y: scroll;
 
             @apply pb-64;
-            & .student-info {
+            & .student-info,
+            & .mentor-info {
                 @apply bg-backgroundColorA;
                 @apply flex;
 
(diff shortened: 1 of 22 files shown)


ErikSchierboom (approved):
Great!

PR #493, opened 2021-02-17
Refactor API with better namespaces and student mentoring endpoints

iHiD:
This does two things:
1. It groups API endpoints into three categories:
  a. `/solutions/...`
  b. `/mentoring/...`
  c. Everything else (e.g. profiles, notifications, etc) at the apex.

2. It  starts to add the mentoring endpoints for the student side of mentoring via the `solutions/.../discussions/.../posts` endpoints. At the moment there is some undesirable duplication in them, but this can get refactored out further down the line once the divisions are clear.

---

This doesn't really need reviewing if the tests pass.


ErikSchierboom:
> This doesn't really need reviewing if the tests pass.

Unfortunately, they don't :(

PR #507, opened 2021-02-21
Hide pagination when totalPages <= 0

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/124.

This should've been a change in just one place, but our other components haven't used our <SearchableList /> component yet. I will do this refactoring another time.

### DIFF
--- a/app/javascript/components/common/SearchableList.tsx
+++ b/app/javascript/components/common/SearchableList.tsx
@@ -131,7 +131,7 @@ const Results = ({
           setOrder={setOrder}
         />
       ) : null}
-      {latestData ? (
+      {latestData && latestData.meta.totalPages > 1 ? (
         <Pagination
           current={query.page}
           total={latestData.meta.totalPages}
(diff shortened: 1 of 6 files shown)


app/javascript/components/common/SearchableList.tsx
```diff
@@ -131,7 +131,7 @@ const Results = ({
           setOrder={setOrder}
         />
       ) : null}
-      {latestData ? (
+      {latestData && latestData.meta.totalPages > 0 ? (
```
iHiD: Should this not be `> 1`? If there's only one result, totalPages is still `1` I think?

Also, can we add a test for this pls.

ErikSchierboom: I think you're right :)

iHiD (changes requested):
See my inline comment :)

app/javascript/components/common/SearchableList.tsx
kntsoriano: Oh yes!

kntsoriano:
@iHiD I'm opening a new PR based on the new `main` branch.

PR #510, opened 2021-02-21
Don't allow submitting changed solutions

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/109.

### DIFF
--- a/app/javascript/components/Editor.tsx
+++ b/app/javascript/components/Editor.tsx
@@ -424,7 +424,10 @@ export function Editor({
           <RunTestsButton onClick={runTests} />
           <SubmitButton
             onClick={submit}
-            disabled={submission?.testRun?.status !== TestRunStatus.PASS}
+            disabled={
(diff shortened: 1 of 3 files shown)


app/javascript/components/Editor.tsx
```diff
@@ -424,7 +424,10 @@ export function Editor({
           <RunTestsButton onClick={runTests} />
           <SubmitButton
             onClick={submit}
-            disabled={submission?.testRun?.status !== TestRunStatus.PASS}
+            disabled={
+              submission?.testRun?.status !== TestRunStatus.PASS ||
+              !isEqual(submissionFilesRef.current, files)
+            }
```
ErikSchierboom, line 430: Should we add a test to verify this behavior?

ErikSchierboom (approved):

app/javascript/components/Editor.tsx
kntsoriano: Thanks. I've added the test.

PR #513, opened 2021-02-22
Fix jumpy lists

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/129.

## Description
In order to make the lists less jumpy, we need to do two things:

1. Use a paginated query instead of a normal query. The paginated query stores the `resolvedData` while it fetches in the background. When the request resolves, `resolvedData` is repopulated.
2. Instead of hiding the pagination footer while fetching, disable it instead.

### DIFF
--- a/app/javascript/components/common/SearchableList.tsx
+++ b/app/javascript/components/common/SearchableList.tsx
@@ -50,10 +50,17 @@ export const SearchableList = ({
   const { request, setPage, setCriteria, setQuery, setOrder } = useList({
     endpoint: endpoint,
   })
-  const { status, resolvedData, latestData, error } = usePaginatedRequestQuery<
-    PaginatedResult,
-    Error | Response
-  >(cacheKey, request, isMountedRef)
+  const {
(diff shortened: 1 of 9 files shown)


app/javascript/components/common/SearchableList.tsx
```diff
@@ -73,6 +80,7 @@ export const SearchableList = ({
           value={request.query.criteria || ''}
           placeholder={placeholder}
         />
+        {isFetching ? <span>Fetching</span> : null}
```
kntsoriano, line 83: @iHiD This is where we can put the spinner. I just stubbed it in for now.

iHiD: OK. Thanks.

iHiD:
Yeah, this is much better. Thanks! Tests are failing here too, but again I think it's due to flakey ones.

iHiD (approved):

ErikSchierboom (approved):

PR #528, opened 2021-02-25
Update mandate improve mocha guards

iHiD:


Gemfile
```diff
@@ -24,7 +24,7 @@ gem 'rest-client' # http
 gem 'octokit' # GitHub
 
 # General-purpose gems
-gem 'mandate', '0.4.0.beta.1'
+gem 'mandate', '1.0.0.beta1'
```
iHiD, line 27: This fixes a bug where memoizing a private method made it public.

app/commands/git/sync_concept.rb
```diff
@@ -9,13 +9,13 @@ def initialize(concept)
     end
 
     def call
-      return concept.update!(synced_to_git_sha: head_git_concept.commit.oid) unless concept_needs_updating?
+      return concept.update!(synced_to_git_sha: head_git_concept.normalised_git_sha) unless concept_needs_updating?
```
iHiD: `commit` is a private method that doesn't need to be accessed.

iHiD:
@ErikSchierboom Updated to sync'd git sha.

---

@kntsoriano Worth you taking a look at this PR just to be aware of the extra guard :)

PR #532, opened 2021-02-26
Fix JS async issue

iHiD:
@kntsoriano This test is a problematic one, so I've spent some time taking a look at it.

There's a couple of problems.
1. Sometimes the onYes function gets called (which you'd expect as the "yes" button is being clicked). The function isn't passed in, so it then blows up.
2. Other times the button is found before it's been disabled.

Basically I think there are a lot of async race conditions that mean a different env (CI) will have differnet outcomes to locally. Specifically the fact that we're relying on the time when the buttons are checked to be:
- After React has updated them to be disabled
- Before the server has responded.

It's clear from (1) above that sometimes the server **does** respond first. 

This therefore tries to fix those things by adding a `onYes` function and by using `waitFor` to check the button. ie, it tries to make it less racey, and deal with things if they happen out of order. In the existing test the `await` just awaits getting the element, which is already in the DOM (ie it doesn't do any real waiting). This new version using `waitFor` should keep replaying the block until the button has been disabled.

It's hard to know if this fixes it, as I only got sporadic failures of it locally, but I think it's saner.

Thanks to @SleeplessByte for explaining some of the `async`/`waitFor` logic to me. But this is my code, so if it sucks, it reflects on my lack of JS, not his 🙂 

---

Note, if this does work, you probably need to spend some time updating other tests that use a server to have similar behaviour as the sporadic failures are probably also caused by the same things.

### DIFF
--- a/test/javascript/setupTests.js
+++ b/test/javascript/setupTests.js
@@ -1,0 +1,15 @@
+import '@testing-library/jest-dom/extend-expect'
+import { waitFor } from '@testing-library/react'
+import { queryCache } from 'react-query'
+
+afterEach(async () => {
+  queryCache.cancelQueries()
+  queryCache.clear()
+})
+
+afterEach(async () => {
+  // waitFor is important here. If there are queries that are being fetched at
+  // the end of the test and we continue on to the next test before waiting for
+  // them to finalize, the tests can impact each other in strange ways.
+  await waitFor(() => expect(queryCache.isFetching).toBe(0))
+})
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -223,6 +223,18 @@ jobs:
           bundle exec setup_exercism_local_aws
 
       ###
+      # Precompile JS
+      - name: Precompile JS
+        env:
+          EXERCISM_ENV: test
+          EXERCISM_CI: true
+          AWS_PORT: ${{ job.services.aws.ports['4566'] }}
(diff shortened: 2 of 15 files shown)


iHiD:
(also cc @ErikSchierboom)

iHiD:
Update - I fixed another test with the same thing.

I also removed `toBeInDocument()` from the end of `expect(await screen.findByText('Unable to end discussion'))`. It's a pointless exception because the expect only resolves when the test is found. So I think we can remove that in most places too?

kntsoriano:
@iHiD Yes! That's what I've also discovered while fixing the tests yesterday. I will add commits to this PR to fix the other test failures.

kntsoriano (approved):

test/javascript/setupTests.js
```diff
@@ -0,0 +1,15 @@
...
+
+afterEach(async () => {
+  queryCache.cancelQueries()
+  queryCache.clear()
+})
+
+afterEach(async () => {
+  // waitFor is important here. If there are queries that are being fetched at
+  // the end of the test and we continue on to the next test before waiting for
+  // them to finalize, the tests can impact each other in strange ways.
```
ErikSchierboom, line 13: Ouch! Great catch. Was there are reason for this code to be in a separate `afterEach` block? I sort of feel like it belongs to the code in the other `afterEach` block.

kntsoriano: Yes. You're correct! I've noticed that some of the tests fail because of this. In the other PRs I've opened, I've combined them into one block.

PR #547, opened 2021-03-01
Hide pagination if totalPages <= 1

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/124.


iHiD:
How come this is so much simpler than the other PR?

iHiD (approved):

kntsoriano:
@iHiD We now always show the Pagination component to prevent the page from being jumpy, except in this case when there are no pages to show.

PR #553, opened 2021-03-02
Remove HTML comments from rendered Markdown

ErikSchierboom:


iHiD:
@ErikSchierboom Thanks :) 

Could you add this to the `docs` too please.

I'd like our preferred method to be markdown comments still, but I'm ok with supporting this as there are HTML comments in legacy documents.

(cc @neenjaw as an FYI)

neenjaw:
That is an awesome package name `Loofah` 😂

PR #558, opened 2021-03-02
Add mentoring dropdown

iHiD:
@kntsoriano This adds the mentoring dropdown.

It looks like this:
Screenshot 2021-03-02 at 23 11 57

---

You can see it at http://local.exercism.io:3020/temp/modals/mentoring_dropdown

---

If is the thing that appears when you click on the RHS of the combo button added in https://github.com/exercism/website/pull/557

https://user-images.githubusercontent.com/286476/109695721-244f2c00-7b84-11eb-8f92-83979a55a667.png

---

- I think we should probably preload the data rather than looking it up at click-time. 
- The data should be populated as per the controller. It's the student's discussions for a given exercise. If any are "in progress" the the top bit should change to the alternative state.
- If you want to just add some mock data to the Helper::ReactComponent, I don't mind adding the real data myself if that's easier.
- Also there is some stuff to avoid n+1s. I've linked to some code for that in the HAML
- The discussion row is also used in the main dashboard (https://exercism.lol/dashboard) so you can extract that bit as a common component (I've added a comment for that too)
- I don't think it's really a "menu" like the notifications or dropdown. I'd treat it more like a modal in terms of behaviour, although it's not really a modal as it doesn't block focus on other things.
- Clicking on the combo-button arrow should open it. Clicking the combo-button arrow again should close it. Clicking anywhere else on the page should close it too.

---

You can build on this PR, or merge this PR and do another one on top.


app/javascript/components/common/CopyToClipboardButton.tsx
```diff
@@ -10,10 +10,6 @@ const KEY_NAMES_LEGACY = Object.freeze({
 })
 
 export function CopyToClipboardButton({ textToCopy }: { textToCopy: string }) {
-  if (window.navigator.clipboard === undefined) {
```
ErikSchierboom, line 13: Is this because we will only support modern browsers?

app/controllers/temp/modals_controller.rb
```diff
@@ -45,5 +45,10 @@ def reputation
 
       @tokens = User::ReputationToken::Search.(current_user).map(&:rendering_data)
     end
+
+    def mentoring_dropdown
+      @active_mentoring_discussion = false
+      @discussions = Solution::MentorDiscussion.limit(2)
```
ErikSchierboom, line 51: Do we need ordering here?

ErikSchierboom (approved):

app/controllers/temp/modals_controller.rb
iHiD: This is just a temp controller.

app/javascript/components/common/CopyToClipboardButton.tsx
iHiD: Yes, but also because it doesn't work on localhost without this, which can introduce bugs. If we can guard this so it's there for production, that'd be good. But we need to fallback to something else instead, rather than just leave it blank.

I think the real solution is to render the widget regardless, then have this guard on the functionality that uses it, not on the whole component.

iHiD:
@kntsoriano Merging this - you can implement it on a separate branch.

PR #560, opened 2021-03-03
Sync authors and contributors

ErikSchierboom:
This PR allows (re-)syncing the authors and contributors of _all_ exercises.

The prime use case for this is when an author/contributor previously did not have an account at Exercism, but then later creates on. In that case, the user should retroactively be credited with their contributions.

### DIFF
--- a/lib/tasks/sync_authors_and_contributors.rake
+++ b/lib/tasks/sync_authors_and_contributors.rake
@@ -1,0 +1,4 @@
+desc 'Sync authors and contributors'
+task sync_authors_and_contributors: :environment do
+  Git::SyncAuthorsAndContributors.call
+end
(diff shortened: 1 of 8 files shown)


iHiD (approved):
Looks good!

lib/tasks/sync_authors_and_contributors.rake
```diff
@@ -0,0 +1,4 @@
+desc 'Sync authors and contributors'
+task sync_authors_and_contributors: :environment do
+  Git::SyncAuthorsAndContributors.call
```
ErikSchierboom:
```suggestion
  Git::SyncAuthorsAndContributors.()
```

PR #559, opened 2021-03-03
Add websockets to iteration summary

kntsoriano:
1. In this PR, we now have two version of `<IterationSummary />` one with websockets and one without. In the Rails app, when rendering the `<IterationSummary />` directly, we use the version with websockets.

PR #564, opened 2021-03-03
Add v1 API

iHiD:
This PR does two things:
- It splits the v1 and v2 APIs out
- It adds support for the new GET api/v2/solutions/xxx.

Be aware that all the React endpoints are now on `api/v2`, not `api/v1`. `v1` remains purely for the CLI.
 
@kntsoriano This implements https://github.com/exercism/website/pull/559/files/56ececfe2fa233ff5d6f1e5cb3fd2dd043230c09#r586067118 so you can use this instead when it's merged. The endpoint is `api_solution_path(solution.uuid, sideload: [:latest_iteration])`

The `sideloads` functionality allows us to specify objects to sideload next to things on the API. This allows API methods to remain relatively RESTful, while giving us the option to add extra things in specific situations where we need them in React.

(cc @SleeplessByte to this sideloads thing as I know you have API opinions 🙂 - Feel free to just wave it through though!)


SleeplessByte:
Are you going to keep v1 forever? If not, I'd actually would argue to _not_ have a v2 namespace, and only move v1 somewhere. 

You can then move forward with media-type versioning and *never* need doing this again

SleeplessByte:
Is the sideloads things similar to `fields=[list, of, fields]&includes=[list, of, things, to, embed]` / GraphQL?

iHiD:
> Are you going to keep v1 forever? If not, I'd actually would argue to not have a v2 namespace, and only move v1 somewhere.
> You can then move forward with media-type versioning and never need doing this again

Interesting. It's a `scope` as opposed to a `namespace`, so none of the controllers are named as `v2` and none of the routes include `_v2_` so it's a one-liner to change. Sooooo, at some point before we launch, let's discuss this properly so I understand the media stuff properly.

> Is the sideloads things similar to fields=[list, of, fields]&includes=[list, of, things, to, embed] / GraphQL?

Sort of maybe unsure.

Basically,:
**GET /solution/xyz**

```
{
  solution: {...}
}
```

**GET /solution/xyz?sideload[]=latest_iteration**

```
{
  solution: {...}
  latest_iteration: {...}
}
```

SleeplessByte:
> Interesting. It's a scope as opposed to a namespace, so none of the controllers are named as v2 and none of the routes include _v2_ so it's a one-liner to change. Sooooo, at some point before we launch, let's discuss this properly so I understand the media stuff properly.

I'm _very_ against versioning inside the URL (for anything that's current / v3 related). If we do this right, and I think we do, you'll never need to "rebuild everything", from an API standpoint.

SleeplessByte:
> Sort of maybe unsure.

Yeah. It exists in many apis. As long as you sanitize it, I'm fine with it, no probs. We can _replace_ the sideloads with media types later (which is what they're great at).

PR #565, opened 2021-03-04
Add iteration page

kntsoriano:
https://github.com/exercism/v3-project-management/issues/140

@iHiD This doesn't include the collapsing logic when a new websocket message comes in. Could you explain it further? I'm a bit confused.


app/channels/solution_channel.rb
```diff
@@ -10,6 +10,7 @@ def unsubscribed; end
   def self.broadcast!(solution)
```
kntsoriano, line 10: @iHiD I didn't know where to put the webhook for the iterations. I just put this here for now.

app/controllers/temp/tracks/exercises_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Temp
```
kntsoriano, line 1: This is a temporary controller which responds to when the user hits "Start exercise" on the iteration zero state.

PR #569, opened 2021-03-04
Fix incorrect use of Accordion

iHiD:
@kntsoriano I think you missed https://github.com/exercism/website/pull/555 when you did the Iterations Header.

The iterations shouldn't be an accordion as they should not open/close based on the state of the others. As such the Accordion's "tabs" approach doesn't make sense here. Instead we can use the details/summary block instead, which gives us all the a11y for free.

I spent some time earlier converting the Accordion to details/summary too, but then realised that that was wrong, and the way you've done that with tabs is more correct 🙂

### DIFF
--- a/app/javascript/components/student/IterationPage.tsx
+++ b/app/javascript/components/student/IterationPage.tsx
@@ -75,19 +75,25 @@ export const IterationPage = ({
   }
 
   return (
-    <div className="lg-container.container">
+    <div className="lg-container container">
       <section className="iterations">
-        {resolvedData.iterations.map((iteration, i) => {
-          return (
-            <IterationReport
-              key={i}
-              iteration={iteration}
-              exercise={exercise}
-              track={track}
-              links={links}
-            />
-          )
-        })}
+        {resolvedData.iterations
+          .slice()
+          .sort((it1: Iteration, it2: Iteration) => {
+            return it2.idx > it1.idx ? 1 : -1
+          })
+          .map((iteration, i) => {
+            return (
+              <IterationReport
+                key={i}
+                iteration={iteration}
+                exercise={exercise}
+                track={track}
+                links={links}
+                isOpen={i == 0}
+              />
+            )
+          })}
       </section>
     </div>
   )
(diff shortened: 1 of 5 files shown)


app/javascript/components/student/IterationPage.tsx
```diff
@@ -75,19 +75,25 @@ export const IterationPage = ({
...
-              track={track}
-              links={links}
-            />
-          )
-        })}
+        {resolvedData.iterations
+          .slice()
+          .sort((it1: Iteration, it2: Iteration) => {
+            return it2.idx > it1.idx ? 1 : -1
+          })
```
iHiD, line 84: The API + webhook order the iterations from 1 -> n, whereas we want to order from n -> 1 in this Component.

app/javascript/components/student/IterationPage.tsx
```diff
@@ -75,19 +75,25 @@ export const IterationPage = ({
...
+          .map((iteration, i) => {
+            return (
+              <IterationReport
+                key={i}
+                iteration={iteration}
+                exercise={exercise}
+                track={track}
+                links={links}
+                isOpen={i == 0}
```
iHiD, line 93: I think this should be through of as the "initial state" rather than an ongoing state.

app/views/tracks/iterations/index.html.haml
```diff
@@ -1,4 +1,3 @@
 #page-iterations-index
   = render ViewComponents::Track::ExerciseHeader.new(@track, @exercise, @solution, :iterations)
-  = render ViewComponents::ProminentLink.new("Learn more about solving exercises locally", "#")
```
iHiD, line 3: I think this was left over accidently when I rebased.

app/javascript/components/student/IterationPage.tsx
kntsoriano: Okay. We can change it to `defaultIsOpen`.

kntsoriano:
@iHiD LETS GOOOOO

iHiD:
LOOK AT ME DOING REACT!!

PR #573, opened 2021-03-05
Add mentoring request page

iHiD:
This adds the RHS of the student request flow. It follows the same sort of pattern as the mentor one with a Session style component that changes based on whether there is a request or not, and whether there is a discussion or not.

---

1. Going to http://local.exercism.io:3020/tracks/ruby/exercises/lasagna/mentor_request/new will give you the page a student goes to request mentoring. The LHS should be the normal student mentoring LHS, and the RHS should be like this:

Screenshot 2021-03-05 at 16 06 25

---

2. The API endpoint to request mentoring is `api_solution_mentor_request_path(solution.uuid)`. It expects a `comment` (maybe this should be `comment_markdown`? It create a mentor_request, which you'll need to serlaize.

---

3. Once the student has requested mentoring they get this second holding state, which you can get to at http://local.exercism.io:3020/tracks/ruby/exercises/lasagna/mentor_request/

Screenshot 2021-03-08 at 20 49 20


iHiD:
@kntsoriano This is ready to work on :)

kntsoriano:
@iHiD I think this works end to end now, but I still want to do the following things:
1. Write tests for error and loading states.
2. Rename `MentoringRequest` to `MentorRequest`.

I've also sprinkled some notes on the code, as there are some things we still need to resolve such as where to store the user's comment when it's their first time on the track.

Please also just leave some bugs here and I will get to them tomorrow.

app/javascript/components/student/mentoring-session/mentoring-request/MentoringRequestForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+              Start a mentoring discussion on <strong>{exercise.title}</strong>{' '}
+              to discover new and exciting ways to approach it. Expand and
+              deepen your knowledge.
+            </p>
+          </div>
+          <GraphicalIcon icon="graphic-mentoring-header" />
+        </div>
+        {isFirstTimeOnTrack ? (
+          <div className="question">
+            {/*TODO: @iHiD Instead of an h3, I made this a label */}
```
iHiD: Why? For testing? I think from an a11y POV it's better as a heading.

app/javascript/components/student/mentoring-session/mentoring-request/MentoringRequestInfo.tsx
```diff
@@ -0,0 +1,94 @@
...
+  )
+}
+
+const Video = ({ title, date, url }: VideoProps) => {
+  return (
+    <a href={url} className="video">
+      <div className="img" />
+      <div className="info">
+        <div className="title">{title}</div>
+        <div className="date">{timeFormat(date, 'DD MMM YYYY')}</div>
```
iHiD: This should be a `time` tag.

iHiD:
This is looking good man!

app/javascript/components/student/mentoring-session/mentoring-request/MentoringRequestForm.tsx
kntsoriano: You should be able to click the text and move focus on the textarea, since it's a form.

kntsoriano:
@iHiD Should be good. I've held off renaming types for now. Let's revisit once we agree on names.

app/javascript/components/student/mentoring-session/mentoring-request/MentoringRequestForm.tsx
iHiD: Ok. I understand.

I'm going to go with h3 though. I think it's more correct from an a11y POV as the text above isn't really a label in my eyes, its more of a description. I've added aria-labelledby and aria-describedby tabs to the textareas to make it clear that they're linked.

iHiD: Urgh. I've gone back to labels as this breaks the tests, and gone with described by to make it clear the paragraph is linked to the input (as labels are normally supposed to go directly next to inputs)

PR #574, opened 2021-03-05
Add iteration observation

iHiD:
@kntsoriano This code works to keep a tally of how many things are open.


app/javascript/components/student/IterationPage.tsx
```diff
@@ -74,6 +82,10 @@ export const IterationPage = ({
     return <EmptyIterations links={links} />
   }
 
+  useEffect(() => {
+    console.log(numIterationsExpanded)
+  }, [numIterationsExpanded])
+
```
iHiD: This is just to show it's working.

app/javascript/components/student/IterationPage.tsx
```diff
@@ -66,6 +67,13 @@ export const IterationPage = ({
     }
   }, [CACHE_KEY, solutionId])
 
+  /* Only run this the first time that the component loads */
```
ErikSchierboom: Isn't this the definition of what `useEffect` does @kntsoriano ? If so, does the comment add much value?

ErikSchierboom (approved):

app/javascript/components/student/IterationPage.tsx
iHiD: (This is my code)

I wanted to make it clear that we're only wanting to set the initial number of iterations in this starting state, rather than whenever resolvedData updates. But if this is the "normal" usage of `useEffect`, then we should get rid of the comment :)

ErikSchierboom: I'm not entirely sure, which is why I asked @kntsoriano (I knew it was your code :))

iHiD:
@kntsoriano Don't forget this needs finishing.

iHiD:
@kntsoriano Your changes have broken the initial clicking on the things: https://www.loom.com/share/62c12250940947beb4a4478959069926

kntsoriano:
Weird. Adding `preventDefault` fixes it.

iHiD:
@kntsoriano The JS tests now fail.

iHiD:
@kntsoriano Bump :)

iHiD:
@kntsoriano I added a system test to replace the JS one you removed

PR #585, opened 2021-03-10
Refactor to use consistent types across sessions

iHiD:
@kntsoriano This is a PR to the `mentoring-request-flow` branch.

I've tried to DRY things up a little at both the Serialization and React level. I've added serializers in the form of `SerializeMentorSessionXYZ` and React Types in the Form of `MentorSessionXYZ`, e.g.

`SerializeMentorSessionDiscussion` gives you the matching React Types of `MentorSessionDiscussion`


app/javascript/components/student/mentoring-session/mentoring-request/MentoringRequestForm.tsx
```diff
@@ -63,6 +64,15 @@ export const MentoringRequestForm = ({
...
 
+  const handleSubmit = useCallback(
+    (e) => {
+      e.preventDefault()
+
+      mutation()
+    },
+    [mutation]
+  )
+
```
iHiD, line 75: The page was reloading on every submission because we weren't called `e.preventDefault()`. I copied this from another place we use Forms in React.

iHiD:
The system tests fail for me, but I'm not sure if that's because of my changes or not.

app/controllers/tracks/mentor_requests_controller.rb
```diff
@@ -5,6 +5,12 @@ class Tracks::MentorRequestsController < ApplicationController
   def new
     @first_time_on_track = true
     @first_time_mentoring = true
+
+    return redirect_to action: :show if @solution.mentor_requests.pending.exists?
```
iHiD, line 9: Could you add a controller test for this pls?

app/helpers/react_components/mentoring/session.rb
```diff
@@ -8,6 +8,12 @@ def to_s
           "mentoring-session",
           {
             user_id: current_user.id,
+            relationship: SerializeMentorStudentRelationship.(student_mentor_relationship),
+            request: SerializeMentorSessionRequest.(request),
+            discussion: SerializeMentorSessionDiscussion.(discussion, current_user),
+            track: SerializeMentorSessionTrack.(track),
+            exercise: SerializeMentorSessionExercise.(exercise),
```
iHiD, line 15: It's nice having all of these properly serializing now I think. We can properly document them in the API docs too.

For Track and Exercise, we might want to reuse the full serializers for them eventually too. I couldn't make my mind up, so I've created the new "light" ones for now.

app/javascript/components/mentoring/request/MentoringRequestPanel.tsx
```diff
@@ -1,6 +1,10 @@
 import React, { useCallback } from 'react'
 import { SessionProps } from '../Session'
-import { MentorDiscussion, MentoringRequest, Iteration } from '../../types'
+import {
+  MentorSessionDiscussion as Discussion,
+  MentorSessionRequest as Request,
```
iHiD, line 5: If you preferred the `Mentor` prefix, I'd be fine with that too. e.g.
```typescript
  MentorSessionDiscussion as MentorDiscussion,
  MentorSessionRequest as MentorRequest,
```

And/or the `Props` suffix too.
```typescript
  MentorSessionDiscussion as DiscussionProps,
  MentorSessionRequest as RequestProps,
```

But I'd like to start standardising everywhere whatever we choose.

app/javascript/components/student/MentoringSession.tsx
```diff
@@ -50,13 +51,14 @@ export const MentoringSession = ({
   links,
 }: {
   userId: number
-  discussion?: MentorDiscussion
+  discussion?: Discussion
+  mentor?: Mentor
```
iHiD, line 55: I brought `mentor` to be next to `discussion` instead of inside it, so that we can use the same `MentorSessionDiscussion` type for both contexts. The other already had the `student` type next to it.

app/javascript/components/student/MentoringSession.tsx
```diff
@@ -85,6 +83,7 @@ export const MentoringSession = ({
         {discussion ? (
           <DiscussionInfo
             discussion={discussion}
+            mentor={mentor!}
```
iHiD: If we have a discussion, we should always have a `mentor`. You could instead do `discussion && mentor` on L83 if you prefered.

app/javascript/components/mentoring/request/MentoringRequestPanel.tsx
kntsoriano: No preference on this. I can't decide either.

app/javascript/components/student/MentoringSession.tsx
kntsoriano: Sounds good.

app/javascript/components/student/MentoringSession.tsx
kntsoriano: Let's do `discussion && mentor`.

kntsoriano:
Let's goooo! Thanks for this. Feels really great to have everything cleaned up.

iHiD:
Great. Glad you like it! :)

PR #587, opened 2021-03-11
Tidy mentor request dropdown

iHiD:
A few small tweaks including adding a short format to the date helper we use.

@kntsoriano There doesn't seem to be test coverage for the solution-summary in the JS tests. Is that intentional because it's all in system tests, or an oversight?


app/javascript/components/common/MentorDiscussionSummary.tsx
```diff
@@ -30,7 +30,7 @@ export const MentorDiscussionSummary = ({
         />
         {postsCount}
       </div>
-      <time dateTime={createdAt}>{fromNow(createdAt)}</time>
+      <time dateTime={createdAt}>{fromNow(createdAt, true)}</time>
```
ErikSchierboom: I found that the boolean nature of the argument made it hard to understand what its function is. Maybe introduce a separate function: `shortFromNow` or something like that?

iHiD: Done. Thanks.

iHiD:
Merging this in. @kntsoriano Please see my comment in the OP.

kntsoriano:
@iHiD The system tests you wrote covers the tests for that component.

PR #596, opened 2021-03-15
Change how icons are rendered

iHiD:
@ErikSchierboom This needs you to add three commits:
- [ ] Rename all icons in app/javascript/icons/ to have their leading underscore removed and all `_` turned into `-`
- [ ] Rename all icons in app/javascript/graphics/ to have their leading `_graphic_` removed and all `_` turned into `-`
- [ ] Change all icons in both above directories to have their `<symbol>` and `</symbol>` tags changed to `<svg>` and `</svg>` respectively.

At that stage this should start to work.


iHiD:
Nice work so far!

Two more jobs on this that I missed.

1. All the icons should be `.svg`, not `.html`
2. `icons/misc.svg` is actually 24 different icons in one file. Got a magic way of splitting those out into the icons dir as individual things?

iHiD:
@kntsoriano It's worth being aware of this PR. Basically we now render all icons as  tags via webpack. Some of the icon colours are currently "off" but everything else should work.

app/javascript/components/common/Icon.tsx
```diff
@@ -4,20 +4,18 @@ export function Icon({
...
   }
+  const iconFile = require(`../../images/${category || 'icons'}/${icon}.svg`)
 
-  return (
-    <svg className={classNames.join(' ')} role="img">
-      <title>{alt}</title>
-      <use xlinkHref={`#${icon}`} />
-    </svg>
-  )
+  return <img src={iconFile} alt={alt} className={classNames.join(' ')} />
```
iHiD, line 20: @kntsoriano This is the magic that makes this work btw. So you can use `<Icon>` where we did before, but now it renders it from webpack via dynamically requiring it.

(cc @SleeplessByte to this too)

app/javascript/components/common/ExerciseIcon.tsx
```diff
@@ -1,7 +1,23 @@
...
+  }
+
+  return (
+    <img
+      className={classNames.join(' ')}
+      src={iconUrl}
+      alt={title ? `Icon for exercise called ${title}` : undefined}
+      role={title ? undefined : 'presentation'}
+    />
+  )
```
iHiD, line 22: @kntsoriano All exercise and track icons are now serialized as `icon_url` and rendered through these helpers. 

(They're actually currently going via webpack too at the Rails side, but eventually they'll just be hard-assets in that live in a fixed dir on S3, and are stored in website-icons as in v2. That means we don't have to deploy the website or make code changes to allow new exercises to be added)

app/css/application.css
```diff
@@ -81,6 +81,14 @@
     --tabIconColorSelected: var(--c-604FCD);
     --modalBackgroundColor: rgba(251, 252, 254, 0.7);
 
+    /* https://codepen.io/sosuke/pen/Pjoqqp */
+    --notifications-active-filter: invert(76%) sepia(54%) saturate(3406%)
+        hue-rotate(349deg) brightness(92%) contrast(111%);
+    --notifications-inactive-filter: invert(17%) sepia(25%) saturate(368%)
+        hue-rotate(188deg) brightness(93%) contrast(86%);
+    --primary-reputation-filter: invert(88%) sepia(37%) saturate(870%)
+        hue-rotate(355deg) brightness(90%) contrast(96%);
```
iHiD, line 90: See https://exercism-team.slack.com/archives/CARRG4MNC/p1615919428002900 for an understanding of this. 
I'll document this in the README or somewhere properly at some stage. I've reproduced the explanation below:

---

1. Take an SVG. Make it black on transparent.
2. Render the SVG using `` tag in HTML
3. Go to https://codepen.io/sosuke/pen/Pjoqqp, put a color (hex) in and spam "Compute Filters" until the thing at the bottom of the screen says it's close enough (`#E2CB2D` is good to play with)
4. Take the filter and apply it to the img via CSS.

You now have color-changable icons through CSS while getting all the value of the `` tag rather than `<svg>` tag (for context, rendering 100 exercise icons as `<svg>` takes the browser **20 seconds** in comparison with basically instant as ``).

app/css/application.css
```diff
@@ -124,11 +132,6 @@ code {
     @apply leading-code;
 }
 
-[data-react-data] {
-    width: 100%;
-    height: 100%;
-}
```
iHiD, line 130: This started playing up. I've removed it for now, and we can do it on an per-component basis as necessary.

app/css/components/iteration-summary.css
```diff
@@ -2,10 +2,16 @@
...
+    width: 100%;
+    height: 100%;
 }
 
 .c-react-wrapper-track-iteration-summary-slim {
-    min-height: 30px;
+    min-height: 25px;
+    width: stretch;
+    width: 100%;
+    height: 100%;
```
iHiD, line 14: (e.g. the "per-component basis" referred to above)

PR #603, opened 2021-03-16
Update Reputation Rules

ErikSchierboom:


iHiD:
I think we said we'd award 1 point of reputation for merging, because it *is* something that needs doing.

iHiD:
So I think that the merged token should have two `levels` - maybe `janitorial` and `reviewal`

iHiD:
Looks good!

PR #604, opened 2021-03-17
Link code reputation tokens to track when possible

ErikSchierboom:
Closes https://github.com/exercism/v3-project-management/issues/155

### DIFF
--- a/app/models/user/reputation_tokens/code_contribution_token.rb
+++ b/app/models/user/reputation_tokens/code_contribution_token.rb
@@ -5,6 +5,10 @@ class User::ReputationTokens::CodeContributionToken < User::ReputationToken
   levels %i[minor regular major]
   values({ minor: 5, regular: 10, major: 15 })
 
+  before_validation on: :create do
+    self.track = Track.find_by(repo_url: "https://github.com/#{repo}") unless track
+  end
+
   def guard_params
     "PR##{pr_node_id}"
   end
--- a/app/models/user/reputation_tokens/code_contribution_token.rb
+++ b/app/models/user/reputation_tokens/code_contribution_token.rb
@@ -6,7 +6,7 @@ class User::ReputationTokens::CodeContributionToken < User::ReputationToken
   values({ minor: 5, regular: 10, major: 15 })
 
   before_validation on: :create do
-    self.track = Track.find_by(repo_url: "https://github.com/#{repo}") unless track
+    self.track_id = Track.where(repo_url: "https://github.com/#{repo}").pluck(&:id).first unless track
   end
 
   def guard_params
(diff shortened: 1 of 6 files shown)


app/models/user/reputation_tokens/code_contribution_token.rb
```diff
@@ -5,6 +5,10 @@ class User::ReputationTokens::CodeContributionToken < User::ReputationToken
   levels %i[minor regular major]
   values({ minor: 5, regular: 10, major: 15 })
 
+  before_validation on: :create do
+    self.track = Track.find_by(repo_url: "https://github.com/#{repo}") unless track
```
ErikSchierboom: In theory I could also pass the repo URL from the webhook all the way up to this point, but I think the repo url is stable enough not to do that :)

iHiD:
```suggestion
    self.track_id = Track.where(repo_url: "https://github.com/#{repo}").pluck(&:id).first unless track
```

This will save loading a whole model

app/models/user/reputation_tokens/code_merge_token.rb
```diff
@@ -4,6 +4,10 @@ class User::ReputationTokens::CodeMergeToken < User::ReputationToken
   reason :merged_code
   value 2
 
+  before_validation on: :create do
+    self.track = Track.find_by(repo_url: "https://github.com/#{repo}") unless track
```
ErikSchierboom:
```suggestion
    self.track_id = Track.where(repo_url: "https://github.com/#{repo}").pluck(&:id).first unless track
```

app/models/user/reputation_tokens/code_review_token.rb
```diff
@@ -4,6 +4,10 @@ class User::ReputationTokens::CodeReviewToken < User::ReputationToken
   reason :reviewed_code
   value 3
 
+  before_validation on: :create do
+    self.track = Track.find_by(repo_url: "https://github.com/#{repo}") unless track
```
ErikSchierboom:
```suggestion
    self.track_id = Track.where(repo_url: "https://github.com/#{repo}").pluck(&:id).first unless track
```

app/models/user/reputation_tokens/code_review_token.rb
```diff
@@ -5,7 +5,7 @@ class User::ReputationTokens::CodeReviewToken < User::ReputationToken
   value 3
 
   before_validation on: :create do
-    self.track = Track.find_by(repo_url: "https://github.com/#{repo}") unless track
+    self.track_id = Track.where(repo_url: "https://github.com/#{repo}").pluck(&:id).first unless track
```
iHiD: These should be `pluck(:id)` not `pluck(&:id)` - sorry :)

PR #615, opened 2021-03-19
Support inline and widget links

ErikSchierboom:
Closes https://github.com/exercism/v3-project-management/issues/142

- Support inline links
- Support widget links

### DIFF
--- a/app/commands/markdown/render_html.rb
+++ b/app/commands/markdown/render_html.rb
@@ -18,10 +18,17 @@ class Markdown::RenderHTML
     attr_reader :nofollow_links
 
     def link(node)
+      inline_link_match = match_inline_link(node)
+
       out('<a href="', node.url.nil? ? '' : escape_href(node.url), '" target="_blank"')
       out(' title="', escape_html(node.title), '"') if node.title.present?
       out(' rel="nofollow"') if nofollow_links
+      out(" data-react-inline-link=\"#{inline_link_match[:link]}\"") if inline_link_match
       out('>', :children, '</a>')
+    end
+
+    def match_inline_link(node)
+      %r{^(?<url>https://exercism.io)?/tracks/(?<link>[a-zA-Z0-9_-]+/(?<type>concepts|exercises)/[a-zA-Z0-9_-]+)}.match(node.url) # rubocop:disable Layout/LineLength
     end
 
     def code_block(node)
--- a/app/commands/markdown/render_html.rb
+++ b/app/commands/markdown/render_html.rb
@@ -18,29 +18,20 @@ class Markdown::RenderHTML
     attr_reader :nofollow_links
 
     def link(node)
-      widget_link_match = match_widget_link(node)
-      if widget_link_match
-        out("<span data-react-widget=\"#{widget_link_match[:track]}/#{widget_link_match[:type]}s/#{widget_link_match[:subject]}\" class=\"data-react-#{widget_link_match[:type]}-widget\" />") # rubocop:disable Layout/LineLength
-      else
-        inline_link_match = match_inline_link(node)
-        out('<a href="', node.url.nil? ? '' : escape_href(node.url), '" target="_blank"')
-        out(' title="', escape_html(node.title), '"') if node.title.present?
-        out(' rel="nofollow"') if nofollow_links
-        out(" data-react-inline-link=\"#{inline_link_match[:link]}\"") if inline_link_match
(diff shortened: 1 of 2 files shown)


iHiD:
Fun!

app/commands/markdown/render_html.rb
```diff
@@ -21,9 +21,19 @@ def link(node)
...
       out(' rel="nofollow"') if nofollow_links
+      out(link_tooltip_attributes(node))
       out('>', :children, '</a>')
     end
 
+    def link_tooltip_attributes(node)
+      link_match = %r{^(?<url>https://exercism.io)?/tracks/(?<track>[a-zA-Z0-9_-]+)/(?<type>concept|exercise)s/(?<slug>[a-zA-Z0-9_-]+)}.match(node.url) # rubocop:disable Layout/LineLength
+      return unless link_match
+
+      endpoint = Exercism::Routes.send("tooltip_track_#{link_match[:type]}_path", track_id: link_match[:track],
```
iHiD: 1. If we're going to arbitarily send methods like this, we need to ensure that `link_match[:type]` is an expected value, else we open ourselves up to injection. So I'd guard that it's either "exercise" or "concept" above this.
2. I don't **think** you need the `track_id` and `id` keys here. I think you can just pass the values

app/commands/markdown/render_html.rb
```diff
@@ -21,9 +21,19 @@ def link(node)
       out('<a href="', node.url.nil? ? '' : escape_href(node.url), '" target="_blank"')
       out(' title="', escape_html(node.title), '"') if node.title.present?
       out(' rel="nofollow"') if nofollow_links
+      out(link_tooltip_attributes(node))
       out('>', :children, '</a>')
     end
 
+    def link_tooltip_attributes(node)
+      link_match = %r{^(?<url>https://exercism.io)?/tracks/(?<track>[a-zA-Z0-9_-]+)/(?<type>concept|exercise)s/(?<slug>[a-zA-Z0-9_-]+)}.match(node.url) # rubocop:disable Layout/LineLength
```
iHiD: Rather than `[a-zA-Z0-9_-]`, maybe it's safer to use `[^/]`. In case people put unicode in slugs etc (which I believe we don't disallow)

iHiD: (Maybe also add a test with a unicode char in the slug to be sure?)

ErikSchierboom: Good suggestion

app/commands/markdown/render_html.rb
ErikSchierboom: Will do

app/commands/markdown/render_html.rb
```diff
@@ -29,8 +29,7 @@ def link_tooltip_attributes(node)
       link_match = %r{^(?<url>https://exercism.io)?/tracks/(?<track>[a-zA-Z0-9_-]+)/(?<type>concept|exercise)s/(?<slug>[a-zA-Z0-9_-]+)}.match(node.url) # rubocop:disable Layout/LineLength
       return unless link_match
 
-      endpoint = Exercism::Routes.send("tooltip_track_#{link_match[:type]}_path", track_id: link_match[:track],
-                                                                                  id: link_match[:slug])
+      endpoint = Exercism::Routes.send("tooltip_track_#{link_match[:type]}_path", link_match[:track], link_match[:slug])
       " data-tooltip-type=\"#{link_match[:type]}\" data-endpoint=\"#{endpoint}\""
```
iHiD: Also, for this, it's better to use `%{...}` and not escape the `"`

test/commands/markdown/parse_test.rb
```diff
@@ -128,4 +128,51 @@ class Markdown::ParseTest < ActiveSupport::TestCase
   test "removes html comments" do
     assert_equal "\n<p>Regular text</p>\n", Markdown::Parse.("<!-- Comment text -->\n\nRegular text\n", strip_h1: false)
   end
+
+  test "render internal concept link using absolute URL" do
+    expected = "<p><a href=\"https://exercism.io/tracks/ruby/concepts/basics\" target=\"_blank\" data-tooltip-type=\"concept\" data-endpoint=\"/tracks/ruby/concepts/basics/tooltip\">basics</a></p>\n" # rubocop:disable Layout/LineLength
```
iHiD: Use `%{}` here too (etc to below)

PR #623, opened 2021-03-20
Replace "Elixir" with Track Titles

aretoo:
Use variable track title instead of hard-coded "Elixir" for all languages.


iHiD:
Thanks :)

PR #624, opened 2021-03-20
improve responsiveness of concept map

neenjaw:
Several small changes:
- move the invisible border to the L/R of the tool-tip
- removing some transitions that might cause the perception of unresponsiveness
- optimising the discovery of active slugs by indexing the result


iHiD:
Thanks :)

PR #642, opened 2021-03-24
Sync practiced concepts for practice exercises

ErikSchierboom:


app/models/exercise/practiced_concept.rb
```diff
@@ -0,0 +1,6 @@
+class Exercise::PracticedConcept < ApplicationRecord
```
ErikSchierboom, line 1: Let me know what you think of the class name.

app/models/practice_exercise.rb
```diff
@@ -1,6 +1,21 @@
...
+    class_name: "Exercise::PracticedConcept",
+    foreign_key: :exercise_id,
+    inverse_of: :exercise,
+    dependent: :destroy
+
+  has_many :practices,
+    through: :exercise_practiced_concepts,
+    source: :concept
+
+  def self.with_prerequisite(concept)
```
ErikSchierboom, line 12: I added this method, as the existing `that_practice` method was slightly confusing now that we actually have practice concepts.

app/models/practice_exercise.rb
```diff
@@ -1,6 +1,21 @@
 class PracticeExercise < Exercise
-  def self.that_practice(concept)
+  has_many :exercise_practiced_concepts,
+    class_name: "Exercise::PracticedConcept",
+    foreign_key: :exercise_id,
+    inverse_of: :exercise,
+    dependent: :destroy
+
+  has_many :practices,
```
iHiD:
```suggestion
  has_many :practiced_concepts,
```

app/models/exercise/practiced_concept.rb
iHiD: Yep - Like it!

PR #647, opened 2021-03-25
Improve and assign todos

iHiD:


app/models/track.rb
```diff
@@ -41,7 +41,7 @@ def git
     Git::Track.new(synced_to_git_sha, repo_url: repo_url)
   end
 
-  # TODO: Drive from config.json
+  # TODO: Erik: Drive from config.json
```
ErikSchierboom, line 44: I assume that we'll use the `status.concept_exercises` key from the `config.json` for this? Or is there more to it?

ErikSchierboom (approved):

app/models/track.rb
iHiD: Yeah, basically. Although I think we might want to rename the key.

ErikSchierboom: > Although I think we might want to rename the key.

That will go over well, another bulk PR :P

PR #651, opened 2021-03-25
Fix concept tooltip

iHiD:
@kntsoriano This broke in the PR that fixed the blur. Can you add a test to check that the concept tooltip is rendered correctly pls.


iHiD:
Merging this in @kntsoriano - Please follow up with a test :)

PR #653, opened 2021-03-26
Add loading state to exercise tooltip

iHiD:
@kntsoriano I wanted to get this right before I demo later so did this. It works well. 

A few follow-on things:
1. It needs some tests adding
2. I don't know how the FetchingBoundary works or if it's in the right place.
3. We should probably extract the ExerciseWidget into it's own component.
4. We should probably generalise this loading behavoiour across all the tooltips and modals.
5. Ideally the spinner would only show up after 0.2s.


app/helpers/react_components/student/exercise_status_chart.rb
```diff
@@ -12,20 +12,35 @@ def to_s
...
           status = user_track.exercise_status(exercise)
 
           hash[exercise.slug] = status
         end
       end
+
+      def height
+        exercise_per_row = 23.0
+        height_per_row = 33.6
+        bottom_margin = 24
```
ErikSchierboom, line 34: Might not be worth it, but should these be private constants?

app/javascript/components/student/ExerciseStatusChart.tsx
```diff
@@ -92,32 +92,40 @@ const ExerciseStatusDot = ({
       <ReferenceElement
         ref={setButtonElement}
         onMouseEnter={() => setOpen(true)}
-        onMouseLeave={() => setOpen(false)}
+        /*onMouseLeave={() => setOpen(false)}*/
```
ErikSchierboom: Will it ever close without this?

iHiD: GOOD MAN.

app/helpers/react_components/student/exercise_status_chart.rb
iHiD: I decided not as they're exclusive to this method and the two lines below. So I didn't feel they need exposing class-wide. But 🤷

PR #655, opened 2021-03-27
Unify statuses

iHiD:
We previously had a mismatch between various "status" fields.
We now have the following statuses used everywhere on `exercises` and `solutions` (solutions can never be `locked` or `available`):

```ruby
%i[locked available started iterated completed published]
```

- `available` means the exercise is available but there is no `solution` record for it.
- `started` means someone has downloaded or started work in the editor. (there is a `solution` record)
- `iterated` means someone has added an iteration
- We use `unlocked` to group all of the statuses other than `locked`. e.g. both `available` and `published` are `unlocked?`
- We use `in_progress` to mean `started` or `iterated`


iHiD:
(@kntsoriano @ErikSchierboom See this FYI)

PR #663, opened 2021-03-29
Use mentoring status

iHiD:
@kntsoriano This is a PR that's not intended to be merged, but intended to show you what to as part of https://github.com/exercism/v3-project-management/issues/165

We will remove the two fields from the serializer then use the enum throughout instead of having the individual flags.

I think all the changes I've made here are correct (so you may wish to work from this anyway) but I'm presuming you'll want to go through and work on this yourself :)


app/javascript/components/student/SolutionSummary.tsx
```diff
@@ -33,21 +33,21 @@ export type SolutionSummaryRequest = {
...
 }
 
-export type SolutionSummarySolution = {
-  id: string
-  hasMentorDiscussionInProgress: boolean
-  hasMentorRequestPending: boolean
-  completedAt?: string
+export type Track = {
+  title: string
+  medianWaitTime: string
```
ErikSchierboom, line 38: What is the unit in which the wait time is measured? It might be nice to add the unit to the name to reduce confusion.

app/javascript/components/student/solution-summary/MentoringComboButton.tsx
```diff
@@ -49,11 +47,11 @@ export const MentoringComboButton = ({
   return (
     /* TODO: Extract into a common component in the future */
     <div className={classNames.join(' ')}>
-      {hasMentorDiscussionInProgress && links.inProgressDiscussion ? (
+      {mentoringStatus == 'in_progress' && links.inProgressDiscussion ? (
```
ErikSchierboom: A previous file used `===` instead of `==`, which is a good idea I think.

app/javascript/components/student/solution-summary/MentoringComboButton.tsx
```diff
@@ -49,11 +47,11 @@ export const MentoringComboButton = ({
...
         <a href={links.inProgressDiscussion} className="--primary-segment">
           Continue mentoring
         </a>
-      ) : hasMentorRequestPending ? (
+      ) : mentoringStatus == 'requested' ? (
```
ErikSchierboom: Same comment as above.

ErikSchierboom (approved):

app/javascript/components/student/SolutionSummary.tsx
iHiD: That's a good question. I actually think we should format this server-side, so this is actually just a string that should be rendered by JS rather than anything intelligent. So it's probably named ok?

app/javascript/components/student/solution-summary/MentoringComboButton.tsx
iHiD: (Bump @kntsoriano)

iHiD:
(This needs rebasing now @kntsoriano)

app/javascript/components/student/SolutionSummary.tsx
ErikSchierboom: Then it's perfectly fine.

PR #669, opened 2021-03-30
Fix unicode rendering of output

ErikSchierboom:
image

### DIFF
--- a/app/serializers/serialize_submission_test_run.rb
+++ b/app/serializers/serialize_submission_test_run.rb
@@ -37,7 +37,10 @@ class SerializeSubmissionTestRun
   def output
     return if test_run.output.blank?
 
-    Ansi::To::Html.new(test_run.output).to_html
+    # The ansi-to-html library does not support unicode escape sequence
+    # See https://github.com/rburns/ansi-to-html/issues/48
+    sanitized_output = test_run.output.gsub("\e\[K", '')
+    Ansi::To::Html.new(sanitized_output).to_html
   end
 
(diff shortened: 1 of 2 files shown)


app/serializers/serialize_submission_test_run.rb
```diff
@@ -37,7 +37,10 @@ def message
   def output
     return if test_run.output.blank?
 
-    Ansi::To::Html.new(test_run.output).to_html
+    # The ansi-to-html library does not support unicode escape sequence
+    # See https://github.com/rburns/ansi-to-html/issues/48
+    sanitized_output = test_run.output.gsub(/\e\[K/, '')
```
iHiD: Does this need to be a regexp? Could you use a string (and/or `tr`) instead?

test/serializers/serialize_submission_test_run_test.rb
```diff
@@ -107,7 +107,7 @@ class SerializeSubmissionTestRunTest < ActiveSupport::TestCase
       ops_status: 403,
       raw_results: {
         version: version,
-        output: "\e[31mHello\e[0m\e[34mWorld\e[0"
+        output: "\e[31mHello\e[0m\e[K\e[34mWorld\e[0"
```
iHiD, line 110: Nice :)

app/serializers/serialize_submission_test_run.rb
ErikSchierboom: I've used a string now

PR #677, opened 2021-03-31
Community solutions

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/166.

@iHiD The CSS is broken because when we render a react component it is wrapped with a div.


app/commands/solution/search_community_solutions.rb
```diff
@@ -9,18 +9,22 @@ def self.default_per
...
       @exercise = exercise
       @page = page.present? && page.to_i.positive? ? page.to_i : DEFAULT_PAGE # rubocop:disable Style/ConditionalAssignment
       @per = per.present? && per.to_i.positive? ? per.to_i : self.class.default_per # rubocop:disable Style/ConditionalAssignment
+      @criteria = criteria
     end
 
     def call
       @solutions = exercise.solutions.published
+
+      @solutions = solutions.joins(:user).where("users.handle LIKE ?", "%#{criteria}%") if @criteria.present?
```
kntsoriano, line 23: @iHiD Just temporary for the tests to pass. You might want to change this to have better performance.

iHiD:
Added one small fix, but otherwise great. Thanks @kntsoriano! 💙

PR #680, opened 2021-03-31
Remove various todos

ErikSchierboom:


app/models/track.rb
```diff
@@ -63,14 +62,20 @@ def top_10_contributors
       to_a
   end
 
-  # TODO: Erik: Implement
   def num_code_contributors
-    10
+    User::ReputationToken.
+      where(track_id: id, type: User::ReputationTokens::CodeContributionToken).
```
ErikSchierboom, line 67: I didn't know exactly what the criterium for a code contributor was. I just went with the `CodeContributionToken`. One could also argue for including the authors and contributors here.

ErikSchierboom:
I think the rubocop failures are addressed in https://github.com/exercism/website/pull/675/files

app/models/track.rb
iHiD: I think this is fine :)

iHiD:
This was great to back to this evening. Thank you 💙

PR #688, opened 2021-04-01
Apparently empty alt attrs are better than no alt attrs

iHiD:


test/serializers/serialize_mentor_requests_test.rb
```diff
@@ -16,8 +16,8 @@ class SerializeMentorRequestsTest < ActiveSupport::TestCase
         id: request.uuid,
 
         track_title: track.title,
-        track_icon_url: track.icon_url,
         exercise_title: exercise.title,
+        exercise_icon_url: exercise.icon_url,
```
iHiD:
```suggestion
        track_icon_url: track.icon_url,
        exercise_title: exercise.title,
```

SleeplessByte (commented):
They mean something different. 

- Empty alt === `role="presentation"` (decorative)
- No alt === "you interpret it". Some tooling may, for example, use AI to read out what it sees on the image if no alt is present.

PR #691, opened 2021-04-02
Order exercises by position

ErikSchierboom:
Closes https://github.com/exercism/v3-project-management/issues/17


db/migrate/20210402085227_add_position_to_exercises.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    add_column :exercises, :position, :integer, null: true, after: :blurb
+
+    Exercise.find_each.with_index do |exercise, position|
+      exercise.update!(position: position)
+    end 
+
+    change_column_null :exercises, :position, false
+
+    add_index :exercises, [:track_id, :type, :position], name: 'index_track_exercise_position'
```
ErikSchierboom: I've added this index to make returning the sorted exercise very fast. As we have a relatively small number of tracks (70-ish) with a relatively small number of exercises (10-200), the total amount will be something like 70*200=14000. I think the cost of the index is probably negligible and the gains more important.

This is the `EXPLAIN` output for `select * from exercises where track_id = 1 order by type, position`:

id? | select_type? | table? | partitions? | type? | possible_keys? | key? | key_len? | ref? | rows? | filtered? | Extra?
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | SIMPLE | exercises | NULL | ref | index_exercises_on_track_id_and_uuid,index_exercises_on_track_id,index_track_exercise_position | index_track_exercise_position | 8 | const | 167 | 100.00 | Using index condition

ErikSchierboom: Update: the `type` has now been removed.

db/migrate/20210402085227_add_position_to_exercises.rb
```diff
@@ -0,0 +1,13 @@
+class AddPositionToExercises < ActiveRecord::Migration[6.1]
+  def change
+    add_column :exercises, :position, :integer, null: true, after: :blurb
```
iHiD:
```suggestion
    add_column :exercises, :position, :integer, null: true, after: :blurb, index: true
```

db/migrate/20210402085227_add_position_to_exercises.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    Exercise.find_each.with_index do |exercise, position|
+      exercise.update!(position: position)
+    end 
+
+    change_column_null :exercises, :position, false
+
+    add_index :exercises, [:track_id, :position]
```
iHiD:
```suggestion
```

PR #700, opened 2021-04-02
Stop the frinkin sporadic tests!

iHiD:
This does three things:
1. Cleanup unnecessary sleeps
2. Add some extra config that seems useful
3. **Flunks if there are any JS errors**.

I strongly suspect that the sporadically failing tests are actual failures that occur occasionally, rather than problems with the tests. We get the a sporadic failure in both a JS test and a system test for the same bit of code, which suggests something is actually wrong there. To try and avoid these hidden issues, this branch will now flunk on any severe JS errors, which indicate things that might cause sporadic failures. It also prints the errors out.

@kntsoriano Can you fix up any of the errors that appear in CI as a top priority please?


SleeplessByte (approved):

test/application_system_test_case.rb
```diff
@@ -59,11 +59,11 @@ def teardown
...
 
-  driven_by :selenium, using: :chrome, screen_size: [1400, 1400]
-  # driven_by :selenium, using: :headless_chrome do |driver_option|
-  #   # Without this argument, Chrome cannot be started in Docker
-  #   driver_option.add_argument('no-sandbox')
-  # end
+  # driven_by :selenium, using: :chrome, screen_size: [1400, 1400]
+  driven_by :selenium, using: :headless_chrome do |driver_option|
+    # Without this argument, Chrome cannot be started in Docker
+    driver_option.add_argument('no-sandbox')
```
SleeplessByte: Yep. We often selenum with firefox btw. Might be worthwhile too, should you run into more chrome-ish issues.

test/system/components/editor_test.rb
```diff
@@ -145,50 +153,54 @@ class EditorTest < ApplicationSystemTestCase
...
+    #       tests: [{ name: :test_a_name_given, status: :pass, output: "Hello" }]
+    #     }
+
+    #   use_capybara_host do
+    #     sign_in!(user)
+    #     visit test_components_editor_path(solution_id: solution.id)
+
+    #     assert_text "1 test passed"
+    #   end
+    # end
```
iHiD, line 176: Why is this commented out?

iHiD:
Merged. Nice work 🙂  

Opened this for the sporadic failure: https://github.com/exercism/v3-project-management/issues/183

test/system/components/editor_test.rb
iHiD: Moved to https://github.com/exercism/v3-project-management/issues/184

PR #706, opened 2021-04-04
Add track notifications dot to list page

iHiD:
Screenshot 2021-04-04 at 18 50 44


iHiD:
@kntsoriano @ErikSchierboom There's a couple of patterns I introduce here that I'll be using in general, so it's worth giving the PR a read (and firing any questions over if anything's unclear) even though I'm merging it now :)

PR #707, opened 2021-04-04
Add notifications for exercises serialized on solutions

iHiD:
Implements `has_notifications?` efficently on `SerializeSolution` and `SerializeSolutions`


app/javascript/components/types.ts
```diff
@@ -77,7 +77,6 @@ type ExerciseCore = {
   blurb: string
   difficulty: ExerciseDifficulty
   isRecommended: boolean
-  isCompleted: boolean
```
iHiD, line 80: This was legacy and unused. I've also removed it from the tests below.

PR #739, opened 2021-04-09
Add <Profile.TestimonialSection />

kntsoriano:
https://github.com/exercism/v3-project-management/issues/192.

@iHiD I don't know what to use for accessibility here. Is this a tab?


iHiD:
I think for accessibility, we'll probably remove the interactiveness and just have it as a list. Let's come back to this later. For now, I'll merge as this is good :)

PR #754, opened 2021-04-11
Add solutions to profiles

iHiD:


iHiD:
@kntsoriano Nice work. I don't think the tracks data needs to go via a request. I think its fine just pass it into the original react component, so I've added a commit to do that. I'll merge this if it goes green, but feel free to update with any improvements to my code :)

PR #786, opened 2021-04-19
Add timeout before determining label positions

kntsoriano:
https://github.com/exercism/v3-project-management/issues/204

I really can't replicate this bug reliably. This is the best I could come up with to make sure that the labels aren't positioned weirdly.


iHiD:
Yeah. I think this is probably the most sensible idea. The other would be to check whether the chart has rendered. I couldn't replicate it locally too - only in production. I've merged it - let's see what happens :)

kntsoriano:
@iHiD I tried waiting for the chart to render, but it doesn't look good since the only callback available is when the chart animation finishes. I can show you how this works then we can decide whether we want to do this.

PR #787, opened 2021-04-19
Fix bugs when revealing a testimonial

kntsoriano:
https://github.com/exercism/v3-project-management/issues/205


iHiD:
Nice work man :)

PR #788, opened 2021-04-19
Store scroll position for RHS panel in editor

kntsoriano:
@SleeplessByte How do you recommend we debounce the `onScroll` event?

https://github.com/exercism/v3-project-management/issues/207

### DIFF
--- a/app/javascript/components/Editor.tsx
+++ b/app/javascript/components/Editor.tsx
@@ -40,8 +40,9 @@ import { useFileRevert, RevertStatus } from './editor/useFileRevert'
 import { isEqual } from 'lodash'
 import { sendRequest, sendPostRequest, APIError } from '../utils/send-request'
 import { TabContext } from './common/Tab'
+import { StoredScrollPanel } from './editor/StoredScrollPanel'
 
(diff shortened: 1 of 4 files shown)


SleeplessByte:
I can't see the linked issue. Can you copy it in?

iHiD:
@SleeplessByte https://www.loom.com/share/46e28bb5943a47c0ac3ac4af8832848f

iHiD:
One general thing to add to this is that I'm going to add `scroll-behavior: smooth` generally to the CSS so we'll have to get around that for this. ~I'll push that onto main for you so its there.~ _(Was already there)_

SleeplessByte:
The far easier, cross-browser, ánd accessible solution is to not replace the _content_ of the scrollable pane, but have the three panes mounted / exist in DOM (but invisible), and have the tabs change which one is showing.

iHiD:
@kntsoriano I agree with what @SleeplessByte is saying here. Do you want to refactor to do that (three panels, with one having an `active` css class at a time). And I can tweak any CSS that breaks as a result of it?

SleeplessByte:
(It also reduces the JS by a lot 🗡️)

PR #795, opened 2021-04-19
Show unseen tokens first

iHiD:
@kntsoriano This shows unseen tokens first in the dropdown. However, it has some weird side effects as the loom below shows. Can you fix these please? (basically the dropdown contents should never change while its open).

https://www.loom.com/share/35cfcd1adfb24dba964552ae92cb05c8?focus_title=1&muted=1&from_recorder=1

### DIFF
--- a/app/commands/user/reputation_token/search.rb
+++ b/app/commands/user/reputation_token/search.rb
@@ -63,6 +63,8 @@ class User
 
       def sort!
         case order&.to_sym
+        when :unseen_first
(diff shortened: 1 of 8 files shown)


kntsoriano:
@iHiD Is this okay? We only updated the notifications when the user closes the dropdown. I just did the bare minimum fix for this one as I don't know how the reputation dropdown should really behave.

iHiD:
@kntsoriano Yeah - this works well! Will merge. Nice work :)

PR #800, opened 2021-04-20
Various mentoring improvements

iHiD:


iHiD:
@kntsoriano I **think** this is going to go green but I'm merging it anyway so I can go to bed without you having merge conflicts to fight. If it's red I'll fix it in the morning :)

PR #822, opened 2021-04-24
Mentoring sessions modal

kntsoriano:
https://github.com/exercism/v3-project-management/issues/219


iHiD:
@kntsoriano Nice work on this. 

I think we should get https://github.com/exercism/website/pull/610 merged, and then rebase this on top of it. I think if we merge this first, we're going to get horrific merge conflicts on #610.

app/serializers/serialize_mentor_discussion.rb
```diff
@@ -0,0 +1,55 @@
...
+      },
+      track: {
+        title: discussion.track.title,
+        icon_url: discussion.track.icon_url
+      },
+      exercise: {
+        title: discussion.exercise.title,
+        icon_url: discussion.exercise.icon_url
+      },
+      is_finished: context == :student ? discussion.finished_for_student? : discussion.finished_for_mentor?,
```
iHiD: Could you pull all of these out into methods on the serializer pls.

e.g. 
```ruby
def is_finished
  context == :student ? discussion.finished_for_student? : discussion.finished_for_mentor?,
end
```

This one is ok but the other if/else/end ones make reading this all quite hard :)

test/helpers/react_components/mentoring/session_test.rb
```diff
@@ -52,9 +52,10 @@ class SessionTest < ReactComponentTestCase
             avatar_url: student.avatar_url,
             reputation: student.formatted_reputation,
             is_favorite: false,
-            num_previous_sessions: 15,
+            num_previous_sessions: 0,
```
iHiD: Could you add a previous session, so that we can check this works pls?

app/controllers/api/mentoring/previous_discussions_controller.rb
```diff
@@ -0,0 +1,17 @@
+module API
+  class Mentoring::PreviousDiscussionsController < BaseController
+    def index
+      discussions = ::Mentor::Discussion::Retrieve.(
+        current_user,
+        :finished,
+        page: params[:page]
+      )
```
iHiD, line 8: I don't think this is right. This will return all previous discussions, not just those between a mentor and a student?

app/serializers/serialize_mentor_discussion.rb
kntsoriano: Thanks. I've done this :)

test/helpers/react_components/mentoring/session_test.rb
kntsoriano: Done.

app/controllers/api/mentoring/previous_discussions_controller.rb
kntsoriano: Yep! This is just a stub for now. Can you replace it to have the correct functionality? I'm also unsure whether we need a separate controller for this.

kntsoriano:
@iHiD Additionally, do we need this functionality when viewing the discussion as a student as well? If yes, I think I'd like to do that in a separate PR since this is too big right now.

iHiD:
Fixed at https://github.com/exercism/website/pull/837

PR #826, opened 2021-04-24
background-patch

neenjaw:
I think this will fix a visual bug that DJ found.


neenjaw:
forgot to pull before branching

PR #827, opened 2021-04-24
Bump active-background version

neenjaw:
bump active background version


neenjaw:
there was a visual bug which DJ experienced.  I think this will fix it?

neenjaw:
this should probably be a minor contribution :)

iHiD:
We haven't hadn't added those labels here yet so looks like it counts as a normal one ;)

PR #830, opened 2021-04-25
Update Profile TrackSwitcher Icon onChange

neenjaw:
Previous to this patch, the [track switcher on the profile page](https://exercism.lol/profiles/neenjaw) did not update the icon based on the track:

Screenshot from 2021-04-24 18-57-48

With this patch, it updates when the value changes (1 line fix)

Screenshot from 2021-04-24 18-58-04

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
     'plugin:react-hooks/recommended',
     'plugin:@typescript-eslint/eslint-recommended',
     'plugin:@typescript-eslint/recommended',
+    'plugin:testing-library/react',
     'prettier',
     'prettier/@typescript-eslint',
   ],
@@ -21,11 +22,32 @@ module.exports = {
     ecmaVersion: 12,
     sourceType: 'module',
   },
-  plugins: ['jest', 'react', '@typescript-eslint'],
+  plugins: ['jest', 'react', '@typescript-eslint', 'testing-library'],
   rules: {},
   settings: {
     react: {
       version: 'detect',
     },
   },
+  overrides: [
+    {
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        'react/prop-types': 'off',
+      },
+    },
+    {
+      files: ['*.test.ts', '*.test.tsx'],
+      rules: {
+        '@typescript-eslint/no-empty-function': 'off',
+        '@typescript-eslint/no-unused-vars': 'off',
+      },
+    },
+    {
+      files: ['*.test.tsx'],
+      rules: {
+        'jest/expect-expect': 'off', // testing-library's getBy___ queries are in themselves assertions
+      },
+    },
+  ],
 }
--- a/test/javascript/components/profile/contribution-summary/track-switcher.test.tsx
+++ b/test/javascript/components/profile/contribution-summary/track-switcher.test.tsx
@@ -1,0 +1,66 @@
+import * as React from 'react'
+
+import { render, fireEvent, act, screen, within } from '@testing-library/react'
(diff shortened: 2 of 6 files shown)


iHiD:
@neenjaw Thanks! Want to throw a test in for it?

.eslintrc.js
```diff
@@ -21,11 +22,32 @@ module.exports = {
...
+    {
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        'react/prop-types': 'off',
+      },
+    },
+    {
+      files: ['*.test.ts', '*.test.tsx'],
+      rules: {
+        '@typescript-eslint/no-empty-function': 'off',
```
neenjaw, line 42: it is sometimes easy to use an empty function for testing without creating a complex requirement

.eslintrc.js
```diff
@@ -21,11 +22,32 @@ module.exports = {
...
+        '@typescript-eslint/no-unused-vars': 'off',
```
neenjaw, line 43: just my opnion here but in test files I don't think this matters too much.  Maybe downgrading it to a warning rather than an error?

.eslintrc.js
```diff
@@ -21,11 +22,32 @@ module.exports = {
...
   settings: {
     react: {
       version: 'detect',
     },
   },
+  overrides: [
+    {
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        'react/prop-types': 'off',
```
neenjaw, line 36: prop-types aren't used in typescript.  I think this is raising an error just because we have a mix of js and ts in the project

iHiD:
@neenjaw Thanks!
@kntsoriano Could you review pls? :)

kntsoriano (approved):
Looks good!

test/javascript/components/profile/contribution-summary/track-switcher.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const currentTrack = tracks[0]
+
+    render(
+      <TrackSwitcher tracks={tracks} value={currentTrack} setValue={() => {}} />
+    )
+
+    const button = screen.getByLabelText('Open the track switcher')
+
+    await act(async () => {
+      await fireEvent.click(button)
```
kntsoriano, line 55: I usually use `userEvent.click` instead of `fireEvent.click`. Is there any advantage to use `fireEvent` instead?

test/javascript/factories/ContributionSummaryFactory.ts
```diff
@@ -0,0 +1,66 @@
+import { RepresenterFeedback } from '../../../app/javascript/components/mentoring/session/RepresenterFeedback'
```
kntsoriano, line 1: Nice :) I should do something like this from now on.

test/javascript/components/profile/contribution-summary/track-switcher.test.tsx
neenjaw: I use fireEvent because that's what I know, but looking for a reference to think points out that using userEvent might be a better way to go about this. Reference: https://github.com/testing-library/eslint-plugin-testing-library/issues/162#issuecomment-645135887

iHiD: @neenjaw Want to change or should I merge? :)

neenjaw: They are essentially the same thing, so this works and covers it, but if I'm in this piece of code again I would change it. I think good to merge

PR #831, opened 2021-04-25
Add profile modal mockup

iHiD:
@kntsoriano Can you move this into a React component pls, so that we can have this at the top of the profile:

```haml
- if params[:first_time]
  = ReactComponents::Modals::ProfileFirstTime.new(@profile)
```


iHiD:
Nice :)

PR #847, opened 2021-04-28
Add <Expander /> component

kntsoriano:
https://github.com/exercism/v3-project-management/issues/229

@iHiD Does this work for you? http://local.exercism.io:3020/test/components/common/expander to test it.


iHiD:
@kntsoriano Yes. Thanks. I added a commit to see it used on the profile.

One change:
There should be two button texts - one for when it is compressed (`button_text_compressed`) and one for expanded (`button_text_expanded`)

kntsoriano:
@iHiD This should be ready.

iHiD:
Nice. You can see a good example on the C# track page (incognito)

PR #875, opened 2021-04-29
[WIP] Don't award reputation for v3 migration PRs

ErikSchierboom:


app/commands/user/reputation_token/award_for_pull_request_author.rb
```diff
@@ -8,6 +8,7 @@ class AwardForPullRequestAuthor
       def call
         return unless has_author?
         return unless merged?
+        return if v3_migration_pr?
```
iHiD, line 11: I'd rather this was a generic "no-reputation" label. Otherwise we'll end up with loads of rules here.

ErikSchierboom: That makes sense. It will be a bit more work updating all the existing PRs though

PR #880, opened 2021-04-30
Add polling back up if websockets fail

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/232

### DIFF
--- a/app/helpers/react_components/student/nudge.rb
+++ b/app/helpers/react_components/student/nudge.rb
@@ -11,7 +11,7 @@ module ReactComponents
             median_wait_time: solution.track.median_wait_time
           },
           discussions: discussions,
-          iteration: iteration,
+          request: request,
           exercise_type: exercise_type,
           links: links
         })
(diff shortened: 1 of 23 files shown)


app/helpers/react_components/student/nudge.rb
```diff
@@ -24,8 +24,17 @@ def exercise_type
...
-      def iteration
-        SerializeIteration.(solution.iterations.order(id: :desc).first)
+      def request
+        {
+          endpoint: Exercism::Routes.api_solution_url(solution.uuid, sideload: [:iterations]),
+          options: {
+            initialData: {
+              iterations: solution.
+                iterations.
+                map { |iteration| SerializeIteration.(iteration) }
```
iHiD: Changing this from `iteration` to `iterations` is quite a lot of extra data. In the JS we only seem to use `iteration.last.status`. Do we need to pass all this other data here or have you only done that because we have a matching API endpoint? If so, I'd rather we added an extra endpoint that just gets the latest iteration's status, rather than having to pass so much data round. If you agree, could you add a temporary endpoint that just returns iterations.status, and get the JS talking to that, and then I'll tweak it to be in the right place etc? (I need to think through where I want it)

iHiD:
This looks good in principle, but the tests are failing everywhere :D

app/helpers/react_components/student/nudge.rb
kntsoriano: Yes. I just tweaked it because I wanted the endpoints to match. I will introduce a new endpoint and a new channel for the latest iteration status.

app/models/submission/test_run.rb
```diff
@@ -5,6 +5,8 @@ class Submission::TestRun < ApplicationRecord
 
   scope :ops_successful, -> { where(ops_status: 200) }
 
+  delegate :solution, to: :submission
```
iHiD, line 8: @kntsoriano This should be an association not a delegate (I'll update, but just be aware of these as they cause n+1s otherwise)

PR #894, opened 2021-05-04
Fix bug where concepts and exercises created together failed

iHiD:


iHiD:
@ErikSchierboom Any idea why the Ruby test fails? Could you get it green pls?

ErikSchierboom:
@iHiD I know why this fails. Let's discuss on Slack how to handle this

PR #901, opened 2021-05-04
Add hydrated and hide removed explainers

iHiD:
This:
- Adds a `--hydrated` class to all React components once they are hydrated
- For the introducer, hides the (outer) react component once it's (inner) introducer has been removed.

### DIFF
--- a/app/javascript/packs/react-bootloader.jsx
+++ b/app/javascript/packs/react-bootloader.jsx
@@ -25,6 +25,7 @@ const renderComponents = (mappings) => {
     document.querySelectorAll(selector).forEach((elem) => {
       const data = JSON.parse(elem.dataset.reactData)
       render(elem, generator(data, elem))
+      elem.classList.add('--hydrated')
     })
   }
 }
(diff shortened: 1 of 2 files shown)


app/javascript/packs/react-bootloader.jsx
```diff
@@ -25,6 +25,7 @@ const renderComponents = (mappings) => {
     document.querySelectorAll(selector).forEach((elem) => {
       const data = JSON.parse(elem.dataset.reactData)
       render(elem, generator(data, elem))
+      elem.classList.add('--hydrated')
```
ErikSchierboom: Do we also want to add `--hydrated` to the tooltips? If so, this call either needs to be duplicated in the `renderTooltips` function, or moved to the shared `render` function.

ErikSchierboom (approved):

app/javascript/packs/react-bootloader.jsx
iHiD: Nice. Thanks. I've moved it into `render`

PR #922, opened 2021-05-06
Add share solution dropdown

iHiD:
@kntsoriano This should be a dropdown that's mounted on the public solution page.

Here's a loom with some info: 
https://www.loom.com/share/894f1a37fcec4910be2fb9c3a44c7f04

Once you've added that, I'll dry up the CSS.


kntsoriano:
@iHiD This looks good now. The only problem is that the CSS on the testimonial is broken.

PR #928, opened 2021-05-07
Allow user to upload their profile photo

kntsoriano:
https://github.com/exercism/v3-project-management/issues/233

## Questions
1. Should we turn the whole form into a whole React page as well? Currently, it updates the user record and attaches the avatar directly after it is uploaded.
2. This doesn't include the AWS setup that we need on production. Should be straightforward once we get everything to a good place.


app/javascript/components/profile/avatar-selector/CropFinishedStep.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const [submit, { status, error }] = useMutation(
+    () => {
+      if (!state.croppedImage) {
+        throw new Error('Cropped image was expected')
+      }
+
+      const formData = new FormData()
+      formData.append('user[avatar]', state.croppedImage, 'avatar.jpg')
+
+      /* TODO: Use our standard sendRequest library */
```
kntsoriano, line 35: I tried doing this but it ended up in a pretty deep rabbit hole. I've just left this as a TODO for now.

app/controllers/api/users_controller.rb
```diff
@@ -0,0 +1,21 @@
+module API
```
kntsoriano, line 1: This is just a temporary controller. I don't know whether we want the avatar attached to the user or the profile.

iHiD:
@kntsoriano I've added styling and made a few changes. It was more complex because Tai designed it to be partially in a modal. I'll merge this in, but you might want to check my code as I've probably not done it all idiomatically. I've added this Loom to explain: https://www.loom.com/share/5e9e0729eb6c4a22b95d3ba4464ef4b9

PR #932, opened 2021-05-07
Add solution stars

iHiD:
@kntsoriano This adds the basic solution stars code.

The star button on the [https://exercism.lol/tracks/ruby/exercises/two-fer/community_solutions/ihid](community solution page) should be a React component that uses this API to set itself. Add states of "starred" and "unstarred" depending on the base state.

https://www.loom.com/share/ed9fc024bae3420d9b15bf321fe86320


kntsoriano:
@iHiD Should be ready.

db/schema.rb
```diff
@@ -10,7 +10,35 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_05_05_220831) do
+ActiveRecord::Schema.define(version: 2021_05_07_230054) do
+
+  create_table "active_storage_attachments", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
```
kntsoriano: Was this accidentally added?

iHiD: Good spot. Yes. It's from the profiles PR I was working on.

PR #933, opened 2021-05-10
Unify journey page

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/236.

@iHiD I needed to move the tabs outside the hero component. We will need to fix the CSS for this. We also need to fix the anchoring when we hit the back button.


iHiD:
> @iHiD I needed to move the tabs outside the hero component. We will need to fix the CSS for this. 

Done.

> We also need to fix the anchoring when we hit the back button.

I actually don't want the anchoring. That was just a temporary thing. I'm fine with how it is right now. 

---

The tests are red. I'll see if I can get them fixed quickly, otherwise that's one for you to sort :)

PR #935, opened 2021-05-11
Update editor styling

iHiD:
Screenshot 2021-05-11 at 18 46 06
Screenshot 2021-05-11 at 18 47 55


iHiD:
@kntsoriano Can you check this over and get it green pls. I've noticed that sometimes the tests change to show as "timed out" a while after they've actually passed, so that one failing JS might be real.

PR #936, opened 2021-05-11
Add history function to <Mentoring.Queue />

### LINKED ISSUE #753: Edit URL when options change
On components where there are page-level components (for example on mentor-queue or mentor-inbox or the tracks page) when someone filters something the page URL should update, so that we can use the back button to return to that.

The actual interception of the query params will be intercepted by Rails and the component rendered with that state set. So this issue is purely for adding/changing query string params when things change.

To start with, work on the filters on the mentor-queue and mentor-inbox.

(Also cc @SleeplessByte in case he knows of the "standard" ways to do this in React)

kntsoriano:
Closes https://github.com/exercism/website/issues/753.

### DIFF
--- a/app/javascript/hooks/use-history.ts
+++ b/app/javascript/hooks/use-history.ts
@@ -24,17 +24,7 @@ function pushState<TParams>(params: TParams) {
   history.pushState(history.state, '', toQuery(params))
 }
 
-function replaceState<TParams>(params: TParams) {
-  history.replaceState(history.state, '', toQuery(params))
-}
-
-export function useHistory<TParams>({
-  pushOn,
-  replaceOn,
-}: {
-  pushOn: TParams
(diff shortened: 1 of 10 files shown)


app/javascript/hooks/use-history.ts
```diff
@@ -43,15 +33,8 @@ export function useHistory<TParams>({
       return
     }
 
+    console.log('pushing', pushOn)
+
```
ErikSchierboom: Should this be removed?

```suggestion
```

ErikSchierboom (approved):

app/javascript/hooks/use-history.ts
iHiD:
```suggestion
```

iHiD:
Nice work :)

PR #938, opened 2021-05-11
Disable Run tests button on editor

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/237.
Closes https://github.com/exercism/v3-project-management/issues/186.


app/controllers/api/solutions/submissions_controller.rb
```diff
@@ -15,6 +15,7 @@ def create
       # rescues SubmissionFileTooLargeError exceptions
       return render_error(400, :file_too_large) if files.any? { |file| file[:content].size > 1.megabyte }
 
+      # TODO: Allow rerunning of tests if previous submission was an error / ops error / timeout
```
kntsoriano, line 18: @iHiD I added this a todo because I don't know how we want to handle this.

iHiD: If they tests fail, then we should enable the button and let them run them again.

kntsoriano: @iHiD The React bit works already. We just need to conditionally raise `DuplicateSubmissionError`.

kntsoriano:
@iHiD Should be good :)

app/controllers/api/solutions/submissions_controller.rb
iHiD: I can ignore this now, right?

iHiD:
Nice work, dude. So as you probably discovered mouseleave doesn't fire on leaving disabled buttons. One bug in the way you had it was that if you exited the button on any side other than the LHS (where there was a little extra margin) the tooltip would then never disappear. Instead I've a hidden overlay that appears when the button is disabled, which the tooltip is mounted on instead. That's a pattern we might need to use on other disabled buttons too, so it's worth checking my commits just so you're clear on it :)

app/controllers/api/solutions/submissions_controller.rb
kntsoriano: @iHiD Oh, I don't think so yet. We need to ignore raising the error if the last submission errored. Currently, the React component works correctly by not disabling the button, but the backend will still return the `DuplicateSubmissionError`.

iHiD: Ah. I see. But it is still a duplicate submission, so that error is correct. What we need is for a new test-run to init off the old submission I guess. I'll have a think about how best to do this.

PR #945, opened 2021-05-11
Add specific last_touched_at column

iHiD:


iHiD:
~@kntsoriano Could you get this green as part of https://github.com/exercism/v3-project-management/issues/240 pls.~

I've hopefully done this.

test/models/user_track_test.rb
```diff
@@ -46,6 +46,39 @@ class UserTrackTest < ActiveSupport::TestCase
...
   end
 
+  test "touching only changes updated_at" do
+    user_track = freeze_time { create :user_track }
+    original_time = user_track.updated_at
+
+    travel 1.day do
+      user_track.touch
+      assert_equal Time.current, user_track.updated_at
+      assert_equal original_time, user_track.last_touched_at
```
ErikSchierboom, line 56: So `touch` basically sets `last_touched_at` to the old `updated_at` value?

ErikSchierboom (approved):

test/models/user_track_test.rb
iHiD: So the `touch` on `solution.rb` updates both `last_touched_at` and `updated_at`. But a normal touch just touches `updated_at`

Basically I want a column that monitors when the user has done something on their track, and another which monitors the actual state of the db record.

PR #947, opened 2021-05-11
[Blocked by dependency] Bump gems

iHiD:


iHiD:
@kntsoriano If this goes red, can you take a look at it pls. It's the bumped version of devise and omniauth.

PR #955, opened 2021-05-13
Tweak introducers on profiles

iHiD:


app/javascript/components/common/Introducer.tsx
```diff
@@ -30,17 +31,13 @@ export const Introducer = ({
...
       onSuccess: () => {
-        setHidden(true)
+        ref.current!.parentElement!.classList.add('hidden')
       },
     }
   )
 
-  if (hidden) {
-    return null
-  }
```
iHiD, line 40: We don't actually want the content to stop showing as we fade out the parent container instead, and we also need to persist the `ref` to access the container, so we continue to render the div.

app/javascript/components/common/Introducer.tsx
```diff
@@ -30,17 +31,13 @@ export const Introducer = ({
     },
     {
       onSuccess: () => {
-        setHidden(true)
+        ref.current!.parentElement!.classList.add('hidden')
```
iHiD, line 33: DJ suggested doing this in `useLayoutEffect`, and keeping the hidden state, but as this is directly powered by this callback and the state cannot be changed back, the extra indirection felt unnecessary to me. However, there may be good reasons to do that instead of what I've suggested so @kntsoriano please feel free to refactor and @SleeplessByte please feel free to shout at me for not doing exactly what you said :)

SleeplessByte: In this case the `layoutEffect` was only necessary to sync the state of the "hidden" variable. IN this case we don't need that because the effect is imperative. What you did here is fine.

iHiD: Ok great. Thank you 🙂

app/javascript/packs/react-bootloader.jsx
```diff
@@ -10,8 +10,15 @@ export const initReact = (mappings) => {
...
 
 const render = (elem, component) => {
-  ReactDOM.render(<React.StrictMode>{component}</React.StrictMode>, elem)
-  elem.classList.add('--hydrated')
+  ReactDOM.render(
+    <React.StrictMode>{component}</React.StrictMode>,
+    elem,
+    () => {
+      setTimeout(() => {
+        elem.classList.add('--hydrated')
```
SleeplessByte, line 18: If you really want this and not be bound to a flaky `setTimeout`, move it into the component.

```ts
function MyComponent() {
  useLayoutEffect(() => {
  	ref.current!.parentElement.classList.add('--hydrated')

	// technically you could even un-hydrate it on unmount:
	const parent = ref.current!.parentElement
    return (): void => {
      parent.classList.remove('--hydrated')
    }
  }, [])
}

iHiD: Ah, ok. That's really helpful to know. Will save that code for other places. Thanks.

PR #966, opened 2021-05-14
Support average run time

ErikSchierboom:


app/helpers/react_components/editor.rb
```diff
@@ -23,6 +23,7 @@ def to_s
           files: SerializeFiles.(solution.solution_files),
           ace_language: track.ace_language,
           highlightjs_language: track.highlightjs_language,
+          average_test_run_time: track.average_test_run_time,
```
iHiD: Could we change `time` to `duration_s` everywhere pls?

app/models/track.rb
```diff
@@ -89,6 +89,10 @@ def highlightjs_language
     git.highlightjs_language || slug
   end
 
+  def average_test_run_time
+    git.average_test_run_time + INFRASTRUCTURE_TIME
```
iHiD:
```suggestion
    git.average_test_run_time + INFRASTRUCTURE_DURATION_S
```

And this too

PR #980, opened 2021-05-19
Switch hello world order

iHiD:


iHiD:
(cc @ErikSchierboom - This is what you suggested)

PR #982, opened 2021-05-20
Replace editor with CodeMirror

kntsoriano:
https://github.com/exercism/v3-project-management/issues/248

No vim bindings for CodeMirror 6 :( https://github.com/codemirror/codemirror.next/issues/79


iHiD:
I get these messages btw:
```
Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-methods since the "loose" mode option was set to "true" for @babel/plugin-proposal-class-properties.
assets   | The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
assets   | 	["@babel/plugin-proposal-private-methods", { "loose": true }]
assets   | to the "plugins" section of your Babel config.
```

iHiD:
This is good:
1. Can you add the other languages etc (JS, Java, Rust, Python, C++ all have the “new” stuff, the others all use that legacy package - I added JS to test it all and commented it out on the branch).
2. Can you add an extra option on the menu for “Enable Tab” with On/Off, set to Off by default. If it’s on, then it adds the extra Tab support you added.

kntsoriano:
@iHiD Should be ready.

app/javascript/components/misc/CodeMirror/languageCompartment.ts
```diff
@@ -0,0 +1,34 @@
...
+    case 'rust':
+      return compartment.of(rust())
+    case 'python':
+      return compartment.of(python())
+    case 'elixir':
+      return compartment.of(StreamLanguage.define(elixir))
+    default:
+      return []
+  }
+}
```
iHiD, line 41: @ErikSchierboom This file needs all the languages in from all the repos. So we should some how co-ordinate the list you get on the tracks, with the libs here, etc.

PR #987, opened 2021-05-20
Add helper methods

iHiD:


iHiD:
@ErikSchierboom @kntsoriano This means you can do `Solution.find('ihid', 'ruby', 'two-fer')`

PR #991, opened 2021-05-21
Add Bugsnag to client side

kntsoriano:
https://github.com/exercism/v3-project-management/issues/139

### DIFF
--- a/app/javascript/components/ErrorBoundary.tsx
+++ b/app/javascript/components/ErrorBoundary.tsx
@@ -6,6 +6,7 @@ import {
   ErrorBoundaryPropsWithComponent,
 } from 'react-error-boundary'
 import { APIError } from './types'
+import Bugsnag from '@bugsnag/js'
 
 const ERROR_MESSAGE_TIMEOUT_IN_MS = 500
 
@@ -20,10 +21,18 @@ export const ErrorMessage = ({
 }: {
   error: Error | unknown
   defaultError: Error
-}) => {
+}): null => {
   useErrorHandler(error, { defaultError: defaultError })
 
   return null
+}
+
+const handleError = (error: Error) => {
+  if (error.name === 'HandledError') {
+    return
+  }
+
+  Bugsnag.notify(error)
 }
 
 export const ErrorBoundary = ({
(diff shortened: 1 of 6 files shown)


app/javascript/components/ErrorBoundary.tsx
```diff
@@ -20,19 +21,31 @@ export const ErrorMessage = ({
...
-}) => {
+}): null => {
   useErrorHandler(error, { defaultError: defaultError })
 
   return null
 }
 
+const handleError = (error: Error) => {
+  if (error.name === 'HandledError') {
+    return
```
ErikSchierboom, line 32: Does this swallow any errors now? As in: do errors still get reported in the console?

config/webpack/environment.js
```diff
@@ -1,12 +1,16 @@
...
-} else {
-  ASSET_HOST = ''
-}
+environment.plugins.prepend(
+  'Environment',
+  new webpack.EnvironmentPlugin(
+    JSON.parse(
+      JSON.stringify({
+        // TODO: @iHiD Should we get this from Exercism::Config?
+        BUGSNAG_API_KEY: '938ae3d231c5455e5c6597de1b1467af',
```
iHiD, line 10: Yes please. I'll merge this for now, but then put a PR together that reads it from Exercism::Config and I'll do the magic in the AWS end to get it all working.

config/webpack/environment.js
```diff
@@ -1,12 +1,16 @@
 const { environment } = require('@rails/webpacker')
+const webpack = require('webpack')
 
-let ASSET_HOST
-
-if (process.env.NODE_ENV === 'production') {
-  ASSET_HOST =
-    'https://exercism-assets-staging.s3.eu-west-2.amazonaws.com/packs/'
```
iHiD, line 7: @kntsoriano I feel like this shouldn't have been removed? I think this gets used for all the chunking stuff. Unless you're certain otherwise?

app/javascript/components/ErrorBoundary.tsx
kntsoriano: This is something I want to check. Currently, on development, the errors still get reported. But on production, I'm not entirely sure. I've been reading up and see that the webpack dev server raises the errors raised by the error boundary in the console.

config/webpack/environment.js
kntsoriano: Yes. We don't need this anymore.

app/javascript/components/ErrorBoundary.tsx
ErikSchierboom: Aha

iHiD:
Deploying this - let's see what happens :)

PR #992, opened 2021-05-21
Only debounce the text field

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/250.

### DIFF
--- a/app/javascript/components/mentoring/Inbox.jsx
+++ b/app/javascript/components/mentoring/Inbox.jsx
@@ -9,13 +9,16 @@ import { usePaginatedRequestQuery } from '../../hooks/request-query'
 import { useIsMounted } from 'use-is-mounted'
 import { ResultsZone } from '../ResultsZone'
 import { useHistory } from '../../hooks/use-history'
-import { useDebounce } from '../../hooks/use-debounce'
 
 export function Inbox({ tracksRequest, sortOptions, ...props }) {
-  const { request, setCriteria, setOrder, setPage, setQuery } = useList(
-    props.discussionsRequest
-  )
-  const debouncedQuery = useDebounce(request.query, 500)
+  const [criteria, setCriteria] = useState('')
+  const {
+    request,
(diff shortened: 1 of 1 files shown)


app/javascript/components/mentoring/Inbox.jsx
```diff
@@ -9,13 +9,16 @@ import { usePaginatedRequestQuery } from '../../hooks/request-query'
 import { useIsMounted } from 'use-is-mounted'
 import { ResultsZone } from '../ResultsZone'
 import { useHistory } from '../../hooks/use-history'
-import { useDebounce } from '../../hooks/use-debounce'
 
 export function Inbox({ tracksRequest, sortOptions, ...props }) {
-  const { request, setCriteria, setOrder, setPage, setQuery } = useList(
-    props.discussionsRequest
-  )
-  const debouncedQuery = useDebounce(request.query, 500)
```
ErikSchierboom, line 18: What's the difference between the old code and the new code?

iHiD: Previously everything was being routed via the debounce (including changing pagination-pages, changing tabs), etc. Now only the textual search goes via it.

app/javascript/components/mentoring/Inbox.jsx
```diff
@@ -24,12 +27,22 @@ export function Inbox({ tracksRequest, sortOptions, ...props }) {
...
+    ['mentor-discussion-list', request.endpoint, request.query],
     request,
     isMountedRef
   )
 
-  useHistory({ pushOn: debouncedQuery })
+  useEffect(() => {
+    const handler = setTimeout(() => {
+      setRequestCriteria(criteria)
+    }, 500)
```
iHiD:
```suggestion
    }, 200)
```

app/javascript/components/mentoring/Inbox.jsx
ErikSchierboom: Aha, okay!

PR #994, opened 2021-05-21
Import GitHub issues

ErikSchierboom:


app/commands/github/issue/create_or_update.rb
```diff
@@ -0,0 +1,43 @@
...
+          i.opened_at = attributes[:opened_at]
+          i.opened_by_username = attributes[:opened_by_username]
+        end
+
+        issue.update!(
+          number: attributes[:number],
+          title: attributes[:title],
+          status: status,
+          repo: attributes[:repo],
+          opened_at: attributes[:opened_at],
```
ErikSchierboom, line 23: I took the liberty of also adding this field (it wasn't in our "spec")

app/commands/github/issue/destroy.rb
```diff
@@ -0,0 +1,14 @@
+module Github
+  class Issue
+    class Destroy
```
ErikSchierboom, line 3: We use this only when we get the `deleted` event from the webhook. We could also treat deleted issues as closed 🤷

app/controllers/webhooks/issue_updates_controller.rb
```diff
@@ -0,0 +1,23 @@
+# This controller listens for Github pull request webhook events,
+# for which the "Issues" event type must be enabled in GitHub.
+# See https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#issues
+module Webhooks
+  class IssueUpdatesController < BaseController
```
ErikSchierboom, line 5: To get this working, remember to set a global webhook that listens for the "Issues" event type and points to `/webhooks/issue_updates`

app/models/github/issue.rb
```diff
@@ -0,0 +1,15 @@
+class Github::Issue < ApplicationRecord
+  enum status: { open: 0, closed: 1 }, _prefix: true
+
+  has_many :labels,
+    dependent: :destroy,
+    inverse_of: :issue,
+    class_name: "Github::IssueLabel",
+    foreign_key: "github_issue_id"
+
+  scope :open, -> { where(status: :open) }
```
ErikSchierboom, line 10: I assumed that we usually only want to deal with open issues, hence the scope

config/sidekiq-schedule.yml
```diff
@@ -7,3 +7,8 @@ fetch_and_sync_all_pull_requests_reputation:
   cron: "0 1 * * 1"
   class: "FetchAndSyncAllPullRequestsReputationJob"
   queue: dribble
+
+sync_issues:
+  cron: "0 2 * * 1"
```
ErikSchierboom: I've started this job one hour after the other job, which is enough time for it to finish I think.

iHiD: is it worth doing on a different day so that we don't overload our token?

ErikSchierboom: Good idea

ErikSchierboom:
```suggestion
  cron: "0 1 * * 2"
```

ErikSchierboom: It now runs a day later. Theoretically it shouldn't really matter as tokens reset each hour, but it is better to spread these jobs around regardless.

PR #995, opened 2021-05-21
Use .ruby-version rather than manually specifying versions

dougal:
* No need to separately specify Ruby version for each action.
* No need to load rubocop gems manually.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
   ##################
   rubocop:
     # TODO: Change this to website when merging
-    if: github.repository == 'exercism/website'
+    # if: github.repository == 'exercism/website'
 
     runs-on: ubuntu-latest
 
@@ -31,7 +31,7 @@ jobs:
   ##### JS TESTS ####
   ###################
   js-tests:
-    if: github.repository == 'exercism/website'
+    # if: github.repository == 'exercism/website'
 
     runs-on: ubuntu-latest
     steps:
@@ -68,7 +68,7 @@ jobs:
   ##### RUBY TESTS ####
   #####################
   ruby-tests:
-    if: github.repository == 'exercism/website'
+    # if: github.repository == 'exercism/website'
 
     runs-on: ubuntu-latest
     services:
@@ -156,7 +156,7 @@ jobs:
   ##### SYSTEM TESTS ####
   #######################
   system-tests:
-    if: github.repository == 'exercism/website'
+    # if: github.repository == 'exercism/website'
 
     runs-on: ubuntu-latest
     services:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -19,19 +19,13 @@ jobs:
...
-          gem install rubocop -v 1.7.0
-          gem install rubocop-minitest -v 0.11.0
-          gem install rubocop-performance -v 1.10.1
-          gem install rubocop-rails -v 2.9.1
+          ruby-version: .ruby-version
+          bundler-cache: true
 
       - name: Run Rubocop
-        run: rubocop --except Metrics
+        run: bundle exec rubocop --except Metrics
```
iHiD: So I deliberately wasn't using bundler here but I think that was because the tooling libraries (e.g. test runner) don't use Bundler as it makes things too slow. For the website, it makes sense I think.

.github/workflows/tests.yml
```diff
@@ -199,9 +193,9 @@ jobs:
       ###
       # Setup Ruby - this needs to match the version in the Gemfile
       - name: Set up Ruby
-        uses: ruby/setup-ruby@a699edbce608a2c128dedad88e3b6a0e28687b3c
+        uses: ruby/setup-ruby@b161cb92f0e2f28d992f04f38fff204d84132c47
         with:
-          ruby-version: 2.6.6
+          ruby-version: .ruby-version
```
iHiD, line 204: This is nice.

iHiD:
(The reason the checks weren't running was that all first-time contributions have to be checked now before GHA runs)

.github/workflows/tests.yml
dougal: Discovered this earlier today. Got to throw away a bunch of hacks which did the same thing.

.github/workflows/tests.yml
dougal: Reasoning to not use makes sense.

I have wondered about putting Rubocop in it's own Rubygems group and installing just that.

dougal: Reverted to original.

.github/workflows/tests.yml
iHiD: That's great.

.github/workflows/tests.yml
iHiD: That's not a bad shout. For the test-runner stuff, I mainly discovered that running the tests with `rake test` takes 0.5s, and with `bundle exec rake test` it takes 2s. When the "product" is running tests fast, that was a ridiculous difference. Then I think these all just copied from that world.

iHiD:
(I forced pushed to your repo FYI)

iHiD:
Thanks dude :)

PR #997, opened 2021-05-21
Link analyzer/representer/test-runner reputation tokens to corresponding track

ErikSchierboom:


app/models/user/reputation_tokens/code_contribution_token.rb
```diff
@@ -6,7 +6,8 @@ class User::ReputationTokens::CodeContributionToken < User::ReputationToken
   values({ minor: 5, regular: 12, major: 30 })
 
   before_validation on: :create do
-    self.track_id = Track.where(repo_url: "https://github.com/#{repo}").pick(:id) unless track
+    normalized_repo = repo.gsub(/-(test-runner|analyzer|representer)$/, '')
```
iHiD, line 9: You could guard this whole thing with `return unless track` rather than having it on the final line.

iHiD (approved):

app/models/user/reputation_tokens/code_contribution_token.rb
ErikSchierboom: Yeah, that's better. I'll update

PR #1004, opened 2021-05-21
Use solution SHA for serializing assignment

### LINKED ISSUE #1000: Fix incorrect sha use for assignments
@ErikSchierboom This should probably take an optional `solution` param, and then use that in the serializer if present.

https://github.com/exercism/website/blob/5b45b949181672c7c87fc261034fcb08bd5afd1d/app/helpers/react_components/editor.rb#L13

ErikSchierboom:
Closes https://github.com/exercism/website/issues/1000


iHiD:
I hadn't appreciated that someone *has* to have a solution for us to do this. That makes it easier :)

app/serializers/serialize_exercise_assignment.rb
```diff
@@ -87,7 +87,7 @@ def instructions_doc
 
   memoize
   def hints_doc
-    Markdown::Render.(exercise.git.hints, :doc)
+    Markdown::Render.(solution.git_exercise.hints, :doc)
```
iHiD, line 90: I think this should be cached in a `git_exercise` method so we don't init everything 3 times.

ErikSchierboom: It's a memoized method:

```
memoize
  def git_exercise
    Git::Exercise.for_solution(self)
  end
```

PR #1005, opened 2021-05-21
Dynamically import languages

ErikSchierboom:


tsconfig.json
```diff
@@ -5,10 +5,10 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "lib": ["es6", "dom", "es2018.promise", "es2017"],
-    "module": "es6",
+    "module": "es2020",
```
ErikSchierboom, line 8: This is required to be able to do dynamic imports.

tsconfig.json
```diff
@@ -5,10 +5,10 @@
...
     "moduleResolution": "node",
     "sourceMap": true,
-    "target": "es5",
+    "target": "es6",
```
ErikSchierboom, line 11: This is just something tangential that I noticed. As we don't really support older browsers, we [can use es6](https://caniuse.com/?search=es6) as the target. This means that the code that is generated can be more compact.

app/javascript/components/misc/CodeMirror/languageCompartment.ts
```diff
@@ -1,39 +1,39 @@
...
-import { javascript } from '@codemirror/lang-javascript'
-import { cpp } from '@codemirror/lang-cpp'
-import { java } from '@codemirror/lang-java'
-import { rust } from '@codemirror/lang-rust'
-import { python } from '@codemirror/lang-python'
-
 const compartment = new Compartment()
 
-export const languageCompartment = (language: string): Extension => {
+export const importLanguageCompartment = async (
```
ErikSchierboom: This has to be an `async` method now that we're using dynamic imports.

app/javascript/components/misc/CodeMirror.tsx
```diff
@@ -71,31 +71,32 @@ export const CodeMirror = ({
...
-          indentUnit.of(useSoftTabs ? '  ' : '	'),
-          wrapCompartment.of(wrap ? EditorView.lineWrapping : []),
-          themeCompartment.of(
-            EditorView.theme({}, { dark: theme === Themes.DARK })
-          ),
-          keymap.of(commands),
-        ],
-      }),
-      parent: textarea,
+    importLanguageCompartment(language).then((languageCompartment) => {
```
ErikSchierboom: @kntsoriano I haven't spent a lot of time on this, but my code is not working. This is most likely due to the fact that the language compartment is now retrieved asynchronously. I assume this means that this `useEffect` call is returning _before_ the language compartment has been retrieved, which might mess with subsequent `useEffect` calls. I'd be happy to look into this a bit more, but if you feel like you can solve this quickly, be my guest.

tsconfig.json
ErikSchierboom: @SleeplessByte Is there any downside to using `es2020` as the module system?

tsconfig.json
ErikSchierboom: @SleeplessByte This is safe to do, right? We only support modern browsers.

joshgoebel:
@ErikSchierboom Please don't forget Wren now via https://github.com/exercism/codemirror-lang-wren.

ErikSchierboom:
@joshgoebel I found an issue while trying to add Wren support: https://github.com/exercism/codemirror-lang-wren/issues/4 @SleeplessByte will be fixing it. Once fixed, it will be easy to add support for Wren.

tsconfig.json
SleeplessByte: Not really, because you'll be using babel to "downgrade" it again :). All good.

iHiD:
Boom. Let's do this!

tsconfig.json
SleeplessByte: You also don't use typescript to "output" the code. You use babel (via webpack). So changing this doesn't change the output. 

The idea is good!

SleeplessByte:
@exercism/codemirror-lang-wren 0.6.0 should fix the issues you encountered @ErikSchierboom

PR #1015, opened 2021-05-22
Fix mentor ordering

iHiD:


iHiD:
@kntsoriano In this PR I ended up refactoring the student tooltip to work the same as the exercise one. This is a good example of where I think we need Tippy though, as it doesn't work very well as a UX thing not following your mouse. You'll see on production.

PR #1020, opened 2021-05-25
Add Ansi::RenderHTML command

### LINKED ISSUE #37: [Page: Editor] ANSI colors work for version 1 test runners, but not version 2
This is an example output of the Prolog test runner (version 1):
Screen Shot 2021-05-22 at 14 18 23

It has colors.

After turning colors in Elixir's test runner (version 2), the output looks like this:

Screen Shot 2021-05-23 at 08 30 50

I would expect both `message` (seen on the screenshot) and `output` to support ANSI colors.

ErikSchierboom:
Closes https://github.com/exercism/v3-beta/issues/37

image

### DIFF
--- a/app/javascript/components/editor/TestSummary.tsx
+++ b/app/javascript/components/editor/TestSummary.tsx
@@ -60,13 +60,13 @@ export function TestSummary({
         {isPresent(test.message) ? (
           <div className="--info">
             <h3>{messageLabels[test.status]}</h3>
-            <pre>{test.message}</pre>
+            <pre dangerouslySetInnerHTML={{ __html: test.messageHtml }}></pre>
           </div>
         ) : null}
         {isPresent(test.output) ? (
           <div className="--info">
             <h3>Your Output</h3>
-            <pre>{test.output}</pre>
+            <pre dangerouslySetInnerHTML={{ __html: test.outputHtml }}></pre>
           </div>
         ) : null}
       </div>
--- a/app/serializers/serialize_submission_test_run.rb
+++ b/app/serializers/serialize_submission_test_run.rb
@@ -12,9 +12,9 @@ class SerializeSubmissionTestRun
       version: test_run.version,
       status: status,
       message: message,
-      message_html: message_html,
+      message_html: Ansi::RenderHTML.(message),
       output: test_run.output,
-      output_html: output_html,
+      output_html: Ansi::RenderHTML.(test_run.output),
       tests: test_run.test_results,
       highlightjs_language: test_run.solution.track.highlightjs_language,
       links: {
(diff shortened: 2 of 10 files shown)


app/javascript/components/editor/TestSummary.tsx
```diff
@@ -44,19 +47,19 @@ export function TestSummary({
         {isPresent(test.testCode) ? (
           <div className="--info">
             <h3>Code Run</h3>
-            <pre>{test.testCode}</pre>
+            <pre dangerouslySetInnerHTML={{ __html: test.testCodeHtml }}></pre>
```
ErikSchierboom: I've also rendered the test code as HTML, as I could imagine that using syntax highlighting colors there might be really beneficial for tracks that can do that.

app/javascript/components/editor/TestSummary.tsx
```diff
@@ -44,19 +47,19 @@ export function TestSummary({
...
             <h3>{messageLabels[test.status]}</h3>
-            <pre>{test.message}</pre>
+            <pre dangerouslySetInnerHTML={{ __html: test.messageHtml }}></pre>
           </div>
         ) : null}
         {isPresent(test.output) ? (
           <div className="--info">
             <h3>Your Output</h3>
-            <pre>{test.output}</pre>
+            <pre dangerouslySetInnerHTML={{ __html: test.outputHtml }}></pre>
```
ErikSchierboom, line 69: I've kept the output as HTML, as student could theoretically use colors in their output, which the test runner could capture.

app/serializers/serialize_submission_test_run.rb
```diff
@@ -31,10 +31,7 @@ def status
   end
 
   def message_html
-    return nil unless message
-
-    sanitized_message = message.gsub("\e\[K", '')
-    Ansi::To::Html.new(sanitized_message).to_html
+    Ansi::RenderHTML.(message)
```
iHiD: This probably doesn't need a method now.

app/serializers/serialize_submission_test_run.rb
```diff
@@ -45,12 +42,7 @@ def message
   end
 
   def output_html
-    return if test_run.output.blank?
-
-    # The ansi-to-html library does not support unicode escape sequence
-    # See https://github.com/rburns/ansi-to-html/issues/48
-    sanitized_output = test_run.output.gsub("\e\[K", '')
-    Ansi::To::Html.new(sanitized_output).to_html
+    ::Ansi::RenderHTML.(test_run.output)
```
iHiD: This probably doesn't need a method now (you can just inline it)

iHiD:
Also still red :)

iHiD:
That's all much nicer! Thanks 🙂

PR #1022, opened 2021-05-25
Split concepts and their exercises so both are clickable

iHiD:
We want to allow both concepts and their exercises to be hoverable and clickable, and both have tooltips. It's quite hard to get a feel for how this works without making it, so this is a spike to test it. Here's a loom I made for @taiyab that demonstrates it: https://www.loom.com/share/278bc7d80e3b4f83bd9731bf87b11fdc


iHiD:
I'll merge this and we can try it and see how it feels.

PR #1025, opened 2021-05-25
Add a11y tab support to CodeMirror

iHiD:
@kntsoriano @ErikSchierboom @SleeplessByte Pls give this a look over.

My aim with this PR is:
- Enable us to have tab as a whitespace inserter by default
- Clearly communicate to a user without a pointing-device that there is an escape hatch.
- Continue to enable support for a user to disable the tab as a whitespace-inserter.

To achieve this, this PR adds an assertive aria-live area that announces to the user that they can press ESC and then tab to leave the editor.

I initially wrote this using the built-in CodeMirror announce functionality (I'll include this in a comment for reference). This uses `aria-live=polite`. However, due to the way things get read out (by VoiceOver at least), the "polite" aria-live region is often ignored or comes after VoiceOver has read **all** the code in the editor twice(?!). If you've tabbed into the editor and want to get out, this is incredibly annoying and effectively means we trap the user for at least 10 seconds. I considered the downsides of having assertive, and I can't really think of any as it only happens in direct response to a user pressing tab (rather than something in the background interrupting the user). It currently only makes the announcement at most every 5 seconds. I'm not sure what the correct frequency is - I wonder if a backoff is good (so 5s between the first, 20s between the second, then every minute or so after that).

I've also changed the tab mode option to be this:

Screenshot 2021-05-25 at 23 51 56

I suggest that we add a third option `Editor (Silent)` so that the tab can work a whitespace-inserter and not have this announcement at all, for those who are happy to remember the ESC-tab and don't want the constant announcements.

With regards to the code, I've commented it as CodeMirror isn't particularly intuitive without reading all the docs (but that complexity means it's also very modular, which is nice). There's a couple of TS-ignore bits (aria and one typing bit that seems wrong). Help on that would be appreciated. Then any other general advice too. This is the first TypeScript I've written that's not just editing @kntsoriano's existing work.

If we're all happy with this, we can extract it into an npm package that others can use, and I'll inform the CodeMirror team about it.

---

~Also, the escape hatch doesn't actually work. It's fixed in the latest version but all the types break if I install that because different things seem to rely on different types, so I'll wait until `@codemirror/basic-setup` is updated to install that (or we'll fix it if we get to launch and it's still broken)~ **Update:** I worked this out.

### DIFF
--- a/app/javascript/components/editor/header/Settings.tsx
+++ b/app/javascript/components/editor/header/Settings.tsx
@@ -19,9 +19,9 @@ const WRAP = [
   { label: 'Off', value: 'off' },
 ]
 
-const TAB_ENABLED = [
-  { label: 'On', value: 'captured' },
-  { label: 'Off', value: 'default' },
+const TAB_MODE = [
+  { label: 'Editor', value: 'captured' },
+  { label: 'Accesibility', value: 'default' },
 ]
 
 const Setting = React.forwardRef<
@@ -175,9 +175,9 @@ export function Settings({
               {...itemAttributes(2)}
             />
             <Setting
-              title="Tab enabled"
+              title="Tab mode"
               value={tabBehavior}
-              options={TAB_ENABLED}
+              options={TAB_MODE}
               set={(tabBehavior) => setTabBehavior(tabBehavior as TabBehavior)}
               {...itemAttributes(3)}
             />
--- a/app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
+++ b/app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
@@ -31,21 +31,45 @@ function tabPressed(changes: any): boolean {
   return changes.inserted.some((insert) => insert.text == '\t')
 }
 
-// This changes the desired value every time tab is pressed.
+// This cycles though the desired value every time tab is pressed.
 // This is useful for ensuring that screen-readers don't just
 // think the text is the same as it previously was and ignore it.
-const text1 = 'Press Escape then Tab to exit the editor'
-const text2 = 'To exit the editor, press Escape then tab'
+//
+// Currently this cycles whenever tab is pressed, whether or not we announce
+// it via the aria-live region (ie whether the box is blank or not at the time)
+// We may want to change this but I've not yet worked out how.
+//
+// Announcements only seem to happen once per prompt however many
+// times it is cycled into the box (I presume that's a non-spamming feature)
+// so for now I have 5 different prompts which means this is announced a maximum
+// of 5 times. There's also a chance this is just down to something weird
+// in VoiceOver (which seems to be quite random to me) so this might be entirelty
+// unncessary and we can just switch between twodifferent ones as per a previous
+// commit in the PR that adds this.
+const prompts = [
+  'Press Escape then Tab to exit the editor',
+  'To exit the editor, press Escape then tab',
+  'Press Escape followed by tab to exit the editor',
+  'If you want to exit the editor, press Escape then Tab',
+  'In order to exit the editor, press Escape then Tab',
+]
 const a11yTabBindingState = StateField.define<string>({
-  create: () => text1,
+  create: () => '',
   update(value, tr) {
-    if (tabPressed(tr.changes)) {
-      if (value == text1) {
-        value = text2
-      } else {
-        value = text1
-      }
+    if (!tabPressed(tr.changes)) {
+      return value
     }
+    let found = false
+    const newValue = prompts.find((prompt) => {
+      if (found) {
+        return value
+      }
+
+      if (value == prompt) {
+        found = true
+      }
+    })
+    value = newValue || prompts[0]
     return value
   },
 })
--- a/app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
+++ b/app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
@@ -1,0 +1,55 @@
+import { Text } from '@codemirror/text'
+import { EditorView, KeyBinding } from '@codemirror/view'
+import { Panel } from '@codemirror/panel'
+import { insertTab, indentSelection } from '@codemirror/commands'
+import { StateCommand } from '@codemirror/state'
+
+export const a11yTabBindingPanelTheme = EditorView.baseTheme({
+  '.cm-a11y-panel': {
+    position: 'absolute',
+    width: '1px',
+    height: '1px',
+    padding: '0',
+    margin: '-1px',
+    overflow: 'hidden',
+    clip: 'rect(0, 0, 0, 0)',
+    whiteSpace: 'nowrap',
+    borderWidth: '0',
+  },
+})
+
+export function a11yTabBindingPanel(view: EditorView): Panel {
+  let dom = document.createElement('div')
+  dom.textContent = ''
+  console.log(dom)
+  // @ts-ignore */
+  dom.ariaLive = 'assertive'
+  dom.className = 'cm-a11y-panel'
+
+  return {
+    top: true,
+    dom,
+    update(update) {
+      if (!update.docChanged) {
+        return
+      }
+      // @ts-ignore */
+      if (update.changes.inserted.some((insert) => insert.text == '\t')) {
+        dom.textContent = 'Press ESC then Tab to exit the editor'
+        setTimeout(() => {
+          dom.textContent = ''
+        }, 5000)
+      }
+    },
+  }
+}
+
+export const customInsertTab: StateCommand = ({ state, dispatch }) => {
+  return insertTab({ state, dispatch })
+}
+
+export const a11yTabBinding: KeyBinding = {
+  key: 'Tab',
+  run: customInsertTab,
+  shift: indentSelection,
+}
--- a/app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
+++ b/app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
@@ -1,9 +1,15 @@
-import { Text } from '@codemirror/text'
-import { EditorView, KeyBinding } from '@codemirror/view'
+import { EditorView } from '@codemirror/view'
 import { Panel } from '@codemirror/panel'
-import { insertTab, indentSelection } from '@codemirror/commands'
-import { StateCommand } from '@codemirror/state'
+import { StateField } from '@codemirror/state'
+import { showPanel } from '@codemirror/panel'
 
+// See the note on this near the bottom of the file
+// import { insertTab, indentSelection } from '@codemirror/commands'
+// import { KeyBinding } from '@codemirror/view'
+// import { StateCommand } from '@codemirror/state'
+
+// This is taken from Tailwind's `sr-only` class.
+// It hides the panel from view on visual devices
 export const a11yTabBindingPanelTheme = EditorView.baseTheme({
   '.cm-a11y-panel': {
     position: 'absolute',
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "exercism",
   "private": true,
   "dependencies": {
-    "@codemirror/basic-setup": "^0.18.1",
+    "@codemirror/basic-setup": "^0.18.2",
```
iHiD, line 6: Adding `@codemirror/view` and specifying 0.18.13 explicitely here breaks Typescript quite badly.

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
```diff
@@ -0,0 +1,144 @@
...
+// This creates a CodeMirror Panel, adds its classes and sets it
+// to be an assertive ariaLive area. It is assertive as it only
+// changes in direct response to a user pressing the tab key.
+function createA11yTabBindingPanel(view: EditorView): Panel {
+  let dom = document.createElement('div')
+  dom.textContent = ''
+  dom.className = 'cm-a11y-panel'
+
+  // @ts-ignore */
+  dom.ariaLive = 'assertive'
```
ErikSchierboom, line 86: According to [this page](https://caniuse.com/?search=ariaLive), `ariaLive` is not supported on Firefox. Same goes for `ariaAtomic`: https://caniuse.com/?search=ariaatomic.

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
```diff
@@ -0,0 +1,144 @@
...
+//
+// Currently this cycles whenever tab is pressed, whether or not we announce
+// it via the aria-live region (ie whether the box is blank or not at the time)
+// We may want to change this but I've not yet worked out how.
+//
+// Announcements only seem to happen once per prompt however many
+// times it is cycled into the box (I presume that's a non-spamming feature)
+// so for now I have 5 different prompts which means this is announced a maximum
+// of 5 times. There's also a chance this is just down to something weird
+// in VoiceOver (which seems to be quite random to me) so this might be entirelty
```
ErikSchierboom, line 46:
```suggestion
// in VoiceOver (which seems to be quite random to me) so this might be entirely
```

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
```diff
@@ -0,0 +1,144 @@
...
+// unncessary and we can just switch between twodifferent ones as per a previous
```
ErikSchierboom, line 47:
```suggestion
// unncessary and we can just switch between two different ones as per a previous
```

ErikSchierboom (approved):

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
iHiD: I think this is wrong. From Googling around lots of people discuss it like it works (e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1198336)

@SleeplessByte might know more? I can also test it later (but I don't have firefox installed atm)

SleeplessByte: Neither is wrong.

The API that Erik linked is indeed not supported in Firefox, but it's not what Jeremy *needs* to be using.

> The ariaLive property of the Element interface reflects the value of the aria-live attribute.

The aria-live _attribute_ works in the _assistive technologies_. 

The `ariaLive` _property_ maps to that attribute and is irrelevant here if using the attribute.

SleeplessByte: So instead, use `.setAttribute` instead.

package.json
iHiD: Fixed.

iHiD:
I'm going to merge this so everyone can test it in production, but I'd still entirely appreciate reviews etc!

app/javascript/components/editor/header/Settings.tsx
```diff
@@ -19,9 +19,9 @@ const WRAP = [
   { label: 'Off', value: 'off' },
 ]
 
-const TAB_ENABLED = [
-  { label: 'On', value: 'captured' },
-  { label: 'Off', value: 'default' },
+const TAB_MODE = [
+  { label: 'Editor', value: 'captured' },
+  { label: 'Accesibility', value: 'default' },
```
SleeplessByte, line 24:
```suggestion
  { label: 'Accessibility', value: 'default' },
```

I'd recommend **not** using `default` as the value, if that's possible. It's unclear from the code that by `default` it means "tab moves focus" and `captures` means "tab inserts whitespace". Both are names of *how* it's achieved and not *what* the mode represents.

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
```diff
@@ -0,0 +1,144 @@
...
+    padding: '0',
+    margin: '-1px',
+    overflow: 'hidden',
+    clip: 'rect(0, 0, 0, 0)',
+    whiteSpace: 'nowrap',
+    borderWidth: '0',
+  },
+})
+
+// The typing for this seems to be wrong in CodeMirror.
```
SleeplessByte, line 27: The types included don't match the codemirror version you use.

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
```diff
@@ -0,0 +1,144 @@
...
+// Changes doesn't have `inserted` defined on it
+function tabPressed(changes: any): boolean {
+  // @ts-ignore */
+  return changes.inserted.some((insert) => insert.text == '\t')
```
SleeplessByte, line 31: You might want to do `===`. It doesn't matter here.

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
```diff
@@ -0,0 +1,144 @@
...
+  'In order to exit the editor, press Escape then Tab',
+]
+const a11yTabBindingState = StateField.define<string>({
+  create: () => '',
+  update(value, tr) {
+    if (!tabPressed(tr.changes)) {
+      return value
+    }
+    let found = false
+    const newValue = prompts.find((prompt) => {
```
SleeplessByte, line 63: I don't understand what you're trying to do in this function.

Are you trying to find the first `prompt` that matches `value` and return that?

```typescript
const newValue = prompts.find((prompt) => prompt === value)
```

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
```diff
@@ -0,0 +1,144 @@
...
+// changes in direct response to a user pressing the tab key.
+function createA11yTabBindingPanel(view: EditorView): Panel {
+  let dom = document.createElement('div')
+  dom.textContent = ''
+  dom.className = 'cm-a11y-panel'
+
+  // @ts-ignore */
+  dom.ariaLive = 'assertive'
+  // @ts-ignore */
+  dom.ariaAtomic = 'true'
```
SleeplessByte, line 88: Use `.setAttribute` instead

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
```diff
@@ -0,0 +1,144 @@
...
+    top: true,
+    dom,
+    update(update) {
+      // Only announce things if the user has pressed a tab
+      if (!update.docChanged || !tabPressed(update.changes)) {
+        return
+      }
+
+      // Only announce things if the aria-live area has been reset
+      if (dom.textContent != '') {
```
SleeplessByte, line 100: `!==`

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
```diff
@@ -0,0 +1,144 @@
...
+        return
+      }
+
+      // Set the text content, which announces via the aria-live functionality.
+      dom.textContent = view.state.field(a11yTabBindingState)
+
+      // Reset the text content five seconds later.
```
SleeplessByte, line 107: I'm assuming you do this because you need to re-announce it the next time? The 5second rules might seem safe, but I can configure my screen reader to s p e a k v  e  r  y  s    l    o    w    e   l   y.

I suspect that what you want to do can *actually* be achieved by changed the `id` of this element when you want to start announcing. I _think_ a new ID will act as if it's a new element. But we should try this in VoiceOver, NVDA and JAWS.

SleeplessByte (commented):
Overall looks like a solid approach. Comments added where I think we can improve.

I also would suggest you look again at the dialog that is created in the settings box here and make sure the following things are true:

- The dialog must be closable by a focusable element (so not just escape). You can do this by adding a `sr-only` button to close the dialog. This is my usual approach.
- You've annotated the dialog title using `aria-title`, but I recommend `aria-labelledby="dialog1Title" aria-describedby="dialog1Desc"` instead. You can make those elements `sr-only` too. `aria-title` on the dialog doesn't do the same thing in all screen readers.
- When the dialog appears on the screen, keyboard focus (whose control depends upon the dialogs purpose) should be moved to the default focusable control inside the dialog.
- After the dialog is dismissed, keyboard focus should be moved back to where it was before it moved into the dialog.
- For most dialogs, the expected behavior is that the dialog's tab order wraps, which means that when the user tabs through the focusable elements in the dialog, the first focusable element will be focused after the last one has been reached. In other words, the tab order should be contained within and by the dialog. This is **a requirement** if the dialog is a modal, but I recommend doing it for most complexer dialogs.

app/javascript/components/editor/header/Settings.tsx
iHiD: `default` is the codemirror key here. We could probably add another label of abstraction, but I think right now, that gets proxied straight to codemirror.

SleeplessByte: I'd add it as a comment :P

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
iHiD: If finds the first prompt *after* the one that matches the value.

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
iHiD: I updated this on main since your suggestion. So I'll resolve this one. Thanks :)

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
iHiD: The 5 second thing more because if someone does presses `tab tab tab`, it would restart each time (because of assertive: true)

> I suspect that what you want to do can actually be achieved by changed the id of this element when you want to start announcing. I think a new ID will act as if it's a new element. But we should try this in VoiceOver, NVDA and JAWS.

This might solve lots. I'll try this :)

iHiD:
@SleeplessByte Thanks 💙 

I'll add the small suggestions in another PR. 

Re the dialog, I've added https://github.com/exercism/website/issues/1046 for that. @kntsoriano has written all this so far so I'll leave it to him to improve it. The plan is to do a full pass once we have all the bits in place on the pages, so we can ensure we standardise everything (in components) where possible considering all the permutations.

app/javascript/components/misc/CodeMirror/a11yTabBinding.ts
SleeplessByte:
```javascript
const newValueIndex = prompts.indexOf(value)
// this returns -1 if it can't be found, so it will become -1 + 1 == 0
const newValue = prompts[(newValueIndex + 1) % prompts.length]
```

SleeplessByte:
All good! Ping me when I need to review that :)

PR #1030, opened 2021-05-26
Fix indentation in file viewer

### LINKED ISSUE #25: [Page: Editor] Inconsistent tab size.
I found this accidentally and am not even sure if one consider it a bug.

While solving a rust exercise and playing around with the setting "Tab enabled" I ended up submitting a file with different indentation styles (mixed spaces and tab) in the editor the file looks like this (Line 20 has tab indentation)

image

And in the "Your iterations" tab it looked like this (Line 20):

image

IMHO both options are fine but they should be consistent.

ErikSchierboom:
Closes https://github.com/exercism/v3-beta/issues/25

### DIFF
--- a/app/javascript/components/mentoring/session/FileViewer.tsx
+++ b/app/javascript/components/mentoring/session/FileViewer.tsx
@@ -4,9 +4,11 @@ import { useHighlighting } from '../../../utils/highlight'
 
 export const FileViewer = ({
   language,
+  indentSize,
   file,
 }: {
   language: string
+  indentSize: number
   file: File
 }): JSX.Element => {
   const parentRef = useHighlighting<HTMLPreElement>()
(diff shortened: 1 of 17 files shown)


app/javascript/utils/highlight.ts
```diff
@@ -49,7 +49,7 @@ function getLinesCount(text: string) {
 
 function wrapLineNumbers(code: string, start = 1) {
   const element = document.createElement('div')
-  element.innerHTML = code.trim()
+  element.innerHTML = code.replace(/(^[\r\n]+|[\r\n]+$)/g, '')
```
ErikSchierboom: This was causing the leading spaces/tabs to be removed. I've changed this to just removing newlines from the start and end.

app/javascript/components/mentoring/session/FileViewer.tsx
```diff
@@ -17,8 +19,10 @@ export const FileViewer = ({
         className={language}
         data-highlight-line-numbers={true}
         data-highlight-line-number-start={1}
-        dangerouslySetInnerHTML={{ __html: file.content }}
-      />
+        style={{ tabSize: indentSize }}
```
ErikSchierboom, line 22: As far as I could tell, this property is supported in all major browsers.

app/javascript/utils/highlight.ts
iHiD: Could you add a comment to this line explaining what this does pls?

iHiD (approved):

PR #1034, opened 2021-05-27
Add contributing section

iHiD:
This adds a stub of the first contribution dashboard page and an implemented contributors list that will get converted into React.

This adds a ReputationPeriod model, which is an aggregate of the ReputationTokens for specific periods. More details are in [this loom](https://www.loom.com/share/14895434285545dd8d20c59032db4fa7). This is all a little complex and full of optimisations but it works very well.

I'll merge this in the morning if it's green and no-one has any comments, and then it **should** (hopefully!) just run moving forward for any new data (PRs merges etc)


app/helpers/view_components/contributing/header.rb
```diff
@@ -0,0 +1,92 @@
...
+          end
+        ]
+      end
+
+      def tab_class(tab)
+        "c-tab-2 #{'selected' if tab == selected_tab}"
+      end
+
+      def guard!
+        raise "Incorrect track nav tab" unless TABS.include?(selected_tab)
```
ErikSchierboom, line 89: Will this show an error page or a 404?

ErikSchierboom (approved):

app/helpers/view_components/contributing/header.rb
iHiD: This will hit bugsnag. It should absolutely never happen and if it does, something somewhere has gone very wrong. And if the tests haven't caught this, I deserve shooting :)

ErikSchierboom: :D

app/models/user/reputation_token.rb
```diff
@@ -23,6 +23,9 @@ class User::ReputationToken < ApplicationRecord
     self.reason = self.class.instance_variable_get("@reason")
     self.category = self.class.instance_variable_get("@category")
 
+    # TODO: This should be set correctly in the command not here
+    self.earned_on = Time.current.to_date unless earned_on
```
iHiD, line 27: @ErikSchierboom We're going to want to set this to be the point at which a PR is merged (or equivalent for other things)

app/commands/user/reputation_period/mark_for_new_token.rb
```diff
@@ -0,0 +1,24 @@
+class User::ReputationPeriod
+  class MarkForNewToken
+    include Mandate
+
+    initialize_with :token
+
+    def call
+      args = { user_id: token.user_id, user_handle: token.user.handle, dirty: true }
```
ErikSchierboom, line 8: This marks the records that need to be updated with a `dirty` flag. Nice,

app/commands/user/reputation_period/search.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      results = @rows.order(reputation: :desc).
+        select(:user_id).
+        page(page).
+        per(self.class.requests_per_page).
+        without_count
+
+      # TODO: Cache this with the expiry key as the id of the last added reputation token
+      # and the other keys as the composite parts of the query
+      # Unless we specify a user_handle, in which case don't bother
```
ErikSchierboom, line 37: 👍

app/commands/user/reputation_period/search.rb
```diff
@@ -0,0 +1,84 @@
...
+      user_ids = results.map(&:user_id)
+      users = User.where(id: user_ids).
+        order(Arel.sql("FIND_IN_SET(id, '#{user_ids.join(',')}')")).to_a
+
+      Kaminari.paginate_array(users, total_count: total_count).
+        page(page).per(self.class.requests_per_page)
+    end
+
+    # This uses a little Rails magic to check the period
+    # is valid and if not, defualt to 0 (the general one).
```
ErikSchierboom:
```suggestion
    # is valid and if not, default to 0 (the general one).
```

app/commands/user/reputation_period/search.rb
```diff
@@ -0,0 +1,84 @@
...
+    # It breaks without the to_s for nil as nil is coverted to NULL
```
ErikSchierboom:
```suggestion
    # It breaks without the to_s for nil as nil is converted to NULL
```

app/commands/user/reputation_period/search.rb
```diff
@@ -0,0 +1,84 @@
...
+    def filter_period!
+      @rows = @rows.where(period: period.to_s)
+    end
+
+    # This uses a little Rails magic to check the category
+    # is valid and if not, defualt to 0 (the general one).
```
ErikSchierboom:
```suggestion
    # is valid and if not, default to 0 (the general one).
```

app/commands/user/reputation_period/search.rb
```diff
@@ -0,0 +1,84 @@
...
+    # It breaks without the to_s for nil as nil is coverted to NULL
```
ErikSchierboom:
```suggestion
    # It breaks without the to_s for nil as nil is converted to NULL
```

app/commands/user/reputation_period/search.rb
```diff
@@ -0,0 +1,84 @@
...
+    def filter_category!
+      @rows = @rows.where(category: category.to_s)
+    end
+
+    # We set track_ids to be 0 rather than null.
+    # This is because mysql can't have unique indexes that contain NULL values.
```
ErikSchierboom, line 63: It can't? I thought I could. Hmmm.

app/commands/user/reputation_period/update_reputation.rb
```diff
@@ -0,0 +1,50 @@
...
+      # Update the period row. We do this all in one SQL command
+      # to avoid any race conditions caused by a new token being added.
+      update_sql = @tokens.select("SUM(value)").to_sql
+      User::ReputationPeriod.where(id: period.id).
+        update_all("dirty = false, reputation = IFNULL((#{update_sql}), 0)")
+
+      # Delete any rows that have been taken down to zero
+      # Ensure we guard for {dirty: false} here in case something
+      # else is making it dirty again. This saves us having to do locking.
+      User::ReputationPeriod.where(id: period.id, reputation: 0, dirty: false).delete_all
```
ErikSchierboom, line 22: Nice to guard against the possibility that `dirty` would be set to `true` after the previous query finished.

app/commands/user/reputation_token/calculate_contextual_data.rb
```diff
@@ -0,0 +1,84 @@
...
+      @user_ids = Array(user_ids)
+      @track_id = track_id
+      @category = category
+
+      if earned_since
+        @earned_since = earned_since
+      elsif period
+        case period.to_sym
+        when :week
+          @earned_since = Time.zone.today - 6.days
```
ErikSchierboom, line 18: I wonder if doing it the same way it is done in the above class reads a tiny bit nicer:

```suggestion
          @earned_since = Time.zone.today - 1.week + 1.day
```

app/controllers/contributing/contributors_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Contributing
+  class ContributorsController < ApplicationController
+    skip_before_action :authenticate_user!, only: %i[index]
+
+    def index
+      response.set_header('Link', '<https://exercism.io/profiles>; rel="canonical"')
+
+      @featured_contributor = User.first
+      @latest_contributor = User.second
```
ErikSchierboom, line 10: Should this have a TODO?

app/controllers/contributing/contributors_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+        user_handle: params[:user_handle],
+        page: params[:page]
+      )
+      contextual_data = User::ReputationToken::CalculateContextualData.(
+        users.map(&:id),
+        period: params[:period],
+        category: params[:category],
+        track_id: track_id
+      )
+      @contributors = SerializeContributors.(users, starting_rank: 1, contextual_data: contextual_data)
```
ErikSchierboom, line 26: What value is returned by this method. `@contributors`? If so, why assign it only to return it directly afterwards?

app/models/user/reputation_token.rb
ErikSchierboom: Okay. Want me to add that to this PR?

ErikSchierboom (approved):

app/commands/user/reputation_period/search.rb
iHiD: It can have them, but it treats each row as different, because in MySQL, `NULL != NULL`

app/commands/user/reputation_period/update_reputation.rb
iHiD: Thanks :)

app/commands/user/reputation_token/calculate_contextual_data.rb
```diff
@@ -0,0 +1,84 @@
...
+        when :month
+          @earned_since = Time.zone.today - 29.days
+        when :year
+          @earned_since = Time.zone.today - 364.days
```
iHiD, line 22:
```suggestion
          @earned_since = Time.zone.today - 1.year + 1.day
```

app/controllers/contributing/contributors_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Contributing
+  class ContributorsController < ApplicationController
+    skip_before_action :authenticate_user!, only: %i[index]
+
+    def index
+      response.set_header('Link', '<https://exercism.io/profiles>; rel="canonical"')
+
+      @featured_contributor = User.first
```
iHiD, line 9:
```suggestion
      # TODO: Set these correctly
      @featured_contributor = User.first
```

app/controllers/contributing/contributors_controller.rb
iHiD: `@contributors` is used in the view to render the table. Once it's been converted to React this whole bit can be removed.

iHiD: (Instance variables set in controllers are available in views)

app/models/user/reputation_token.rb
iHiD: I think a second PR.

app/commands/user/reputation_token/calculate_contextual_data.rb
iHiD: I'm actually going to go the other way, purely because I can't actually use `1.month` as the value isn't 30, which is what I expected. But thank you for the nudge on consistency.

PR #1036, opened 2021-05-28
Add tippy to <RunTestsButton />

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/148.

@iHiD Is this all we need? It was really easy to set up!


iHiD:
@kntsoriano I simplified it even more! :)

I also added a `LazyTippy` version for anywhere where we don't want to do the AJAX call until it's mounted.

PR #1037, opened 2021-05-28
Fix back button in mentoring queue

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/253.

### DIFF
--- a/app/javascript/hooks/use-history.ts
+++ b/app/javascript/hooks/use-history.ts
@@ -35,7 +35,6 @@ export function useHistory<TParams>({ pushOn }: { pushOn: TParams }): void {
   }, [pushOn])
 
   useEffect(() => {
-    console.log(history)
     if (!isMounted.current) {
       isMounted.current = true
       return
--- a/app/javascript/hooks/use-history.ts
+++ b/app/javascript/hooks/use-history.ts
@@ -1,8 +1,8 @@
-import { useRef, useEffect } from 'react'
+import { useRef, useEffect, useState } from 'react'
 import { decamelizeKeys } from 'humps'
 import { stringify } from 'qs'
 
-function removeEmpty<TParams>(obj: TParams) {
+function removeEmpty<TParams>(obj: TParams): TParams {
   return Object.entries(obj)
     .filter(([_, v]) => {
       if (typeof v === 'string') {
@@ -11,13 +11,13 @@ function removeEmpty<TParams>(obj: TParams) {
         return v !== null
       }
     })
-    .reduce((acc, [k, v]) => ({ ...acc, [k]: v }), {})
+    .reduce((acc, [k, v]) => ({ ...acc, [k]: v }), {}) as TParams
 }
 
 function toQuery<TParams>(params: TParams): string {
-  const state = removeEmpty(params)
-
-  return `?${stringify(decamelizeKeys(state), { arrayFormat: 'brackets' })}`
+  return `?${stringify(decamelizeKeys(params as Record<string, unknown>), {
+    arrayFormat: 'brackets',
+  })}`
 }
 
 function pushState<TParams>(params: TParams) {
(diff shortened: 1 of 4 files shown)


app/javascript/hooks/use-history.ts
```diff
@@ -11,28 +11,36 @@ function removeEmpty<TParams>(obj: TParams) {
...
   const isMounted = useRef(false)
 
   useEffect(() => {
+    const newHistory = removeEmpty(pushOn)
+
+    setHistory(newHistory)
+  }, [pushOn])
+
+  useEffect(() => {
+    console.log(history)
```
ErikSchierboom: Was the console.log intentional?

```suggestion
```

app/javascript/hooks/use-history.ts
```diff
@@ -11,28 +11,36 @@ function removeEmpty<TParams>(obj: TParams) {
...
     if (!isMounted.current) {
       isMounted.current = true
       return
     }
 
-    pushState<TParams>(pushOn)
-  }, [pushOn])
+    pushState<TParams>(history)
+  }, [JSON.stringify(history)])
```
ErikSchierboom: @SleeplessByte Is there any downside to using a function as the dependency?

SleeplessByte: You're not using a function as the dependency.

What are you trying to achieve?

iHiD:
(Loom FYI: https://www.loom.com/share/8c96381e77254e65ad5e5b614506abe5?focus_title=1&muted=1&from_recorder=1)

app/javascript/hooks/use-history.ts
```diff
@@ -11,28 +11,36 @@ function removeEmpty<TParams>(obj: TParams) {
         return v !== null
       }
     })
-    .reduce((acc, [k, v]) => ({ ...acc, [k]: v }), {})
+    .reduce((acc, [k, v]) => ({ ...acc, [k]: v }), {}) as TParams
```
SleeplessByte:
```suggestion
    .reduce<TParams>((acc, [k, v]) => ({ ...acc, [k]: v }), {})
```

This doesn't cast it, but only stops inferring the type. Using this will still show you errors if it's not correct (can't be narrowed), whereas the cast shows that only if it has no overlap (so no narrows and no widens).

iHiD (approved):
@kntsoriano Can you check if the comments need actioning before merging pls :)

app/javascript/hooks/use-history.ts
```diff
@@ -11,28 +11,36 @@ function removeEmpty<TParams>(obj: TParams) {
...
 }
 
 function toQuery<TParams>(params: TParams): string {
```
SleeplessByte:
```suggestion
function toQuery<TParams extends Record<string, unknown>>(params: TParams): string {
```

I think this is what you meant?

app/javascript/hooks/use-history.ts
```diff
@@ -11,28 +11,36 @@ function removeEmpty<TParams>(obj: TParams) {
...
-  const state = removeEmpty(params)
-
-  return `?${stringify(decamelizeKeys(state), { arrayFormat: 'brackets' })}`
+  return `?${stringify(decamelizeKeys(params as Record<string, unknown>), {
+    arrayFormat: 'brackets',
+  })}`
 }
 
 function pushState<TParams>(params: TParams) {
```
SleeplessByte:
```suggestion
function pushState<TParams extends Record<string, unknown>>(params: TParams) {
```

app/javascript/hooks/use-history.ts
```diff
@@ -11,28 +11,36 @@ function removeEmpty<TParams>(obj: TParams) {
...
   history.pushState(history.state, '', toQuery(params))
 }
 
 export function useHistory<TParams>({ pushOn }: { pushOn: TParams }): void {
```
SleeplessByte:
```suggestion
export function useHistory<TParams extends Record<string, unknown>>({ pushOn }: { pushOn: TParams }): void {
```

app/javascript/hooks/use-history.ts
```diff
@@ -11,28 +11,36 @@ function removeEmpty<TParams>(obj: TParams) {
...
+  const [history, setHistory] = useState(removeEmpty(pushOn))
```
SleeplessByte: I don't quite understand what this is supposed to do. Can you tell me why there's this extra internal state?

app/javascript/hooks/use-history.ts
kntsoriano: Oh. Good catch!

app/javascript/hooks/use-history.ts
kntsoriano: Thanks. I've rewritten the code and removed it.

PR #1050, opened 2021-06-01
Send notification when org-wide file has changed

ErikSchierboom:
CC @SaschaMann


app/commands/webhooks/process_push_update.rb
```diff
@@ -15,12 +15,23 @@ def call
...
     private
     def pushed_to_main?
       ref == "refs/heads/#{Git::Repository::MAIN_BRANCH_REF}"
     end
+
+    def org_wide_file_changed?(track)
+      return false unless track
+
+      commits.to_a.any? do |commit|
+        commit[:added].concat(commit[:removed], commit[:modified]).any? do |file|
```
iHiD: `concat`. Posh.

ErikSchierboom: Oh, that's just my functional background :D

app/jobs/notify_track_org_wide_file_change_job.rb
```diff
@@ -0,0 +1,13 @@
+class NotifyTrackOrgWideFileChangeJob < ApplicationJob
```
iHiD:
```suggestion
class NotifyTrackSyncerAboutTrackChangesJob < ApplicationJob
```

app/commands/webhooks/process_push_update.rb
iHiD: It's nice. As purely an FYI, I'd have done `Set.new([*commit[:added], *commit[:removed], *commit[:modified]])` probably.

app/jobs/notify_track_syncer_about_track_changes_job.rb
```diff
@@ -0,0 +1,13 @@
...
+  queue_as :default
+
+  def perform(track)
+    body = {
+      event_type: 'track_changes',
+      client_payload: {
+        tracks: ["exercism/#{track.slug}"]
+      }
+    }.to_json
+    Exercism.octokit_client.post("https://api.github.com/repos/exercism/org-wide-files/dispatches", body)
```
iHiD: The only one "downside" of this is that it means the Sidekiq server needs internet access, which it doesn't currently have (and I'd rather it didn't). I wonder we could move into a command rather than a job and run it inline?

iHiD (approved):
I think this is all good now!

SaschaMann:
I don't know any Ruby but it looks correct :)

PR #1059, opened 2021-06-02
Allow searching issues

ErikSchierboom:


app/commands/github/issue/search.rb
```diff
@@ -0,0 +1,75 @@
...
+  class Issue::Search
+    include Mandate
+
+    # Use class method rather than constant for
+    # easier stubbing during testing
+    def self.requests_per_page
+      20
+    end
+
+    def initialize(action: nil, knowledge: nil, module_: nil, size: nil, type: nil, repo: nil, order: nil, page: 1)
```
ErikSchierboom: The `module` keyword is reserved. What would your preferred name be?

app/commands/github/issue/search.rb
```diff
@@ -0,0 +1,75 @@
...
+      paginate!
+    end
+
+    private
+    attr_reader :action, :knowledge, :module_, :size, :type, :repo, :order, :page, :issues
+
+    def filter_repo!
+      return if repo.blank?
+
+      @issues = @issues.where(repo: repo)
```
ErikSchierboom: Issues are not linked to a track, but are instead linked to a repo (e.g. `exercism/ruby`). I could replace the `repo` argument with a `track` parameter, but with the current setup we can filter on issues for other repos to (like configlet).

app/commands/github/issue/search.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+
+    def filter_unclaimed!
+      @issues = @issues.without_label('x:status/claimed')
+    end
+
+    %w[action knowledge module_ size type].each do |label|
+      normalized_label = label.delete('_')
```
ErikSchierboom: This is needed for the `_` suffix on the `module_` value. This can be removed if we decide on another name.

app/commands/github/issue/search.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      define_method "filter_#{normalized_label}!" do
+        return if send(label).blank?
+
+        @issues = @issues.with_label(Github::IssueLabel.send("for_#{normalized_label}", send(label)))
```
ErikSchierboom: `Github::IssueLabel` contains methods to convert a symbol into its fully namespace tag. So `Github::IssueLabel::for_action(:proofread)` returns `x:action/proofread`. I was unsure about how to name these methods though. Any ideas?

app/models/github/issue.rb
```diff
@@ -7,6 +7,22 @@ class Github::Issue < ApplicationRecord
     class_name: "Github::IssueLabel",
     foreign_key: "github_issue_id"
 
+  scope :with_label, lambda { |label|
```
ErikSchierboom: The generated SQL looked good to me. Is this the right approach?

app/models/github/issue_label.rb
```diff
@@ -3,4 +3,20 @@ class Github::IssueLabel < ApplicationRecord
     inverse_of: :labels,
     foreign_key: "github_issue_id",
     class_name: "Github::Issue"
+
+  %w[action knowledge module size status type].each do |label|
+    define_singleton_method "for_#{label}" do |val|
+      return unless const_get(label.pluralize.upcase.to_sym).include?(val)
+
+      "x:#{label}/#{val.to_s.tr('_', '-')}"
```
ErikSchierboom: The `_` to `-` translation is due to symbols not supporting dashes.

app/commands/github/issue/search.rb
iHiD: `area`?

app/commands/github/issue/search.rb
iHiD: We need some wider functionality here, because if someone searches for "Ruby" I want to include *all* the ruby repo_urls (for main repo, tooling, etc). So I think we're better off tagging each **issue** with a `track_id` and then filtering on that.

I think we can **also** filter on `repo` but I think that should be an additional, seperate option to searching by track_id. Also, I think the param (coming in at least) should be `repo_url`.

app/models/github/issue_label.rb
iHiD: Symbols do support dashes but you have to use a special syntax: `:"foo-bar"`

Where are the symbols here?

app/models/github/issue_label.rb
```diff
@@ -3,4 +3,20 @@ class Github::IssueLabel < ApplicationRecord
...
+    end
+  end
```
iHiD, line 27: I think this all gets easier if this is one method `for_label(type, value)` rather than lots of individual methods. The consumer will find this hard to work with as you've seen in the command. The constants below could then (potentially) be one hash, rather than multiple constants (I've not read through the whole PR yet, so that might not be a good idea)

app/models/github/issue.rb
iHiD: Maybe 🤷 

Using `WHERE IN` instead of joins is often fine, because MySQL optimises. But you may well find you have a wider search space as you'll get a temporary table for each inner github_labels that then joins. It's definitely easier for the `not.in` bit below.

So yeah, I'd go with this and then we just check it in production once we're deployed. There's not really a huge amount of data in all this so it should be fine.

app/models/github/issue_label.rb
ErikSchierboom: Yep, much better. I think I've left my go a bit overboard with the metaprogramming stuff :)

app/commands/github/issue_label/create_or_update.rb
```diff
@@ -6,7 +6,7 @@ class CreateOrUpdate
       initialize_with :issue, :label
 
       def call
-        ::Github::IssueLabel.create_or_find_by!(issue: issue, label: label)
+        ::Github::IssueLabel.create_or_find_by!(issue: issue, name: label)
```
ErikSchierboom, line 9: I've decided to rename the column to match the github property name and it is a little clearer. Label.label is weird

app/commands/github/task/search.rb
```diff
@@ -0,0 +1,85 @@
...
+      @tasks = @tasks.where(area: areas) if areas.present?
+    end
+
+    def filter_sizes!
+      @tasks = @tasks.where(size: sizes) if sizes.present?
+    end
+
+    def filter_types!
+      @tasks = @tasks.where(type: types) if types.present?
+    end
```
ErikSchierboom: I _could_ do metaprogramming, but it found that it was actually hard to read than just having these plain methods.

app/models/github/issue.rb
```diff
@@ -10,4 +12,8 @@ class Github::Issue < ApplicationRecord
   def status
     super.to_sym
   end
+
+  def github_url
```
ErikSchierboom, line 14: Convenience method to (re-)construct the issue url

app/models/github/task.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  belongs_to :track, optional: true
+
+  enum action: {
+    create: 0,
+    fix: 1,
+    improve: 2,
+    proofread: 3,
+    sync: 4
+  }, _suffix: true
```
ErikSchierboom, line 12: I added the suffix as rails would otherwise complain about the "create" method already being defined

app/models/track.rb
```diff
@@ -13,6 +13,9 @@ class Track < ApplicationRecord
   has_many :concept_exercises # rubocop:disable Rails/HasManyOrHasOneDependent
   has_many :practice_exercises # rubocop:disable Rails/HasManyOrHasOneDependent
 
+  # TODO: Pre-launch: remove dependent: :destroy
+  has_many :tasks, class_name: "Github::Task", dependent: :destroy
```
ErikSchierboom, line 17: The dependent bit was required according to Rails. I tried it without and got an error.

app/serializers/serialize_github_task.rb
```diff
@@ -0,0 +1,22 @@
...
+  def call
+    {
+      title: task.title,
+      track: task.track.present? ? {
+        slug: task.track.slug,
+        title: task.track.title,
+        icon_url: task.track.icon_url
+      } : nil,
+      opened_by_username: task.opened_by_username,
+      opened_at: task.opened_at.iso8601,
```
ErikSchierboom: The design showed these two fields so I added them here too

db/migrate/20210603065037_create_github_tasks.rb
```diff
@@ -0,0 +1,21 @@
...
+      t.string :repo, null: false
+      t.string :issue_url, null: false, index: { unique: true }
+      t.string :opened_by_username, null: true
+      t.datetime :opened_at, null: false
+
+      t.column :action, :tinyint, null: true, default: 0
+      t.column :knowledge, :tinyint, null: true, default: 0
+      t.column :area, :tinyint, null: true, default: 0
+      t.column :size, :tinyint, null: true, default: 0
+      t.column :type, :tinyint, null: true, default: 0
```
ErikSchierboom, line 14: We _might_ want to add an index to speed this up, although there won't be that many records so the benefit might not that great. But I can look at the queries later

app/commands/github/task/create_or_update.rb
```diff
@@ -0,0 +1,37 @@
...
+          title: attributes[:title],
+          repo: attributes[:repo],
+          opened_at: attributes[:opened_at],
+          opened_by_username: attributes[:opened_by_username],
+          action: attributes[:action],
+          knowledge: attributes[:knowledge],
+          area: attributes[:area],
+          size: attributes[:size],
+          type: attributes[:type]
+        )
```
iHiD: I might make a helper for this pattern that encapsulates doing this. We have it in a few places and it's annoying to have this duplication but there's no easy way to avoid it that I know of.

app/commands/github/task/destroy.rb
```diff
@@ -0,0 +1,14 @@
+module Github
+  class Task
+    class Destroy
+      include Mandate
+
+      initialize_with :issue_url
+
+      def call
+        task = ::Github::Task.find_by(issue_url: issue_url)
+        task&.destroy
```
iHiD:
```suggestion
        ::Github::Task.find_by(issue_url: issue_url)&.destroy
```

iHiD: You could also do `::Github::Task.where(issue_url: issue_url).destroy_all` if you want to avoid the null check. It's effectively the same as we have a unique index on the column. Whichever you prefer is fine by me.

app/commands/github/task/search.rb
iHiD: Yeah, I'd probably reduce this to just one method. I think the multiple methods don't add much when they're all identical, and also probably not worth metaprogramming a new method for each either.

So maybe:
```ruby
    def filter!
      %i[track_id repo_url actions knowledge areas sizes types].each do |filter|
        value = send(filter)
        @tasks = @tasks.where(filter => value) if value.present?
      end
    end
```

app/commands/github/task/search.rb
```diff
@@ -0,0 +1,85 @@
...
+      filter_track!
+      filter_repo!
+      filter_actions!
+      filter_knowledge!
+      filter_areas!
+      filter_sizes!
+      filter_types!
+
+      sort!
+      paginate!
```
iHiD: I think either inline pagination here, or return `@tasks` at the end. The reason is that we might end up having an option not to paginate in search (for example, when we want to see how many issues are open for a filter and want to use search but without the pagination). Then we might do a premature return from paginate, and then end up returning nil from the `call` method. So it's better to be explicit in the return here I think.

```suggestion
      paginate!
      
      @tasks
```

app/commands/github/task/sync_task.rb
```diff
@@ -0,0 +1,50 @@
...
+      include Mandate
+
+      initialize_with :issue
+
+      def call
+        if closed? || claimed?
+          destroy_task
+        else
+          create_or_update_task
+        end
```
iHiD: I prefer early return guards. Anything to avoid more indentation :)

I also like "bang" methods for anything that takes action in commands.

```suggestion
        return destroy_task! if closed? || claimed?
        
        create_or_update_task!
```

app/commands/github/task/sync_task.rb
```diff
@@ -0,0 +1,50 @@
...
+      private
+      def closed?
+        issue.status == :closed
+      end
+
+      def claimed?
+        issue.labels.any? { |label| label.name == Github::IssueLabel.for_type(:status, :claimed) }
+      end
+
+      def create_or_update_task
```
iHiD:
```suggestion
      def create_or_update_task!
```

app/commands/github/task/sync_task.rb
```diff
@@ -0,0 +1,50 @@
...
+          opened_by_username: issue.opened_by_username,
+          action: find_label_of_type(:action),
+          knowledge: find_label_of_type(:knowledge),
+          area: find_label_of_type(:module),
+          size: find_label_of_type(:size),
+          type: find_label_of_type(:type)
+        )
+      end
+
+      def destroy_task
```
iHiD:
```suggestion
      def destroy_task!
```

app/commands/github/task/create_or_update.rb
```diff
@@ -0,0 +1,37 @@
...
+          t.title = attributes[:title]
+          t.repo = attributes[:repo]
+          t.opened_at = attributes[:opened_at]
+          t.opened_by_username = attributes[:opened_by_username]
+          t.action = attributes[:action]
+          t.knowledge = attributes[:knowledge]
+          t.area = attributes[:area]
+          t.size = attributes[:size]
+          t.type = attributes[:type]
+        end
```
iHiD, line 11: I think you can just do `t.attributes = attributes` here, as the LHS and RHS are identical.

app/commands/github/task/sync_task.rb
```diff
@@ -0,0 +1,50 @@
...
+        end
+      end
+
+      private
+      def closed?
+        issue.status == :closed
+      end
+
+      def claimed?
+        issue.labels.any? { |label| label.name == Github::IssueLabel.for_type(:status, :claimed) }
```
iHiD: If I'm understanding this correctly this is going to load all the labels into memory then check each for the name.

I think maybe SQL would be better to avoid the extract weight of query and object allocation:
```ruby
issues.labels.where(name: Github::IssueLabel.for_type(:status, :claimed)).exists?
```

However, because we also use `issue.labels` again below, it might be better to have what you have here and memoize `issue.labels` into a method for `labels` in this command? So your code is fine, but the `find_label_of_type` method is expensive as it'll rebuild the `issue.labels` each time.

app/commands/github/task/sync_task.rb
```diff
@@ -0,0 +1,50 @@
...
+          type: find_label_of_type(:type)
+        )
+      end
+
+      def destroy_task
+        Github::Task::Destroy.(issue.github_url)
+      end
+
+      def find_label_of_type(type)
+        label = issue.labels.find { |l| l.of_type?(type) }
```
iHiD: When you memoize this, do it as `issues.labels.to_a`. I'm surprised the code below works as `find` is an AR method, as well as an array method. I guess AR checks to see if there's a block and if so uses the array version. But I found it a little confusing because of this.

```suggestion
        label = issue.labels.find { |l| l.of_type?(type) }
```

You can also merge the two lines in this method.

app/commands/github/task/sync_tasks.rb
```diff
@@ -0,0 +1,22 @@
...
+        rescue StandardError => e
+          Rails.logger.error "Error syncing task for issue #{issue.repo}/#{issue.number}: #{e}"
+        end
+
+        nil
+      end
+
+      private
+      def issues
+        Github::Issue.left_joins(:labels)
```
iHiD:
```suggestion
        Github::Issue.includes(:labels)
```

If you're just looking to avoid `n+1s` as opposed to using the join in SQL, you're better using `includes` than `joins`, as it simplifies the SQL but has the same n+1 guarding. (It might be `include` - I can never remember)

app/models/github/issue_label.rb
```diff
@@ -3,4 +3,37 @@ class Github::IssueLabel < ApplicationRecord
     inverse_of: :labels,
     foreign_key: "github_issue_id",
     class_name: "Github::Issue"
+
+  def self.type_prefix(type)
+    "x:#{type}"
```
iHiD:
```suggestion
    "x:#{type}/"
```

app/models/github/issue_label.rb
```diff
@@ -3,4 +3,37 @@ class Github::IssueLabel < ApplicationRecord
...
+  end
+
+  def self.for_type(type, val)
+    return unless TYPES.include?(type)
+    return unless TYPES[type].include?(val)
+
+    "#{type_prefix(type)}/#{val}"
```
iHiD:
```suggestion
    "#{type_prefix(type)}#{val}"
```

app/models/github/issue_label.rb
iHiD: I think it's marginally safer to include the `/` here, in case (stupidly) one label's name becomes a subset of the other. It's probably a meaningless change in reality, but just gives a tiny bit more safety in `of_type?` below.

app/models/github/issue_label.rb
```diff
@@ -3,4 +3,37 @@ class Github::IssueLabel < ApplicationRecord
...
+  end
+
+  def of_type?(type)
+    name.start_with?(Github::IssueLabel.type_prefix(type))
+  end
+
+  def prefix
```
iHiD: May as well memoize this and `value` below. Save us a little bit of allocation in the commands where we repeatidly iterate through the label names/values.

app/models/github/task.rb
```diff
@@ -0,0 +1,62 @@
...
+        gsub(%r{/issues/\d+}, '').
+        gsub(/-(test-runner|analyzer|representer)$/, '')
+
+      self.track_id = Track.where(repo_url: repo_url).pick(:id)
+    end
+  end
+
+  %i[action knowledge area size type].each do |type|
+    define_method type do
+      super().to_sym
```
iHiD:
```suggestion
      super.to_sym
```

iHiD: Unless this doesn't work for some reason I don't know?

app/models/track.rb
iHiD: Rails or Rubocop?

db/migrate/20210603065037_create_github_tasks.rb
iHiD: Sure :)

app/models/github/task.rb
ErikSchierboom: It doesn't: https://stackoverflow.com/questions/2572322/implicit-argument-passing-of-super-from-method-defined-by-define-method-is-not

app/commands/github/task/destroy.rb
ErikSchierboom: I like that.

app/commands/github/task/create_or_update.rb
ErikSchierboom: I've updated. It looks like you can just pass the hash to `update!`, which makes it very little code anyways

app/commands/github/task/search.rb
ErikSchierboom: Agreed. I was just copying what other commands did :D

app/commands/github/task/sync_tasks.rb
ErikSchierboom: Yep, works well

app/models/track.rb
ErikSchierboom: I don't remember :D

app/assemblers/assemble_tasks.rb
```diff
@@ -0,0 +1,45 @@
...
+  memoize
+  def unscoped_total
+    ::Github::Task::Search.(
+      actions: params[:actions],
+      knowledge: params[:knowledge],
+      areas: params[:areas],
+      sizes: params[:sizes],
+      types: params[:types],
+      repo_url: params[:repo_url],
+      track_id: params[:track_id],
```
iHiD: Could you DRY this up into a helper method pls, so we don't accidently add it to one and not the other.

iHiD: Hmm, I'm confused about this in general actually.

test/controllers/api/contributors_controller_test.rb
```diff
@@ -1,7 +1,7 @@
 require_relative './base_test_case'
 
 module API
-  class CommunitySolutionsControllerTest < API::BaseTestCase
+  class ContributorsControllerTest < API::BaseTestCase
```
iHiD, line 4: Thanks :)

PR #1062, opened 2021-06-02
Add Assemblers

iHiD:
@ErikSchierboom @kntsoriano New concept here for you to familaisie yourself with: `Processors`.

_Firstly, I spent ages trying to choose a less shit name, but naming is hard 🙂_

The purpose of these processors is to allow us to respond to things in exactly the same way regardless of whether the request is coming via the API or via a normal http request. There are two reasons for this:
1. I dislike the pattern of "load the page, then AJAX in the data" as it means you need two server requests to get data (which is nearly always going to be slower for the user) and also the content isn't just there for SEO - it relies on Google waiting for us to load in everything, which is unreliable at best.
2. It means we can continue the work Karlo started in updating URLs when we change options in React (e.g. select a dropdown and it adds a query string), and that the pages then work exactly the same whether you hit that URL via React or via the website (ie if you refresh a page that has query params, you don't get a broken state).

So these Processors take the responsibility of parsing/using the params and then serializing the data into the correct format. They are designed to be called as per the Contributors list example - once in the API, and once in the React helper. 

I've also done the mentor-requests one but it doesn't yet work in React as the initialData isn't honoured. I think possibly because of the debounce logic. @kntsoriano can you look at this please and add a commit to this to make it work. 

When moving forward, let's use this pattern everywhere. Older things that don't currently have this pattern (e.g. Tracks) can be migrated over time.

### DIFF
--- a/app/assemblers/assemble_contributors.rb
+++ b/app/assemblers/assemble_contributors.rb
@@ -1,4 +1,4 @@
-class ProcessContributors
+class AssembleContributors
   include Mandate
 
   initialize_with :params
--- a/app/assemblers/assemble_mentor_requests.rb
+++ b/app/assemblers/assemble_mentor_requests.rb
@@ -1,4 +1,4 @@
-class ProcessMentorRequests
+class AssembleMentorRequests
   include Mandate
 
   initialize_with :params, :current_user
--- a/app/controllers/api/contributors_controller.rb
+++ b/app/controllers/api/contributors_controller.rb
@@ -4,7 +4,7 @@ module API
     before_action :authenticate_user
 
     def index
-      render json: ProcessContributors.(params)
+      render json: AssembleContributors.(params)
     end
   end
 end
--- a/app/controllers/api/contributors_controller.rb
+++ b/app/controllers/api/contributors_controller.rb
@@ -4,29 +4,7 @@ module API
     before_action :authenticate_user
 
     def index
-      track_id = Track.find(params[:track]).id if params[:track].present?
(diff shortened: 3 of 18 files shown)


kntsoriano:
@iHiD I like this! It makes sense to extract it right now, as we have this pattern done so much in the codebase. I think what slightly confuses me, though, is the difference between a Processor and a Serializer. I have it vaguely in my head, but can't seem to put it into words. Maybe you could talk a bit more about that here.

ErikSchierboom (approved):
Looks good. The only thing I'm not sure of is the name. Maybe instead of process `Fetch`?

iHiD:
I've renamed to "Assemblers". Both definitions of the word come together nicely for us here:

Screenshot 2021-06-03 at 11 33 09

@kntsoriano The difference between these and serializers as as follows:
- Serializers are responsible for **taking a data set** and efficiently **parsing it into a standardised form**.
- Assemblers are responsible for **taking a query**, **filtering it** and **returning the data to the user** in the right format.

Assemblers rely on Commands to do the work of filtering and Serializers to do the work of putting the data in the right format. In situations where the filtering is simple (one command) then we may not need an Sssembler, and can just use the Command followed by the Serializer, but for situations where there are multiple Commands daisy-chained together, Assemblers dry up the duplication.

We *could* have called out to a serializers from within the command, but I felt that that would messy things further and having a new thing responsible for the knowledge about how a set of things should be filtered and serialized felt cleaner.

That make any more sense?

PR #1072, opened 2021-06-03
Set stale time to 30 seconds

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/256.

Currently, `react-query` refetches the data every time the query mounts because the stale time is set to 0. I've set it to 30 seconds for now. I'm unsure of what a good value is

A stale time of 30 seconds means that `react-query` won't refetch if the query mounts again, in the case of the video above, when toggling through pages, if we have the data in our cache. After 30 seconds, `react-query` will refetch if ever the query mounts again.

### DIFF
--- a/app/helpers/react_components/mentoring/inbox.rb
+++ b/app/helpers/react_components/mentoring/inbox.rb
@@ -35,14 +35,16 @@ module ReactComponents
             criteria: params[:criteria],
             page: params[:page] ? params[:page].to_i : 1,
             track: params[:track]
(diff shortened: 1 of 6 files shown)


iHiD:
@kntsoriano Is this configurable? So I'd like this to be the default, but for mentor-inbox and mentor-discussions I think it should stay at 0.

kntsoriano:
@iHiD Yep. We can do that. I'll add it to this PR.

PR #1074, opened 2021-06-03
Add git syncing modal

kntsoriano:
https://github.com/exercism/v3-project-management/issues/258


app/commands/git/generate_diff_between_exercise_versions.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      diffs = []
+      current = nil
+
+      raw_diff.lines.each do |line|
+        if line.starts_with?("diff --git")
+          diffs << current if current
+
+          current = Diff.new(
+            line.split(" b/#{new_git.dir}/").last.strip,
```
iHiD: This probably needs to be `new_git.dir` or `old_git.dir`

iHiD: I might calculate this later actually. Seems safer.

iHiD:
@kntsoriano Great work. Here's the next part: https://www.loom.com/share/76d36a72f1c648eb9be507b0bcf367c7?focus_title=1&muted=1&from_recorder=1

Here's the command I mentioned in the video:
```ruby
reload!;s = Solution.for('ihid', 'csharp', 'lucians-luscious-lasagna');s.update(git_sha: 'ad34b6a023ff79e301173d392feae58896f26263');Git::GenerateDiffBetweenExerciseVersions.(s.exercise, s.git_slug, s.git_sha)
```

yarn.lock
```diff
@@ -6531,10 +6546,15 @@ hex-color-regex@^1.1.0:
   resolved "https://registry.yarnpkg.com/hex-color-regex/-/hex-color-regex-1.1.0.tgz"
   integrity sha512-l9sfDFsuqtOqKDsQdqrMRk0U85RZc0RtOR9yPI7mRVOa4FsR/BVnZ0shmQRM96Ji99kYZP/7hn1cedc1+ApsTQ==
 
-highlight.js@^10.5.0:
-  version "10.5.0"
-  resolved "https://registry.yarnpkg.com/highlight.js/-/highlight.js-10.5.0.tgz"
-  integrity sha512-xTmvd9HiIHR6L53TMC7TKolEj65zG1XU+Onr8oi86mYa+nLcIbxTTWkpW7CsEwv/vK7u1zb8alZIMLDqqN6KTw==
+highlight.js@10.7.2:
```
kntsoriano, line 6549: @iHiD This is still in the lockfile, but we're not using this. We're using highlight JS 11 as the syntax highlighter for our diffs. I can't seem to remove this from the lockfile even though I specify `--ignore-optional` when adding the `diff2html` library.

kntsoriano:
Screen Shot 2021-06-04 at 4 37 24 PM

There is no way to hide the portion in the red rectangle. I've tried all the options possible. Maybe we can just hide it via CSS?

yarn.lock
iHiD: @joshgoebel Help us oh wise one :)

iHiD: @SleeplessByte Any ideas too?

iHiD:
@kntsoriano I'm going to merge this. Really great work. We can work out how to remove v10 of HighlightJs with some help from @SleeplessByte and @joshgoebel in a separate PR :)

yarn.lock
joshgoebel: If the version specifiers of both libraries can't be resolved to a single version then you're going to be stuck with two (that's my understanding) since each library will use it's own version.  The best thing to do is to get the upstream to upgrade to version 11.  Perhaps @SleeplessByte has some better ideas though?

iHiD: Thanks, Josh. It seems weird that it's marked as an "optional dependency" but that the `--ignore-optional` flag doesn't do anything.

joshgoebel: > It seems weird that it's marked as an "optional dependency" 

Well it seems to allow injecting your own highlighter instance (which would make bundling it optional).  One might bring their own custom packaged version, or even a similar library that matched our API, etc... or even just a pre-configured object with many settings/plugins, etc.

iHiD: Yeah. I mean it's an optional dependency in the `yarn.lock` though. Look here: https://github.com/exercism/website/pull/1074/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR5158-R5159

joshgoebel: Yeah, that's a question for yarn people. :-) I don't know that one.

SleeplessByte: It should have been a _peer_ dependency.

Anyway, this can be solves using yarns _resolutions_. See: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

You'll want to target `diff2html/highlight.js` For example: `"diff2html/highlight.js": "./node_modules/highlight.js"`

PR #1083, opened 2021-06-04
Fix tasks

ErikSchierboom:


app/commands/github/task/sync_task.rb
```diff
@@ -20,6 +21,16 @@ def claimed?
...
       end
 
+      def labeled?
+        types = %i[action knowledge module size type]
+
+        labels.any? do |label|
+          types.any? do |type|
+            label.of_type?(type)
+          end
+        end
```
ErikSchierboom, line 31: I could use `.product` here, but I don't know which one you prefer.

iHiD: Once we validate the label values we'll also need to check whether it actually has correct label values, and if not destroy it.

iHiD: In fact, I think we can **only** do this after the labels have been resynced, else we're analyzing off old data at this stage.

app/models/github/issue_label.rb
```diff
@@ -10,20 +10,48 @@ def self.for_type(type, val)
...
 
-  def of_type?(type)
-    name.start_with?(Github::IssueLabel.type_prefix(type))
+  def of_type?(type_to_check)
+    return unless value && type && namespace
+
+    namespace == :exercism &&
+      type == type_to_check &&
+      TYPES.include?(type_to_check) &&
+      TYPES[type_to_check].include?(value)
```
iHiD:
```suggestion
      TYPES[type_to_check]&.include?(value)
```

app/models/github/issue_label.rb
```diff
@@ -10,20 +10,48 @@ def self.for_type(type, val)
     return unless TYPES.include?(type)
     return unless TYPES[type].include?(val)
 
-    "#{type_prefix(type)}#{val}"
+    "x:#{type}/#{val}"
   end
 
-  def of_type?(type)
-    name.start_with?(Github::IssueLabel.type_prefix(type))
+  def of_type?(type_to_check)
```
iHiD, line 16: I think there's two methods getting pulled together here:
- `of_type?`
- `valid_for_type?`

The second can always call the first, but I think seperating them will make this more readable.

app/models/github/issue_label.rb
```diff
@@ -10,20 +10,48 @@ def self.for_type(type, val)
...
+  def namespace
+    return unless components
+
+    case components[:namespace]
+    when 'x'
+      :exercism
+    else
+      :unknown
+    end
+  end
```
iHiD, line 48: And this could probably be `exercism_namespace?`which is just `components&.fetch(:namespace) == :exercism`. As we don't use `:unknown`, we probably don't need to have it here.

PR #1086, opened 2021-06-07
Add walkthrough

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/260.


iHiD:
@kntsoriano Thanks. I'll add styling today. I've made this as well: https://github.com/exercism/website/issues/1088

test/system/flows/student_views_walkthrough_test.rb
```diff
@@ -0,0 +1,43 @@
...
+require_relative "../../support/capybara_helpers"
+
+module Flows
+  class StudentViewsWalkthroughTest < ApplicationSystemTestCase
+    include CapybaraHelpers
+
+    test "student views walkthrough" do
+      use_capybara_host do
+        sign_in!
+        visit standalone_temp_walkthrough_url
```
iHiD: @kntsoriano These should be in the testing namespace, not the temp namespace.

app/javascript/components/modals/CLIWalkthroughModal.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { CLIWalkthrough, GraphicalIcon } from '../common'
+import { Modal, ModalProps } from './Modal'
+
+export const CLIWalkthroughModal = ({
+  html,
+  ...props
+}: Omit<ModalProps, 'className'> & { html: string }): JSX.Element => {
```
iHiD, line 8: @kntsoriano The `className` should always be specified within the modal itself (see L10) rather than passed into the Modal. So when you add new modals, can you follow this pattern pls.

PR #1087, opened 2021-06-07
Add submit keyboard shortcut

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/261.

This overwrites the default search and replace keyboard shortcut of CodeMirror, though.


iHiD:
Cool. I think that's ok that ours take priority, but we might want to think carefully about what they are.

PR #1097, opened 2021-06-07
Add badge modal

iHiD:
Screenshot 2021-06-08 at 00 50 00


ErikSchierboom (approved):
That looks fantastic!

iHiD:
@kntsoriano Nice work 🙂 

So the modal should also open if you click on the badge once it's been revealed, but the modal will have slightly different text. If you could add the modal to the other buttons too with a flag passed into the modal, I'll do any changes to the text etc.

PR #1099, opened 2021-06-08
Set earned on date for PR-related reputation tokens

ErikSchierboom:
- Stored closed at for pull requests
- Working on earned on for authors
- Set earned on date for PR-related reputation tokens

### DIFF
--- a/app/javascript/components/journey/Contribution.tsx
+++ b/app/javascript/components/journey/Contribution.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { fromNow } from '../../utils/time'
+import { fromNow } from '../../utils/date'
 import { GraphicalIcon, TrackIcon, Reputation } from '../common'
 import { Contribution as ContributionProps } from '../types'
 
--- a/app/javascript/components/journey/Contribution.tsx
+++ b/app/javascript/components/journey/Contribution.tsx
@@ -9,7 +9,7 @@ export const Contribution = ({
   iconUrl,
   internalUrl,
   externalUrl,
-  awardedAt,
+  earnedOn,
   track,
 }: ContributionProps): JSX.Element => {
   const url = internalUrl || externalUrl
@@ -39,7 +39,7 @@ export const Contribution = ({
           ) : (
             <div className="generic">Generic</div>
           )}
-          <time dateTime={awardedAt}>{fromNow(awardedAt)}</time>
+          <time dateTime={earnedOn}>{fromNow(earnedOn)}</time>
         </div>
       </div>
       <Reputation value={`+ ${value}`} type="primary" />
--- a/app/models/user/reputation_tokens/code_contribution_token.rb
+++ b/app/models/user/reputation_tokens/code_contribution_token.rb
@@ -1,11 +1,13 @@
 class User::ReputationTokens::CodeContributionToken < User::ReputationToken
-  params :repo, :pr_node_id, :pr_number, :pr_title
+  params :repo, :pr_node_id, :pr_number, :pr_title, :merged_at
   category :building
   reason :contributed_code
   levels %i[tiny small medium large massive]
   values({ tiny: 3, small: 5, medium: 12, large: 30, massive: 100 })
 
   before_validation on: :create do
+    self.earned_on = self.merged_at unless earned_on
+
     unless track
       normalized_repo = repo.gsub(/-(test-runner|analyzer|representer)$/, '')
       self.track_id = Track.where(repo_url: "https://github.com/#{normalized_repo}").pick(:id)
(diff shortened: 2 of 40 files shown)


ErikSchierboom:
@iHiD Should the `earned_on` date be shown instead of the `awarded_at` value on the https://exercism.lol/profiles/ErikSchierboom/contributions page?

iHiD:
Yes it should :)

app/javascript/components/journey/Contribution.tsx
```diff
@@ -39,7 +39,7 @@ export const Contribution = ({
           ) : (
             <div className="generic">Generic</div>
           )}
-          <time dateTime={awardedAt}>{fromNow(awardedAt)}</time>
+          <time dateTime={earnedOn}>{fromNow(earnedOn)}</time>
```
iHiD, line 42: Can you double check this works on now that it's being passed a date, not a datetime pls. Like, it doesn't think that the date is midnight or something.

(I made `earned_on` a date, because it makes the searching faster, which was all it was used for at the time)

app/models/user/reputation_tokens/code_contribution_token.rb
```diff
@@ -1,11 +1,13 @@
 class User::ReputationTokens::CodeContributionToken < User::ReputationToken
-  params :repo, :pr_node_id, :pr_number, :pr_title
+  params :repo, :pr_node_id, :pr_number, :pr_title, :merged_at
   category :building
   reason :contributed_code
   levels %i[minor regular major]
   values({ minor: 5, regular: 12, major: 30 })
 
   before_validation on: :create do
+    self.earned_on = self.merged_at unless earned_on
```
iHiD, line 9: The `unless earned_on` will stop us from updating these records. Probably not an issue moving forward but something to be aware of if you do this then rerun the script to try and get the right data in. Probably best left as you have it but I just wanted to raise it.

app/models/user/reputation_tokens/exercise_contribution_token.rb
```diff
@@ -7,6 +7,7 @@ class User::ReputationTokens::ExerciseContributionToken < User::ReputationToken
   before_validation on: :create do
     self.track = contributorship.exercise.track unless track
     self.exercise = contributorship.exercise unless exercise
+    self.earned_on = self.exercise.updated_at unless earned_on
```
iHiD: 1. Not `created_at`?
2. Why do you need `self.`?

app/models/user/reputation_tokens/mentored_token.rb
```diff
@@ -6,6 +6,10 @@ class User::ReputationTokens::MentoredToken < User::ReputationToken
   reason :mentored
   value 5
 
+  before_validation on: :create do
+    self.earned_on = self.discussion.finished_at unless earned_on
```
iHiD, line 10: We might need to think about this, as reputation is actually probably awarded when the testimonial is left. But this is fine for now.

iHiD (approved):
Nice work. It's nice that you didn't have to rip stuff up to do this - was all nicely additive. A couple of comments to check before you merge, but then feel free 🙂

app/models/user/reputation_tokens/mentored_token.rb
ErikSchierboom: Aha. Good to come back at.

app/models/user/reputation_tokens/code_contribution_token.rb
ErikSchierboom: Yeah I considered it to. The reasoning was that `merged_at` shouldn't ever change.

app/models/user/reputation_tokens/exercise_contribution_token.rb
ErikSchierboom: 1. Updated
2. I don't. Removed.

At some point we should probably improve this a bit more to have the earned on data be set to the date at which the new contributor was synced.

app/javascript/components/journey/Contribution.tsx
ErikSchierboom: So `earned_on` being a date has the result that if you set it to the current time, it will display "X hours ago", where X is the number of hours today. It would be nicer if it would say "today", "yesterday", "x days ago", "x weeks ago", "x months ago" and "x years ago". Is that something you'd like me to look at?

iHiD: Yes please :)

iHiD: Maybe make a new `fromNowDate()` so we can use that explicitly?

ErikSchierboom: Will do

iHiD (approved):

PR #1103, opened 2021-06-08
Use updated size labels for reputation

ErikSchierboom:
This PR changes the reputation awarding for PRs to not look at the `reputation/code_contribution/{major,minor}` labels, but to instead use the `x:size/{tiny/small/medium/large/massive}` labels.

The reputation for merging a PR is unaffected as we didn't use the existing size labels. We currently have `values({ janitorial: 1, reviewal: 5 })` there. Do we also want to differentiate between sizes for merges?

I've made this a draft as merging this will remove reputation temporarily, until I've run a script to convert the labels.

### DIFF
--- a/app/commands/user/reputation_token/award_for_pull_request_author.rb
+++ b/app/commands/user/reputation_token/award_for_pull_request_author.rb
@@ -40,10 +40,10 @@ class User
       end
 
       def reputation_level
-        return :major if params[:labels].include?('reputation/contributed_code/major')
-        return :minor if params[:labels].include?('reputation/contributed_code/minor')
-
-        :regular
+        # Sort descendingly to award greatest possible reputation
+        %i[massive large medium small tiny].find do |type|
+          params[:labels].include?(Github::IssueLabel.for_type(:size, type))
+        end || :medium
       end
     end
   end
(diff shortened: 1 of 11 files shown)


app/commands/user/reputation_token/award_for_pull_request_author.rb
```diff
@@ -40,10 +40,10 @@ def has_author?
       end
 
       def reputation_level
-        return :major if params[:labels].include?('reputation/contributed_code/major')
-        return :minor if params[:labels].include?('reputation/contributed_code/minor')
-
-        :regular
+        # Sort descendingly to award greatest possible reputation
```
iHiD, line 43: Nice.

app/models/user/reputation_tokens/code_review_token.rb
```diff
@@ -2,8 +2,8 @@ class User::ReputationTokens::CodeReviewToken < User::ReputationToken
   params :repo, :pr_node_id, :pr_number, :pr_title
   category :maintaining
   reason :reviewed_code
-  levels %i[minor regular major]
-  values({ minor: 2, regular: 5, major: 10 })
+  levels %i[tiny small medium large massive]
+  values({ tiny: 1, small: 2, medium: 5, large: 10, massive: 20 })
```
iHiD, line 6: I really need to refactor this duplication out at some point :)

iHiD (approved):

app/models/user/reputation_tokens/code_review_token.rb
ErikSchierboom: :)

PR #1106, opened 2021-06-08
Add email sending on notifications

iHiD:


app/models/user.rb
```diff
@@ -51,6 +52,10 @@ class User < ApplicationRecord
     self.name = self.handle if self.name.blank?
   end
 
+  after_create_commit do
+    create_communication_preferences
```
ErikSchierboom, line 56: This ensures that after the record is created, a communication preferences record is also created?

test/commands/user/notification/send_email_test.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  test "updates email status" do
+    notification = create :notification, :unread
+    refute notification.email_sent?
+    User::Notification::SendEmail.(notification)
+    assert notification.email_sent?
+  end
+
+  test "does not sent if preference set to false" do
```
ErikSchierboom:
```suggestion
  test "does not send if preference set to false" do
```

ErikSchierboom (approved):

app/models/user.rb
iHiD: Correct, yes.

PR #1109, opened 2021-06-10
Implement tasks

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/267.

@iHiD Needs some styling left. I'm pretty sure you'll find some bugs as well! Let me know. I can also add the history function when we get this merged.

I've also got to a good point where I can make `<TrackSwitcher />` a common component. I will refactor once we get this merged.


app/assemblers/assemble_track_switcher.rb
```diff
@@ -0,0 +1,28 @@
+class AssembleTrackSwitcher
```
kntsoriano: This is for assembling the track switcher parameters.

ErikSchierboom (approved):
Great!

iHiD:
@kntsoriano Really great work with this. There's a couple of demarcations I'm not quite happy with around the track-selector still, but we can improve that in a separate PR, as you said. It'd be worth us discussing before you do that.

If you can do the ff, we can merge :)
- The "All" button has changed to be a checkbox row, and when clicked then closed it doesn't update correctly. There's also a related React warning.
- Rename `ExercismSelect` to `SingleSelect`
- Rename `ExercismMultipleSelect` to `MultipleSelect`
- Get it green

app/assemblers/assemble_track_switcher.rb
iHiD: I prefer these things to return an array, which we can then assign to a hash key, rather than returning a hash. I've pushed a commit with an update :)

PR #1120, opened 2021-06-14
End mentoring via keyboard shortcut

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/262.


iHiD:
Nice :)

PR #1121, opened 2021-06-14
Add Site Updates

iHiD:
This adds site updates, renders them on tracks, and adds a very basic admin page for them. The admin UI is really temporary (thus no tests) and the HAML will end up as React eventually, but the rest should be solid.

This currently only has support for New Exercises and New Concepts, but there will be other types too.

They look like this:

Screenshot 2021-06-14 at 21 21 11

### DIFF
--- a/app/controllers/maintaining/site_updates_controller.rb
+++ b/app/controllers/maintaining/site_updates_controller.rb
@@ -1,0 +1,23 @@
+class Maintaining::SiteUpdatesController < ApplicationController
+  def index
+    @updates = SiteUpdate.order(published_at: :desc).page(params[:page]).per(20)
+  end
+
+  def edit
+    @update = SiteUpdate.find(params[:id])
+  end
+
+  def update
+    @update = SiteUpdate.find(params[:id])
+
+    if @update.editable_by?(current_user)
+      @update.update!(
+        params.require(:site_update).permit(
+          :title, :description, :pull_request_number
+        ).merge(author: current_user)
+      )
+    end
+
+    redirect_to action: :index
+  end
+end
--- a/app/controllers/maintaining/site_updates_controller.rb
+++ b/app/controllers/maintaining/site_updates_controller.rb
@@ -1,6 +1,6 @@
 class Maintaining::SiteUpdatesController < ApplicationController
   def index
-    @updates = SiteUpdate.order(published_at: :desc).page(params[:page]).per(20)
+    @updates = SiteUpdate.sorted.page(params[:page]).per(20)
   end
 
   def edit
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -31,6 +31,7 @@ class TracksController < ApplicationController
       end
 
       @recent_solutions = UserTrack::RetrieveRecentlyActiveSolutions.(@user_track)
+      @updates = SiteUpdate.published.where(track: @track).order(published_at: :desc).limit(10)
 
       render "tracks/show/joined"
     else
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -31,7 +31,7 @@ class TracksController < ApplicationController
       end
 
       @recent_solutions = UserTrack::RetrieveRecentlyActiveSolutions.(@user_track)
-      @updates = SiteUpdate.published.where(track: @track).order(published_at: :desc).limit(10)
+      @updates = SiteUpdate.published.for_track(@track).sorted.limit(10)
 
       render "tracks/show/joined"
     else
--- a/app/models/concerns/is_paramaterised_sti.rb
+++ b/app/models/concerns/is_paramaterised_sti.rb
@@ -68,12 +68,11 @@ module IsParamaterisedSTI
 
     cattr_accessor :class_suffix, :i18n_category
 
-    belongs_to :user
     belongs_to :track, optional: true
     belongs_to :exercise, optional: true
 
     before_create do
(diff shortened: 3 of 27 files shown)


app/controllers/tracks_controller.rb
```diff
@@ -31,6 +31,7 @@ def show
       end
 
       @recent_solutions = UserTrack::RetrieveRecentlyActiveSolutions.(@user_track)
+      @updates = SiteUpdate.published.where(track: @track).order(published_at: :desc).limit(10)
```
iHiD: This will end up a React Helper via an assembler eventually.

app/controllers/maintaining/site_updates_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  def index
+    @updates = SiteUpdate.sorted.page(params[:page]).per(20)
+  end
+
+  def edit
+    @update = SiteUpdate.find(params[:id])
+  end
+
+  def update
+    @update = SiteUpdate.find(params[:id])
```
ErikSchierboom, line 11: Can `find` return `nil`? (also applies to the `edit` method)

app/models/concerns/is_paramaterised_sti.rb
```diff
@@ -68,12 +68,11 @@ module IsParamaterisedSTI
 
     cattr_accessor :class_suffix, :i18n_category
 
-    belongs_to :user
     belongs_to :track, optional: true
     belongs_to :exercise, optional: true
 
     before_create do
-      self.uniqueness_key = "#{user_id}|#{type_key}|#{guard_params}"
+      self.uniqueness_key = generate_uniqueness_key!
```
ErikSchierboom, line 75: Better!

app/models/site_update.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  scope :published, -> { where('published_at < ?', Time.current) }
+  scope :for_track, ->(track) { where(track: track) }
+  scope :sorted, -> { order(published_at: :desc) }
+
+  belongs_to :author, optional: true, class_name: "User"
+  belongs_to :pull_request, optional: true, class_name: "Github::PullRequest"
+
+  before_validation only: :create do
+    self.published_at = Time.current + 3.hours unless published_at
```
ErikSchierboom, line 16: The default is for an update to be published 3 hours later?

app/models/site_update.rb
```diff
@@ -0,0 +1,77 @@
...
+      }
+    end
+
+    if pull_request
+      d[:pull_request] = {
+        title: pull_request.title,
+        number: pull_request.number,
+        url: "https://github.com/#{pull_request.repo}/pull/#{pull_request.number}",
+        merged_by: pull_request.merged_by_username,
+        merged_at: pull_request.updated_at.iso8601 # TODO: Use merged_at?
```
ErikSchierboom, line 62: Note: `merged_at` is not guaranteed to be set. You could do `merged_at || closed_at`  which should give data, although it might not at the moment due to older PRs being imported without the closed_at date.

app/models/site_updates/new_concept_update.rb
```diff
@@ -0,0 +1,44 @@
...
+    )
+  end
+
+  def icon_type
+    :concept
+  end
+
+  # TODO: This is pretty gross
+  def icon_url
+    concept.name[0, 2]
```
ErikSchierboom, line 28: Eh, just the first three characters? How does that work?

Should we do something like with exercises:

```ruby
def icon_name
  config[:icon] || exercise_slug.to_s
end
```

(that's defined it `Git::Exercise` btw)

ErikSchierboom (approved):

app/controllers/maintaining/site_updates_controller.rb
iHiD: No, `find` raises an `ActiveRecord::NotFound` exception if it can't find the record, which causes the page to 404.

app/models/site_update.rb
iHiD: Yeah, so my thinking here is that we give maintainers 3 hours to add a fuller version of the post before it gets published. You think that's a sensible timeframe?

iHiD: (They can still do it afterwards, but this gives them space to get the announcement right when the first people see it)

app/models/site_update.rb
iHiD: I didn't find either property on the model. Do I need to dig into the PR data to get those properties or did I miss something?

app/models/site_updates/new_concept_update.rb
iHiD: Concepts don't have actual icons, the first two chars just get rendered in HTML and decorated in a box.

So this gives us the data for this:

Screenshot 2021-06-15 at 12 44 28

I feel like my comment here is very accurate 😉

app/controllers/maintaining/site_updates_controller.rb
ErikSchierboom: Okay cool

app/models/site_update.rb
ErikSchierboom: I think that's reasonable. Publishing would imply that it's "done" anyways.

app/models/site_update.rb
ErikSchierboom: You do

app/models/site_updates/new_concept_update.rb
ErikSchierboom: It is :)

PR #1134, opened 2021-06-16
Unify track select

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/274.

@iHiD I kept the css class to `c-track-switcher` still. I think it'll be easier for you to fix the CSS if you're the one who will change it.

Also, I've decided to keep the mentoring queue's track switcher as its own component. I'm still thinking of whether to generalize it or not as it includes an extra option to show a modal.


iHiD:
@kntsoriano The `format` method on the `assembler` should be a serializer (as it takes something and transforms it into a hash for rendering). So I've refactored this out. Take a look at the commit and see if that makes sense.

iHiD:
@kntsoriano Nice work with this :)

- I've added some styling.
- I think you missed the one in `app/javascript/components/mentoring/queue/TrackFilterList.tsx`. I've styled this correctly now (using the new styles). If you don't want to move it to the new format for some reason, that's ok, but if we could it'd be nice. I'd say that's a lower priority and it's not blocking though. We could merge this and do that later.
- Tests need fixing.

iHiD:
@kntsoriano I'm going to merge this. Could you add an issue to do the `app/javascript/components/mentoring/queue/TrackFilterList.tsx` at some point in the future pls?

PR #1135, opened 2021-06-16
Add maintainers

ErikSchierboom:
- Add models for contributor teams
- Create commands to create contributor team and membership


app/commands/contributor_team/add_reviewer.rb
```diff
@@ -0,0 +1,14 @@
...
+  class AddReviewer
+    include Mandate
+
+    initialize_with :user, :attributes
+
+    def call
+      team = ContributorTeam.find_by!(track: nil, type: :reviewers)
+      Github::Team::AddMember.(team.github_name, user.github_username)
+      ContributorTeam::Membership::CreateOrUpdate.(user, team, attributes)
+      user.update(roles: user.roles + [:reviewer]) unless user.roles.include?(:reviewer)
```
iHiD: Does `roles` return symbols or strings?

app/commands/contributor_team/add_track_maintainer.rb
```diff
@@ -0,0 +1,27 @@
...
+  class AddTrackMaintainer
+    include Mandate
+
+    initialize_with :user, :track, :attributes
+
+    def call
+      team = ContributorTeam.find_by!(track: track, type: :track_maintainers)
+      ContributorTeam::Membership::CreateOrUpdate.(user, team, attributes)
+
+      Github::Team::AddMember.(team.github_name, user.github_username)
```
iHiD: This should probably move inside the `ContributorTeam::Membership::CreateOrUpdate`

app/commands/contributor_team/membership/create_or_update.rb
```diff
@@ -0,0 +1,16 @@
...
+    initialize_with :user, :team, :attributes
+
+    def call
+      membership = ContributorTeam::Membership.create_or_find_by!(user: user, team: team) do |m|
+        m.attributes = attributes
+      end
+
+      membership.update!(attributes)
+      membership
+    end
```
iHiD:
```suggestion
    def call
      membership = ContributorTeam::Membership.create!(user: user, team: team, **attributes)
      # Side effects of creating a new membership (e.g. adding to GH team)
    rescue ActiveRecord::RecordNotUnique
      ContributorTeam::Membership.find_by!(user: user, team: team).tap do |membership|
        membership.update!(attributes)
      end
    end
```

app/models/contributor_team/membership.rb
```diff
@@ -0,0 +1,19 @@
...
+    senior: 2
+  }
+
+  belongs_to :team,
+    class_name: "ContributorTeam",
+    foreign_key: :contributor_team_id,
+    inverse_of: :memberships
+
+  belongs_to :user,
+    inverse_of: :team_memberships
```
iHiD, line 14: This should be `members` not `memberships`

config/routes.rb
```diff
@@ -239,6 +239,7 @@
     resources :exercise_representations
     resources :tracks, only: [:show]
     resources :site_updates, except: [:destroy]
+    resources :teams, except: [:destroy]
```
iHiD: Probably worth naming this `contributor_teams` rather than `teams`, as once we have "real" Exercism teams back in play, this will get confusing. 

_Queue Erik muttering something about C# being better than Ruby 😉_

ErikSchierboom:
@iHiD I was already renaming locally :D

app/controllers/maintaining/contributor_teams/memberships_controller.rb
```diff
@@ -1,2 +1,59 @@
...
+  before_action :use_seniority_options, only: %i[new create edit update]
+
+  def show; end
+
+  def new
+    @membership = ContributorTeam::Membership.new
+  end
+
+  def create
+    @membership = ContributorTeam::Membership.new(membership_params)
```
iHiD: If you do this:

```suggestion
    @membership = @team.memberships.build(membership_params)
```

Then you can remove the setting of the team in the santized params at the bottom of this file.

app/views/maintaining/contributor_teams/memberships/edit.html.haml
```diff
@@ -0,0 +1,15 @@
+#page-team-new
+  .sm-container.container
+    %header
+      %h1 Update #{@membership.user.github_username} contributor in team #{@team.name}
+      %hr.c-divider
+
+    %section.form-section
+      = form_for @membership, url: maintaining_contributor_team_contributor_team_membership_path(@team, @membership), as: :contributor_team_membership do |f|
```
iHiD:
```suggestion
      = form_for @membership, url: maintaining_contributor_team_contributor_team_membership_path(@team, @membership) do |f|
```

I suspect you don't need this. It may change the param name in the controller, but I think it'll make it shorter, not longer.

app/views/maintaining/contributor_teams/memberships/new.html.haml
```diff
@@ -0,0 +1,19 @@
+#page-team-new
+  .sm-container.container
+    %header
+      %h1 Add contributor to team #{@team.name}
+      %hr.c-divider
+
+    %section.form-section
+      = form_with url: maintaining_contributor_team_contributor_team_memberships_path, method: :post, model: @membership do
```
iHiD:
```suggestion
      = form_with url: maintaining_contributor_team_contributor_team_memberships_path, model: @membership do
```

I don't think you need this.

app/views/maintaining/contributor_teams/new.html.haml
```diff
@@ -6,17 +6,17 @@
 
     %section.form-section
       = form_with url: maintaining_contributor_teams_path, method: :post, model: @team do
-        = fields_for @team do |fu|
+        = fields_for @team do |f|
```
iHiD: You shouldn't need this. You can elevate the `|f|` to the line above. The fields for is only if you have nested models in one form.

app/commands/github/organization_member/remove.rb
```diff
@@ -0,0 +1,13 @@
+module Github
+  class OrganizationMember
+    class Remove
```
iHiD: I don't think this is used anywhere yet.

app/models/track.rb
```diff
@@ -41,6 +41,10 @@ def git
     Git::Track.new(synced_to_git_sha, repo_url: repo_url)
   end
 
+  def repo
```
iHiD:
```suggestion
  def repo_name
```

I get confused about whether `repo` is the repo object (which it is in some places), a URL (https://githbub...) or the name (this?). I think we should try and always suffix the url and the name.

app/views/maintaining/contributor_teams/new.html.haml
```diff
@@ -0,0 +1,21 @@
+#page-team-new
+  .sm-container.container
+    %header
+      %h1 Create a new contributor team
+      %hr.c-divider
```
iHiD: Nice!

app/views/maintaining/contributor_teams/new.html.haml
```diff
@@ -0,0 +1,21 @@
...
+      = form_with url: maintaining_contributor_teams_path, model: @team do |f|
+        .field
+          = f.label :name, "Name"
+          = f.text_field :name, placeholder: "The team's name", required: true
+        .field
+          = f.label :github_name, "GitHub Name"
+          = f.text_field :github_name, placeholder: "The team's GitHub name", required: true
+        .field
+          = f.label :type, "Type"
+          = f.select :type, @type_options, required: true
```
iHiD:
```suggestion
          .c-select
            = f.select :type, @type_options, required: true
```

This will look nicer for you.

app/views/maintaining/contributor_teams/show.html.haml
```diff
@@ -0,0 +1,35 @@
+#page-maintaining-track
+  .lg-container.container
+    %h1 Maintaining #{@team.name}
+
+    .data
+      %div
+        GitHub name: #{@team.github_name}
```
iHiD:
```suggestion
      %div GitHub name: #{@team.github_name}
```

(etc in the other places)

app/views/maintaining/dashboard/show.html.haml
```diff
@@ -9,6 +9,11 @@
           = link_to "All Submissions", maintaining_submissions_path
         %li
           = link_to "All Representations", maintaining_exercise_representations_path
+        %li
+          = link_to "Site Updates", maintaining_site_updates_path
+        - if current_user.admin?
+          %li
+            = link_to "Contributor Teams", maintaining_contributor_teams_path
```
iHiD: - We may want all maintainers to be able to see the data, and only admins to be able to edit it.
- If we only want admins to see it, we should move this whole section into an `app/views/admin` namespace.

I'm not sure which is right.

db/migrate/20210616073326_create_contributor_teams.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.belongs_to :contributor_team, null: false, foreign_key: true
+      t.belongs_to :user, null: false, foreign_key: true
+      
+      t.column :visible, :boolean, null: false, default: true
+      t.column :seniority, :tinyint, null: false, default: 0
+
+      t.index [:contributor_team_id, :user_id], name: "index_contributor_team_memberships_on_team_id_and_user_id", unique: true
+
+      t.timestamps
+    end
```
iHiD: It's best to generally split seperate table creates into seperate migrations.

The reason for this is that if the first succeeds, but the second fails, you'll be in a state where you can't re-run the migration, and can't rollback easily either. 

If you want to still use redo/rollback, you can do `rails db:migrate:redo STEP=2`

config/routes.rb
```diff
@@ -239,6 +239,9 @@
     resources :exercise_representations
     resources :tracks, only: [:show]
     resources :site_updates, except: [:destroy]
+    resources :contributor_teams, except: [:destroy] do
+      resources :contributor_team_memberships, controller: "contributor_teams/memberships"
```
iHiD:
```suggestion
      resources :memberships, controller: "contributor_teams/memberships"
```

app/commands/github/organization_member/remove.rb
ErikSchierboom: Not yet, but it will be once I modify the controllers to use the commands.

app/models/track.rb
ErikSchierboom: Good idea

app/views/maintaining/dashboard/show.html.haml
ErikSchierboom: > We may want all maintainers to be able to see the data, and only admins to be able to edit it.

This might be the nicest option. I'll update

db/migrate/20210616073326_create_contributor_teams.rb
ErikSchierboom: Yeah I got bitten by that. Will update.

app/commands/contributor_team/membership/destroy.rb
```diff
@@ -7,6 +7,8 @@ class Destroy
     def call
       ContributorTeam::Membership.where(user: user, team: team).destroy_all
 
+      # TODO: remove member from organization if no team memberships remaining
```
iHiD: :shipit:

app/commands/contributor_team/add_track_maintainer.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def call
+      @team = ContributorTeam.find_by!(track: track, type: :track_maintainers)
+      ContributorTeam::Membership::CreateOrUpdate.(user, team, attributes)
+
+      if well_maintained?
+        github_reviewers_team.remove_from_repository(track.repo_name)
+      else
+        github_reviewers_team.add_to_repository(track.repo_name, :push)
+      end
```
iHiD: This.

app/commands/contributor_team/membership/create_or_update.rb
```diff
@@ -0,0 +1,19 @@
+class ContributorTeam::Membership
+  class CreateOrUpdate
+    include Mandate
+
+    initialize_with :user, :team, :attributes
+
+    def call
+      ContributorTeam::Membership.create!(user: user, team: team, **attributes).tap do |_membership|
```
iHiD:
```suggestion
      ContributorTeam::Membership.create!(user: user, team: team, **attributes).tap do
```

app/commands/contributor_team/membership/destroy.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      user.update(roles: user.roles - [team.role_for_members]) unless remains_in_team_role?
+
+      team.github_team.remove_membership(user.github_username)
+    end
+
+    private
+    memoize
+    def remains_in_team_role?
+      user.teams.any? { |t| t.role_for_members == team.role_for_members }
```
iHiD:
```suggestion
      user.teams.pluck(:role_for_members).include?(team.role_for_members }
```

That will reduce object initialization and data transfer for someone like you who is in lots of teams, etc. (Code untested)

app/models/github/organization.rb
```diff
@@ -0,0 +1,127 @@
...
+  end
+
+  memoize
+  def active?
+    name.present?
+  end
+
+  def remove_membership(_github_username)
+    Exercism.octokit_client.remove_organization_membership(name, username)
+  end
```
iHiD, line 20: Bug here - mismatch between `github_username` and `username`. (Also check why the test missed it)

app/models/github/organization.rb
```diff
@@ -0,0 +1,127 @@
+class Github::Organization
+  include Mandate
```
iHiD, line 2: You might also like to `include Singleton` here and use `.instance` instead of `.new` in other places.

app/models/github/organization.rb
```diff
@@ -0,0 +1,127 @@
...
+        organization(login: "#{name}") {
+          teams(userLogins: ["#{github_username}"]) {
+            totalCount
+          }
+        }
+      }
+    QUERY
+
+    response = Exercism.octokit_client.post("https://api.github.com/graphql", { query: query }.to_json).to_h
+    response.dig(:data, :organization, :teams, :totalCount)
```
iHiD, line 34: These two lines are nice.

app/models/github/organization.rb
```diff
@@ -0,0 +1,127 @@
...
+      handle_rate_limit(response.dig(:data, :rateLimit))
+    end
+  end
+
+  def fetch_member_usernames(cursor)
+    query = <<~QUERY.strip
+      {
+        organization(login: "#{name}") {
+          membersWithRole(first: 100
+                          #{%(, after: "#{cursor}") if cursor}) {
```
iHiD: If rubocop will let you fit these two lines into one, it would read clearer IMO.

app/models/github/organization.rb
```diff
@@ -0,0 +1,127 @@
...
+        }
+      }
+    QUERY
+
+    Exercism.octokit_client.post("https://api.github.com/graphql", { query: query }.to_json).to_h
+  end
+
+  def team_member_usernames
+    cursor = nil
+    users = [].to_set
```
iHiD:
```suggestion
    users = Set.new
```

Unless there's some other reason not to do that?

app/models/github/organization.rb
```diff
@@ -0,0 +1,127 @@
...
+
+    loop do
+      response = fetch_team_member_usernames(cursor)
+
+      response.dig(:data, :organization, :teams, :nodes).each do |team|
+        users |= team.dig(:members, :nodes).pluck(:login)
```
iHiD:
```suggestion
        users.merge(team.dig(:members, :nodes).pluck(:login))
```

`|=` creates a new set everytime, where as `merge` keeps the same set with just extra members added, so it's a bit more efficient.

app/models/github/team.rb
```diff
@@ -0,0 +1,68 @@
...
+      privacy: :closed,
+      parent_team_id: parent_team&.team_id,
+      accept: 'application/vnd.github.hellcat-preview+json')
+  end
+
+  memoize
+  def team_id
+    # The octokit client does not surface the API method to retrieve a team
+    # by its github team name, so we just call it directly
+    Exercism.octokit_client.get("https://api.github.com/orgs/#{organization.name}/teams/#{name}")[:id]
```
iHiD, line 61: Nice.

app/models/track.rb
```diff
@@ -41,6 +41,11 @@ def git
     Git::Track.new(synced_to_git_sha, repo_url: repo_url)
   end
 
+  memoize
+  def repo_name
+    slug
+  end
```
iHiD: You should get probably get this from the `repo_url`, not the `slug`. Sometimes they get mismatched, which could be bad for us here. You might not actually want this on `track` as `repo_url` is already exposed. I don't actually think you use this anywhere unless I'm missing it.

app/commands/contributor_team/membership/destroy.rb
```diff
@@ -15,7 +15,7 @@ def call
     private
     memoize
     def remains_in_team_role?
-      user.teams.any? { |t| t.role_for_members == team.role_for_members }
+      user.teams.map(&:role_for_members).include?(team.role_for_members)
```
iHiD, line 18: Using `map` here is the same as using `any` above in terms of it making all the Team objects. Using `pluck` just retrieves that one column from the database.

The only time this would be preferably if `user.teams` is already loaded elsewhere. Is that why you went for map?

ErikSchierboom:
> Using map here is the same as using any above in terms of it making all the Team objects. Using pluck just retrieves that one column from the database.

I know, but `role_for_members` is _not_ a DB column, hence `pluck` not working. And I liked `map` better than the previous version.

iHiD:
Ah - ok! 🙂

app/commands/webhooks/process_organization_member_update.rb
```diff
@@ -0,0 +1,31 @@
...
+      when 'member_added'
+        add_member!
+      when 'member_removed'
+        remove_member!
+      end
+    end
+
+    private
+    def add_member!
+      ::Github::OrganizationMember.create!(username: user_name)
```
iHiD: - [ ] Set `alumus: false`
- [ ] Also should be `create_or_find_by!

app/jobs/sync_team_permissions_job.rb
```diff
@@ -0,0 +1,14 @@
...
+# against one of the GitHub webhook calls failing, which would result in
+# repo permissions not being set correctly
+class SyncTeamPermissionsJob < ApplicationJob
+  queue_as :dribble
+
+  def perform
+    ContributorTeam.find_each do |team|
+      ContributorTeam::UpdateReviewersTeamPermissions.(team)
+    rescue StandardError => e
+      Rails.logger.error "Error updating reviewers team permissions for team #{team.github_name}: #{e}"
```
iHiD, line 11: Open GH issue here. Internal repo? Dunno.

iHiD: exercism/exercism - and add `@exercism/maintainers-admin`

app/commands/github/organization_member/remove_when_no_team_memberships.rb
```diff
@@ -0,0 +1,20 @@
...
+  class OrganizationMember
+    class RemoveWhenNoTeamMemberships
+      include Mandate
+
+      initialize_with :github_username
+
+      def call
+        return if organization.team_membership_count_for_user(github_username) >= 1
+
+        Exercism.octokit_client.remove_organization_member(organization.name, github_username)
```
iHiD: Use your modal for this - not octokit directly

iHiD:
Needs a rebase. Then we should get this merged I think.

ErikSchierboom:
@iHiD I've rebased

app/views/maintaining/contributor_teams/show.html.haml
```diff
@@ -0,0 +1,44 @@
...
+            %th Name
+            %th Visible
+            %th Seniority
+            %th
+        %tbody
+          - @team.memberships.each do |membership|
+            %tr
+              %td= membership.user.handle
+              %td= membership.visible
+              %td= membership.seniority.to_s.humanize
```
ErikSchierboom, line 24: Add rep for maintaining this track

PR #1147, opened 2021-06-21
Fix pagination typing issue

iHiD:
@SleeplessByte - Your thoughts on this one would be appreciated.

https://www.loom.com/share/e26f255228cd430fa8a4d08cb04b9420

### DIFF
--- a/app/javascript/components/common/Pagination.tsx
+++ b/app/javascript/components/common/Pagination.tsx
@@ -19,6 +19,10 @@ export function Pagination({
     return null
   }
 
+  current = Number(current)
+  around = Number(around)
+  total = Number(total)
+
   const range = createRange(
     Math.max(current - around, 1),
     Math.min(current + around, total)
@@ -64,9 +68,12 @@ export function Pagination({
             >
               1
             </button>
-            <div className="--pagination-more">…</div>
           </>
         ) : null}
+        {current - around > 2 ? (
+          <div className="--pagination-more">…</div>
+        ) : null}
+
         {range.map((page) => {
           return (
             <button
@@ -83,9 +90,13 @@ export function Pagination({
             </button>
           )
         })}
+
+        {total - current > around + 1 ? (
+          <div className="--pagination-more">…</div>
+        ) : null}
+
         {total - current > around ? (
           <>
-            <div className="--pagination-more">…</div>
             <button
               key={total}
               onClick={() => {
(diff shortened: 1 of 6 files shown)


app/javascript/components/common/Pagination.tsx
```diff
@@ -19,6 +19,10 @@ export function Pagination({
     return null
   }
 
+  current = Number(current)
+  around = Number(around)
+  total = Number(total)
```
iHiD, line 24: The bit below this is fine. It's just this bit that is "controversial".

SleeplessByte: What's likely happening is that the URL param "page" isn't actually converted to a number when it's read, and that's what you see as "string value".

The better solution would be, if that's the problem, to type convert that value to a number instead of here.

That said, using Number to ensure its a number is the right thing and you definitely.dont want the parseXXX family because you know that it's a number and it should break when it's not. parseXXX doesn't break and gives you unexpected values instead.

iHiD: > The better solution would be, if that's the problem, to type convert that value to a number instead of here.

Yeah - agreed. What's confusing me is that it is definitely a number in the JSON coming out of Rails and we pretty much just read that here as I understand it. @kntsoriano Could you try and trace it through to see where this is coming from pls. I don't mind leaving this here as a final guard (maybe just "current" and "total") - but it would be good to understand the underlying problem pls.

SleeplessByte: Again I don't think the issue is the JSON, but the URL param override.

Can you check that one?

iHiD: I don't think we read from the URL param. As far as I can tell in the code, we read from the JSON when the page loads.

This seems to be the code that does that.

Screenshot 2021-06-21 at 14 54 27

I don't know how any of this works though, and I may well be missing a pathway that does what you're saying, so @kntsoriano will need to check 🙂

SleeplessByte: (These were all just educated guesses without looking at the code. Might as well be something different, but it's were I would look first)

iHiD:
@kntsoriano Could you also get this green pls :)

kntsoriano:
@iHiD I pushed a fix, but for this specific component only. We might need to type each request's parameters.

PR #1148, opened 2021-06-21
Add settings pages

### LINKED ISSUE #49: No feedback when clicking .c-copy-text-to-clipboard button
```html
<button type="button" class="c-copy-text-to-clipboard " aria-label="Copy &quot;exercism download --exercise=bob --track=wren&quot; to the cliboard"><div class="text">exercism download --exercise=bob --track=wren</div><img src="https://exercism-assets-staging.s3.eu-west-2.amazonaws.com/packs/media/images/icons/clipboard-a14bc00a6ce65fbe85a1f8d6f58e8faa.svg" alt="Copy to clipboard" class="c-icon"></button>
```
Is this actually working? It may be but I can't tell.

Running Chrome Version 91.0.4472.77 (Official Build) (64-bit) on Ubuntu Linux

iHiD:
Fixes https://github.com/exercism/v3-beta/issues/49


app/views/settings/show.html.haml
```diff
@@ -22,35 +22,37 @@
...
+              = f.text_field "pronouns[2]", placeholder: "e.g. their"
+              insight was great. I’d definitely recommend
+              = f.text_field "pronouns[1]", placeholder: "e.g. them"
+              to others!
+
+            .commonly-used
+              .info Use a common option:
+              %button she/her/her
+              %button he/him/his
+              %button they/their/them
```
ErikSchierboom, line 58: I'm not saying you should do this, but I could imagine people taking offence with the order of these options. You could possibly randomize them. 🤷

ErikSchierboom (approved):

app/views/settings/show.html.haml
iHiD: I think if I randomised them, we'd be more likely to get complaints about he/him coming first 1/3 of the time. I take your point though.

ErikSchierboom: Yeah I considered that. Now you'll get men complaining about she/her/her being the first option 😉

iHiD: I can cope with that more :)

PR #1149, opened 2021-06-22
Convert Journey page to React

kntsoriano:


iHiD:
Nice work, dude :)

app/assemblers/assemble_contributions_summary.rb
```diff
@@ -0,0 +1,162 @@
...
+  include ActionView::Helpers::NumberHelper
+
+  initialize_with :user
+
+  def call
+    {
+      tracks: [
+        SerializeTrackForSelect::ALL_TRACK.merge(categories: categories_data),
+        tracks.map { |track| SerializeTrackForSelect.(track).merge(categories: categories_data(track.id)) }
+      ].flatten,
```
iHiD:
```suggestion
        *tracks.map { |track| SerializeTrackForSelect.(track).merge(categories: categories_data(track.id)) }
      ],
```

I prefer this way of doing it. You?

iHiD:
Great start :) I've pushed a branch with a couple of styling tweaks. A few comments:

1. I think I’d rather this first page was just one long component in JS, rather than having the individual sections as helpers. (The individual badges/reputation/etc are fine as individual things - just this overview should all be one thing I think)
2. Where it says "iHiD has xyz reputation" and "See iHiD's contributions ->" - these should be "You have xxx reputation" and "See your contributions ->"

app/javascript/components/profile/contributions-summary/use-chart.ts
```diff
@@ -31,30 +31,38 @@ const padReputation = (reputation: number[]): number[] => {
...
-  options: {
-    aspectRatio: 1,
-    elements: {
-      line: { borderWidth: 3 },
+const generateConfig = (
+  canvas: HTMLCanvasElement
+): ChartConfiguration<'radar'> => {
+  const borderColor = getComputedStyle(canvas).getPropertyValue(
+    '--chartBorderColor'
+  )
```
iHiD, line 39: @kntsoriano Check the commit I added that changed this to make sure you're happy with it pls.

kntsoriano: Looks good.

app/assemblers/assemble_contributions_summary.rb
kntsoriano: Yeah :)

iHiD:
@kntsoriano This bit seems like it still needs doing?

Screenshot 2021-06-30 at 22 19 24

iHiD:
@kntsoriano I've added a commit that does the ff:
- Adds the real smoothed ProgressGraph that @neenjaw made
- Moves the Journey-specific types out of the main types.ts into journey/types.ts - as these types don't have "proper" serializers I don't think they should be in the main types file. Otherwise that types file is huge and difficult to work out what is what. I originally renamed them all to start with Journey but that became too messy.
- Added `totals` to the mentoring details. The logic of adding all the students up to get a total client-side was incorrect as you might mentor the same student on different tracks. I also added the "ranks" to the same bit.
- Added some tests for the assembler.

iHiD:
And another commit that generates the relevant data to power the pretty charts 🙂

PR #1153, opened 2021-06-24
Add medium and hard difficulties

### LINKED ISSUE #57: [Track/Exercises] All Exercises Flagged "Easy"
On  https://exercism.lol/tracks/python/exercises, and subsequently on the detail page https://exercism.lol/tracks/python/exercises/bank-account

This exercise is flagged "Easy", but in `config.json`, it is labelled as `difficulty 8`:

image

_____

Since the track only has one other exercise of higher difficulty (at 9), this exercise should be labeled "challenging" or "hard" or "high".

Here is another from the JS track (https://exercism.lol/tracks/javascript/exercises & https://exercism.lol/tracks/javascript/exercises/crypto-square) This exercise is labeled as "Easy" in the UI, but is `difficulty 9`, in `config.json`:

image

iHiD:
Closes https://github.com/exercism/v3-beta/issues/57


app/javascript/components/types.ts
```diff
@@ -183,7 +183,7 @@ type ExerciseCore = {
   isExternal: boolean
 }
 
-export type ExerciseDifficulty = 'easy'
+export type ExerciseDifficulty = 'easy' | 'medium' | 'hard'
```
ErikSchierboom, line 186: Should this be `DifficultyCategory` to be consistent with the naming used in the Ruby model? I can do a rename factor if you agree.

```suggestion
export type ExerciseDifficulty = 'easy' | 'medium' | 'hard'
```

app/commands/solution/publish.rb
```diff
@@ -27,7 +27,7 @@ def award_reputation!
         solution.user,
         :published_solution,
         solution: solution,
-        level: solution.exercise.difficulty_description
+        level: solution.exercise.difficulty_category
```
ErikSchierboom, line 30: Should level be `difficulty_category` to match the model's naming? The latter is a better name than level.

app/serializers/serialize_exercise.rb
```diff
@@ -13,7 +13,7 @@ def call
       type: exercise.tutorial? ? "tutorial" : exercise.git_type,
       title: exercise.title,
       icon_url: exercise.icon_url,
-      difficulty: "easy", # TOOD
+      difficulty: exercise.difficulty_category,
```
ErikSchierboom, line 16: Should the property be `difficulty_category`?

app/javascript/components/types.ts
iHiD: Yeah - that'd be helpful 🙂  I think it should be `ExerciseDifficultyCategory`.

app/commands/solution/publish.rb
iHiD: I think `level` as a keyword is quite hard coded into the reputation system so I don't think we can change it. Or at least, it's going to be more work to make it customisable per reputation type, so let's leave this one for now.

app/serializers/serialize_exercise.rb
iHiD: Yeah - good shout. It'll also need updating at the JS side (camelcased) too. So either we do that now, or we add a TODO for it for later?

iHiD:
@ErikSchierboom I'm going to merge this and we can tidy up the JS to match when we have chance (if you want to do it as a follow-on that's great - or we do it further down the line)

PR #1160, opened 2021-06-25
Fix bug with incorrect guessing of mentoring session

iHiD:


app/helpers/react_components/mentoring/session.rb
```diff
@@ -51,16 +52,15 @@ def links
...
+          comment_counts = discussion.posts.
             group(:iteration_id, :seen_by_mentor).
             count
         end
 
         solution.iterations.map do |iteration|
           counts = discussion ? comment_counts.select { |(it_id, _), _| it_id == iteration.id } : nil
           num_comments = discussion ? counts.sum(&:second) : 0
-          unread = discussion ? counts.reject { |(_, seen), _| seen }.present? : 0
+          unread = discussion ? counts.reject { |(_, seen), _| seen }.present? : false
```
ErikSchierboom, line 63: Is this the same as:

```suggestion
          unread = discussion && counts.reject { |(_, seen), _| seen }.present?
```

?

ErikSchierboom (approved):

app/helpers/react_components/mentoring/session.rb
iHiD: Good shout but no (I had to try it to realise this myself). If discussion is `nil` then the expression returns `nil` not `false`.

ErikSchierboom: Okay!

PR #1144, opened 2021-06-21
Add reset and delete button

kntsoriano:

PR #1163, opened 2021-06-27
Add settings page

kntsoriano:
https://github.com/exercism/v3-project-management/issues/282


kntsoriano:
@iHiD We just need to add the buttons from https://github.com/exercism/website/pull/1144 here.

iHiD:
@kntsoriano Great. Could you merge that PR into this PR pls. (The Rails-side react components etc can then be deleted as they'll only get called from React). I'll then finish that styling tomorrow and deploy.

kntsoriano:
@iHiD I've merged it, but didn't remove the Rails-side react components. I've structured it in a way where the different components are separate. Let me know if you want them merged into just one component after you've finished styling it. It should be easy.

app/controllers/settings/api_controller.rb
```diff
@@ -0,0 +1,5 @@
+module Settings
+  class APIController < ApplicationController
+    def show; end
+  end
+end
```
iHiD: Can you move this back into the `def api` on settings pls. I don't want a whole extra controller for it. You can rename it to `api_token` if there's a naming conflict.

Add also add `communication_preferences` as an action on `settings`, rather than having a separate controller pls.

iHiD:
@kntsoriano Could you add API controller tests for this pls? Please take a look at the existing tests to work out the approach that we've been using throughout (I can't remember if you've done many of the API controller tests before).

app/controllers/settings/api_controller.rb
iHiD: I've done this.

iHiD:
@kntsoriano I've done everything to get this mergable, except for the fact I've somehow broken the delete account modal. It never sends a request to delete any more. Could you check pls.

kntsoriano:
@iHiD I've fixed it already, but could you check the reset account flow if it's correct? After resetting your account, should you be redirected to the home page?

app/commands/user/reset_account.rb
```diff
@@ -0,0 +1,45 @@
...
+        became_mentor_at: nil
+      )
+    end
+
+    def reset_tracks!
+      user.user_tracks.each do |user_track|
+        UserTrack::Destroy.(user_track)
+      end
+
+      Mentor::Request.where(student_id: user.id).pending.delete_all
```
ErikSchierboom, line 28: Shouldn't all mentoring request be removed, not just the pending ones?

iHiD:
> @iHiD I've fixed it already, but could you check the reset account flow if it's correct? After resetting your account, should you be redirected to the home page?

Yeah - it's correct. It's just messy atm because the homepage isn't a real homepage.

app/controllers/api/settings_controller.rb
```diff
@@ -0,0 +1,39 @@
+module API
+  class SettingsController < BaseController
+    def update
+      permitted = params.require(:user).permit(
+        :name, :location, :bio,
+        pronoun_parts: []
```
ErikSchierboom, line 6: What does this line do? Would it reset the pronoun parts?

iHiD: This is telling Rails to expect `params[:pronoun_parts]` to be an array, and cope with that.

app/commands/user/reset_account.rb
iHiD: It should only delete the pending ones, but it should probably change the user of the historic ones to be ghosts. (We can't delete the old ones without deleting the discussions too)

iHiD: Updated. Thanks.

app/controllers/api/settings_controller.rb
ErikSchierboom: Aha, okay

PR #1167, opened 2021-06-28
Fix dropdown stale time

### LINKED ISSUE #82: [Navigation] Notification dropdowns do not open after using navigation
When I refresh the page, the notification dropdowns work. After I use the top navigation to change pages, they don't open anymore. I have to refresh the page.

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/82.


iHiD:
Thanks!

Could you explain why this fixes it pls? Just so I understand 🙂

kntsoriano:
This happens because we want to refetch every time the we open the dropdown. Otherwise, when we invalidate the query on closing the dropdown, the data gets deleted.

PR #1171, opened 2021-06-29
Reset pages when filtering

### LINKED ISSUE #95: [Mentoring/Queue] `page` query param is not reset when filtering by language or exercise
When I change the filters with the programming language or exercise, the "page" query param doesn't get reset to 1. This makes it look like the filtering is broken because there are no visible results, but the filters have numbers next to them that show how many results there should be.

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/95.

@iHiD I haven't tidied up the journey page as we are planning to change the filters on there in the future.


iHiD:
Great. Thanks :)

PR #1175, opened 2021-06-29
Render backtrace as separate lines

### LINKED ISSUE #99: Stack traces in sync issues are on one line.
Compare the stack trace in https://github.com/exercism/website/issues/1118 (via Bugsnag) to the one in https://github.com/exercism/vbnet/issues/103 (our code).

I think we should be render a string, not an array (so `backtrace.join("\n")` or something)

ErikSchierboom:
See https://github.com/exercism/csharp/issues/1548 for an example of how the backtrace will now be rendered.

Closes https://github.com/exercism/v3-beta/issues/99

### DIFF
--- a/app/commands/github/issue/open_for_sync_failure.rb
+++ b/app/commands/github/issue/open_for_sync_failure.rb
@@ -23,7 +23,7 @@ module Github
           ```
           #{error.message}
 
-          #{error.backtrace}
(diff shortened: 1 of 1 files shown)


iHiD:
Great!

iHiD (approved):

PR #1215, opened 2021-07-06
Standardize params

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/283.

### DIFF
--- a/app/assemblers/assemble_tasks.rb
+++ b/app/assemblers/assemble_tasks.rb
@@ -30,6 +30,6 @@ class AssembleTasks
 
   memoize
   def track_id
-    Track.find_by!(slug: params[:track_slug]).id if params[:track_slug].present?
+    Track.find(params[:track_slug]).id if params[:track_slug].present?
   end
 end
--- a/app/assemblers/assemble_tasks.rb
+++ b/app/assemblers/assemble_tasks.rb
@@ -30,6 +30,6 @@ class AssembleTasks
 
   memoize
   def track_id
-    Track.find(params[:track]).id if params[:track].present?
+    Track.find_by!(slug: params[:track_slug]).id if params[:track_slug].present?
   end
 end
--- a/app/javascript/components/journey/Solution.tsx
+++ b/app/javascript/components/journey/Solution.tsx
@@ -1,10 +1,10 @@
 import React from 'react'
 import { fromNow } from '../../utils/time'
-import { GraphicalIcon, Icon, TrackIcon, ExerciseIcon } from '../common'
+import { GraphicalIcon, TrackIcon, ExerciseIcon } from '../common'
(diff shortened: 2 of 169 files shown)


app/models/user/reputation_token.rb
```diff
@@ -88,7 +88,7 @@ def non_cacheable_rendering_data
 
   def cacheable_rendering_data
     data = {
-      id: uuid,
+      uuid: uuid,
```
kntsoriano, line 91: @iHiD I updated the `cacheable_rendering_data` for this model.

app/assemblers/assemble_tasks.rb
```diff
@@ -30,6 +30,6 @@ def tasks
 
   memoize
   def track_id
-    Track.find(params[:track]).id if params[:track].present?
+    Track.find_by!(slug: params[:track_slug]).id if params[:track_slug].present?
```
iHiD: I think this should remain as `Track.find` rather than `Track.find_by` as track uses friendly-id to handle renamed things.

If you wanted to be explicit that these take a slug, you could change them to `Track.friendly.find(...)` instead

```suggestion
    Track.find(params[:track_slug]).id if params[:track_slug].present?
```

app/models/github/task.rb
```diff
@@ -59,4 +59,8 @@ class Github::Task < ApplicationRecord
       super()&.to_sym
     end
   end
+
```
kntsoriano, line 62: @iHiD I added a `uuid` to this model.

app/models/submission/test_run.rb
```diff
@@ -14,6 +14,7 @@ class Submission::TestRun < ApplicationRecord
     self.message = raw_results[:message] unless self.message
     self.output = raw_results[:output] unless self.output
     self.status = raw_results.fetch(:status, :error) unless self.status
+    self.uuid = SecureRandom.uuid unless self.uuid
```
kntsoriano, line 17: @iHiD I added a `uuid` to this model.

app/assemblers/assemble_tasks.rb
kntsoriano: Cool. I will do this.

app/javascript/components/journey/Solution.tsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react'
 import { fromNow } from '../../utils/time'
-import { GraphicalIcon, Icon, TrackIcon, ExerciseIcon } from '../common'
+import { GraphicalIcon, TrackIcon, ExerciseIcon } from '../common'
```
iHiD, line 3: Did you find this manually or use some tool?

iHiD:
Epic PR. Nice work @kntsoriano!

PR #1217, opened 2021-07-06
Open issue for prerequisite cycles

### LINKED ISSUE #1210: TrackHasCyclicPrerequisiteError in tracks/concepts#index
## Error in v3

**TrackHasCyclicPrerequisiteError** in **tracks/concepts#index**
TrackHasCyclicPrerequisiteError

[View on Bugsnag](https://app.bugsnag.com/exercism/v3/errors/60e36940eb20fe00076401e8?event_id=60e36940007f9561c4320000&i=gh&m=ci)

## Stacktrace

    app/commands/track/determine_concept_map_layout.rb:63 - node_levels
    app/commands/track/determine_concept_map_layout.rb:43 - concept_levels
    app/commands/track/determine_concept_map_layout.rb:19 - call
    app/controllers/tracks/concepts_controller.rb:9 - index

[View full stacktrace](https://app.bugsnag.com/exercism/v3/errors/60e36940eb20fe00076401e8?event_id=60e36940007f9561c4320000&i=gh&m=ci)

ErikSchierboom:
This PR adds functionality to automatically open an issue when a dependency cycle is found in a track's concept map.
For an example, see https://github.com/ErikSchierboom/go/issues/10. 

Let me know what you think of the issue body. I did not include the stacktrack because it wasn't really helpful. Once you're happy with the text, I'll create an issue on the Go repo (which was the repo with the cycles).

I've also refactored a bit to allow sharing the open/reopen issue on GitHub functionality between commands.

Closes https://github.com/exercism/website/issues/1210

### DIFF
--- a/app/commands/github/issue/open.rb
+++ b/app/commands/github/issue/open.rb
@@ -1,0 +1,41 @@
+module Github
+  class Issue
+    class Open
+      include Mandate
+
+      initialize_with :repo, :title, :body
+
+      def call
+        if missing?
+          create_issue
+        elsif closed?
+          reopen_issue
+        end
+      end
+
+      private
+      def missing?
+        issue.nil?
+      end
+
+      def closed?
+        issue.state == "closed"
+      end
+
+      def create_issue
+        Exercism.octokit_client.create_issue(repo, title, body)
+      end
+
+      def reopen_issue
+        Exercism.octokit_client.reopen_issue(repo, issue.number)
+      end
+
+      memoize
+      def issue
+        # TODO: Elevate this into exercism-config gem
+        author = "exercism-bot"
+        Exercism.octokit_client.search_issues("\"#{title}\" is:issue in:title repo:#{repo} author:#{author}")[:items]&.first
+      end
+    end
+  end
+end
--- a/app/commands/track/determine_concept_map_layout.rb
+++ b/app/commands/track/determine_concept_map_layout.rb
@@ -19,6 +19,8 @@ class Track
         levels: concept_levels,
         connections: concept_connections
       }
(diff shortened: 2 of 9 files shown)


app/commands/track/determine_concept_map_layout.rb
```diff
@@ -19,6 +19,8 @@ def call
         levels: concept_levels,
         connections: concept_connections
       }
+    rescue TrackHasCyclicPrerequisiteError
+      Github::Issue::OpenForDependencyCycle.(track)
```
iHiD: This should needs wrapping as a job, as this is happening in-progress for the user.

I think we should also then re-raise the exception (you can use `raise` with no args), and catch and handle it in the UI.

app/commands/github/issue/open.rb
```diff
@@ -0,0 +1,41 @@
...
+      include Mandate
+
+      initialize_with :repo, :title, :body
+
+      def call
+        if missing?
+          create_issue
+        elsif closed?
+          reopen_issue
+        end
```
ErikSchierboom, line 13: There is a risk of duplicate issues being created if this is called in parallel. Do we want to guard against that?

app/jobs/open_issue_for_dependency_cycle_job.rb
```diff
@@ -0,0 +1,7 @@
+class OpenIssueForDependencyCycleJob < ApplicationJob
+  queue_as :default
```
ErikSchierboom, line 2: Is this the right queue?

iHiD: Yeah, it's fine.

PR #1219, opened 2021-07-07
Fix cancel button behavior

### LINKED ISSUE #73: [Mentoring/Discussion] Cancel button in weird place

### LINKED ISSUE #76: [Mentoring/Discussion] "Cancel" discards my whole comment without warning
When I click "cancel", I lose everything I have written in the comment. This might be surprising to some users, and might also annoy those that clicked the button accidentally. That is exactly what happened to me the first time... I would like to propose that clicking this button should produce a warning similarly like on GitHub.

Screen Shot 2021-06-25 at 10 54 57

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/287.
Closes https://github.com/exercism/v3-beta/issues/76.
Closes https://github.com/exercism/v3-beta/issues/73.


app/css/components/markdown-editor.css
```diff
@@ -25,7 +25,12 @@
                 @apply hidden;
             }
         }
+
+        & .CodeMirror-line {
+            visibility: hidden;
+        }
```
iHiD, line 32: What's this do?

kntsoriano: @iHiD If a user types in the textbox and hits cancel, the text they typed goes over the "Add a comment" pseudo element you added. So I just hide the text they typed in when the editor is compressed.

iHiD: Thanks. I'll check this out to understand it better :)

iHiD: > If a user types in the textbox and hits cancel, the text they typed goes over the "Add a comment" pseudo element you added. So I just hide the text they typed in when the editor is compressed.

I don't understand this. Someone can't press cancel until the text box is empty, so how are they able to still have text in the text box and press cancel?

iHiD:
@kntsoriano Have a watch of this loom: https://www.loom.com/share/28c3c2cfa8a64917ae7033833640fe56

This is the summary:

New comment:
- [x] Only one of save/cancel should show

Edit comment:
- [x] Inline, not modal
- [x] Both buttons show and are always enabled.
- [x] Button text is "Update"
- [x] If all text is deleted from the comment, button text changes to "Delete"

app/css/components/markdown-editor.css
kntsoriano: It is because we have localstorage. Refreshing the page has the saved content from localstorage on the compressed editor.

kntsoriano:
@iHiD Should be ready. Took a really long time to do this, but I think I've got it to a point where the code looks good.

iHiD:
Great job @kntsoriano. 

I've added a commit with one change, as I think you misunderstood where I wanted the edit box to be. My version is not very DRY but I don't think it matters.

Finally got there with this - thanks for persisting!

kntsoriano:
I think the way you did it is good! I would've probably done the same.

iHiD:
Great :)

PR #1228, opened 2021-07-08
End-to-end working subscription setup

iHiD:
@kntsoriano https://www.loom.com/share/111faf0850094d49a87bb8deca5e6e7e

---

This is a WIP first-pass of the subscription setup. The flow is:
1. In React, present a form that a user puts card details into [1]
2. When the card is valid according to the JS, we send a request to the server to get a "payment intent". This API call creates a Stripe Customer object if one doesn't exist, a Stripe Subscription and (implicitly), the "payment intent". 
3. We then pass the payment intent's secret back to the browser.
4. The browser then tries to hit Stripe's API with the payment intent secret, and the card details the user has entered.
5. If the card fails, it errors. If it succeeds, it says so.

Try it at http://local.exercism.io:3020/donate


app/javascript/components/stripe/Form.tsx
```diff
@@ -0,0 +1,148 @@
...
+    'subscription'
+  )
+
+  const handleAmountChange = useCallback((e) => {
+    console.log(e)
+    setAmountInDollars(e.target.value)
+    for (let child of e.target.closest('.amounts').children) {
+      child.classList.remove('selected')
+    }
+    e.target.closest('.c-faux-input, button').classList.add('selected')
```
kntsoriano: It's better to handle the class names via React rather than changing the DOM.

app/javascript/components/stripe/StripeForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+type PaymentIntent = {
+  id: string
+  clientSecret: string
+}
+export type PaymentIntentType = 'payment' | 'subscription'
+
+// Make sure to call `loadStripe` outside of a component’s render to avoid
+// recreating the `Stripe` object on every render.
+const stripePromise = loadStripe(
+  'pk_test_51IDGMXEoOT0Jqx0UcoKlkvB7O0VDvFdCBvOCiWiKv6CkSnkZn7IG6cIHuCWg7cegGogYJSy8WsaKzwFHQqN75T7b00d56MtilB'
```
kntsoriano: We can probably put this into a meta tag so that we can store the secret key inside Rails.

app/javascript/components/stripe/StripeForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+  paymentIntentType: PaymentIntentType
+  onSuccess: () => void
+  amountInDollars: number
+}) {
+  const [succeeded, setSucceeded] = useState(false)
+  const [error, setError] = useState<string | undefined>()
+  const [processing, setProcessing] = useState(false)
+  const [cardValid, setCardValid] = useState(false)
+
+  const createPaymentIntentEndpoint = '/api/v2/donations/payment_intents'
```
kntsoriano, line 59: Is there a Stripe JS library instead of sending bare JSON requests? Might be more reliable to do that.

app/javascript/components/stripe/StripeForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+    const endpoint = cancelPaymentIntentEndpoint.replace(
+      '$ID',
+      paymentIntent.id
+    )
+    return fetchJSON(endpoint, {
+      method: 'PATCH',
+    })
+  }, [])
+
+  const getPaymentRequest = useCallback(() => {
```
kntsoriano: Let's use `useMutation` for this one.

app/javascript/components/stripe/StripeForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+    return fetchJSON(createPaymentIntentEndpoint, {
+      method: 'POST',
+      body: JSON.stringify({
+        type: paymentIntentType,
+        amount_in_dollars: amountInDollars,
+      }),
+    }).then((data: any) => data.paymentIntent)
```
kntsoriano: We could type this instead of using any as we know what we want from the API.

app/javascript/components/stripe/Form.tsx
iHiD: Yeah - I figured you'd rewrite this bit :)

app/javascript/components/stripe/StripeForm.tsx
iHiD: Agreed

app/javascript/components/stripe/StripeForm.tsx
iHiD: I tried and failed at that. So yes please :)

iHiD:
@ErikSchierboom Can you check you're happy with the GitHub webhooks setup pls.

ErikSchierboom (commented):
The webhooks changes LGTM

iHiD:
@kntsoriano This is ready for you to work on.

kntsoriano:
@iHiD Should be ready now. We just need to style the buttons on the settings page.

Do we need this, btw?

From [@stripe/stripe-js](https://www.npmjs.com/package/@stripe/stripe-js)
> Note: To be PCI compliant, you must load Stripe.js directly from https://js.stripe.com. You cannot include it in a bundle or host it yourself. This package wraps the global Stripe function provided by the Stripe.js script as an ES module.

iHiD:
> Do we need this, btw?

Yeah - what we have is fine. It's just very confusingly worded! This blog post explains more. https://dev.to/stripe/importing-stripe-js-as-an-es-module-1ni - Thanks for checking though :)

I'll go through the PR now

iHiD:
@kntsoriano https://www.loom.com/share/e1d336b31ed14cb5951714ef9ded6afe

iHiD:
@kntsoriano Merging this. Please deal with the loom in a new PR. Thanks! :)

PR #1230, opened 2021-07-09
Update comment count whenever a user posts

kntsoriano:


iHiD:
@kntsoriano I've added a commit to this which moves the `DiscussionPostProps` into the main types as `MentorDiscussionPost`. The reason was that I'm trying to not have non-common things being imported into common as it makes it harder to split out the packs.

I'll merge this when it's green. Nice work 🙂

PR #1231, opened 2021-07-09
Update org-wide-files repository dispatch

### LINKED ISSUE #37: Consider adding the actor to the repo dispatch event payload

ErikSchierboom:
@SaschaMann and I have been working on re-enabling the org-wide-files repository dispatch event, which is the last step in the org-wide-syncing setup for it to fully work. Without this, any changes made within the repositories (e.g. to an appends file), won't result in any changes as the org-wide-files workflow won't be automatically run.

See https://github.com/exercism/org-wide-files/pull/27

Closes https://github.com/exercism/org-wide-files/issues/37


db/schema.rb
```diff
@@ -52,6 +52,28 @@
     t.index ["type"], name: "index_badges_on_type", unique: true
   end
 
+  create_table "contributor_team_memberships", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
```
SaschaMann: I know nothing about this repo, but these changes seem unrelated

ErikSchierboom: Yeah, I sometimes get that. Removed.

PR #1234, opened 2021-07-09
Exclude wip and deprecated exercises

### LINKED ISSUE #119: [Exercise/Overview] "You've unlocked X exercises" - too many exercises?
I feel like there's something wrong with the logic that counts and displays "You've unlocked X concepts/exercises" on the exercise overview page, but I find it hard to describe what the actual behavior is.

What I would expect form this page is that:
- the count of concepts and exercises in the sentence is always equal to the count of concepts and exercises displayed below the sentence
- the "concepts" that I unlock by doing a specific concept exercise that teaches concepts A & B are a sum of all the concepts taught by concept exercises that: 
	- have concepts A and/or B in prerequisites
	- have all of their other prerequisites already unlocked
- the "exercises" that I unlock by doing a specific concept exercise that teaches concepts A & B:
	- have concepts A and/or B in prerequisites
	- have all of their other prerequisites already unlocked

This means that I would NOT expect to see there exercises that I happen to have available because I did them in v2 or in independent mode of v3, but I didn't actually unlock their prerequisite concepts yet.

Now, here's an example of why I am confused. After completing a concept exercise, I got this modal. The modal shows exactly what I would expect: the next concepts that I just unlocked and their concept exercise. After the modal, I go back to the exercise overview page and I would expect to see exactly the same concepts and exercises that I unlocked. Instead, it says there's 13 unlocked concepts even though it shows only 3, and then 70 unlocked exercises? That makes no sense, on the list of exercises are exercises whose prerequisites I did not unlock yet, eg. strings for "forth" and "phone number".

Screen Shot 2021-07-10 at 16 14 05

Screen Shot 2021-07-10 at 16 14 27

Here another example of a different concept exercise, again the count is wrong and somehow "Bob" shows up even though it also requires (still locked) strings:

Screen Shot 2021-07-11 at 12 39 52

ErikSchierboom:
Closes https://github.com/exercism/v3-beta/issues/119


app/commands/exercise/search.rb
```diff
@@ -10,11 +10,21 @@ def initialize(track, user_track: nil, criteria: nil)
...
       @exercises
     end
 
+    def filter_status!
+      if !user_track || user_track.external?
+        @exercises = @exercises.where.not(status: %i[deprecated wip])
+      else
+        @exercises = @exercises.where.not(status: %i[deprecated wip]).
+          or(@exercises.where(id: user_track.solutions.pluck(:exercise_id)))
+      end
```
ErikSchierboom: I had to duplicate the `where.not` bit for Rails to correctly pick up the `.or`, but maybe there is a better way?

ErikSchierboom:
I found that there was a class named `GenerateConceptExerciseMapping` (see https://github.com/exercism/website/blob/main/app/commands/user_track/generate_concept_exercise_mapping.rb) but that there are no references to that class, expect for a tests file. @neenjaw do you know anything about that, seeing that you appear to be the author (https://github.com/exercism/website/blame/main/app/commands/user_track/generate_concept_exercise_mapping.rb)?

neenjaw:
It was being used by a command which generated the data for the ConceptMap, not sure where it would have gone

app/views/tracks/exercises/show/_concepts_section.html.haml
```diff
@@ -8,15 +8,15 @@
...
             = ReactComponents::Student::ExerciseStatusDot.new(exercise, user_track)
-          - concept.practice_exercises.each do |exercise|
+          - concept.practice_exercises.enabled(user_track).each do |exercise|
             = ReactComponents::Student::ExerciseStatusDot.new(exercise, user_track)
 
   = render ViewComponents::ProminentLink.new("See how your concept map has changed", track_concepts_path(track), with_bg: true)
 
-  - unlocked_concepts = exercise.unlocked_concepts.to_a
-  - unlocked_exercises = exercise.unlocked_exercises.to_a
+  - unlocked_concepts = exercise.unlocked_concepts(user_track).to_a.filter { |unlocked_concept| user_track.concept_unlocked?(unlocked_concept) }
```
ErikSchierboom: Originally the code below had a `next unless` statement, which meant that some records were not shown even though they _were_ included in the count. It is possible that we want to use this change in more places.

app/helpers/react_components/student/exercise_status_chart.rb
```diff
@@ -18,8 +18,10 @@ def to_s
       end
 
       private
+      attr_reader :user_track
+
       def exercises_data
-        user_track = UserTrack.for(current_user, track)
+        @user_track = UserTrack.for(current_user, track)
```
iHiD, line 25: Make this into a memoized method.

app/commands/track/determine_concept_map_layout.rb
```diff
@@ -6,9 +6,9 @@ class Track
   class DetermineConceptMapLayout
     include Mandate
 
-    def initialize(track)
+    def initialize(track, user_track)
```
iHiD: I'd probably *just* pass user_track if we do the external bit.

app/commands/user_track/generate_exercise_status_mapping.rb
```diff
@@ -16,7 +16,7 @@ def mapping
       mapping = Hash.new { |k, v| k[v] = Set.new }
 
       Exercise::TaughtConcept.joins(:exercise, :concept).
-        where('exercises.track_id': track.id).
+        where('exercises.id': user_track.concept_exercises).
```
iHiD, line 19: Can you check the logs to ensure this does an inner query pls (rather than passing all the ids in). 

(You can see the SQL when you run the command in Rails c)

ErikSchierboom: Yep, it uses a subquery:

```sql
SELECT `track_concepts`.`slug`, `exercises`.`slug`
FROM `exercise_taught_concepts`
INNER JOIN `exercises` ON `exercises`.`id` = `exercise_taught_concepts`.`exercise_id`
INNER JOIN `track_concepts` ON `track_concepts`.`id` = `exercise_taught_concepts`.`track_concept_id`
WHERE `exercises`.`id` IN (SELECT `exercises`.`id` FROM `exercises` WHERE `exercises`.`type` = 'ConceptExercise' AND `exercises`.`track_id` = 1 AND (`exercises`.`status` IN (2, 1) OR `exercises`.`id` IN (SELECT `solutions`.`exercise_id` FROM `solutions` INNER JOIN `exercises` ON `exercises`.`id` = `solutions`.`exercise_id` WHERE `solutions`.`user_id` = 4 AND `exercises`.`track_id` = 1)))
 ```

Comment (iHiD) [reply app/commands/user_track/generate_exercise_status_mapping.rb:19]: Thanks!

File: app/controllers/tracks/exercises_controller.rb [hunk 6]
```diff
@@ -45,7 +45,7 @@ def start
   def edit
     return if @solution
 
-    if @user_track && !@user_track.external?
+    if !@user_track.external?
```
iHiD:
```suggestion
    unless @user_track.external?
```

iHiD: In fact, change this to be a guard `redirect_to action: :show if @user_track.external?`

app/models/user_track.rb
```diff
@@ -29,7 +29,7 @@ def self.for!(user_param, track_param)
     )
   end
 
-  def self.for(user_param, track_param, external_if_missing: false)
+  def self.for(user_param, track_param, external_if_missing: true)
```
iHiD: I suggest removing `external_if_missing` entirely now.

app/models/user_track.rb
```diff
@@ -41,6 +41,30 @@ def self.for(user_param, track_param, external_if_missing: false)
     end
   end
 
+  def exercises
+    enabled_exercises
```
iHiD: I'd suggest memoizing this directly (and inlining `filter_enabled_exercises(track.exercises)`) rather than calling out to the other method.

app/models/user_track.rb
ErikSchierboom: Agreed. I wanted to save that for last :0

app/controllers/tracks_controller.rb
```diff
@@ -19,7 +19,10 @@ def index
   def show
     @user_track = UserTrack.for(current_user, @track)
 
-    if @user_track
+    if @user_track.external?
+      @showcase_exercises = @user_track.exercises.order("RAND()").limit(3).to_a
+      render "tracks/show/unjoined"
+    else
```
iHiD:
```suggestion
      return render "tracks/show/unjoined"
    end
```

(and remove the end below)

app/models/user_track.rb
```diff
@@ -29,18 +29,47 @@ def self.for!(user_param, track_param)
...
+    filter_enabled_exercises(concept.practice_exercises) if concept.present?
+  end
+
+  def unlocked_concepts_for(exercise: nil)
+    exercise.unlocked_concepts.to_a.filter { |c| concept_unlocked?(c) } if exercise.present?
+  end
+
+  def unlocked_exercises_for(exercise: nil)
+    exercise.unlocked_exercises.to_a.filter { |e| exercise_unlocked?(e) } if exercise.present?
+  end
```
iHiD: For all of these it seems like the concept/exercise is required. So should we remove the `nil` in the method definition and remove the `if` statements?

app/views/tracks/exercises/show/_action_box_join_track.html.haml
```diff
@@ -1,14 +1,14 @@
...
     and
-    = link_to "#{track.num_exercises} exercises", track_exercise_path(track)
+    = link_to "#{user_track.num_exercises} exercises", track_exercise_path(user_track.track)
     ready for you to dig into!
 
   .buttons
-    = button_to "Join the #{track.title} track", join_track_path(track), method: :post, class: "btn-primary btn-m"
-    = link_to "See all #{track.title} exercises", track_exercise_path(track), class: 'btn-secondary btn-m'
+    = button_to "Join the #{user_track.track.title} track", join_track_path(user_track.track), method: :post, class: "btn-primary btn-m"
+    = link_to "See all #{user_track.track.title} exercises", track_exercise_path(user_track.track), class: 'btn-secondary btn-m'
```
iHiD: I think for this, it might be better to set `- track = user_track.track` at the top of the file, rather than making these changes everything. Same for the other templates.

iHiD (approved):

PR #1236, opened 2021-07-11
Load correct mentor requests for default track correctly

### LINKED ISSUE #111: [Mentoring/Queue] Not correctly filtered by track at the start
Screenshot 2021-07-05 at 18 15 05

The solutions on the LHS when the page loads aren't the Ruby solutions that they should be according to the dropdown on the RHS.

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/111.

### DIFF
--- a/app/helpers/react_components/mentoring/queue.rb
+++ b/app/helpers/react_components/mentoring/queue.rb
@@ -64,13 +64,10 @@ module ReactComponents
       end
 
       def queue_request
-        query = {
-          order: params[:order],
-          criteria: params[:criteria],
-          page: params[:page],
+        query = params.merge(
           track_slug: default_track[:slug],
           exercise_slug: default_exercise.try(:[], :slug)
-        }.compact
+        ).compact
 
         {
           endpoint: Exercism::Routes.api_mentoring_requests_path,
--- a/app/helpers/react_components/mentoring/queue.rb
+++ b/app/helpers/react_components/mentoring/queue.rb
@@ -64,17 +64,19 @@ module ReactComponents
       end
 
       def queue_request
+        query = {
+          order: params[:order],
+          criteria: params[:criteria],
+          page: params[:page],
+          track_slug: default_track[:slug],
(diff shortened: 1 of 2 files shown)


iHiD:
@kntsoriano Nice. I've added a commit to use (params.merge) so that if we add an extra param in the future, we don't have to remember to add it here too. Hopefully that'll work with Strong Params. Will merge if it goes green 🙂

kntsoriano:
@iHiD I think I prefer being explicit about what params we pass instead of using merge. It might be harder to change, but I like how clear it is when looking at the code. Maybe we could agree on what level we should enforce being explcit about the parameters being passed in. Should it be at the assembler level, the command level, etc?

iHiD:
Come on Slack if you're around 🙂

iHiD:
> Maybe we could agree on what level we should enforce being explcit about the parameters being passed in. Should it be at the assembler level, the command level, etc?

So the point of the assemblers is that the external interface (controllers, react helpers, etc) don't have to know/care about the user's input, they just pass it in and the assembler works stuff out. This means we have one central point of truth (inside the assembler) about what `params` keys should be called, and one place to sync up with the React.

So there should basically be an identical use of `params` in these two places:
https://github.com/exercism/website/blob/4b5c45ab6f0e7a00e65a6032db10513f4ece6e75/app/controllers/api/mentoring/requests_controller.rb#L13
https://github.com/exercism/website/blob/4b5c45ab6f0e7a00e65a6032db10513f4ece6e75/app/helpers/react_components/mentoring/queue.rb#L77

So the way I think you're suggesting would be specify a series of kwargs in the assembler for each param, rather than taking the params hash directly. But that would undo the DRYness of the assembler by moving the same knowledge (e.g. of the key names) into two separate places (controller/react helper), and then make it much easier to introduce bugs as we could use two different param names in two different places. By having the central point of truth inside the params has, we give ourselves loads of protection.

iHiD:
@kntsoriano I'll merge this and you can add the extra guards you mentioned seperately :)

PR #1239, opened 2021-07-12
Update progress graph line algorithm

neenjaw:
previously a sliding window algorithm was being used which found the line connect the prev/next points and used as parallel line to create the bezier control points.  This was leading to a "dipping" behaviour when there were extreme changes in the datapoints.  This is because in order for the bezier line to curve to accomodate all of the point, it sometimes must curve the opposite way to accomodate the extremeness of a curve.

Screenshot from 2021-07-11 20-58-51

> Example of the "Dip"

Now using a catmullrom spline to bezier curve algorithm (http://schepers.cc/svg/path/catmullrom2bezier.js) to produce a more predictable curve, and then when there is an extreme curve, points are being interpolated to ease it in.  There is a chance that this causes the same "dipping" after the curve rather than preceed it, but this is the limitation of using a spline to graph datapoints rather than computing a best-fit line.

Screenshot from 2021-07-06 00-33-12
> Using catmullrom spline (translated to bezier) with interpolated points

I think if there are further aesthetic concerns, probably would be good consider creating a high-resolution (many many data points) poly-line so that it appears smooth, but is more reflective of the true data.

### DIFF
--- a/app/javascript/components/common/ProgressGraph.tsx
+++ b/app/javascript/components/common/ProgressGraph.tsx
@@ -2,10 +2,10 @@ import * as React from 'react'
 import { useRef } from 'react'
 
 import {
+  mapToSvgDimensions,
   minMaxNormalize,
-  drawSegmentPath,
-  drawSmoothPath,
-  DataPoint,
+  smoothDataPoints,
+  transformCatmullRomSplineToBezierCurve,
 } from './svg-graph-util'
 
 interface IProgressGraph {
@@ -19,7 +19,6 @@ export const ProgressGraph: React.FC<IProgressGraph> = ({
   data,
   height,
   width,
-  smooth = false,
 }) => {
   const randomIdRef = useRef<null | number>(null)
 
@@ -36,18 +35,11 @@ export const ProgressGraph: React.FC<IProgressGraph> = ({
   const aspectRatio = width / height
   const hInset = height * 0.1
   const vInset = Math.round(hInset / aspectRatio)
-  const vBuffer = height * 0.05 // due to smoothing, some curves can go outside the boundaries without this buffer
 
-  const step = width / (data.length - 1)
   const normalizedData = minMaxNormalize(data)
-  const dataPoints = normalizedData.map(
-    (n, i) =>
-      ({
-        x: i * step,
-        y: height - vBuffer - (height - vBuffer) * n, // SVG coordinates start from the upper left corner
-      } as DataPoint)
-  )
-  const path = smooth ? drawSmoothPath(dataPoints) : drawSegmentPath(dataPoints)
+  const dataPoints = mapToSvgDimensions(normalizedData, height, width)
+  const smoothedDataPoints = smoothDataPoints(dataPoints, height)
+  const path = transformCatmullRomSplineToBezierCurve(smoothedDataPoints)
 
   return (
     <svg
@@ -57,6 +49,15 @@ export const ProgressGraph: React.FC<IProgressGraph> = ({
       }`}
     >
       <defs>
+        <mask
+          id={`progress-graph-color-mask-${getRandomId()}`}
+          x="0"
+          y="0"
+          width={width}
+          height={height}
+        >
+          <path d={path} stroke="white" fill="transparent" />
+        </mask>
         <linearGradient
           id={`progress-graph-color-gradient-${getRandomId()}`}
           x1="0%"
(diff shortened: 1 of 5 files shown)


app/javascript/components/common/ProgressGraph.tsx
```diff
@@ -57,6 +49,15 @@ export const ProgressGraph: React.FC<IProgressGraph> = ({
...
       <defs>
+        <mask
+          id={`progress-graph-color-mask-${getRandomId()}`}
+          x="0"
+          y="0"
+          width={width}
+          height={height}
+        >
+          <path d={path} stroke="white" fill="transparent" />
+        </mask>
```
neenjaw, line 60: As it turns out, when a vertical gradient is applied to a horizontal (flat) line, the svg breaks and shows nothing.  This method is that the line creates a mask over a rectangle that is colored with the gradient, and the mask only shows the resulting line and thus works for horizontal lines.

app/javascript/components/common/svg-graph-util.ts
```diff
@@ -7,119 +7,221 @@ export type Line = {
   length: number
   angle: number
 }
-
-const SMOOTHING_RATIO = 0.2
+const SIMILAR_THRESHOLD_DELTA = 0.01
```
neenjaw, line 10: if points are within 1% of each other of each other, they will be considered equal for the purposes of smoothing.

app/javascript/components/common/svg-graph-util.ts
```diff
@@ -7,119 +7,221 @@ export type Line = {
...
+const SMOOTHING_THRESHOLD_DELTA = 0.2
```
neenjaw, line 11: If points are more than 20% different from eachother (with respect to the total height of the graph) they will be considered extreme enough to be smoothed

app/javascript/components/common/svg-graph-util.ts
```diff
@@ -7,119 +7,221 @@ export type Line = {
...
+const INTERPOLATION_STEPS = 3
+const SEGMENT_PARTS = INTERPOLATION_STEPS + 1
 
 export const minMaxNormalize = (data: Array<number>): Array<number> => {
   const min = Math.min(0, ...data)
   const max = Math.max(...data)
+
+  if (max === 0 && min === 0) {
+    return data
+  }
```
neenjaw, line 21: Cover the case of a list of 0's `[0,0,0,0,0]`

app/javascript/components/common/svg-graph-util.ts
```diff
@@ -7,119 +7,221 @@ export type Line = {
...
+
   return data.map((n) => (n - min) / (max - min))
 }
 
-export const drawSegmentPath = (dataPoints: Array<DataPoint>): string => {
+export const mapToSvgDimensions = (
```
neenjaw, line 26: function extracted from the component for clarity and ease of comprehension

app/javascript/components/common/svg-graph-util.ts
```diff
@@ -7,119 +7,221 @@ export type Line = {
...
-      switch (idx) {
-        case 0:
-          return `${drawM(dataPoint)}`
-        case 1:
-          return `${path} ${drawC(dataPoints, idx)}`
-        default:
-          return `${path} ${drawS(dataPoints, idx)}`
+      const prevPoint = dataPoints[index - 1]
+      const nextPoint = dataPoints[index + 1]
+      const nextNextPoint = dataPoints[index + 2]
```
neenjaw, line 66: if the current index is near the beginning or the end, these may be undefined, which is why they are being handled carfully on lines 69 and 77

app/javascript/components/common/svg-graph-util.ts
```diff
@@ -7,119 +7,221 @@ export type Line = {
...
-  const x1 = xPrev + (x - xPrev) / 2
-  const y1 = yPrev
-  const x2 = x - (x - xPrev) / 2
-  const y2 = y
+const getTimeAtStep = (step: number): number => (1 / SEGMENT_PARTS) * (step + 1)
 
-  return `C ${x1} ${y1}, ${x2} ${y2}, ${x} ${y}`
-}
+const quartEaseInFunction = (t: number): number => t * t * t * t
+const quartEaseOutFunction = (t: number): number => 1 - --t * t * t * t
```
neenjaw, line 164: https://gist.github.com/gre/1650294

iHiD:
Great work again, @neenjaw. My only request would be to add some of those comments inline as they'll be useful in the future. And https://gist.github.com/gre/1650294 specifically as we should reference it in the code as a source. I'll gratefully merge this when that's done! :)

iHiD:
@neenjaw I'm going to merge this. If you'd like to follow up with another PR for adding the comments, that'd be great :) Thanks again!

PR #1240, opened 2021-07-12
Add redirect overlay

kntsoriano:
https://github.com/exercism/v3-project-management/issues/286


app/javascript/packs/application.tsx
```diff
@@ -490,5 +491,17 @@ document.addEventListener('turbolinks:load', () => {
   highlightAll()
 })
 
+document.addEventListener('turbolinks:visit', () => {
```
kntsoriano: @iHiD This is how I plan to do this, but I might just trigger this to turn on only on actions that redirect, instead of  the overlay appearing every time it is loading between pages.

PR #1244, opened 2021-07-12
Fix text selection on split panels

### LINKED ISSUE #115: [Editor] Click + hold + move mouse doesn't select text in right column
In the editor, selecting text in the right column is not obvious. The usual method that you use (click, hold down the mouse button and move the mouse) doesn't work. It only works to double-click a text block.

In the code editor itself, selecting text works fine.

I think all usual ways of selecting text should work everywhere on the website, especially in that right column with instructions and test results.

It's crucial to be able to copy the compilation errors and google for them, copy some unknown English words and translate them if you're not a native speaker, or copy and part of the instructions and send it to a friend/mentor asking for clarifications.

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/115.

### DIFF
--- a/app/javascript/components/common/SplitPane.tsx
+++ b/app/javascript/components/common/SplitPane.tsx
@@ -30,28 +30,36 @@ export const SplitPane = ({
     [dragging]
   )
 
-  const onMouseDown = useCallback((e: React.MouseEvent) => {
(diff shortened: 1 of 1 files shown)


iHiD:
Nice :)

PR #1248, opened 2021-07-13
Minimize notification channel broadcasts

kntsoriano:
https://github.com/exercism/v3-beta/issues/82

@iHiD I'm trying to see if ActionCable is what's causing the notifications dropdown to be buggy.


iHiD:
So if this does work, it won't be a "fix", but it will tell us there's some race condition that's hurting us?

PR #1256, opened 2021-07-14
Guard against mentoring your own solutions

ErikSchierboom:
- Guard against student mentoring own solution
- Handle student self-mentoring
- Remove todo
- Add todo for styling of error messages

### DIFF
--- a/config/initializers/exceptions.rb
+++ b/config/initializers/exceptions.rb
@@ -19,6 +19,8 @@ class SubmissionFileTooLargeError < RuntimeError; end
 
 class SolutionLockedByAnotherMentorError < RuntimeError; end
 
+class StudentCannotMentorOwnSolutionError < RuntimeError; end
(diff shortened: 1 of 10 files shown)


config/initializers/exceptions.rb
```diff
@@ -19,6 +19,8 @@ class SubmissionFileTooLargeError < RuntimeError; end
 
 class SolutionLockedByAnotherMentorError < RuntimeError; end
 
+class StudentCannotMentorThemselvesError < RuntimeError; end
```
ErikSchierboom, line 22: Let me know if this name is okay

iHiD:
Note to self: I need to style this.

config/initializers/exceptions.rb
iHiD: Yep - fine :)

PR #1258, opened 2021-07-15
Fix bug where sendRequest returns undefined when component unmounts

kntsoriano:
https://github.com/exercism/v3-beta/issues/82
https://github.com/exercism/v3-beta/issues/112

I just found out that when a component unmounts, our `sendRequest` function returns `undefined`. This is why tooltips don't load on hovering out of them quickly. Also, I think this is why the notification dropdown doesn't work, since we've unmounted the component when we change pages.

There are some things that might break from this, and these might at the top of my head:

1. Cancelling a test run.

I will handle this case at the same time as this issue: https://github.com/exercism/v3-beta/issues/123

2. Scratchpad 

When I was looking at the scratchpad code, I noticed that we weren't using `useRequestQuery`. I will make a separate issue for this.

### DIFF
--- a/app/javascript/components/mentoring/session/Scratchpad.tsx
+++ b/app/javascript/components/mentoring/session/Scratchpad.tsx
@@ -1,10 +1,6 @@
-import React, { useCallback, useState, useRef, useEffect } from 'react'
-import {
-  MarkdownEditor,
-  MarkdownEditorHandle,
-} from '../../common/MarkdownEditor'
+import React, { useCallback, useState, useEffect } from 'react'
+import { MarkdownEditor } from '../../common/MarkdownEditor'
 import { sendRequest } from '../../../utils/send-request'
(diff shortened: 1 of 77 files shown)


app/javascript/components/mentoring/session/Scratchpad.tsx
```diff
@@ -38,45 +34,48 @@ export const Scratchpad = ({
...
+        }
+      },
+    }
+  )
+
+  const handleSubmit = useCallback(
+    (e) => {
+      e.preventDefault()
+
+      setError('')
```
iHiD, line 65: Thoughts on doing `setError(undefined)` instead of `''`?

iHiD:
@kntsoriano Should we also remove `use-is-mounted` from the package.json?

kntsoriano:
@iHiD We still use it in some cases.

iHiD:
OK, great :)

PR #1272, opened 2021-07-16
Add tasks tooltip

kntsoriano:
https://github.com/exercism/v3-project-management/issues/292


app/javascript/components/misc/ExercismTippy.tsx
```diff
@@ -0,0 +1,17 @@
...
+  return (
+    <LazyTippy
+      animation="shift-away-subtle"
+      followCursor="horizontal"
+      maxWidth="none"
+      plugins={[followCursor]}
+      {...props}
+    />
+  )
+}
```
iHiD, line 17: Nice!!

iHiD:
@kntsoriano Sorry man - I've totally changed my mind on this after thinking it all through more. I think this **should** be in React instead. Could you basically set it up so:
- Hover opens tooltip
- Tooltip contains the following for `Knowledge`, `Size`, `Module`, `Action` and `Type` (this example is for knowledge and size. We already have the `...Tag` bits - but we'll need the `...Details` bits adding.

```
<div className="info">
  <KnowledgeTag knowledge={task.tags.knowledge} />
  <KnowledgeDetails knowledge={task.tags.knowledge} />
</div>
<div className="info">
   <SizeTag size={task.tags.size} />
  <SizeDetails size={task.tags.size} />
</div>
```

etc. 

And then above it an extra new `Summary` one that takes the `tags` and will make a sentence from them.

That should give us the details for this effectively:
Screenshot 2021-07-19 at 22 57 06

kntsoriano:
@iHiD I've done this. I haven't handled edge cases such as when some tags are blank, but what I have pushed should get you started with the styling.

PR #1276, opened 2021-07-18
Add empty state for tracks

iHiD:
@kntsoriano Can you please add a commit that sets implements the button. It should clear the search bar and the filters.


iHiD:
@kntsoriano Oops. Merged this forgetting it still needed your work. Pls add another PR with the button uimplemented.

PR #1280, opened 2021-07-20
Fix test warnings

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/293.

@iHiD I think I'm going to stop here. There are still some flaky tests and its really complicated since we're dealing with complex components such as `<SimpleMDE />`. I'm thinking of stubbing those kinds of components out in the future, but I'm pretty happy to where I got to now. I think I've spent too much time on this now.


app/javascript/components/modals/PublishSolutionModal.tsx
```diff
@@ -19,6 +19,7 @@ export const PublishSolutionModal = ({
   endpoint: string
   iterations: readonly Iteration[]
 }): JSX.Element => {
+  console.log('test')
```
iHiD:
```suggestion
```

iHiD:
@kntsoriano Great. I have no idea but if you're happy, I'm happy! So I'll merge. Thanks for taking pride in this :)

PR #1283, opened 2021-07-20
Keep track of hidden introducers

ErikSchierboom:


app/models/user.rb
```diff
@@ -175,6 +177,14 @@ def ghost?
     id == GHOST_USER_ID
   end
 
+  def hide_introducer(name)
+    introducers.create_or_find_by!(name: name)
+  end
+
+  def introducer_hidden?(name)
+    introducers.where(name: name).exists?
+  end
```
ErikSchierboom: I made these methods on the `User` model. I could also make them commands, but I felt that it was nice to be able to do `current_user&.introducer_hidden?('scratchpad')`, but I have no idea if this is something we want.

iHiD:
This is great. Thanks for taking it on.

The one thing I'd like to change is to have the model as `User::DismissedIntroducer` and then `user.dismissed_introducers`. (And then maybe change `hidden_*` to `dismissed_*` too?)

The current reading sounds to me like users have made introducers etc rather than dismissed them.

ErikSchierboom:
@iHiD Updated

app/models/user.rb
```diff
@@ -175,6 +177,14 @@ def ghost?
     id == GHOST_USER_ID
   end
 
+  def dismiss_introducer(name)
```
iHiD:
```suggestion
  def dismiss_introducer!(name)
```

db/migrate/20210720075622_create_user_dismissed_introducers.rb
```diff
@@ -0,0 +1,13 @@
+class CreateUserDismissedIntroducers < ActiveRecord::Migration[6.1]
+  def change
+    create_table :user_dismissed_introducers do |t|
+      t.belongs_to :user, null: false, foreign_key: true
+
+      t.string :name, null: false
```
iHiD: At the risk of you hating me, I think we should change this to be `slug` rather than `name`. I think we could use `name` to provide a human readable name for this later, which we show in the settings (to turn introducers back on maybe). And `name` is always human-readable and `slug` is generally this format.

ErikSchierboom: No I don't mind. Good argument

iHiD (approved):
Added a commit to use slug not id. I'll merge when green, or you feel free to if I forget! :)

PR #1284, opened 2021-07-20
Max two locks per mentor

ErikSchierboom:


app/commands/mentor/request/lock.rb
```diff
@@ -5,17 +5,14 @@ class Lock
...
-      # hits 'start discussion' that should remove the lock.
-
       def call
         ActiveRecord::Base.transaction do
           # This is a DB lock (The naming is confusing!)
           request.lock!
 
           # Guard against not being lockable
           raise SolutionLockedByAnotherMentorError unless request.lockable_by?(mentor)
+          raise MentorSolutionLockLimitReachedError if mentor.locks.size >= Mentor::RequestLock::MAX_LOCKS_PER_MENTOR
```
ErikSchierboom: Is `size` the right method to call here?

app/models/mentor/request_lock.rb
```diff
@@ -3,4 +3,6 @@ class Mentor::RequestLock < ApplicationRecord
   belongs_to :locked_by, class_name: "User"
 
   scope :expired, -> { where("locked_until < ?", Time.current) }
+
+  MAX_LOCKS_PER_MENTOR = 2
```
ErikSchierboom: Is this a good location for the locks?

app/commands/mentor/discussion/create.rb
```diff
@@ -18,6 +18,8 @@ def call
           # whole load need to fail.
           request.fulfilled!
 
+          request.locks.destroy_all
```
iHiD:
```suggestion
          request.locks.where(locked_by: mentor).destroy_all
```

Just in case two people have somehow managed to lock the same discussion?

app/commands/mentor/request/lock.rb
iHiD: Yeah, it'll proxy to `count` as locks hasn't been loaded as an array. So you could explicitly use `count` but I think `size` is fine too.

app/models/mentor/request_lock.rb
iHiD: Yeah, I think it's the right file. Pop the constant to the top of the class though. (Above the `belongs_to` calls)

app/commands/mentor/discussion/create.rb
ErikSchierboom: 👍

iHiD (approved):
~I fixed my other comment (moving the constant). Merge it if it goes green :)~

@erikSchierboom Actually one more thing. Could you rename this to `mentor_locks`, not `locks` on `User` pls.

PR #1285, opened 2021-07-20
Guard against solution already being completed

ErikSchierboom:


app/commands/solution/complete.rb
```diff
@@ -5,7 +5,7 @@ class Complete
     initialize_with :solution, :user_track
 
     def call
-      # TODO: (Required) Guard against already being completed
+      raise SolutionAlreadyCompletedError if solution.completed?
```
ErikSchierboom: @iHiD Do we want to raise an exception or just do an early return?

PR #1286, opened 2021-07-20
Add guard to publishing/completing in solutions API

ErikSchierboom:
- Add API tests for publishing
- Add API tests for unpublishing
- Return 400 when trying to complete/publish solution without iteration
- Complete solution when publishing uncompleted solution
- Remove redundant todo

### DIFF
--- a/app/commands/solution/publish.rb
+++ b/app/commands/solution/publish.rb
@@ -2,10 +2,12 @@ class Solution
   class Publish
     include Mandate
 
-    initialize_with :solution, :iteration_idx
+    initialize_with :solution, :user_track, :iteration_idx
 
     def call
       solution.with_lock do
+        Solution::Complete.(solution, user_track) unless solution.completed?
+
         return if solution.published?
 
         ActiveRecord::Base.transaction do
--- a/app/commands/solution/publish.rb
+++ b/app/commands/solution/publish.rb
@@ -5,9 +5,9 @@ class Solution
     initialize_with :solution, :user_track, :iteration_idx
 
     def call
-      solution.with_lock do
-        Solution::Complete.(solution, user_track) unless solution.completed?
+      Solution::Complete.(solution, user_track) unless solution.completed?
 
+      solution.with_lock do
         return if solution.published?
 
         ActiveRecord::Base.transaction do
--- a/test/controllers/api/solutions_controller_test.rb
+++ b/test/controllers/api/solutions_controller_test.rb
@@ -436,12 +436,12 @@ class API::SolutionsControllerTest < API::BaseTestCase
     )
   end
 
-  test "publish renders 400 when solution has no iterations" do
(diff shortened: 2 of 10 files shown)


test/controllers/api/solutions_controller_test.rb
```diff
@@ -347,7 +375,204 @@ class API::SolutionsControllerTest < API::BaseTestCase
...
         ]
       }.to_json
+
       assert_equal expected, response.body
     end
   end
+
+  ############
+  # Publish #
+  ############
```
ErikSchierboom, line 385: There was no test coverage, so I added some :)

test/controllers/api/solutions_controller_test.rb
```diff
@@ -347,7 +375,204 @@ class API::SolutionsControllerTest < API::BaseTestCase
...
+      solution.reload
+      assert_equal Time.current, solution.completed_at
+      assert_equal Time.current, solution.published_at
+      assert_equal :published, solution.status
+    end
+  end
+
+  ############
+  # Unpublish #
+  ############
```
ErikSchierboom, line 505: There was no test coverage, so I added some :)

app/commands/solution/publish.rb
```diff
@@ -2,10 +2,12 @@ class Solution
   class Publish
     include Mandate
 
-    initialize_with :solution, :iteration_idx
+    initialize_with :solution, :user_track, :iteration_idx
```
ErikSchierboom, line 5: I pass in the user_track instead of doing `solution.user_track` to be consistent with `Solution::Complete`

iHiD: Is that preferable to doing solution.user_track inside `Solution::Complete` do you think?

ErikSchierboom: Well, we already have the `user_track` instantiated before `Solution::Complete` is called to do another check. I was slightly hesistant to use `solution.user_track` due to this comment: https://github.com/exercism/website/blob/main/app/models/solution.rb#L12

iHiD: Yeah, pass it in.

app/commands/solution/publish.rb
```diff
@@ -2,10 +2,12 @@ class Solution
...
 
     def call
       solution.with_lock do
+        Solution::Complete.(solution, user_track) unless solution.completed?
```
iHiD: You probably need to move this outside the lock block it to work in production. Otherwise you'll try and lock it both here and in complete and I *think* that might break in production (tests don't use transactions so things like this sometimes pass in the tests and fail for real).

ErikSchierboom: Updated

iHiD (approved):
I rebased this up with the other completing PR and force-pushed. Merge when green! :)

PR #1288, opened 2021-07-20
Fix introducer animations

### LINKED ISSUE #132: Introducers leave a white space when closed.
https://github.com/exercism/website/commit/2729b1f4ae9a70eb8db4ecfd2fe3f28370cdf86b#r53738397

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/132.

With the wrapper component introduced, we need to fix the introducer on the scratchpad.
Screen Shot 2021-07-21 at 7 01 11 AM

### DIFF
--- a/app/javascript/components/mentoring/session/Scratchpad.tsx
+++ b/app/javascript/components/mentoring/session/Scratchpad.tsx
@@ -117,20 +117,20 @@ export const Scratchpad = ({
 
   return (
     <>
-      <Introducer
-        icon="scratchpad"
-        endpoint={scratchpad.links.hideIntroducer}
-        hidden={scratchpad.isIntroducerHidden}
-        size="small"
-      >
-        <h2>Introducing your scratchpad</h2>
-        <p>
-          A <a href={scratchpad.links.markdown}>Markdown-supported</a> place for
-          you to write notes and add code snippets you’d like to refer to during
-          mentoring.
-        </p>
-      </Introducer>
-
+      {scratchpad.isIntroducerHidden ? (
+        <Introducer
+          icon="scratchpad"
+          endpoint={scratchpad.links.hideIntroducer}
+          size="small"
+        >
+          <h2>Introducing your scratchpad</h2>
+          <p>
+            A <a href={scratchpad.links.markdown}>Markdown-supported</a> place
+            for you to write notes and add code snippets you’d like to refer to
+            during mentoring.
+          </p>
+        </Introducer>
+      ) : null}
       <div className="title">
         Your notes for <strong>{exercise.title}</strong> in
         <TrackIcon iconUrl={track.iconUrl} title={track.title} />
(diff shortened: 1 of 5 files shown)


iHiD:
> @iHiD Hey man. We shouldn't rely on adding styles to react wrappers. As we don't have a way to know whether a component is rendered directly by our bootloader or rendered inside another component. In this case, the introducer is rendered inside the scratchpad and not via the bootloader, thus it is unreliable to call the parentElement because we never know what it is.

@kntsoriano This won't fix the animations problem. By changing the css to be on the rendered React component, everything on the page jumps down when the React renders. The CSS has to be on the parent container to avoid that jump. And then parent container has to then have the class set on it to remove it.

I don't see another way to achieve what we need other than setting the class on the outer wrapper container. This is a limitation of the fact we're sprinkling arbitrary react components around rather that having an app where everything renders at the same time. 

So I think we need to add back in the `ref.current!.parentElement!.classList.add('hidden')` bit to achieve this. We could add a check in there that it has the expected class (the `react-container-...` one).

iHiD:
@kntsoriano You can see the bug that the branch introduces by changing the CSS: https://www.loom.com/share/f573dd0f2d7d4b52ab88c06d5f0f9dcc?from_recorder=1&focus_title=1

Everything jumps dump when the introducer loads.

kntsoriano:
@iHiD I don't get what the difference this has from the previous solution we had. I added a wrapper to our introducer instead of relying on the external wrapper that react bootloader adds.

iHiD:
The problem is that the element needs to take up space **before** React renders, else the page jumps when React loads. So the external container div has a height set. So styling/removing/etc anything inside that container div (in React) is irrelevant. So even if you entirely removed the introducer when you click close, it's still in a div that's in the HTML that has a height.  So you have to remove the container div as well, which is what that LOC does. 

There are three scenarios:
1. Everything jumps when the React loads (see the loom). That's what this PR does.
2. There's a big whitespace gap in the page when the introducer closes (that's what your previous commit that broke this does).
3. The original code that doesn't jump and removes the whitespace gap.

app/javascript/components/mentoring/session/Scratchpad.tsx
```diff
@@ -117,20 +117,20 @@ export const Scratchpad = ({
...
-      >
-        <h2>Introducing your scratchpad</h2>
-        <p>
-          A <a href={scratchpad.links.markdown}>Markdown-supported</a> place for
-          you to write notes and add code snippets you’d like to refer to during
-          mentoring.
-        </p>
-      </Introducer>
-
+      {scratchpad.isIntroducerHidden ? (
```
iHiD: @kntsoriano This logic was the wrong way round. I've fixed it.

iHiD:
@kntsoriano I've added a commit that:
1. Fixes some logic in the scratchpad
2. Readds the functionality that used to work.

I'm going to merge this. Happy for you to work out a different way to try and solve this to try and achieve no jumping, unwanted whitespace, and no reference to divs outside of React, but let's wait until we've launched v3 and we have time, as this is of low importance atm 🙂

PR #1289, opened 2021-07-21
Refactor Editor to allow cancelled Test Runs

kntsoriano:
https://github.com/exercism/v3-beta/issues/123


app/controllers/api/solutions/submission_cancellations_controller.rb
```diff
@@ -0,0 +1,11 @@
+module API
+  module Solutions
+    class SubmissionCancellationsController < BaseController
+      def create
+        # TODO
```
kntsoriano: This is where the code goes to signal the test runners to cancel, etc.

app/javascript/components/editor/TestRunSummary.tsx
```diff
@@ -198,6 +198,9 @@ const TestRunSummaryContent = ({
         </div>
       )
     }
+    case TestRunStatus.CANCELLED: {
```
kntsoriano: I added this. Do we want a state to show to the user that they've cancelled their test?

app/javascript/components/editor/TestRunSummaryContainer.tsx
```diff
@@ -54,23 +54,17 @@ export const TestRunSummaryContainer = ({
...
     clearTimeout(timer.current)
 
     channel.current?.disconnect()
-  }, [channel, timer])
+  }, [channel])
+
   const cancel = useCallback(() => {
     setTestRun({ ...testRun, status: TestRunStatus.CANCELLED })
-  }, [testRun, setTestRun])
+
```
kntsoriano, line 56: I made this an optimistic change. I mark the test run as cancelled even though we're not sure if the backend has acknowledged it yet.

iHiD:
@kntsoriano Added a commit for cancelling the test runs. When the test run is cancelled we should do:
1. Revert back to previous test results (or the zero state if there weren't any)
2. Add a div that says "Test Run Cancelled" at the top. I'll add styling to make this float over the top of the content. 
3. Fade out the "Test Run Cancelled" div after 3s.

FYI, there's also an edge case atm where the test run is finished before the cancelled button reaches the API. This means it looks cancelled, but when you refresh the cancelled test results appear. This is really common due to how fast everything is happening, so I'll work out how we should fix that server-side - it shouldn't need anything in React.

iHiD:
I've added another commit around only retrieving non-cancelled test runs, and only comparing changed files with non-cancelled/exceptioned files. This should fix the issue with rerunning exceptioned/cancelled tests too. I'll add some extra tests for this before merging.

app/javascript/components/editor/TestRunSummaryContainer.tsx
iHiD: Yeah, that's fine.

kntsoriano:
@iHiD Should be ready. Just needs styling.

iHiD:
Really great work, man!

PR #1290, opened 2021-07-21
Add mentoring stats

ErikSchierboom:
- Add columns for num solutions mentored and satisfaction rating
- Add ViewComponentTestCase base class
- Show solutions mentored and satisfaction rating in mentor header
- Recalculate number of solutions mentored
- Recalculate satisfaction rating mentored
- Add discussion scope tests
- Use symbols in discussion scopes


app/models/mentor/discussion.rb
```diff
@@ -33,7 +33,9 @@ class Mentor::Discussion < ApplicationRecord
   scope :in_progress_for_student, -> { where(status: %i[awaiting_student awaiting_mentor mentor_finished]) }
   scope :finished_for_student, -> { where(status: :finished) }
   scope :finished_for_mentor, -> { where(status: %i[mentor_finished finished]) }
-  scope :not_negatively_rated, -> { where(rating: [nil, 3, 4, 5]) }
+  scope :not_negatively_rated, -> { where(rating: [nil, :acceptable, :good, :great]) }
+  scope :satisfactory_rated, -> { where(rating: %i[acceptable good great]) }
```
ErikSchierboom: Is `satisfactory_rated` the right word? It feels awkward, but I don't know what a better word would be.

test/helpers/view_components/view_component_test_case.rb
```diff
@@ -0,0 +1,13 @@
+require "test_helper"
+
+class ViewComponentTestCase < ActionView::TestCase
```
ErikSchierboom: I've created this helper class for the ActionView tests to allow signing in the current user, which I needed for the mentor header test

db/migrate/20210721064506_add_mentoring_stats_to_user.rb
```diff
@@ -0,0 +1,6 @@
+class AddMentoringStatsToUser < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :num_solutions_mentored, :integer, limit: 3, null: false, default: 0
```
ErikSchierboom, line 3: This allows for 8388607 solutions to be mentored before overflowing, which I don't think we'll reach anytime soon. 32767 seemed like a too low limit.

db/migrate/20210721064506_add_mentoring_stats_to_user.rb
```diff
@@ -0,0 +1,6 @@
...
+    add_column :users, :satisfaction_rating, :integer, limit: 1, null: false, default: 100
```
ErikSchierboom: The satisfaction rating is a value between 0 and 100, so it is actually a percentage. I consided renaming it, but was unsure if `satisfaction_rating_percentage` would be better.

I also considered making this a nullable field, but I've opted for not doing that and defaulting to 100% satisfaction rating instead. Let me know what you think.

app/models/mentor/discussion.rb
ErikSchierboom: This scope is only used in the recalculating of the satisfaction rating. Want me to inline it? I didn't yet because I could imagine it being useful somewhere else later on.

app/models/mentor/discussion.rb
```diff
@@ -33,7 +33,9 @@ class Mentor::Discussion < ApplicationRecord
...
+  scope :rated, -> { where.not(rating: nil) }
```
ErikSchierboom: This scope is only used in the recalculating of the satisfaction rating. Want me to inline it? I didn't yet because I could imagine it being useful somewhere else later on.

app/models/mentor/discussion.rb
```diff
@@ -33,7 +33,9 @@ class Mentor::Discussion < ApplicationRecord
   scope :in_progress_for_student, -> { where(status: %i[awaiting_student awaiting_mentor mentor_finished]) }
   scope :finished_for_student, -> { where(status: :finished) }
   scope :finished_for_mentor, -> { where(status: %i[mentor_finished finished]) }
-  scope :not_negatively_rated, -> { where(rating: [nil, 3, 4, 5]) }
+  scope :not_negatively_rated, -> { where(rating: [nil, :acceptable, :good, :great]) }
```
ErikSchierboom, line 36: I replaced the integer constants with their symbols.

app/models/mentor/discussion.rb
```diff
@@ -159,4 +163,34 @@ def awaiting_mentor!
...
+  def update_satisfaction_rating!
+    satisfactory_rated_count_sql = Arel.sql(
+      Mentor::Discussion.where(mentor_id: mentor.id).satisfactory_rated.select("COUNT(*)").to_sql
+    )
+
+    rated_count_sql = Arel.sql(
+      Mentor::Discussion.where(mentor_id: mentor.id).rated.select("COUNT(*)").to_sql
+    )
+
+    satisfaction_rating_sql = "CEIL((#{satisfactory_rated_count_sql}) / (#{rated_count_sql}) * 100)"
```
ErikSchierboom: I'm using `CEIL` to round up, which I felt was the friendly thing to do :)

app/models/mentor/discussion.rb
iHiD: 1.`satisfactorily` is probably the word you're after. But I'd probably have `rated_acceptable_or_better` or something.
2. Yeah - I think inline it.

test/helpers/view_components/view_component_test_case.rb
iHiD: :+1:

db/migrate/20210721064506_add_mentoring_stats_to_user.rb
iHiD: I think naming it as `mentor_satisfaction_percentage` and having it as `nullable`. Sorry! 🙂 

Then in `app/helpers/view_components/mentor/header.rb` only showing it if it's not null.

iHiD: (I think if someone's not done any mentoring we want to be able to know that via this rating too, basically)

app/models/mentor/discussion.rb
```diff
@@ -159,4 +163,34 @@ def awaiting_mentor!
       awaiting_student_since: nil
     )
   end
+
+  def update_num_solutions_mentored!
+    count_sql = Arel.sql(
+      Mentor::Discussion.where(mentor_id: mentor.id).finished_for_mentor.select("COUNT(*)").to_sql
```
iHiD:
```suggestion
      Mentor::Discussion.where(mentor: mentor).finished_for_mentor.select("COUNT(*)").to_sql
```

app/models/mentor/discussion.rb
```diff
@@ -55,6 +57,8 @@ def self.between(mentor:, student:)
 
   after_save_commit do
     solution.update_mentoring_status! if previous_changes.key?('status')
+    update_num_solutions_mentored! if previous_changes.key?('status')
+    update_satisfaction_rating! if previous_changes.key?('rating')
```
iHiD: So a couple of things:
1. These are going to need to go in a job (you can put both in one job called `UpdateMentorStatsJob` or something). I've checked on Bastion, and they'd add 60ms to the request, which I think is too long to have inline.
2. I don't think all the logic in `update_num_solutions_mentored!` etc belongs in discussion. I think each method should be it's own command. I think they can both be at the apex of `app/commands/mentor` (e.g. `Mentor::UpdateSatisfactionRating`)

app/models/mentor/discussion.rb
iHiD: Just slightly shorter. Makes no practical difference.

test/helpers/view_components/view_component_test_case.rb
iHiD: I'd consider adding this to `ActionView::TestCase` itself I think. (See how we do the same for integration tests in `test_helper.rb`). The its available everywhere rather than having to change the base class.

app/helpers/view_components/mentor/header.rb
```diff
@@ -35,10 +35,11 @@ def bottom
...
         tag.div(class: "stats") do
-          # TODO: (Required)
-          tag.div("131 solutions mentored", class: "stat") +
-            # TODO: (Required)
-            tag.div("95% satisfaction", class: "stat")
+          tag.div("#{current_user.num_solutions_mentored} solutions mentored", class: "stat") +
+            (if current_user.mentor_satisfaction_percentage
+               tag.div("#{current_user.mentor_satisfaction_percentage}% satisfaction",
+                 class: "stat")
+             end)
```
ErikSchierboom: Is this an acceptable way of doing this?

test/models/mentor/discussion_test.rb
```diff
@@ -268,4 +270,81 @@ class Mentor::DiscussionTest < ActiveSupport::TestCase
...
+
+      discussion_1.student_finished!
+      assert_equal 1, mentor.reload.num_solutions_mentored
+
+      discussion_2.finished!
+      assert_equal 2, mentor.reload.num_solutions_mentored
+
+      discussion_3.finished!
+      assert_equal 3, mentor.reload.num_solutions_mentored
+    end
```
ErikSchierboom, line 293: I've kept this test as an integration test that things are continually being updated

app/helpers/view_components/mentor/header.rb
iHiD: If it works!

I'd probably have switched to `safe_join` for readiability:
```ruby
safe_join(
  [
    tag.div("#{current_user.num_solutions_mentored} solutions mentored", class: "stat"),
    (tag.div("#{current_user.mentor_satisfaction_percentage}% satisfaction", class: "stat") if current_user.mentor_satisfaction_percentage)
  ].compact
)
```

But it all sort of depends on line-length as to what works.

test/models/mentor/discussion_test.rb
iHiD: Good decision. I think it would also make sense to add tests for the commands too (as I can see them being used outside of this scenario). They could just be copied/pasted from here. Not blocking merge though.

iHiD (approved):

app/helpers/view_components/mentor/header.rb
ErikSchierboom: rubocop rewrites that to:

```ruby
safe_join(
            [
              tag.div("#{current_user.num_solutions_mentored} solutions mentored", class: "stat"),
              (if current_user.mentor_satisfaction_percentage
                 tag.div("#{current_user.mentor_satisfaction_percentage}% satisfaction",
                   class: "stat")
               end)
            ].compact
          )
```

Which I still like better than my version

iHiD: OK :)

PR #1293, opened 2021-07-21
Add automated feedback and solving exercises locally links

ErikSchierboom:
See https://github.com/exercism/docs/pull/181

### DIFF
--- a/app/views/tracks/exercises/show/_action_box_completed.html.haml
+++ b/app/views/tracks/exercises/show/_action_box_completed.html.haml
@@ -26,4 +26,4 @@
   %h4 Work locally (via CLI)
   = ReactComponents::Common::CopyToClipboardButton.new(exercise.download_cmd)
 
-  = render ViewComponents::ProminentLink.new("Learn more about solving exercises locally", "#")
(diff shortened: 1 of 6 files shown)


app/views/tracks/exercises/show/_action_box_completed.html.haml
```diff
@@ -26,4 +26,4 @@
   %h4 Work locally (via CLI)
   = ReactComponents::Common::CopyToClipboardButton.new(exercise.download_cmd)
 
-  = render ViewComponents::ProminentLink.new("Learn more about solving exercises locally", "#")
+  = render ViewComponents::ProminentLink.new("Learn more about solving exercises locally", Exercism::Routes.doc_path('solving-exercises', 'locally'))
```
ErikSchierboom: I don't necessarily mind the duplication of the link, but there are now five different places where the same link is used using the same magic strings. Do you mind?

iHiD: Agreed. Makes sense to add a `Exercism::Routes.solving_exercises_locally_path` method to `config/initializers/routes.rb` which calls this.

PR #1295, opened 2021-07-22
Fix v1 API

ErikSchierboom:
https://github.com/exercism/website/pull/1215 normalized the parameter names, but unfortunately this broke the CLI as that requires very specific parameters to be used.

### DIFF
--- a/app/controllers/api/v1/solutions_controller.rb
+++ b/app/controllers/api/v1/solutions_controller.rb
@@ -71,14 +71,6 @@ module API
       end
 
       private
-      def set_track
-        @track = Track.find_by!(slug: params[:track_id])
-      end
-
-      def set_exercise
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/solutions_controller.rb
```diff
@@ -68,14 +71,6 @@ def update
...
 
       private
-      def set_track
-        @track = Track.find(params[:track_slug])
-      end
-
-      def set_exercise
-        @exercise = @track.exercises.find(params[:exercise_slug])
-      end
-
```
ErikSchierboom, line 78: These methods were not used

iHiD:
(FYI @kntsoriano in case we do further refactoring on this just so you know not to change these files - you weren't to know though so doesn't matter)

iHiD (approved):

PR #1300, opened 2021-07-24
Fix editor dark theme

kntsoriano:
@iHiD I had to delete some CSS to make this work. Could you check it?

### DIFF
--- a/app/css/pages/editor.css
+++ b/app/css/pages/editor.css
@@ -445,7 +445,6 @@
             @apply text-15 font-mono leading-160;
         }
         & .cm-gutters {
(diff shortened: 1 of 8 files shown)


iHiD:
@kntsoriano Nice work on this. 

1. I think we should disable the dark theme option until the whole site is dark theme (which we'll do quickly after we launch). So can you add a commit to this that hides the button pls?
2. Also, I think the CSS change should be reverted - the reason it needed changing is because the outer body theme doesn't change, but once that does change, the CSS that was there would be correct.

---

Btw - I rebased and force-pushed, so you'll need to run `git reset --hard origin/fix-themes`.

PR #1303, opened 2021-07-24
Add generic tooltip

iHiD:
@kntsoriano We are going to be using a consistent generic tooltip all over the place. I've added a component that wraps what you currently have, and used it in two places. `content` can either be a string or a JSX component. Let me know what you think! :)

### DIFF
--- a/app/javascript/components/editor/RunTestsButton.tsx
+++ b/app/javascript/components/editor/RunTestsButton.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import { GraphicalIcon } from '../common/GraphicalIcon'
-import { ExercismGenericTooltip } from '../misc/ExercismTippy'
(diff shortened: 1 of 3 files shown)


kntsoriano:
@iHiD Am I correct in saying that this is going to be a component on top of `<ExercismTippy />`? This implies that we use `<ExercismTippy />` if we want to create a very special tooltip, but we should use `<GenericTooltip />` if we just want to have tooltips that we sprinkle around the site?

I'm asking about this since we might just want to replace `<ExercismTippy />` with `<GenericTooltip />` if that isn't the case.

iHiD:
> This implies that we use <ExercismTippy /> if we want to create a very special tooltip, but we should use <GenericTooltip /> if we just want to have tooltips that we sprinkle around the site?

That's correct, yeah. So these are specifically ones with arrows that are black background with white writing etc.

PR #1306, opened 2021-07-26
Fix editor scroll position

kntsoriano:
https://github.com/exercism/v3-beta/issues/143

This is a known error for Turbolinks, and I copied the solution here: https://github.com/turbolinks/turbolinks/issues/181#issuecomment-735246518

It seems to me that https://github.com/turbolinks/turbolinks/pull/420 is the PR for this fix, but it won't seem to get merged anytime soon. Should we look into using Hotwire now?

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -7,7 +7,7 @@
     = csp_meta_tag
     = action_cable_meta_tag
 
-    = javascript_packs_with_chunks_tag(*js_packs, 'data-turbolinks-track': 'reload')
+    = javascript_packs_with_chunks_tag(*js_packs)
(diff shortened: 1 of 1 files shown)


iHiD:
Hmmm, it looks like this broke the tests.

iHiD:
@kntsoriano I think we want Turbo, not Hotwire (Hotwire also installs Stimulus, which we don't want to use at least for now).

Spend a day working on replacing Turbolinks with Turbo. See whether it works or not. If it looks like it'll be too painful, we'll just abort, but if we can do it, that'd be good 🙂

iHiD:
@kntsoriano This would be a good one for you to look at next week while I'm away.

iHiD:
@kntsoriano Can we close this now?

PR #1327, opened 2021-07-28
Highlight tests in tests panel

kntsoriano:
https://github.com/exercism/v3-beta/issues/141

@iHiD We just need to style this. Probably just add `width: 100%` to the div.

### DIFF
--- a/app/javascript/components/Editor.tsx
+++ b/app/javascript/components/Editor.tsx
@@ -58,7 +58,6 @@ type EditorPanels = {
   }
   tests?: {
     tests: string
-    language: string
   }
   results: {
     averageTestDuration: number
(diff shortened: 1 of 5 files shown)


app/javascript/components/Editor.tsx
```diff
@@ -378,7 +377,9 @@ export function Editor({
                   <ResultsTab />
                 </div>
                 <InstructionsPanel {...panels.instructions} />
-                {panels.tests ? <TestsPanel {...panels.tests} /> : null}
+                {panels.tests ? (
+                  <TestsPanel language={track.slug} {...panels.tests} />
```
iHiD: @kntsoriano The `track.highlightjs_language` needs to be used for anything with highlightjs, not the `slug`. I've fixed this.

PR #1347, opened 2021-08-02
Add Legacy files

iHiD:
@kntsoriano So, this is for readonly files in the editor.

There are files that a student might have submitted via the CLI, which they need to **see** in the editor, but should **not be able to edit**. This might happen because:
- You can upload anything in the CLI and they uploaded supporting files
- They uploaded a solution and the exercise has since had the expected filename renamed, so they have a new file and an old stub.

There's various possibilities here we need to work with but for now I want to handle the basic one, which is:
- Files get a ` type` key. It can currently be `exercise`, `solution`, `legacy`. (This is done in this PR on the Ruby side)
- Files with `exercise` or `solution` behave as normal.
- Files with `legacy` are readonly. They cannot be edited.
- Files with `legacy` have an extra banner when they open. [1]
- Files with `legacy` can be deleted within the editor. This doesn't do anything outside of the editor, but means in the editor that file no longer exists and when the next submission is made it is without that file (and we should persist this fact in local storage).

Also, we also need to have support for multiple files in the editor, which I'm not sure we do atm. It doesn't seem to work for me. You should be able to submit an iteration then do `Submission.last.files.create(filename: "foobar.rb", content: "# something")` and it should appear in the editor.

Also, you might want to consider https://github.com/exercism/v3-beta/issues/145 at the same time as doing this. I think it's a good idea, but, depending on how "readonly" works it might have a different codepath entirely, as we might not want to lose the cursor position while the tests run maybe 🤷 See what you come up with 🙂 

---

[1]
Screenshot 2021-08-02 at 15 12 35
Screenshot 2021-08-02 at 15 12 37

### DIFF
--- a/app/controllers/api/solutions/last_iteration_files_controller.rb
+++ b/app/controllers/api/solutions/last_iteration_files_controller.rb
@@ -9,6 +9,7 @@ module API
         end
 
         return render_403(:solution_not_accessible) unless current_user.may_view_solution?(solution)
+        return render_400(:no_iterations_submitted_yet) if solution.latest_iteration.blank?
 
         files = solution.latest_iteration.files.map do |file|
           {
--- a/app/css/components/iteration-pane.css
+++ b/app/css/components/iteration-pane.css
@@ -3,12 +3,14 @@
     @apply bg-backgroundColorB;
 
     & .tabs {
+        @apply flex items-center;
         @apply bg-backgroundColorB;
         @apply border-b-1 border-borderColor7;
+
         & .c-tab {
             @apply grid place-items-center;
             @apply bg-backgroundColorB;
-            @apply px-16 py-8;
+            @apply px-16 py-8 mr-0;
             @apply text-13 text-textColor6 leading-paragraph;
             @apply border-r-1 border-borderColor7;
             @apply rounded-none;
--- a/app/css/pages/editor.css
+++ b/app/css/pages/editor.css
@@ -200,6 +200,17 @@
                 }
             }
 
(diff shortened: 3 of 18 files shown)


kntsoriano:
@iHiD This should be ready.

kntsoriano:
@iHiD I'll work on https://github.com/exercism/v3-beta/issues/145 once this is merged.

iHiD:
@kntsoriano This doesn't compile properly (asset tests fail)

iHiD:
~I can't get this branch to work: https://www.loom.com/share/e58011935eda4bd28d5584ddd59a4564~

OK - I got it to work by clearing local storage. I think “reverting to latest iteration” should effectively do the same though. I'll open a seperate issue for this.

iHiD:
- [ ] I think this bit needs improving.

https://www.loom.com/share/12a56f8cb47a4526850d25b6e8cb4d1a

kntsoriano:
Done.

I've also added https://github.com/exercism/v3-project-management/issues/301 to this.

Doing the issue above made me think, though, if we want a "revert to last submission" button too?

iHiD:
> Doing the issue above made me think, though, if we want a "revert to last submission" button too?

I think not for now. I think people can undo for code, and if they want to get a whole file back, then can go back to the iteration. But maybe as a follow-on post launch. Thanks for doing this! :)

iHiD:
Nice work. There's still a little bug, but I'll merge and I've opened the bug as https://github.com/exercism/v3-beta/issues/153

PR #1350, opened 2021-08-03
Handle currency precision on client side better

kntsoriano:
https://github.com/exercism/v3-beta/issues/148

@iHiD Could you check the modal CSS? I might've broken it.


iHiD:
@kntsoriano. Nice work. There's one system test that's still failing that you need to look at as I don't know why it's failing.

PR #1354, opened 2021-08-04
Set editor readonly when tests are running

kntsoriano:
https://github.com/exercism/v3-beta/issues/145


iHiD:
Works great. Nice work :)

PR #1363, opened 2021-08-05
Fix reverting with legacy files

### LINKED ISSUE #153: [Editor] When going back to last iteration, legacy files do not get the legacy bar
https://www.loom.com/share/9a9d3dcfbbf04227a0cdce18f62a51d2

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/153.

@iHiD Let me know if you are okay with this refactoring.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -87,10 +87,12 @@ class Submission < ApplicationRecord
   end
 
   def solution_files
-    files.each_with_object(solution.exercise_solution_files) do |file, files|
-      type = files.key?(file.filename) ? :solution : :legacy
+    # Merge the submission files into the exercise files. If we find a 
+    # file we don't expect, that it as type: :legacy
+    files.each_with_object(solution.exercise_solution_files) do |file, merged_files|
+      type = merged_files.key?(file.filename) ? :solution : :legacy
 
-      files[file.filename] = {
+      merged_files[file.filename] = {
         type: type,
         content: file.content,
         digest: file.digest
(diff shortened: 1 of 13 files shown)


iHiD:
Hey man :)

app/models/submission.rb
```diff
@@ -86,6 +86,18 @@ def viewable_by?(user)
...
+    files.each_with_object(solution.exercise_solution_files) do |file, files|
+      type = files.key?(file.filename) ? :solution : :legacy
+
+      files[file.filename] = {
+        type: type,
+        content: file.content,
+        digest: file.digest
+      }
+    end
+  end
```
iHiD: This is quite confusing because we have an external `files` method and the same named (but different) internal `files` block param.

Also I found the intersection of the two arrays in the `each_with_object` quite confusing to reason about. But I think maybe just a comment that explains that is best.

```suggestion
    # Merge the submission files into the exercise files. If we find a 
    # file we don't expect, that it as type: :legacy
    files.each_with_object(solution.exercise_solution_files) do |file, merged_files|
      type = merged_files.key?(file.filename) ? :solution : :legacy

      merged_files[file.filename] = {
        type: type,
        content: file.content,
        digest: file.digest
      }
    end
  end
```

iHiD:
Yeah, I'm fine with this refactor - it's good. One suggestion to make it a little clearer.

test/support/redirect_helpers.rb
```diff
@@ -0,0 +1,6 @@
+module RedirectHelpers
+  def wait_for_redirect
+    assert_css ".c-loading-overlay.--loading"
+    assert_no_css ".c-loading-overlay.--loading"
+  end
+end
```
iHiD, line 6: Nice.

PR #1368, opened 2021-08-06
Add notifications page

iHiD:


kntsoriano:
@iHiD This should be ready. Here are the following things I've taken into consideration:

1. Instead of updating our collection from the API's repsonse, we instead just refetch the collection again. I decided to do this because we don't want to return a very long API response. Also, if we decide to update things on the client side, this means that our data is not in sync with the server's. This implies though that every time that we fetch, the list is reordered which doesn't lead to a good user experience. This is because the API returns the unread notifications first on the list. I think we should, instead, return notifications chronologically on this page. I haven't done this yet. Let me know what you think.
2. For the "mark as read" and "mark as unread" button, I've put the error messages underneath each button for now. I don't know where to place this and whether we want a centralized space to show all error messages. Let me know what you think.
3. In relation to above, the error message for the "mark all as read" button is within the modal itself.

Other notes:
1. We need to reset the rendering cache data for the notifications.

PR #1376, opened 2021-08-13
Analysis modal

kntsoriano:
https://github.com/exercism/v3-project-management/issues/304

I decided to take the following shortcuts as we need to discuss them further:
1. I decided to make a crude implementation of the controller, but it isn't namespaced properly. Also, it doesn't include checking on whether the user accessing the controller has the right permissions to do so.
2. I didn't include the student info (top right hand side of modal) because it was too complicated for me to include and this led me to think whether it was really necessary to do so. I was looking at the code behind this, and it should allow for "anonymous mode" and the such. If it is necessary, I will include it.

Other notes:
1. I've also seen a placeholder link which we haven't filled in yet. I've added a TODO.
2. This only needs styling left.


kntsoriano:
@iHiD The failing tests for this have been fixed in https://github.com/exercism/website/pull/1363.

PR #1381, opened 2021-08-17
Upgrade to Turbo

kntsoriano:
https://github.com/exercism/v3-beta/issues/143

@iHiD I tried using Turbo Frames and it's a bit more complicated than I expected. Firstly, there aren't events to listen to when a frame loads so we can't load the bootloader JS to render the React components. Also, we should wrap everything within a `turbo_frame_tag 'body'` to rerender just the body. I will spend more time on this when I get through the other bugs.

The good thing is that we're on Turbo now and the issue described above seems to be fixed.


app/controllers/concerns/turbo/redirection.rb
```diff
@@ -0,0 +1,29 @@
+module Turbo
+  module Redirection
+    extend ActiveSupport::Concern
+
+    def redirect_to(url = {}, options = {})
+      turbo = options.delete(:turbo)
+
+      super.tap do
+        visit_location_with_turbo(location, turbo) if turbo != false && request.xhr? && !request.get?
```
iHiD:
```suggestion
        visit_location_with_turbo(location, turbo) if turbo && request.xhr? && !request.get?
```

app/controllers/concerns/turbo/redirection.rb
```diff
@@ -0,0 +1,29 @@
+module Turbo
+  module Redirection
+    extend ActiveSupport::Concern
+
+    def redirect_to(url = {}, options = {})
+      turbo = options.delete(:turbo)
```
iHiD, line 6: What's the reason we `delete` it here, rather than just accessing it?

iHiD:
Nice work, @kntsoriano. I'm taking a look at the frame thing now :)

iHiD:
@kntsoriano I've added a commit with turboframes working. Let me know what you think :)

kntsoriano:
@iHiD Looks good! Let's just make the tests pass and we can merge this.

iHiD:
@kntsoriano Could you get them green pls?

kntsoriano:
@iHiD The reason that some of our tests were failing was that we were loading the site header in the onboarding page. In order to fix this, I decided to not show it on that page, and when the onboarding form submits, we make a full page reload. This makes sure that the header and footer are reloaded along it.

I've been thinking more about how we render frames and such, and I think we need to get it standardized and decide on whether on some pages we just want to hide the header and footer versus not rendering them at all. It would be nice to talk about this in the future.

iHiD:
> @iHiD The reason that some of our tests were failing was that we were loading the site header in the onboarding page. In order to fix this, I decided to not show it on that page, and when the onboarding form submits, we make a full page reload. This makes sure that the header and footer are reloaded along it.

The problem with this is that if someone hits the onboarding page, then navigates off, they get taken to a page with no header, so I'm not sure this will work. Maybe they can't navigate off the page, so it's ok, but that feels brittle to me for us to remember. A lot of the code I wrote before was to have the header on all pages to avoid these problems. I feel really wary relying on remembering to always skip turbo on links. But I'll check the branch out and see! 🙂

iHiD:
@kntsoriano I've added a commit which handles the underlying error by allowing the notifications and reputation API calls to be hit when someone is not onboarded. I think this should fix the issue without us having custom rules around the site header, which I prefer. That ok with you?

iHiD:
> I've been thinking more about how we render frames and such, and I think we need to get it standardized and decide on whether on some pages we just want to hide the header and footer versus not rendering them at all. It would be nice to talk about this in the future.

So previously we didn't render them. I had a commit which then added `target: "_top"` in loads of places, but it just felt super-brittle. So I decided it was better just to render the header on all pages and just not show it if it's not needed. That felt much more robust. Happy to chat it through more though! :)

PR #1387, opened 2021-08-17
Update journey headers

kntsoriano:
@iHiD You can start styling this. ~~I'm opening it as a draft because the "reset filters" button for the solutions list isn't implemented yet.~~

Edit: I've already added the reset filters button to this PR.


iHiD:
@kntsoriano Nice work - this is great. I've added a commit with some styling, and also simplified the HTML somewhat. Worth you taking a look for if we add more of these in the future 🙂

PR #1394, opened 2021-08-18
Fix mentoring inbox track filter

### LINKED ISSUE #182: Mentoring track filter is not applied on refresh
If I go to my mentoring inbox and select a track to filter on, I see the filtering being applied and the URl changes. However, pressing F5 does not result in the filtering being applied. 

image

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/182.


iHiD:
Thanks. Will merge, but please follow up with a new test case so we don't get regressions on this :)

PR #1395, opened 2021-08-18
Hide chevron if contributor has no profile link

### LINKED ISSUE #183: [Contributors List] Some list items are not clickable.
https://exercism.lol/contributing/contributors
If I click on one of the contributors in the list, nothing happens. I would expect to get redirected to the contributors' profile.

Correction: It seems it only happens for some contributors. Probably the ones without public profiles.
Feel free to close the issue if the current state is the intended behavior for now.

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/183.

@iHiD The CSS has to be fixed. This is how it looks right now:
Screen Shot 2021-08-18 at 1 55 00 PM


iHiD:
Thanks :)

PR #1399, opened 2021-08-19
Add donation footer form

kntsoriano:
Closes https://github.com/exercism/v3-project-management/issues/307.

@iHiD This needs styling. You can view the component by going to `local.exercism.io:3020/test/components/donations/footer_form`.


iHiD:
@kntsoriano See https://www.loom.com/share/230daa193a944775acdc7aa32cf589ae

iHiD:
You can also remove the temporary endpoint as it's now in the footer.

iHiD:
@kntsoriano Don't forget this one 🙂

kntsoriano:
@iHiD This was pretty complex! I'm pretty sure there's going to be a bug somewhere. Just let me know what I have to fix.

kntsoriano:
@iHiD System tests are so annoying. I hope it works now.

kntsoriano:
@iHiD I've disabled some of the fade out transitions as I think we need to apply a class when we are fading out. Otherwise, I think we won't be able to disable animations in Capybara which will make our tests flaky.

I've gone through these and they still look file on production, just not as snappy with the fade out transitions.

iHiD:
> @iHiD I've disabled some of the fade out transitions as I think we need to apply a class when we are fading out. Otherwise, I think we won't be able to disable animations in Capybara which will make our tests flaky.

Can you explain a little more what you mean here by "applying a class when we are fading out" pls? I don't really want to start disabling animations and making the site feel less good in order to get the tests to not be flakey - I think we need to work out how to solve this properly instead (I know that's what you've been trying to do 🙂)

iHiD:
I've reenabled animations and going to cope with tests been a little flakey atm. And then we can eventually sort this out post-launch! :) Thanks for your effort in fixing things :)

PR #1401, opened 2021-08-19
Add links to share panel

kntsoriano:
@iHiD This adds the links to the share panel. We need to customize the following though:

1. What do we want as the default message on the user's post when sharing a testimonial?
2. What do we want as the default message on the user's post when sharing a solution?

Also, since we are sharing these both above, we need to add the Facebook and Twitter meta tags to the pages.


iHiD:
What you've done here is fine :)

PR #1407, opened 2021-08-19
Output .meta/config.json as .exercism/config.json

ErikSchierboom:


app/models/git/exercise.rb
```diff
@@ -176,12 +183,12 @@ def important_absolute_filepaths
 
     memoize
     def cli_filepaths
-      special_filepaths = [SPECIAL_FILEPATHS[:readme], SPECIAL_FILEPATHS[:help]]
+      special_filepaths = [SPECIAL_FILEPATHS[:config], SPECIAL_FILEPATHS[:readme], SPECIAL_FILEPATHS[:help]]
```
ErikSchierboom: An alternative would be to do:

```suggestion
      special_filepaths = SPECIAL_FILEPATHS
```

and then delete the hints file on the next lineunless `filepaths.include?(hints_filepath)`

app/models/solution.rb
```diff
@@ -219,9 +219,10 @@ def sync_git!
   end
 
   def read_file(filepath)
-    return Solution::GenerateReadmeFile.(self) if filepath == 'README.md'
-    return Solution::GenerateHelpFile.(self) if filepath == 'HELP.md'
-    return Solution::GenerateHintsFile.(self) if filepath == 'HINTS.md'
+    return Solution::GenerateReadmeFile.(self) if filepath == Git::Exercise::SPECIAL_FILEPATHS[:readme]
+    return Solution::GenerateHelpFile.(self) if filepath == Git::Exercise::SPECIAL_FILEPATHS[:help]
+    return Solution::GenerateHintsFile.(self) if filepath == Git::Exercise::SPECIAL_FILEPATHS[:hints]
+    return git_exercise.read_file_blob(exercise.git.config_filepath) if filepath == Git::Exercise::SPECIAL_FILEPATHS[:config]
```
ErikSchierboom, line 225: I wanted to try using the special file paths here, but I don't think it becomes more readable. Maybe hardcoding is better?

app/models/git/exercise.rb
```diff
@@ -7,11 +7,18 @@ class Exercise
...
-                  instructions_append: ".docs/instructions.append.md",
-                  introduction: ".docs/introduction.md",
-                  introduction_append: ".docs/introduction.append.md",
-                  hints: ".docs/hints.md",
-                  config: ".meta/config.json"
+      instructions_append: ".docs/instructions.append.md",
+      introduction: ".docs/introduction.md",
+      introduction_append: ".docs/introduction.append.md",
+      hints: ".docs/hints.md",
+      config: ".meta/config.json"
```
ErikSchierboom, line 15: This was rubocop

app/models/solution.rb
iHiD: I think this is better code now. And we can make `SPECIAL_FILEPATHS` cleverer later? Like 

`if Git::Exercise::SPECIAL_FILEPATHS.readme?(filepath)`

iHiD:
Do we need to also guard against people uploading `.exercism` or `.meta` dirs?

ErikSchierboom:
> Do we need to also guard against people uploading .exercism or .meta dirs?

We already guarded against the latter. I've added a check for the former.

ErikSchierboom:
Rubocop fails due to a mismatch between the version in the Gemfile.lock and the version used in the `tests.yml` workflow. I've tried to fix this in https://github.com/exercism/website/pull/1413 but hit another issue.

app/models/git/exercise.rb
iHiD: Maybe this?
```
special_filepaths = SPECIAL_FILEPATHS.values_at(:config, :readme, :help)
```

iHiD:
Nice. Happy for this to be merged?

ErikSchierboom:
Yes!

PR #1413, opened 2021-08-20
Sync rubocop version

ErikSchierboom:
**Blocked by a bug in Rubocop - awaiting new version**

---

- Sync rubocop version
- Update Rubocop gem
- Format using rubocop


.github/workflows/tests.yml
```diff
@@ -25,10 +25,16 @@ jobs:
...
         run: |
-          gem install rubocop -v 1.7.0
-          gem install rubocop-minitest -v 0.11.0
-          gem install rubocop-performance -v 1.10.1
-          gem install rubocop-rails -v 2.9.1
+          # Extract the exact versions of the rubocop gems from Gemfile.lock
+          rubocop_version=$(sed -n -E 's/^    rubocop \((.+)\)/\1/p' Gemfile.lock)
+          rubocop_minitest_version=$(sed -n -E 's/^    rubocop-minitest \((.+)\)/\1/p' Gemfile.lock)
+          rubocop_performance_version=$(sed -n -E 's/^    rubocop-performance \((.+)\)/\1/p' Gemfile.lock)
+          rubocop_rails_version=$(sed -n -E 's/^    rubocop-rails \((.+)\)/\1/p' Gemfile.lock)
```
ErikSchierboom, line 32: A little bit of sed magic to ensure that we always use the same rubocop gems here as used to format our code

Gemfile.lock
```diff
@@ -339,16 +339,16 @@ GEM
     reverse_markdown (2.0.0)
       nokogiri
     rexml (3.2.5)
-    rubocop (1.18.4)
+    rubocop (1.19.1)
```
ErikSchierboom: Upgrading fixed an infinite loop formatting issue

ErikSchierboom (commented):
After upgrading, we're now hitting the following bug: https://github.com/rubocop/rubocop/pull/10017

It has been fixed but no new release has yet been built.

.github/workflows/tests.yml
iHiD: Nice.

PR #1445, opened 2021-08-22
Hide favorite button

### LINKED ISSUE #176: Removing as favourite is broken.
https://www.loom.com/share/e1155a79c2f943c2982dbe7c789cfec3

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/176.

### DIFF
--- a/app/commands/mentor/student_relationship/toggle_favorited.rb
+++ b/app/commands/mentor/student_relationship/toggle_favorited.rb
@@ -6,13 +6,17 @@ module Mentor
       initialize_with :mentor, :student, :favorited
 
       def call
-        return unless Mentor::Discussion.between(mentor: mentor, student: student).exists?
+        return unless allowed?
 
         relationship = Mentor::StudentRelationship.create_or_find_by!(
           mentor: mentor,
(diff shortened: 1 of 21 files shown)


app/commands/mentor/student_relationship/toggle_favorited.rb
```diff
@@ -6,14 +6,18 @@ class ToggleFavorited
...
+        return unless allowed?
 
         relationship = Mentor::StudentRelationship.create_or_find_by!(
           mentor: mentor,
           student: student
         )
         relationship.update_column(:favorited, favorited)
       end
+
+      def allowed?
```
kntsoriano, line 18: @iHiD I think the code will be cleaner if this lives within the `Mentor::StudentRelationship` model. I didn't do that right now though and opted for this change as I'll leave that for you to decide.

iHiD: OK, thanks. I'm ok with it here for now, but I agree that if we use it in multiple places it would be worth adding elsewhere :)

PR #1451, opened 2021-08-23
Remove digest in editor files serializer

### LINKED ISSUE #203: `Run Tests` button behaves weirdly after the page refreshes
https://user-images.githubusercontent.com/73734376/130339746-46d5e020-2803-4d7b-b51e-393851f3a672.mp4

So whenever the user runs the tests, the `Run Tests` button is `enabled`, and after the tests are completed, the button is `disabled`. But when the page is refreshed, the button is `enabled` even if the code is the same, and clicking it does nothing and the user cannot submit their code as the `Submit` button is `disabled`

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/203.

### DIFF
--- a/app/helpers/react_components/editor.rb
+++ b/app/helpers/react_components/editor.rb
@@ -7,7 +7,7 @@ module ReactComponents
         "editor",
         {
           default_submissions: submissions,
-          default_files: SerializeFiles.(solution.solution_files),
+          default_files: SerializeEditorFiles.(solution.solution_files),
           default_settings: {
             tab_size: track.indent_size,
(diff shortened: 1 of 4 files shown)


iHiD:
Thanks. How did this fix it?

(I've added a test btw)

kntsoriano:
@iHiD When we check if files changed, we compare the whole files object. With the digest, we won't be able to match it unless we get the digest from the server. I just removed it since we don't need it anyway. An alternative solution would just be to just compare the contents of the files itself versus the whole files object.

iHiD:
Ah, gotcha. Thanks :)

PR #1461, opened 2021-08-24
[WIP] Don't include current discussion in previous sessions modal

### LINKED ISSUE #5750: The prior sessions mentored is perhaps off by one?
image

But then, when I click on it, it shows 2 instead of 1.

image

kntsoriano:
Closes https://github.com/exercism/exercism/issues/5750

@iHiD The major change to this was that it pulls the information from the API instead of relying on the `numDiscussions` variable previously passed. This is because we view this button in two contexts: (1) within a mentoring discussion and (2) within a mentoring request. If we're inside a mentoring request, we shouldn't exclude, and if we're inside a mentoring discussion we should exclude. For each of those contexts, I decided to just create separate endpoints for them as I feel this is the cleaner solution and wouldn't involve any data mismatches between the client and server.


app/controllers/api/mentoring/discussions/previous_sessions_controller.rb
```diff
@@ -0,0 +1,24 @@
+module API
```
kntsoriano, line 1: @iHiD I decided to let the server handle this instead of the excluding the current discussion from the client. If we just exclude the current discussion, it will be hard for us to recreate the correct pagination.

app/helpers/react_components/mentoring/session.rb
```diff
@@ -56,6 +56,11 @@ def mentor_student_relationship
       def links
         {
           mentor_dashboard: Exercism::Routes.mentoring_inbox_path,
+          previous_sessions: if discussion
+                               Exercism::Routes.api_mentoring_discussion_previous_sessions_path(discussion)
+                             else
+                               Exercism::Routes.api_mentoring_discussions_path(student: student.handle, status: :all)
+                             end,
```
iHiD: Extract to method pls?

iHiD:
Cool - looks good. Tests are red though 🙂

iHiD:
Thanks dude. Can we have a chat about this one? There's a lot going on here and I want to be sure I understand it. 

There's a couple of concerns I have:
- We seem to now have an extra API request to be able to render the link to the modal, which also renders all the extra discussions. That feels very heavy for me - extra request for the user + extra load on the server. So I don't think we should have an extra API request until the link is clicked.
- While I get why you want to handle the exclusion of the discussion serverside, I don't really like having the API endpoint as it is. I think it would be better to have the endpoint as a student/discussions endpoint and pass the excluding uuid to that. I think that should still work with pagination etc?

Let's discuss both of these points. But I think let's do it after launch - there are probably more important bugs to fix beforehand?

ErikSchierboom:
Superseded by https://github.com/exercism/website/pull/2704

PR #1467, opened 2021-08-25
Allow custom docs branch

ErikSchierboom:
- Allow passing branch to Git repository
- Allow overriding docs branch via env variable

### DIFF
--- a/app/models/git/repository.rb
+++ b/app/models/git/repository.rb
@@ -20,7 +20,7 @@ module Git
         @repo_url = repo_url || "https://github.com/exercism/#{repo_name}"
       end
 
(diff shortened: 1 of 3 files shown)


app/models/git/repository.rb
```diff
@@ -20,6 +20,8 @@ def initialize(repo_url: nil)
         @repo_url = repo_url || "https://github.com/exercism/#{repo_name}"
       end
 
+      @branch_ref = branch_ref || ENV.fetch("GIT_CONTENT_BRANCH", MAIN_BRANCH_REF)
```
iHiD:
```suggestion
      @branch_ref = branch_ref || ENV["GIT_CONTENT_BRANCH"].presence || MAIN_BRANCH_REF
```

We need to use `presence` here to guard against an empty string. Fetch only checks for `nil`.

PR #1474, opened 2021-08-25
Allow student to cancel mentoring request

kntsoriano:
@iHiD This works fine. Could you check the following please?

1. Confirmation modal copy
2. Backend implementation (API endpoint, `Mentor::Request::Cancel` command)


iHiD:
Will do man!

PR #1483, opened 2021-08-26
Fix mentoring queue

kntsoriano:
When I was playing around with the mentoring queue, I was confused on how to use it as the right things weren't reset when switching tracks and exercises.


iHiD:
Nice :)

PR #1486, opened 2021-08-26
Add guidance link

### LINKED ISSUE #254: Mentor notes has temporary text
This is seen in all mentor requests and is not from a specific exercise.
Screen Shot 2021-08-26 at 12 14 01 PM

iHiD:
@kntsoriano See the two TODOs for you here.

- Firstly, `exercise.slug` doesn't seem to be set, but `exercise.id` is set (which is unexpected)
- Secondly, the URL should probably come from Rails in the normal way you do things.

The tests will also need fixing.

---

Fixes https://github.com/exercism/v3-beta/issues/254


kntsoriano:
@iHiD This should be ready.

iHiD:
@kntsoriano Tests fail and needs rebasing.

iHiD:
@kntsoriano Think you missed this one :)

app/models/exercise.rb
```diff
@@ -93,6 +93,8 @@ def download_cmd
     "exercism download --exercise=#{slug} --track=#{track.slug}".freeze
   end
 
+  def notes; end
```
iHiD: Is this left over accidentally? I can't see it being used anywhere.

app/models/exercise.rb
```diff
@@ -108,6 +110,10 @@ def icon_url
     "#{Exercism.config.website_icons_host}/exercises/#{icon_name}.svg"
   end
 
+  def mentoring_notes_url
```
iHiD: Nice refactor :)

```suggestion
  def edit_mentoring_notes_url
```

(as this takes you directly into GH's edit interface)

kntsoriano: Thanks!

app/models/exercise.rb
kntsoriano: Yeah. I forgot to remove it.

app/css/components/mentor/discussion.css
```diff
@@ -511,6 +511,9 @@
         }
         & #panel-guidance {
             @apply py-16 px-24;
+
+            & .notes-cta {
+            }
```
iHiD:
```suggestion
```

PR #1493, opened 2021-08-26
Fix CSS issue with legacy banner

### LINKED ISSUE #237: [Editor] Old file dialog cuts off end of file; disappears without clicking on it
I have a lot of solved Elixir exercises with old files. There's this dialog at the top of the file. Before making a decision on the dialog, I wanted to see the whole file, but it was impossible to do normal scrolling to reach the end of the page. So what I did instead is to press and hold a mouse button inside of the file and then drag it down. That revealed the rest of the file, but when I tried to scroll back up again to see the dialog, it was gone. Reloading the page didn't help.

iHiD:
Fixes https://github.com/exercism/v3-beta/issues/237


iHiD:
@kntsoriano I've changed from hidden attribute to hidden css class, because the hidden attribute gets overriden if you set a css class in a stylesheet, so it doesn't work when we start styling those elements.

Could you update the JS tests to work with this pls.

PR #1495, opened 2021-08-26
Remove old statuses

iHiD:
@kntsoriano I think these are both legacy and handled more prettily elsewhere. Can you confirm pls? For the two that are remaining, are they used? If so, how do I replicate so I can style?

### DIFF
--- a/app/css/pages/editor.css
+++ b/app/css/pages/editor.css
@@ -14,6 +14,10 @@
     width: 100%;
     height: 100%;
 
+    & .editor-status {
+        @apply px-20;
+    }
+
     & .header {
         @apply bg-backgroundColorA;
         @apply border-b-1 border-borderColor6;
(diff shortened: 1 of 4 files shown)


kntsoriano:
@iHiD These stauses are still used, so I've reverted your previous commit. In order for you to see how this look inside the editor, I've added comments to the relevant tests so you could replicate them.

The other statuses `CREATE_SUBMISSION_FAILED` and `REVERT_FAILED` appear in the same place as these messages. We currently don't have a way to replicate them via system tests and they are reserved only as error messages if something unknown to us happens (server shutting down, etc).

iHiD:
@kntsoriano The submitting one definitely shouldn't display. That needs to be removed as it shows up every time you submit an exercise. I'm less fussed about the others.

kntsoriano:
@iHiD This should be ready. I just added a small padding on the message in order for it to have space between the run tests button.

iHiD:
@kntsoriano Thanks. I've added a delay so that you don't get a flash if it takes <0.3s to make the request.

PR #1496, opened 2021-08-26
Add missing files to mentoring

iHiD:
@kntsoriano I realised we were missing the instructions and tests files on the mentoring dashboard, so I've added them.

Screenshot 2021-08-26 at 22 11 20

I think this implementation is a bit gross, but as it's blocking launch, I figured it's ok for now, and its something you can refactor later. I'll let you check it over though before merging :)


app/javascript/components/mentoring/session/FilePanel.tsx
```diff
@@ -60,6 +76,32 @@ export const FilePanel = ({
...
+          {instructions ? (
+            <Tab.Panel
+              key="instructions"
+              id="instructions"
+              context={TabsContext}
+            >
+              <div
+                className="p-16 c-textual-content --small"
+                dangerouslySetInnerHTML={{ __html: instructions }}
+              />
```
iHiD, line 88: Note that this isn't a highlightJS file so I couldn't just handle it by adding it to the files array (and even when I did initially try that it got horribly messy fast)

iHiD:
I'll look at these test failures tomorrow unless anyone quickly understands them.

kntsoriano:
@iHiD I've fixed the tests. But should we catch the error for when a solution doesn't have tests? Or do we check for that before an exercise gets published?

iHiD:
> But should we catch the error for when a solution doesn't have tests? Or do we check for that before an exercise gets published?

Yeah - they should always have tests. Good thinking though 🙂

iHiD:
Thanks for fixing the tests!

PR #1497, opened 2021-08-26
Add readonly editor files

### LINKED ISSUE #275: Concept Exercise "Editor" files Not Copied/Appearing in Code Editor
For the Python [`cater-waiter` (`sets`) exercise](https://github.com/exercism/python/tree/main/exercises/concept/cater-waiter), a couple of helper files were specified in the exercise [`editor` key](https://github.com/exercism/python/blob/main/exercises/concept/cater-waiter/.meta/config.json) for use in coding/testing.

For the Elixir track, [`need for speed` (`alias` and `import`)](https://github.com/exercism/elixir/tree/main/exercises/concept/need-for-speed) has a similar need/usage.  

These files aren't displaying in the UI as read-only, and are not being copied to the Test-Runner or Analyzer.

iHiD:
We have not yet added support for readonly "editor" files. This PR does that. I need to add some tests to this but I think it'll work. I also renamed `solution_files` to `files_for_editor` as part of this.

Closes https://github.com/exercism/v3-beta/issues/275 (Take a read for an example exercise to test with)

### DIFF
--- a/app/controllers/api/solutions/initial_files_controller.rb
+++ b/app/controllers/api/solutions/initial_files_controller.rb
@@ -10,7 +10,7 @@ module API
 
         return render_403(:solution_not_accessible) unless current_user.may_view_solution?(solution)
 
-        files = SerializeFiles.(solution.exercise_solution_files)
+        files = SerializeFiles.(solution.exercise_files_for_editor)
 
         render json: { files: files }
       end
--- a/app/controllers/api/solutions/last_iteration_files_controller.rb
+++ b/app/controllers/api/solutions/last_iteration_files_controller.rb
@@ -11,7 +11,7 @@ module API
         return render_403(:solution_not_accessible) unless current_user.may_view_solution?(solution)
         return render_400(:no_iterations_submitted_yet) if solution.latest_iteration.blank?
 
-        files = solution.latest_iteration.solution_files
+        files = solution.latest_iteration.files_for_editor
 
         render json: { files: SerializeFiles.(files) }
       end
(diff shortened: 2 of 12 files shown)


kntsoriano:
Looks good! I think I don't need to do anything here :)

app/controllers/api/solutions/initial_files_controller.rb
```diff
@@ -10,7 +10,7 @@ def index
 
         return render_403(:solution_not_accessible) unless current_user.may_view_solution?(solution)
 
-        files = SerializeFiles.(solution.exercise_solution_files)
+        files = SerializeFiles.(solution.exercise_files_for_editor)
```
ErikSchierboom, line 13: I feel like the name (`exercise_files_for_editor`) is slightly odd here, as we're in the context of the API returning files for the CLI. Maybe alias the method to `exercise_files_for_api`? Or, alternatively, `exercise_files_for_editing`?

app/controllers/api/solutions/last_iteration_files_controller.rb
```diff
@@ -11,7 +11,7 @@ def index
         return render_403(:solution_not_accessible) unless current_user.may_view_solution?(solution)
         return render_400(:no_iterations_submitted_yet) if solution.latest_iteration.blank?
 
-        files = solution.latest_iteration.solution_files
+        files = solution.latest_iteration.files_for_editor
```
ErikSchierboom, line 14: Same comment as above.

ErikSchierboom (approved):

iHiD:
Going to address Erik's comments post-launch (as I agree with the problem, but not sure what the best solution is)

Altairian7:
Frontend UI logic addition.

PR #1525, opened 2021-08-28
Add editor disabled button

kntsoriano:
https://github.com/exercism/v3-beta/issues/299

@iHiD Just needs styling. Tests fail because `editorEnabled` is hardcoded to `false`.


iHiD:
@kntsoriano One bug with this, the little arrow doesn't follow the mouse on the tooltip on the primary segment when disabled. See this loom as an example: https://www.loom.com/share/9f4f1130b8224f378f40dda0ef3ac824

iHiD:
Added styling and implement `has_test_runner?` (cc @ErikSchierboom as this is a git change that it's worth checking I've done correctly)

app/css/components/combo-button.css
```diff
@@ -16,14 +16,21 @@
         }
     }
 
+    & .--primary-segment.--disabled {
```
kntsoriano, line 19: @iHiD Could you check this? I needed to use a CSS class instead of the `disabled` property because the `disabled` property don't send out mouse events.

PR #1526, opened 2021-08-28
Allow viewing published solutions when logged out

### LINKED ISSUE #295: [Exercise: Commuity Solutions / Solution] "Share" via LinkedIn is broken
It leads me to a broken page.

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/295.

### DIFF
--- a/app/javascript/components/community-solutions/CommentsList.tsx
+++ b/app/javascript/components/community-solutions/CommentsList.tsx
@@ -5,7 +5,6 @@ import { Reminder } from './comments-list/Reminder'
 import { ListContainer } from './comments-list/ListContainer'
 import { ListDisabled } from './comments-list/ListDisabled'
 import { Request } from '../../hooks/request-query'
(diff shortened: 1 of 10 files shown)


app/javascript/components/community-solutions/CommentsList.tsx
```diff
@@ -41,11 +42,12 @@ export const CommentsList = ({
       <Header
```
iHiD: I think add the guard for `isGuest` here, rather than in the Header, then we don't have to pass it into Header?

iHiD:
@kntsoriano Nice work. One suggestion about where to guard one bit. Main comment though: Can you change this to `userSignedIn` rather than `isGuest` and invert it pls? We don't use the word `guest` anywhere in the code base and I'd rather not introduce another term at this stage, without thinking through how it might be used in the future (e.g. whether "guest" blog posts might be a thing or something). Thanks :)

PR #1527, opened 2021-08-28
Redirect to profile when accessing a hidden tab via URL

### LINKED ISSUE #1351: TypeError in /profiles/Deiivid/contributions
## Error in v3 (client side)

**TypeError** in **/profiles/Deiivid/contributions**
Cannot read property 'title' of undefined

[View on Bugsnag](https://app.bugsnag.com/exercism/v3-client-side/errors/6109180adacdc100079f6f14?event_id=6109180a007fa3ba97de0000&i=gh&m=ci)

## Stacktrace

    app/javascript/components/profile/ContributionsList.tsx:50 - ContributionsContent

[View full stacktrace](https://app.bugsnag.com/exercism/v3-client-side/errors/6109180adacdc100079f6f14?event_id=6109180a007fa3ba97de0000&i=gh&m=ci)

### LINKED ISSUE #1512: TypeError in /profiles/OsDroidi/contributions
## Error in v3 (client side)

**TypeError** in **/profiles/OsDroidi/contributions**
Cannot read property 'title' of undefined

[View on Bugsnag](https://app.bugsnag.com/exercism/v3-client-side/errors/6128f2074229090007c66a57?event_id=6128f207007fa54c995c0000&i=gh&m=ci)

## Stacktrace

    app/javascript/components/profile/ContributionsList.tsx:50 - ContributionsContent

[View full stacktrace](https://app.bugsnag.com/exercism/v3-client-side/errors/6128f2074229090007c66a57?event_id=6128f207007fa54c995c0000&i=gh&m=ci)

kntsoriano:
Closes https://github.com/exercism/website/issues/1512.
Closes https://github.com/exercism/website/issues/1351.

### DIFF
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -17,16 +17,22 @@ class ProfilesController < ApplicationController
   end
 
   def solutions
+    redirect_to profile_path(@user) unless @profile.solutions_tab?
+
     @solutions = Solution::SearchUserSolutions.(
       @user,
       status: :published
     )
   end
 
-  def contributions; end
+  def contributions
+    redirect_to profile_path(@user) unless @profile.contributions_tab?
+  end
 
   # TODO: (Optional) Add tests for published scope
   def testimonials
+    redirect_to profile_path(@user) unless @profile.testimonials_tab?
+
     @num_solutions_mentored = @user.mentor_discussions.count
     @num_students_helped = @user.mentor_discussions.joins(:solution).distinct.count(:user_id)
     @num_testimonials = @user.mentor_testimonials.published.count
--- a/app/models/user/profile.rb
+++ b/app/models/user/profile.rb
@@ -17,7 +17,6 @@ class User::Profile < ApplicationRecord
 
   memoize
   def contributions_tab?
-    user.reputation_tokens.where(category: %i[building authoring
-                                              maintaining]).exists? || user.authored_exercises.any? || user.contributed_exercises.any?
+    user.reputation_tokens.where(category: %i[building authoring maintaining]).exists?
   end
 end
--- a/app/models/user/profile.rb
+++ b/app/models/user/profile.rb
@@ -17,6 +17,7 @@ class User::Profile < ApplicationRecord
 
   memoize
   def contributions_tab?
-    user.reputation_tokens.where(category: %i[building authoring maintaining]).exists?
+    user.reputation_tokens.where(category: %i[building authoring
+                                              maintaining]).exists? || user.authored_exercises.any? || user.contributed_exercises.any?
   end
 end
--- a/test/system/flows/profile/user_views_published_solutions_test.rb
+++ b/test/system/flows/profile/user_views_published_solutions_test.rb
@@ -6,6 +6,8 @@ module Flows
     include CapybaraHelpers
 
     test "shows published solutions" do
+      # TODO: Change this stub
+      User::Profile.any_instance.expects(solutions_tab?: true).at_least_once
       author = create :user, handle: "author"
       create :user_profile, user: author
       ruby = create :track, title: "Ruby"
@@ -23,6 +25,8 @@ module Flows
     end
 
     test "shows published solutions for a logged out user" do
+      # TODO: Change this stub
+      User::Profile.any_instance.expects(solutions_tab?: true).at_least_once
       author = create :user, handle: "author"
       create :user_profile, user: author
       ruby = create :track, title: "Ruby"
@@ -45,6 +49,8 @@ module Flows
     end
 
     test "searches published solutions" do
+      # TODO: Change this stub
+      User::Profile.any_instance.expects(solutions_tab?: true).at_least_once
       author = create :user, handle: "author"
       create :user_profile, user: author
       ruby = create :track, title: "Ruby"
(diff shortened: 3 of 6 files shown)


iHiD:
Looks like the tests need updating to add a minimum of 4 solutions etc

app/models/user/profile.rb
```diff
@@ -5,18 +5,25 @@ class User::Profile < ApplicationRecord
...
   memoize
   def testimonials_tab?
-    user.mentor_testimonials.published.count > 3
+    user.mentor_testimonials.published.count >= self.class.count_to_show_tab
   end
 
   memoize
   def contributions_tab?
-    user.reputation_tokens.where(category: %i[building authoring maintaining]).exists?
+    user.reputation_tokens.where(category: %i[building authoring
```
kntsoriano: @iHiD I needed to modify this to make the tests pass. Is this correct? Or should I change the test to have a corresponding reputation token awarded to the exercise authorship?

app/controllers/profiles_controller.rb
```diff
@@ -9,24 +9,30 @@ def index
 
   def show
     # TODO: (Required) Order all these by most prominent first
-    @solutions = @user.solutions.published.first(3)
-    @testimonials = @user.mentor_testimonials.published.first(3)
+    @solutions = @user.solutions.published.first(User::Profile.count_to_show_tab - 1)
+    @testimonials = @user.mentor_testimonials.published.first(User::Profile.count_to_show_tab - 1)
```
iHiD: These should stay as 3 I think. We might change the rules about when the tab shows, but we'd always want to show 3 here regardless.

app/models/user/profile.rb
```diff
@@ -5,18 +5,25 @@ class User::Profile < ApplicationRecord
...
   memoize
   def solutions_tab?
-    user.solutions.published.count > 3
+    user.solutions.published.count >= self.class.count_to_show_tab
   end
 
   memoize
   def testimonials_tab?
-    user.mentor_testimonials.published.count > 3
+    user.mentor_testimonials.published.count >= self.class.count_to_show_tab
```
iHiD: These two aren't the same as each other either. The fact that they're both 3 is a conicidence atm. So either use two constants, but I'd rather we didn't (as this class will quickly get messy) and instead just stub the method in the tests.

app/models/user/profile.rb
iHiD: Change the test pls :)

test/system/flows/profile/user_views_published_solutions_test.rb
```diff
@@ -23,6 +23,7 @@ class UserViewsPublishedSolutionsTest < ApplicationSystemTestCase
     end
 
     test "shows published solutions for a logged out user" do
+      User::Profile.stubs(:count_to_show_tab).returns(1)
```
iHiD: I think it's easier to do this than adding all the extra code to profile:

```suggestion
      User::Profile.any_instance.expects(solutions_tab?: true).at_least_once
```

kntsoriano: Okay. I'll just do it this way and add a TODO to change this. This isn't really a good thing to stub because it's not a direct value.

iHiD: I think in this case, because we're testing that the user is redirected if the `solutions_tab` logic returns false, it's ok to stub that logic out. We're not actually testing the logic of `solutions_tab` itself here, just what happens when the solutions tab shouldn't show. By doing this, if we make that logic more complex in the future (which is quite possible), this test doesn't need to be updated to respect the new rules. This is why I'm a fan of stubbing public methods like this, and then making sure all public methods have unit tests, because it means the tests need changing a lot less often.

kntsoriano: It's a system test. You shouldn't stub these things because they have underlying logic inside. It's okay at this phase, but we need to extract it to something that we shouldn't either stub via creating db models which respect the rules, or just via a setting variable which I did.

This is a system test, we are replicating how the user will be using the system.

app/models/user/profile.rb
kntsoriano: We then need a way to ensure that a reputation token is awarded for exercise authorships which we could add in later. The authorship tab values don't have anything to do with awarded reputation tokens, but rather they are accessing the exercise authorships and contributions model directly.

test/system/flows/profile/user_views_published_solutions_test.rb
iHiD: I agree. In which case, don't stub anything and just create the relevant db records :) 

(I'm not suggesting we stub something here, just that we stub less underlying logic than what you stubbed :P)

app/models/user/profile.rb
iHiD: > We then need a way to ensure that a reputation token is awarded for exercise authorships

Yeah - it is. It's all automated through the GH syncing process. You get two reputation tokens for authoring - one for the authoring itself and one for the PR that creates it.

test/system/flows/profile/user_views_published_solutions_test.rb
iHiD: The reason I care is that I don't like going from one quite clean method (`if x > 3`) to having that method, plus a constant, pls a class method. And doing that twice once for solutions and once for testimonials. That would add an extra 14 LOCs in total, which just makes the code so much harder to parse when you're reading through trying to find stuff later. 

I'm ok with it more in commands, where there's less code in general (e.g where you've done it for pagination) but in models, which can get big and hairy anyway, moving from a three line method to ten lines, for no functional improvement, doesn't feel like a tradeoff I'm happy with.

PR #1544, opened 2021-08-29
Fix button mismatch between React and Rails

### LINKED ISSUE #289: [Exercise] Continue in Editor/ Start in Editor Button Mis-Match

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/289.


app/views/tracks/exercises/show.html.haml
```diff
@@ -56,10 +56,14 @@
           = icon "external-link", "The link opens in a new window or tab", css_class: "action-icon"
 
     .flex-shrink-0.lg:w-1-3.lg:ml-32.mb-32.lg:mb-0
-      - if @solution&.completed?
+      - if @solution&.status == :completed
```
kntsoriano: @iHiD Could you check if you are happy with this change?

iHiD: I rather use methods than test comparison (because Ruby doesn't give us any guarantee against typos on the RHS). So, how about:

```suggestion
      - if @solution&.status_completed?
```

iHiD:
Thanks. I'll need to think through the logic to check this is correct.

PR #1545, opened 2021-08-29
Add error to dif modal

### LINKED ISSUE #1348: TypeError in /tracks/python/exercises/black-jack
## Error in v3 (client side)

**TypeError** in **/tracks/python/exercises/black-jack**
Cannot read property 'filename' of undefined

[View on Bugsnag](https://app.bugsnag.com/exercism/v3-client-side/errors/6108379fcb99b80007e3a590?event_id=6108379f007fa8ce9b830000&i=gh&m=ci)

## Stacktrace

    app/javascript/components/modals/exercise-update-modal/ExerciseUpdateForm.tsx:27 - ExerciseUpdateForm

[View full stacktrace](https://app.bugsnag.com/exercism/v3-client-side/errors/6108379fcb99b80007e3a590?event_id=6108379f007fa8ce9b830000&i=gh&m=ci)

kntsoriano:
Closes https://github.com/exercism/website/issues/1348.

### DIFF
--- a/app/controllers/api/solutions_controller.rb
+++ b/app/controllers/api/solutions_controller.rb
@@ -149,7 +149,8 @@ module API
 
       files = Git::GenerateDiffBetweenExerciseVersions.(solution.exercise, solution.git_slug, solution.git_sha)
 
-      raise "Unable to produce diff for solution: #{solution.uuid}" if files.empty?
+      # TODO: (Optional): Change this to always be a 200 and handle the empty files in React
+      status = files.present? ? 200 : 400
 
       render json: {
         diff: {
@@ -162,7 +163,7 @@ module API
             update: Exercism::Routes.sync_api_solution_url(solution.uuid)
           }
         }
-      }
+      }, status: status
     end
 
     def sync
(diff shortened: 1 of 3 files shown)


iHiD:
@kntsoriano The API should never raise exceptions - it should always return valid JSON, so I've changed it to return a 400 with the response if files is empty. Having thought more about it, I'm not actually a fan of this as I think people could consume the API to see what's changed and that "nothing" is a valid response, so I've added a TODO to change this later, but for now it's fine.

Could you confirm this will still work with the React pls?

kntsoriano:
@iHiD I still think we should raise an exception, because in our code this never should happen. This has a good side-effect of being notified via Bugsnag as well. Let's then catch the exception with a 500 message, because this is an error on our part.

iHiD:
> I still think we should raise an exception, because in our code this never should happen. This has a good side-effect of being notified via Bugsnag as well.

It should only never happen because of our workflow though. ie, it should never happen because we only request this once we've checked the previously that we should show the banner. In a different workflow (say someone checks every day whether any files have changed using this API endpoint) then no files changed would be valid output.

I'm on Slack btw :)

kntsoriano:
@iHiD Still works! Let's get this merged.

iHiD:
Thanks! :)

PR #1574, opened 2021-08-30
Add social details to profile form

### LINKED ISSUE #326: Add options to change social details for profile
@kntsoriano These are some extra settings on the settings page:

Screenshot 2021-08-29 at 23 01 20

- It is part of the name/bio form (extra fields there)
- These fields only render if someone has a profile
- Send them as params[:profile], and do an update on that too in the same action

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/326.

### DIFF
--- a/app/commands/user/update.rb
+++ b/app/commands/user/update.rb
@@ -1,9 +1,9 @@
 class User
   class Update
-    def initialize(user, params)
-      @user = user
-      @params = params
-    end
+    include Mandate
+    include Mandate::Callbacks
+
+    initialize_with :user, :params
 
     def call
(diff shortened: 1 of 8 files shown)


app/commands/user/update.rb
```diff
@@ -0,0 +1,42 @@
+class User
```
kntsoriano, line 1: @iHiD Could you check if you are happy with this?

iHiD: @kntsoriano I refactored it to use Mandate Callbacks 🙂

PR #1585, opened 2021-08-30
Make completion modals not shit

iHiD:
https://github.com/exercism/v3-beta/issues/273


iHiD:
@kntsoriano Please fix the tests.

To test with different exercises, go to http://local.exercism.io:3020/temp/modals/completed_exercise and change the exercise in the haml there. `ruby`/`lasagna` is good to test with.

Before each time you refresh you need to run this to reset it to get completions:
```
Solution.for('ihid', 'ruby', 'lasagna').update(completed_at: nil, published_at: nil)
```

PR #1606, opened 2021-08-31
Render internal links

ErikSchierboom:
We hadn't implemented the internal concept/exercise widgets. It is a little bit more effort to fully support them, so for now we'll just render them as links (with tooltips).

See https://github.com/exercism/v3-beta/issues/251


iHiD:
Brilliant. Thanks.

PR #1612, opened 2021-08-31
Add height to select options

### LINKED ISSUE #329: Track select in contributors list is really long
Screen Shot 2021-08-30 at 10 46 34 AM

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/329.

### DIFF
--- a/app/css/components/multiple-select.css
+++ b/app/css/components/multiple-select.css
@@ -25,6 +25,9 @@
         }
     }
     & > .--options {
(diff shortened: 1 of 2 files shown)


iHiD:
@kntsoriano Too tweaks here. The first lets people with big screens get more longer screens. The second enforces that only `y` should scroll. But very similar to what you had! Thanks :)

PR #1627, opened 2021-08-31
Fix background-image urls

ErikSchierboom:


iHiD:
@ErikSchierboom This doesn't seem to be right on the leaderboard still.

PR #1646, opened 2021-09-01
Match exercise status on journey solutions page

### LINKED ISSUE #190: [Journey/Solutions] Filtering solutions partially broken
There is something wrong with filtering solutions both by status and mentoring status.

## General

- "Showing X solutions" only shows the numbers of solutions on the first page, not all in total, which makes it feel like filtering is broken because it sounds like those are ALL results, and I know there should be much more.
- 

## Status problems
- I get nothing for "completed" but I have A LOT of completed exercises. I assume "completed" is both "completed published" and "completed not published".
- And I get the same results for "in progess" and "not published (completed)" which doesn't make sense because those two statuses are exclusive.

Screen Shot 2021-08-20 at 19 14 58

Screen Shot 2021-08-20 at 19 15 02

Screen Shot 2021-08-20 at 19 15 05

Screen Shot 2021-08-20 at 19 15 09

## Mentoring status problems

- I get the same results for "no mentoring" and "mentoring completed" which doesn't make sense because those two mentoring statuses are exclusive.

Screen Shot 2021-08-20 at 19 15 18

Screen Shot 2021-08-20 at 19 15 21

kntsoriano:
Closes https://github.com/exercism/v3-beta/issues/190.

### DIFF
--- a/app/javascript/components/journey/solutions-list/ExerciseStatusSelect.tsx
+++ b/app/javascript/components/journey/solutions-list/ExerciseStatusSelect.tsx
@@ -1,12 +1,7 @@
 import React from 'react'
 import { SingleSelect } from '../../common'
 
-type Status =
-  | undefined
-  | 'in_progress'
-  | 'completed'
-  | 'published'
-  | 'not_published'
+type Status = undefined | 'started' | 'iterated' | 'completed' | 'published'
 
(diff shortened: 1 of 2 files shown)


iHiD:
Looks good. Have you tested it?

PR #1647, opened 2021-09-01
Show editor error

kntsoriano:
https://github.com/exercism/v3-beta/issues/339

### DIFF
--- a/app/javascript/components/Editor.tsx
+++ b/app/javascript/components/Editor.tsx
@@ -106,18 +106,31 @@ export default ({
         dispatch({ status: EditorStatus.INITIALIZED })
         setSubmissionFiles(files)
       },
-      onError: (error) => {
+      onError: async (error) => {
         let editorError = null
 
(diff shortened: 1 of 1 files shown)


iHiD:
@kntsoriano Will this mean they can resubmit the tests?

kntsoriano:
@iHiD They still won't be able to. But they'll see an error where duplicate submissions aren't allowed. How do you think we should solve that? It's marked as a TODO in `api/submissions#create`

iHiD:
Hmmm. I thought I'd fixed that. Strange. I'll try and remember why I thought that.

kntsoriano:
I was thinking of the same! I think the `solution#last_valid_submission` doesn't include test runs that timed out?

iHiD:
Yeah. I think the issue here is that the solutions were still queued. So maybe we need to think about that post-launch, and say "Your submission is still queued". The React timeout is different from the ops timeout in this situation.

PR #1689, opened 2021-09-01
Sort published solutions

### LINKED ISSUE #371: Sort published solutions on profile page by most started
image

ErikSchierboom:
I know this has no priority but I just did this for fun :)

Closes https://github.com/exercism/v3-beta/issues/371

The query is as follows:

```sql
SELECT `solutions`.* FROM `solutions` WHERE `solutions`.`user_id` = 4 AND `solutions`.`published_at` IS NOT NULL ORDER BY `solutions`.`num_stars` DESC, `solutions`.`updated_at` DESC LIMIT 3
```

This doesn't use any keys, so I don't know how problematic that is.

### DIFF
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -8,8 +8,9 @@ class ProfilesController < ApplicationController
   end
 
   def show
-    # TODO: (Required) Order all these by most prominent first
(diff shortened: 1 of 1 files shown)


iHiD:
This is fine from a performance POV as its scoped to a user (which means it's only got to reorder a couple of thousand rows max). Nice one :)

PR #1694, opened 2021-09-01
Fix minor casing issue

petemcfarlane:
I noticed this error happening a few times on the new exercism v3 site. I'm not sure why it was erroring, but at least this fixes the casing 😆


willfolsom (approved):

iHiD:
Thanks :)

PR #1729, opened 2021-09-02
Fix date formatting issue when date is over 5 years ago

sevenseacat:
I noticed this while browsing my profile (with some old solutions!)

Screen Shot 2021-09-02 at 5 39 07 pm


iHiD:
@sevenseacat Thanks! :) However, for the short one, I'd like to change it to `5y+`. "Over 5y" won't fit in various places. Could you make that change pls? 

@kntsoriano Is this also an issue in JS land?

iHiD:
Could you also sort the rubocop issue pls 🙂

Running `bundle exec rubocop  --except Metrics` is the same as what CI does :)

iHiD:
Thanks :)

PR #1737, opened 2021-09-02
Escape code

### LINKED ISSUE #970: Tests missing attributes
The test files (that I've checked, it may not apply to all) are currently missing their `[<Fact>]` attributes, so are essentially unavailable to the online test runner. So far as I can tell, the user cannot edit these. 

Please see the following three screenshots: 

The standard `hello-world`:
hello-world-f-sharp-error

The second for the `phone-number` exercise
phone-number-f-sharp-error

I reproduced this locally:
rider-f-sharp-error

With `[<Fact>]` reinstated, the tests are available:
rider-f-sharp-pass

Well done and thank you to all involved in the new version 3 of Exercism!

### LINKED ISSUE #5610: Rendering solution preview: angle brackets
The solution previews (as shown on the published solutions page) seems to have a small problem where invalid html tags are stripped, but some languages like C use angle brackets:

An actual solution:
solution actual

The solution preview:
solution preview

ErikSchierboom:
This is how the tests were rendered previously (no escaping loses the `<>` characters):

image

And this is how they're displayed with this change:

image

I don't know if there was any particular reason to use the `dangerouslySetInnerHTML` option here, but the new version still has syntax highlighting so 🤷 . @kntsoriano?

Closes https://github.com/exercism/fsharp/issues/970
Closes https://github.com/exercism/exercism/issues/5610

### DIFF
--- a/app/javascript/components/common/CommunitySolution.tsx
+++ b/app/javascript/components/common/CommunitySolution.tsx
@@ -100,10 +100,9 @@ export const CommunitySolution = ({
         <ProcessingStatusSummary iterationStatus={solution.iterationStatus} />
       </header>
       <pre ref={snippetRef}>
-        <code
-          className={solution.track.highlightjsLanguage}
-          dangerouslySetInnerHTML={{ __html: solution.snippet }}
-        />
+        <code className={solution.track.highlightjsLanguage}>
+          {solution.snippet}
+        </code>
       </pre>
       <footer className="--footer">
         {solution.publishedAt ? (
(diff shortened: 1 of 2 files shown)


joshgoebel (approved):
**Yes, non-dangerously** - which I presume means it is properly escaped now (as any old content in the template should be), which is exactly what we want.

SleeplessByte (approved):

iHiD (approved):

joshgoebel:
Might be worth a quick audit of the codebase for `dangerousSetHTML` in general?

iHiD:
@kntsoriano Could you double check the other uses pls to ensure they're all valid: `git grep dangerouslySetInnerHTML`

PR #1758, opened 2021-09-03
Allow solution download

### LINKED ISSUE #5611: Mentoring UI: download solution by uuid
In v2, it was possible to copy a command to download a student's solution by uuid when mentoring. This is crucial, even when the test runners are there, because I need to be able to try out modifications to the code myself before suggesting them to the student.

It's not possible to comfortably copy-paste the solution from the editor because the output includes a lot of extra newlines.

I would need the same button in v3, or at the very least any way to find the uuid. I can assemble the command myself.

kntsoriano:
Closes https://github.com/exercism/exercism/issues/5611.

## Notes
1. When I was looking into the code, I noticed that `<IterationHeader />`  had lots of conditionally rendered children. As we're now slowing down our pace, I think it's better for us to not add another conditional, but to rather extract a duplicated component. Doing it this way maps our current thinking more accurately and introduces a conversation on whether we want to make things more similar.
2. In order for me to be able to assert that the "Copied" message appears, I needed to change it from a pseudoelement to an element. @iHiD please check if I've broken anything in the CSS.
3. In order for me to assert that the text was copied to the clipboard, I added `<span data-test-clipboard />` to our components. This is a hidden component which contains the text we are sending to the browser's clipboard. Unfortunately, we can't assert the clipboard's value directly as this isn't supported by headless Chrome. I've gone through the hassle of testing on normal Chrome first before finding out that headless Chrome doesn't support it 🤦 
4. When clicking the "Copy solution" button, I assumed that we want the a "Copied" message to appear as well. This mimics how the copy to clipboard button works. Let me know if we are changing this. In line with (1), I've duplicated the code as we aren't sure yet on how it really works.


app/javascript/components/mentoring/session/iteration-view/iteration-header/CopyButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+      return
+    }
+
+    const timer = setTimeout(() => setJustCopied(false), JUST_COPIED_TIMEOUT)
+
+    return () => clearTimeout(timer)
+  }, [justCopied, setJustCopied])
+
+  {
+    /* rODO: Style button */
```
kntsoriano: 😆

app/models/solution.rb
```diff
@@ -79,6 +79,10 @@ def self.for(*args)
 
   delegate :instructions, :introduction, :tests, :source, :source_url, to: :git_exercise
 
+  def external_download_cmd
```
kntsoriano: I don't know how to name this correctly. We currently have two download commands:

`exercism download uuid=uuid` and `exercism download --track=track --exercise=exercise`.

iHiD:
1. Cool. I think you maybe went a little too far by redoing all the websocket stuff so I've reverted that bit in https://github.com/exercism/website/pull/1767 - I think you'll add it's correct so I'll probably merge it and deploy all this.
2. Seems ok.
3. Bit yucky but ok! :) 
4. Thanks :)

iHiD:
I also fixed the failing test.

app/models/solution.rb
iHiD: Same as v2. `mentor_download_cmd`

app/javascript/components/student/mentoring-session/IterationView.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <IterationsList
+            iterations={iterations}
+            onClick={onClick}
+            current={currentIteration}
+          />
+          {discussion ? (
+            <LinkButton value={isLinked} setValue={setIsLinked} />
+          ) : null}
+        </footer>
+      ) : null}
```
iHiD, line 83: I think this should be extracted to a common component.

app/javascript/components/student/mentoring-session/IterationView.tsx
```diff
@@ -0,0 +1,86 @@
...
+                files={resolvedData.files}
+                language={language}
+                indentSize={indentSize}
+                instructions={instructions}
+                tests={tests}
+              />
+            ) : null}
+          </FetchingBoundary>
+        </ResultsZone>
+      )}
```
iHiD, line 71: I think this should be extracted to a common component.

iHiD:
I think where we should be aiming for with these refactors is to keep small chunks as common components, but then put them together like lego in the various places. I've tagged a couple more where I think that's how we should approach it. Else we just have to remember to change two things every time, which isn't great. I also agree about not having lots of conditionals though.

(I'm sure you agree with all this)

PR #1762, opened 2021-09-03
Use correct exercise/concept count for Track upsell

sevenseacat:
Currently the main Track page hardcodes 120 concepts/100 exercises - this uses the real values for a track instead.

Screen Shot 2021-09-03 at 10 47 39 pm

Screen Shot 2021-09-03 at 10 47 48 pm


iHiD:
Thanks :)

PR #1789, opened 2021-09-04
Contributor badge

iHiD:


app/commands/user/acquired_badge/create.rb
```diff
@@ -20,7 +20,11 @@ def call
           User::AcquiredBadge.create!(
             user: user,
             badge: badge
-          )
+          ).tap do |uab|
+            User::Notification::CreateEmailOnly.(user, :acquired_badge, user_acquired_badge: uab) if badge.send_email_on_acquisition?
+
```
iHiD, line 25: Urgh to this whitespace!

PR #1800, opened 2021-09-05
Pass all communication preference keys from Rails

### LINKED ISSUE #5775: Communication preferences should be powered by Rails
We should be able to add keys in the seriazlier ([like this](https://github.com/exercism/website/pull/1793/files#diff-5c7ddc9979ac7a95aad2adb7eeabbcf310d3b65e07e5011454701ec5d523745dR8-R14)) and they should then appear in React.

- The serializer should power what shows in the component, not duplicated in React.
- The copy for each option is avaliable in an i18n file in rails. The serializer should pass this on with the keys.

kntsoriano:
Closes https://github.com/exercism/exercism/issues/5775.

Screenshot:
Screen Shot 2021-09-05 at 3 03 11 PM

### DIFF
--- a/app/models/user/communication_preferences.rb
+++ b/app/models/user/communication_preferences.rb
@@ -1,6 +1,18 @@
 class User::CommunicationPreferences < ApplicationRecord
   belongs_to :user
 
+  def self.keys
+    %i[
+      email_on_mentor_started_discussion_notification
+      email_on_mentor_replied_to_discussion_notification
+      email_on_student_replied_to_discussion_notification
+      email_on_student_added_iteration_notification
+      email_on_acquired_badge_notification
+      email_on_general_update_notification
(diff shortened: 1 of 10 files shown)


app/models/user/communication_preferences.rb
```diff
@@ -1,6 +1,18 @@
 class User::CommunicationPreferences < ApplicationRecord
   belongs_to :user
 
+  def self.keys
```
kntsoriano, line 4: @iHiD I'll leave this to you. Maybe there's a better way to do it dynamically.

iHiD: Nah, this is fine. Thanks!

iHiD:
Sexy work on this one K-dawg!

PR #1801, opened 2021-09-05
Import nim mode correctly

kntsoriano:
https://github.com/exercism/exercism/issues/5703

I've read through the nim package and I think we just didn't import it correctly. Let's merge this and let the nim users test it?


ynfle:
👍 I'm happy to test on merge

iHiD:
Sweet. If this works, there's a number of others we need to check too.

ynfle:
Thanks. I'm assuming it's not deployed yet.

iHiD:
It takes 5-30mins to deploy (depending on what's changed). I'd imagine this is a 30min one.

iHiD:
It worked. Nice work @kntsoriano! :)

PR #1802, opened 2021-09-05
Fix file comparison when disabling editor buttons

### LINKED ISSUE #5701: Can't submit the exercise solution with the tests completed if you leave the page and then return
Steps to reproduce:
1. Go to any exercise and write a correct solution to the problem, then click "Run tests".
2. Without clicking "submit" leave the page (e.g. click "Back to Exercise")
3. Return to the page with the exercise

Now the "Submit" is visible but not active despite all tests showing as passed:
image

The current workaround is to modify the code (since you can't rerun the tests with the same code which makes sense), rerun the tests, modify the code back to the correct one, rerun the tests again and click "Submit".

kntsoriano:
Closes https://github.com/exercism/exercism/issues/5701.

I think this was an error with how I was checking whether the files have changed for the editor.

### DIFF
--- a/app/javascript/components/Editor.tsx
+++ b/app/javascript/components/Editor.tsx
@@ -5,7 +5,7 @@ import React, {
   useEffect,
   createContext,
 } from 'react'
-import { Submission, TestRun, TestRunStatus } from './editor/types'
+import { TestRun, TestRunStatus } from './editor/types'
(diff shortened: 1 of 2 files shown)


iHiD:
Tests seem to be legitimately red on this one.

PR #1803, opened 2021-09-05
Order testimonials correctly

### LINKED ISSUE #5698: Sort of testimonials appears to be incorrect or reporting incorrectly.
image

kntsoriano:
Closes https://github.com/exercism/exercism/issues/5698.

### DIFF
--- a/app/commands/mentor/testimonial/retrieve.rb
+++ b/app/commands/mentor/testimonial/retrieve.rb
@@ -63,9 +63,9 @@ module Mentor
       def sort!
         case order&.to_sym
         when :newest
-          @testimonials = @testimonials.order("mentor_testimonials.id DESC")
+          @testimonials = @testimonials.order("mentor_testimonials.created_at DESC")
         when :oldest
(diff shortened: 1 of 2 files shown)


kntsoriano:
@iHiD You might've forgotten this one.

iHiD:
We don't have indexes on this column so need to remember we can never do this for more than a couple of thousand records.

kntsoriano:
@iHiD Thanks! What should we have done here then if this was a big table? Add an index to the column?

iHiD:
Two things:
1. Yeah - add index if it was big, although then you have to consider if there's searching etc, in which case you might need a compound index on different columns, in which case you you need to consider the order of the columns in the index (this is a good analogy: https://stackoverflow.com/a/24315527/856688).
2. `id` and `created_at` should always be in the same order really. It's only really like this because of ETL, where I didn't maintain that order. So the alternative would have been to have said that it's an ETL issue but people just have to live with it (the reason I waited before merging for a couple of days was that I was considering if this was an option).

PR #1822, opened 2021-09-06
[Waiting for Taiyab] Tweak mentor discussion styling

### LINKED ISSUE #6003: Provide the avatar tooltip bio in the mentoring page that's available in the queue list page
When I am choosing an exercise to mentor, I can hover over each student's avatar and see a bio, or explanation of what their learning goal is. It would be great to have access to that text once you start mentoring the person on that exercise.

I'm not aware of any workarounds to get that information once you start mentoring the exercise.

Their avatar is at the top left of the exercise mentoring page, so perhaps the tooltip could be attached there?

iHiD:
Closes https://github.com/exercism/exercism/issues/6003

**This is waiting on the pronouns design from Taiyab before it can be merged**


iHiD:
@kntsoriano I think maybe you missed this one.

kntsoriano:
@iHiD This should be ready :)

iHiD:
Brilliant, thanks. Tests are red but I've not looked at them.

app/helpers/react_components/student/mentoring_session/mentor.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+
+      class NullRelationship
+        def num_discussions
+          0
+        end
+      end
+    end
+  end
+end
```
iHiD: @kntsoriano I like your instinct here to refactor this to be cleaner. I've made a few changes though:
1. I've inlined it into the class. I want internal helper components (such as this) to be kept internal to the place they're used. Othewise we end up with a mixture of things called publically and internally, and it becomes very hard to navigate around with hundreds of files (similar to what you noted abut the volume of serializers). I've mentioned this on another PR as well.
2. I've changed it to be a serializer, rather than using the `as_json` pattern. Let's stick with the same patterns we use elsewhere rather than introducing new ones. I'm not arguing one is better than the other - simply that I want to keep things consistent.
3. I've removed the stubbing of the attributes. I get why you've done this to make the tests faster, but in doing this we don't actually check anywhere that the correct keys are used, which is the most important part of the test in my opinion. So I changed the one test where those were stubbed to ensure the correct keys were passed, at which point all of the stubbing code became unnecessary.
4. Renamed the bits called `json` to `hash` as they weren't actually doing anything with JSON.

iHiD:
(Waiting on some design stuff from @taiyab before this one gets merged)

app/helpers/react_components/student/mentoring_session/mentor.rb
kntsoriano: 1. I think this is why I put it within the `MentoringSession` namespace as this would imply that it is internal to the component that it is used with.
2. Okay!
3. We check in the system test that the correct attributes are used.
4. Okay!

iHiD: Two thoughts on 3:
1. We check some of them in the system test, but not all of them. That might be a smell (in a good way), because it might mean we should delete some from the serializer that aren't used. But I feel uncomfortable that we're potentially relying on data downstream but not checking for it here. 
2. I really like unit tests! 🙂 So although we might check things in a system test, that for me is a test of last resort - it's our final safety net. I think it's really important that the individual components within Rails get tested at that level too.

kntsoriano: 1. We check all of them in the system tests, and that should be the case because we want the user to see all of these information on their screen.
2. I use system tests as a way to discover how the system will go as there is an overarching goal to get to. It's not really a last resort for me as I think we should test most primarily how the user will use the system and unit tests should test how the units interact with each other. My fear with having unit tests and system tests testing the same thing is that we will be changing multiple tests for a single change. This is one of the frustrations you've raised and I want to avoid this.

taiyab:
image

image

This and this?

@iHiD

iHiD:
@taiyab https://exercism-io.slack.com/archives/D01156WCXR9/p1664022058346459?thread_ts=1632573252.000300&cid=D01156WCXR9

kotp:
[Forums conversation] here.

[Forums conversation]: http://forum.exercism.org/t/floating-over-student-name-in-track-queue-gives-information/534

dem4ron:
Broken out into:
https://github.com/exercism/website/pull/6132
https://github.com/exercism/website/pull/6130
https://github.com/exercism/website/pull/6163

PR #1828, opened 2021-09-07
Fix task text being invalid when using newlines in text

### LINKED ISSUE #172: [Editor] Markdown parsing bug: paragraph after a block quote
I have an exercise (Elixir / City Office) that has a step with this Markdown:

```markdown
## 2. Document filling out fields with blank values

Add documentation and a typespec to the `Form.blanks/1` function. The documentation should read:

> Generates a string of a given length.
>
> This string can be used to fill out a form field that is supposed to have no value.
> Such fields cannot be left empty because a malicious third party could fill them out with false data.

The typespec should explain that the function accepts a single argument, a non-negative integer, and returns a string.
```

This is how GitHub renders it:

Screen Shot 2021-08-08 at 07 57 17

This is how it looks like in the editor. The last paragraph that is supposed to be outside of the block quote gets put into the block quote for some reason:

Screen Shot 2021-08-08 at 07 53 35

### LINKED ISSUE #5780: Instructions rendering issue for several exercises in the Elixir track
I was going to file these in the Elixir track repo, but I think they might be more general and apply to multiple exercises so we'll see?

--- 

I noticed that for code blocks in the City Office exercise, the first line of text is not being rendered inside the code block. And if the block is only a single line, then there's no rendered code block at all!

Source: https://github.com/exercism/elixir/blob/main/exercises/concept/city-office/.docs/instructions.md

How it renders: 

Screen Shot 2021-09-05 at 2 42 27 pm

Screen Shot 2021-09-05 at 2 42 52 pm

---

There's a similar issue for table rendering in the German Sysadmin exercise. Content below the table is being rendered inside the table?

Source: https://github.com/exercism/elixir/blob/main/exercises/concept/german-sysadmin/.docs/instructions.md

How it renders: 

Screen Shot 2021-09-05 at 2 50 40 pm

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/5780
Closes https://github.com/exercism/v3-beta/issues/172

See:

image

### DIFF
--- a/app/serializers/serialize_exercise_assignment.rb
+++ b/app/serializers/serialize_exercise_assignment.rb
@@ -66,7 +66,7 @@ class SerializeExerciseAssignment
       chunk_while { |_, nxt| nxt.type != :header }.
       map do |nodes|
         task_title = parse_title(nodes.first)
-        task_text = Markdown::Parse.(nodes[1..].each.map(&:to_commonmark).join)
+        task_text = Markdown::Parse.(nodes[1..].each.map(&:to_commonmark).join("\n"))
         task_hints = hints[task_title.downcase].to_a
 
         { title: task_title, text: task_text, hints: task_hints }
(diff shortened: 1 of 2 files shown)


Reunina (approved):

sevenseacat:
Will this also fix the table issue mentioned in https://github.com/exercism/exercism/issues/5780 ? I suspect so, just want to double-check.

ErikSchierboom:
@sevenseacat It does:

image

PR #1659, opened 2021-09-01
[WIP] Don't create site updates for wip exercises

ErikSchierboom:
Closes https://github.com/exercism/v3-beta/issues/365

Todo: we need to create a site update when a wip exercise become a beta or active exercise

Note: we don't have wip concepts (yet)

PR #1829, opened 2021-09-07
Fix creation issue with Exercises

iHiD:
@ErikSchierboom We need to force_sync for new exercises to get the authors/contributors to create. I think this needs copying for both types of exercises and concept, and tests adding.

This also guarantees we don't don't issues like the test_runner stauts one from before (in fact we could now actively leave that field out in this bit if we wanted, and set the exercise's status here to `wip` and let the force_sync set the right values)

### DIFF
--- a/app/commands/git/sync_track.rb
+++ b/app/commands/git/sync_track.rb
@@ -78,7 +78,7 @@ module Git
           blurb: git_concept.blurb,
           synced_to_git_sha: head_git_track.commit.oid
         ).tap do |concept|
-          Git::SyncConcept.(concept, force_sync: force_sync)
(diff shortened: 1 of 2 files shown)


ErikSchierboom:
> This also guarantees we don't don't issues like the test_runner stauts one from before (in fact we could now actively leave that field out in this bit if we wanted, and set the exercise's status here to wip and let the force_sync set the right values)

We could leave it out, although it also doesn't really hard. Setting the status to wip here would interfere with https://github.com/exercism/website/pull/1659

ErikSchierboom:
@iHiD I've updated the PR

PR #1845, opened 2021-09-08
Update task's track

ErikSchierboom:
- Update task track on update
- Refactor task

### DIFF
--- a/app/models/github/task.rb
+++ b/app/models/github/task.rb
@@ -44,7 +44,7 @@ class Github::Task < ApplicationRecord
     docs: 4
   }, _suffix: true
 
-  before_validation on: :create do
+  before_validation do
     unless track
       repo_url = issue_url.
         gsub(%r{/issues/\d+}, '').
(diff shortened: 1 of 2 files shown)


app/models/github/task.rb
```diff
@@ -63,4 +57,15 @@ class Github::Task < ApplicationRecord
   before_create do
     self.uuid = SecureRandom.compact_uuid if self.uuid.blank?
   end
+
+  private
+  def update_track!
+    self.track_id = Track.where(repo_url: repo_url).pick(:id)
+  end
```
iHiD, line 64: The `update_xxx!` tends to signify an update will occur (e.g. the method uses `update(...)`). In this situation it doesn't actually update anything (because it's used in the validation).

So I think it shouldn't use "update" in the name, and probably not `!` either. Maybe: `set_track_id_from_repo`.

iHiD (approved):
Approved pending the rename :)

app/models/github/task.rb
ErikSchierboom: Renamed

PR #1867, opened 2021-09-09
Remove cache time from mentoring queue

### LINKED ISSUE #5639: left -right panel sync mismatch in mentoring queue
Mentored a Rust - Clock exercise. After closing the solution the Mentoring Queue showed 1 Clock exercise in the right panel but no solution in the left panel. I refreshed. Now it shows 0 Clock exercises, even though "Only show exercises that need mentoring" is checked. 
err_1

kntsoriano:
Closes https://github.com/exercism/exercism/issues/5639.

This is not a really good solution, but I don't know how else to fix it.

### DIFF
--- a/app/helpers/react_components/mentoring/queue.rb
+++ b/app/helpers/react_components/mentoring/queue.rb
@@ -43,7 +43,8 @@ module ReactComponents
             initial_data: {
               tracks: tracks_data
             },
-            stale_time: 0
+            stale_time: 0,
+            cache_time: 0
           }
         }
       end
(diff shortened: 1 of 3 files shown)


iHiD:
@SleeplessByte Any thoughts on this (the problem as much as the solution)?

kntsoriano:
As a background, when I go to `/mentoring/queue`, the query does not seem to remount. It is forever marked stale and won't refetch. I don't understand why it doesn't remount. I might not understand something about react-query. `fetchOnMount` is also set to true.

iHiD:
Should we merge this anyway @kntsoriano and then follow up when DJ has time? If so, could you make the tests green pls?

kntsoriano:
@iHiD Cool. I'm okay to merge this.

PR #1873, opened 2021-09-09
New user notification

iHiD:


config/locales/notifications/en.yml
```diff
@@ -16,6 +16,12 @@ en:
     added_to_contributors_page:
       1: >
         You now appear on our Contributors page. Thank you for contributing to Exercism!
+
+    # Not actually used
```
iHiD, line 20: We need these here right now else things break, but at some point I'll refactor them out. We don't show them in the UI, because they're email-only-notifications.

app/models/user/notifications/joined_exercism_notification.rb
```diff
@@ -0,0 +1,21 @@
...
+      def url
+        Exercism::Routes.dashboard_url
+      end
+
+      def image_type; end
+
+      def image_url; end
+
+      def guard_params
+        "" # Users should only have this badge once
```
ErikSchierboom, line 13: Is this a TODO?

ErikSchierboom (approved):

ErikSchierboom:
There is a failing test (might be unrelated)

app/models/user/notifications/joined_exercism_notification.rb
iHiD: No. This is correct.

PR #1875, opened 2021-09-09
Refactor sending into emailable and add to payments

iHiD:
This allows objects other than notifications to have emails sent about them, by adding an `emailable` concern. The example use-case also implemented here is donation payments, but this is relevant to lots of things where it doesn't make sense to add a notification, but we do want to send an email.


test/models/user/notifications/donation_subscription_created_notification_test.rb
```diff
@@ -0,0 +1,14 @@
...
+class User::Notifications::DonationSubscriptionCreatedNotificationTest < ActiveSupport::TestCase
+  include ActionView::Helpers::AssetUrlHelper
+  include Webpacker::Helper
+
+  test "keys are valid" do
+    user = create :user
+    payment = create :donation_payment, user: user
+
+    notification = User::Notifications::DonationSubscriptionCreatedNotification.create!(user: user, params: { payment: payment })
+    assert_equal "#{user.id}|acquired_badge|Badge##{badge.id}", notification.uniqueness_key
```
ErikSchierboom: Where does `badge.id` come from?

db/migrate/20210908203439_add_email_to_donations.rb
```diff
@@ -0,0 +1,6 @@
+class AddEmailToDonations < ActiveRecord::Migration[6.1]
+  def change
+    add_column :donations_payments, :email_status, :integer, limit: 1, default: 0, null: false
+    add_column :donations_subscriptions, :email_status, :integer, limit: 1, default: 0, null: false
```
ErikSchierboom, line 4: Will we retroactively send emails to people who donated?

app/models/user/notifications/donation_subscription_created_notification.rb
```diff
@@ -0,0 +1,19 @@
...
+      def url
+        Exercism::Routes.settings_url
+      end
+
+      def image_type; end
+
+      def image_url; end
+
+      def guard_params
+        "Badge##{user_acquired_badge.badge_id}"
```
ErikSchierboom: Why does the guard here use the badge ID? Is that for uniqueness?

ErikSchierboom (approved):

app/models/user/notifications/donation_subscription_created_notification.rb
iHiD: This file should have been deleted :+1:

db/migrate/20210908203439_add_email_to_donations.rb
iHiD: Yeah - I'm going to do that once we push this out.

test/models/user/notifications/donation_subscription_created_notification_test.rb
iHiD: Should also have been deleted! :)

PR #1885, opened 2021-09-10
Fix tutorial modal

### LINKED ISSUE #5856: "Return to  Hello World" does nothing
On the "Hello world" completion modal - "Return to  Hello World" does nothing. It should close the modal.

Screenshot 2021-09-09 at 22 05 36

kntsoriano:
Close https://github.com/exercism/exercism/issues/5856.

Instead of closing the modal as described in the issue, I decided to redirect the user back to the exercise page as this would contain the "completed" state for the exercise.

### DIFF
--- a/app/javascript/components/modals/complete-exercise-modal/TutorialCompletedModal.tsx
+++ b/app/javascript/components/modals/complete-exercise-modal/TutorialCompletedModal.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import { Modal } from '../Modal'
 import { GraphicalIcon } from '../../common'
-import { Track } from '../../types'
 import { ExerciseCompletion } from '../CompleteExerciseModal'
 
 export const TutorialCompletedModal = ({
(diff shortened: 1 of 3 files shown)


app/javascript/components/modals/complete-exercise-modal/TutorialCompletedModal.tsx
```diff
@@ -40,17 +43,25 @@ export const TutorialCompletedModal = ({
...
         ) : (
-          <a href={track.links.exercises} className="btn-primary btn-m">
+          <a
+            href={completion.track.links.exercises}
+            className="btn-primary btn-m"
+          >
             Show me more exercises
           </a>
         )}
-        <button className="btn">Return to “Hello, World!”</button>
```
kntsoriano, line 53: @iHiD I decided to remove the hardcoded exercise names in this modal. I remember that there was a TODO for this in another component. We might want to apply the same here.

iHiD: OK, cool. You need to put `{completion.exercise.title}` back in quotes in the button to match what this originally was.

test/system/flows/complete_tutorial_test.rb
```diff
@@ -22,7 +22,10 @@ class CompleteTutorialTest < ApplicationSystemTestCase
         click_on "Mark as complete"
         find("label", text: "Yes, I'd like to share my solution with the community.").click
         click_on "Confirm"
-        assert_text "You’ve completed “Hello, World!”"
+        assert_text "You’ve completed Hello World"
```
iHiD: This should be changed back.

iHiD:
Cool, that makes sense. Added a couple of comments,

PR #1886, opened 2021-09-10
Show test run and output correctly

kntsoriano:
Closes https://github.com/exercism/ruby/issues/1200.

@iHiD It is an escaping issue.

### DIFF
--- a/app/javascript/components/editor/TestSummary.tsx
+++ b/app/javascript/components/editor/TestSummary.tsx
@@ -60,13 +60,13 @@ export function TestSummary({
         {isPresent(test.message) ? (
           <div className="--info">
             <h3>{messageLabels[test.status]}</h3>
-            <pre dangerouslySetInnerHTML={{ __html: test.messageHtml }}></pre>
+            <pre dangerouslySetInnerHTML={{ __html: test.messageHtml }} />
           </div>
         ) : null}
         {isPresent(test.output) ? (
           <div className="--info">
             <h3>Your Output</h3>
-            <pre dangerouslySetInnerHTML={{ __html: test.outputHtml }}></pre>
+            <pre dangerouslySetInnerHTML={{ __html: test.outputHtml }} />
           </div>
         ) : null}
       </div>
--- a/app/javascript/components/editor/TestRunOutput.tsx
+++ b/app/javascript/components/editor/TestRunOutput.tsx
@@ -14,7 +14,7 @@ export const TestRunOutput = ({
     />
   ) : (
     <pre className="v1-message">
-      <code dangerouslySetInnerHTML={{ __html: testRun.messageHtml }} />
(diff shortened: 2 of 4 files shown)


ErikSchierboom (approved):
@kntsoriano Is the `messageHtml` field properly named then? I would expect HTML to be rendered dangerously, but as the bug is fixed by not escaping it, is the value really HTML?

iHiD:
Yeah, I'm not sure this is correct. `messageHtml` **is** HTML, so this is going to break as soon as there's ansi used in the message I think. @kntsoriano I think the issue is more that things need escaping in `app/commands/ansi/render_html.rb`

iHiD:
Good catch, @ErikSchierboom.

Replacing this with https://github.com/exercism/website/pull/1890.

iHiD:
Nice. @joshgoebel Better?

iHiD:
This is green for me locally 🤷

app/javascript/components/editor/TestSummary.tsx
```diff
@@ -60,13 +60,13 @@ export function TestSummary({
         {isPresent(test.message) ? (
           <div className="--info">
             <h3>{messageLabels[test.status]}</h3>
-            <pre dangerouslySetInnerHTML={{ __html: test.messageHtml }}></pre>
+            <pre dangerouslySetInnerHTML={{ __html: test.messageHtml }} />
```
joshgoebel, line 63: This still seems problematic... what is the content of `test.messageHtml` exactly in this case?  Is there a non-zero chance that it can contain ANY user generated content at all (custom test errors, etc)?  Is any such content  guaranteed to be pre-escaped by the test runner itself?

joshgoebel: So far in my experience `dangerouslySetInnerHTML` should require like the sign-off of 5 different people before it's approved. :-)

joshgoebel: Allowing test runners (which run user code!) to generate HTML at all seems problematic to me from a security standpoint. _Test runners should generate JSON data that the website can interpret in a secure fashion, not raw HTML._

iHiD: > Test runners should generate JSON data that the website can interpret in a secure fashion, not raw HTML.

They do.

iHiD: However, they are preprocessed before being displayed (e.g. they ansi is converted to colours)

joshgoebel: That's fine. But RAW html from test runners should probably not be allowed/escaped.

iHiD:
@kntsoriano This is still red.

iHiD:
I fixed it :)

PR #1896, opened 2021-09-10
Hide lines and views for journey -> solution

### LINKED ISSUE #5862: All my solutions are 0 lines and 0 views?
Screen Shot 2021-09-10 at 1 55 03 AM

Some sort of caching or cron-job issue?

kntsoriano:
Closes https://github.com/exercism/exercism/issues/5862.

Screen Shot 2021-09-10 at 3 40 19 PM

### DIFF
--- a/app/javascript/components/journey/Solution.tsx
+++ b/app/javascript/components/journey/Solution.tsx
@@ -7,11 +7,11 @@ export type SolutionProps = {
   uuid: string
   privateUrl: string
   status: string
-  numViews: number
+  numViews?: number
   numStars: number
   numComments: number
   numIterations: number
-  numLoc: string
+  numLoc?: string
   lastIteratedAt: string
   exercise: {
     title: string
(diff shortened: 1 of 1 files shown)


app/javascript/components/journey/Solution.tsx
```diff
@@ -64,14 +64,18 @@ export const Solution = ({
...
-            {numLoc} lines
-          </div>
-          <div className="stat">
-            <GraphicalIcon icon="views" />
-            {numViews} {pluralize('view', numViews)}
-          </div>
+          {numLoc ? (
+            <div className="stat">
+              <GraphicalIcon icon="loc" />
+              {numLoc} lines
```
joshgoebel, line 70: Are there no languages with single line solutions?  Surely there are.  JavaScript even has some.  I think we need `pluralize` here also...

iHiD: Probably. This PR just disables this bit though.

PR #1908, opened 2021-09-13
Display all exemplar files

### LINKED ISSUE #5871: Exemplar solutions for concept exercises only show one file even if there are multiple that make up the solution
For example (java/remote-control-competition):

image

There are 3 more files that make up the full solution and only one of them is shown.

kntsoriano:
Closes https://github.com/exercism/exercism/issues/5871.

## Screenshot
Screen Shot 2021-09-13 at 5 32 25 PM

It probably needs styling.


app/javascript/components/types.ts
```diff
@@ -655,3 +655,8 @@ export type Notification = {
 }
 
 type NotificationImageType = 'icon' | 'avatar'
+
+export type GitFile = {
```
ErikSchierboom: There's not really anything specific to Git for this file. Maybe call this `FileWithContents` or `NamedFile` or something else?

ErikSchierboom (approved):
One tiny nit. Nice!

app/helpers/react_components/mentoring/session.rb
```diff
@@ -31,7 +31,7 @@ def to_s
               anonymous_mode: discussion&.anonymous_mode?
             ),
             mentor_solution: mentor_solution,
-            exemplar_solution: exercise.exemplar_files.values.first,
+            exemplar_files: SerializeGitFiles.(exercise.exemplar_files),
```
iHiD: @kntsoriano I've inlined the serializing here, as it needs custom rules for this situation (stripping the `.meta`)

app/javascript/components/mentoring/session/Guidance.tsx
```diff
@@ -109,12 +112,18 @@ export const Guidance = ({
...
-              <pre className="overflow-auto">
-                <code
-                  className={language}
-                  dangerouslySetInnerHTML={{ __html: exemplarSolution }}
-                />
-              </pre>
+              {exemplarFiles.map((file) => {
+                return (
+                  <div key={file.filename} className="exemplar-files">
+                    {exemplarFiles.length > 1 ? (
```
iHiD: @kntsoriano I added this so that the filename only shows if there are multiple. That's broken a couple of tests. Could you get them green pls.

app/javascript/components/types.ts
kntsoriano: Ah. Good point. For this, I will be very specific and call it `MentoringSessionExemplarFile` for now. We can then generalize when we have more examples.

app/helpers/react_components/mentoring/session/exemplar_file_list.rb
```diff
@@ -0,0 +1,23 @@
+module ReactComponents
```
kntsoriano: I didn't want to add this to `serializers/` because it's been getting harder for me to navigate that directory. I think we ought to have some form of object that is local to the react component it is for. I'll have a play with this approach and see where it takes us.

test/helpers/react_components/mentoring/session/exemplar_files_test.rb
```diff
@@ -0,0 +1,25 @@
+require_relative "../../../../../app/helpers/react_components/mentoring/session/exemplar_file_list"
```
kntsoriano: I'm testing without requiring Rails. This will make our test suite faster.

iHiD: I'm not sure you actually are. If you don't have Rails then `ActiveSupport::TestCase` below won't exist so it'll fail. I think you're just not referencing test_helper.rb here, which means if you run this file by itself, it'll be fast, but when you run the whole suite, the test_helper will be required for any file, so the whole thing will render normally.

I might be wrong though and it might be faster, but I can't see why or how it would be.

iHiD: So changing this back to require `test_helper` and then running results in this:

Screenshot 2021-09-22 at 16 30 47

So it's already very fast when you just run this as one file. All the slowdown really is when we start to do stuff with databases etc.

app/helpers/react_components/mentoring/session/exemplar_file_list.rb
iHiD: Yep. Serializers should only be used for things that use the API. Other things should be inlined like you've done here.

test/helpers/react_components/mentoring/session/exemplar_files_test.rb
iHiD: @kntsoriano I've reverted this for now. I'm not against doing things like this, but its important we do them systematically across everything rather than just as a one-off variation. So let's work out what we can do to speed everything up and then implement everywhere :)

app/helpers/react_components/mentoring/session/exemplar_file_list.rb
iHiD: @kntsoriano I've inlined this into the React component. We don't need extra files for every helper class that's used - we can add things into the one class instead. This is how Rails organises everything internally and it means you don't need to jump around a code base to understand what's going on, which I much prefer.

kntsoriano: My reason with breaking this up was to be able to test only what we've changed. It's getting frustrating that when we change one thing here, we need to fix three other tests. I think this is a symptom of not testing things in the right level.

test/helpers/react_components/mentoring/session/exemplar_files_test.rb
kntsoriano: It runs a lot faster without requiring the test helper:
Screen Shot 2021-09-27 at 7 11 52 AM

I know that we want to be systematic about things, but having really slow tests and fixing tests that break when we don't want them really makes it frustrating from day to day.

iHiD: > I know that we want to be systematic about things, but having really slow tests and fixing tests that break when we don't want them really makes it frustrating from day to day.

Totally agree with this. But `0.005s` is pretty fast already. Our whole test suite would run in 2s if all the tests were that fast. I don't think this is where the speed improvements are going to come from. There's many tests that take 5s to run, which are 1000x slower than this, which is where I think we need to find solutions.

app/helpers/react_components/mentoring/session/exemplar_file_list.rb
iHiD: Yeah - totally agree with this too. We should only need to change 1 test if this changes. See https://github.com/exercism/website/pull/1975

test/helpers/react_components/mentoring/session/exemplar_files_test.rb
kntsoriano: The difference is though is when you TDD and run the tests constantly. This is where the difference lies in getting instant feedback versus waiting another multiple of a second. I find that I get to focus easier and I get work done a lot faster.

I could just then do this setup locally and when I'm going to commit the changes, I'm going to just require test helper.

iHiD: Yeah, that makes sense.

PR #1934, opened 2021-09-17
Fix reputation token elasped time

### LINKED ISSUE #5620: Wrong elapsed time on reputation drop down list.
A few seconds after publishing a solution I clicked on the reputation points button and the first entry, corresponding to the just-published solution, says it was published 14 hours ago. 14, by the way, is the current hour in my time zone at the time when the bug appeared.

image

kntsoriano:
Closes https://github.com/exercism/exercism/issues/5620.


iHiD:
This is our most reported bug, so nice to have it fixed! 🥳

PR #1954, opened 2021-09-23
Remove student mentor relationships when account is deleted

### LINKED ISSUE #5926: Account deletion option does not work as intended.
Choosing to delete an account seems to reset the account instead. All track progress and public profile are removed but the account still remains.

iHiD:
Should resolve https://github.com/exercism/exercism/issues/5926

### DIFF
--- a/app/commands/user/reset_account.rb
+++ b/app/commands/user/reset_account.rb
@@ -33,6 +33,7 @@ class User
       user.mentor_discussions.update_all(mentor_id: User::GHOST_USER_ID)
       user.mentor_discussion_posts.update_all(user_id: User::GHOST_USER_ID)
       user.mentor_testimonials.update_all(mentor_id: User::GHOST_USER_ID)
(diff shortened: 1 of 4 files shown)


app/commands/user/reset_account.rb
```diff
@@ -33,6 +33,7 @@ def reset_mentoring!
       user.mentor_discussions.update_all(mentor_id: User::GHOST_USER_ID)
       user.mentor_discussion_posts.update_all(user_id: User::GHOST_USER_ID)
       user.mentor_testimonials.update_all(mentor_id: User::GHOST_USER_ID)
+      user.mentor_testimonials.update_all(mentor_id: User::GHOST_USER_ID)
```
ErikSchierboom: Isn't this the exact same line as the line above it?

iHiD: Dammed fat vim fingers. Yes. Thanks.

ErikSchierboom (approved):

PR #1955, opened 2021-09-23
Fix download URL for non-owned solutions

### LINKED ISSUE #5951: meta: The `url` entry in `.exercism/metadata.json` is empty
This prevents the use of `exercism open .`

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/5951#issuecomment-925266187

### DIFF
--- a/app/serializers/serialize_solution_for_cli.rb
+++ b/app/serializers/serialize_solution_for_cli.rb
@@ -41,12 +41,11 @@ class SerializeSolutionForCLI
   def solution_url
     # TODO: Don't let someone download a personal_uuid and
     # end up with the mentor_uuid here. Actively guard this.
-    #
-    # TODO: How is this actually used within the CLI and could
-    # we just have it as nil if the user isn't the requestor?
-    return nil unless solution.user == requester
 
-    Exercism::Routes.private_solution_url(solution)
+    return Exercism::Routes.private_solution_url(solution) if solution.user == requester
+    return Exercism::Routes.mentoring_discussion_url(discussion) if requester.mentor? && discussion
+    return Exercism::Routes.published_solution_url(solution) if solution.published?
+    return Exercism::Routes.published_solution_url(solution) if requester.mentor? && mentor_request_pending?
   end
 
   def instructions_url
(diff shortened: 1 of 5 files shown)


app/serializers/serialize_solution_for_cli.rb
```diff
@@ -41,12 +41,11 @@ def user_handle
...
-    #
-    # TODO: How is this actually used within the CLI and could
-    # we just have it as nil if the user isn't the requestor?
-    return nil unless solution.user == requester
 
-    Exercism::Routes.private_solution_url(solution)
+    return Exercism::Routes.private_solution_url(solution) if solution.user == requester
+    return Exercism::Routes.mentoring_discussion_url(discussion) if requester.mentor? && discussion
+    return Exercism::Routes.published_solution_url(solution) if solution.published?
+    return Exercism::Routes.published_solution_url(solution) if requester.mentor? && mentor_request_pending?
```
ErikSchierboom: I've ordered them like this so save the query used in `mentor_request_pending`. I _could_ do the same for the `discussion` query where I first check if someone is _not_ a mentor. I don't know if that is a common scenario though.

app/serializers/serialize_solution_for_cli.rb
```diff
@@ -79,4 +78,12 @@ def submission
   def track
     @track ||= solution.exercise.track
   end
+
+  def discussion
+    @discussion = Mentor::Discussion.find_by(mentor: requester, solution: solution)
+  end
```
ErikSchierboom: I used the same memoization pattern used in the other methods.

iHiD: I'd probably just replace this and the others with `memoize`. The LOC above doesn't actually memoize - it just resets it every time.

app/serializers/serialize_solution_for_cli.rb
iHiD: Two thoughts:
1. I think we should probably add a guard at the start of this for `return unless solution.viewable_by?(xxx)` so that method is still the central guard.
2. We should probably use the mentor request url for the final one, rather than the published solution one, which which case case we should probably put that first (although I appreciate the logic of what you've done!)

iHiD (approved):

ErikSchierboom:
@iHiD It _looks_ to me like the system test failures are unrelated. Do we have flaky tests?

PR #1961, opened 2021-09-24
Only send welcome email when someone has confirmed or skipped confirmation

iHiD:
It's **really** confusing some people that they get the confirmation email and the welcome email at the same time.

e.g.

Screenshot 2021-09-24 at 14 33 35
Screenshot 2021-09-24 at 14 33 55

(Yes, this is the sort of email I get daily about all sorts of stuff)

### DIFF
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -201,4 +201,20 @@ class UserTest < ActiveSupport::TestCase
     user.dismiss_introducer!('scratchpad')
     assert user.introducer_dismissed?('scratchpad')
   end
+
+  test "email is sent after confirmation" do
+    user = create :user
+
+    User::Notification::CreateEmailOnly.expects(:call).with(user, :joined_exercism, {})
+
+    user.confirm
+  end
+
+  test "email is sent if confirmation is skipped" do
+    user = create :user
+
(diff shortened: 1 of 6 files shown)


test/models/user_test.rb
```diff
@@ -201,4 +201,20 @@ class UserTest < ActiveSupport::TestCase
     user.dismiss_introducer!('scratchpad')
     assert user.introducer_dismissed?('scratchpad')
   end
+
+  test "email is sent after confirmation" do
```
ErikSchierboom: Maybe be slightly more specific?

```suggestion
  test "joined exercism email is sent after confirmation" do
```

test/models/user_test.rb
```diff
@@ -201,4 +201,20 @@ class UserTest < ActiveSupport::TestCase
...
+    user = create :user
+
+    User::Notification::CreateEmailOnly.expects(:call).with(user, :joined_exercism, {})
+
+    user.confirm
+  end
+
+  test "email is sent if confirmation is skipped" do
```
ErikSchierboom: Maybe be slightly more specific?

```suggestion
  test "joined exercism email is sent if confirmation is skipped" do
```

ErikSchierboom (approved):

PR #1963, opened 2021-09-24
Fix problem report

### LINKED ISSUE #5969: Editor bug reports need more meta-data
Currently if someone uses the bug report feature in the editor, no additional data is saved with the report so we don't know what track and language the report is for if it is not explicitly mentioned in the text.

iHiD:
@kntsoriano This is a good example of why I argue so hard that system tests should also test the actual flow works properly, not just what the user gets back. We didn't get the `about` field filled in for problem reports in the editor, so now we have 500 bug reports with no data as to what they're about 🙈 If we'd checked that in the system test, we'd probably have tracked down the bug before it bit us 🙂 

---

Closes https://github.com/exercism/exercism/issues/5969

### DIFF
--- a/app/controllers/api/bug_reports_controller.rb
+++ b/app/controllers/api/bug_reports_controller.rb
@@ -1,10 +1,23 @@
+# TODO: This whole class needs tests
+# TODO: Rename to ProblemReports
 module API
   class BugReportsController < BaseController
     def create
+      begin
+        exercise_slug = params[:bug_report][:exercise_slug]
+        track_slug = params[:bug_report][:track_slug]
+        if track_slug.present? && exercise_slug.present?
+          exercise = Exercise.joins(:track).where('tracks.slug': track_slug).find(exercise_slug)
+        end
+      rescue StandardError
+        # Don't let this being wrong break the bug report
+      end
+
       @report = ProblemReport.create!(
         bug_report_params.merge(
           type: :bug,
-          user: current_user
+          user: current_user,
+          about: exercise
         )
       )
 
--- a/app/helpers/react_components/editor.rb
+++ b/app/helpers/react_components/editor.rb
@@ -35,7 +35,8 @@ module ReactComponents
             slug: track.slug
           },
           exercise: {
-            title: solution.exercise.title
+            title: solution.exercise.title,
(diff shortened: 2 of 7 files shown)


kntsoriano:
Yeah, this is a good point. I think we should test that user input gets to the server or something like that?

iHiD:
I think if a system test is end-to-end then it should test:
1. The user entering the correct data / pressing the right buttons
2. The server doing what is expected
3. The server responding correctly

The current test did (1) and (3), but not (2). By checking a record got created we cover (2).

The real reason this broke though was that there was for two reasons:
1. There was not a controller test, which checked that the correct input etc
2. There's too much responsibility in the action here. This should be in a command (`ProblemReport::Create`), which handles all that functionality. If that had been in place, we wouldn't have had this bug, because all the logic inside the command would have been tested in that.

---

I'm not blaming you for this btw. You stubbed this with a `# TODO: This needs a real implementation` and I just never got round to it. But it's a useful example as to why I feel so strongly about having:
- TDD'd Commands, 
- Controller tests,
- System tests that check at least the basic thing has been created.

kntsoriano:
Let's start wrapping everything in commands then. I think it's the way to go moving forward to ensure that we are able to test the server side functionality.

I'm not really a fan of controller test unless they test edge cases and or if its the direct point of communication with the API.

iHiD:
> Let's start wrapping everything in commands then. I think it's the way to go moving forward to ensure that we are able to test the server side functionality.

Yeah, I agree.

> I'm not really a fan of controller test unless they test edge cases and or if its the direct point of communication with the API.

How come? It makes it very easy to TDD API endpoints if using controller tests, but very slow if using system tests.

kntsoriano:
I just feel that it doesn't really test how we interact with the system. I kind of feel unsafe with it. But that just maybe my biases and I haven't really explored it that much. I will start to write more controller tests moving forward to get a feel of how it works!

iHiD:
Nice. I think it's only relevant for the API controllers, where those controllers are almost the same as system tests (because you only interact with an API using code). I don't do it for normal controllers - I agree we can generally just use system tests for those :)

PR #1990, opened 2021-09-28
Update to use latest tooling jobs logic

iHiD:
This is basically updating the gem, and adding a helper file.

There's one hole in this atm, which is that doing `find` after a job has been processed won't work any more, so we have to use `new, {}` but I'll refactor this later when we're under less time pressures.

### DIFF
--- a/test/models/submission/test_run_test.rb
+++ b/test/models/submission/test_run_test.rb
@@ -101,5 +101,20 @@ class Submission::TestRunTest < ActiveSupport::TestCase
     assert_equal test_as_hash, result.as_json(1, 2, 3) # Test with arbitary args
   end
 
+  test "tooling_job" do
+    submission = create :submission
+    job = create_test_runner_job!(submission)
+    Submission::TestRun::Process.(job)
+    test_run = submission.test_run
+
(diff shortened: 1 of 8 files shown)


test/models/submission/test_run_test.rb
```diff
@@ -101,5 +101,20 @@ class Submission::TestRunTest < ActiveSupport::TestCase
...
+    Submission::TestRun::Process.(job)
+    test_run = submission.test_run
+
+    Exercism::ToolingJob.expects(:new).with(test_run.tooling_job_id, {}).returns(job)
+    job.expects(:stdout)
+    job.expects(:stderr)
+    job.expects(:metadata)
+    test_run.stdout
+    test_run.stderr
+    test_run.metadata
```
ErikSchierboom, line 116: Is this just testing whether this doesn't throw?

ErikSchierboom (approved):

test/models/submission/test_run_test.rb
iHiD: It's testing:
1. The methods are delegated correctly
2. The methods exist (because stubbing missing methods in our testing setup raises an exception)

PR #2001, opened 2021-09-30
Update iterations-footer.css

403-html:
When there are many iterations that overflow, the footer is not responsive, so to make up for this error without a lot of work that the iterations/additional options in the footer are cut off, I added an option \`overflow-x: auto\` to allow the footer to scroll if there are many iterations.


ErikSchierboom:
@403-html Thanks for submitting the PR. Unfortunately, we don't take community contributions to the website code. That said, I've created [a PR](https://github.com/exercism/website/pull/2711/files) to fix the issue (it is almost identical, but it uses the tailwind classes).

PR #2002, opened 2021-09-30
Two fixes for the mentor track selection page

samfundev:
I noticed two problems with the mentor track selection page:
 1. The number of solutions queued is missing.
 2. There should be a space between the two sentences when signing up to be a mentor.

image


kntsoriano (approved):
Looks great @samfundev! Thank you so much!

PR #2009, opened 2021-10-01
Update site updates when exercises change and add wip guard

iHiD:
This moves the guarding and updating into one command.

We currently have a race condition where this happens:
- Exercise created
- Site updated created
- Authors added

This change means that every time the exercise syncs, the site update is refreshed, not just when it's created.

---

Closes https://github.com/exercism/website/pull/2009

### DIFF
--- a/app/commands/site_updates/process_new_exercise_update.rb
+++ b/app/commands/site_updates/process_new_exercise_update.rb
@@ -1,0 +1,40 @@
+module SiteUpdates
+  class ProcessNewExerciseUpdate
+    include Mandate
+
+    initialize_with :exercise
+
+    def call
+      if exercise.wip?
+        destroy!
+      else
+        begin
+          create!
+        rescue ActiveRecord::RecordNotUnique
+          update!
+        end
+      end
+    end
+
+    def destroy!
+      SiteUpdates::NewExerciseUpdate.where(
+        exercise: exercise,
+        track: exercise.track
+      ).destroy_all
+    end
+
+    def create!
+      SiteUpdates::NewExerciseUpdate.create!(
+        exercise: exercise,
+        track: exercise.track
+      )
+    end
+
+    def update!
+      SiteUpdates::NewExerciseUpdate.find_by!(
+        exercise: exercise,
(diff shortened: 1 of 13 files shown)


test/commands/site_updates/process_new_exercise_update_test.rb
```diff
@@ -0,0 +1,40 @@
+require "test_helper"
+
+class ConceptExercise::CreateTest < ActiveSupport::TestCase
```
ErikSchierboom:
```suggestion
class SiteUpdates::ProcessNewExerciseUpdateTest < ActiveSupport::TestCase
```

test/commands/site_updates/process_new_exercise_update_test.rb
```diff
@@ -0,0 +1,40 @@
...
+    exercise = create :concept_exercise, status: :wip
+
+    create :new_exercise_site_update, exercise: exercise
+
+    SiteUpdates::ProcessNewExerciseUpdate.(exercise)
+
+    refute SiteUpdate.exists?
+  end
+
+  test "creates if not wip" do
```
ErikSchierboom: We could consider adding explicit tests for `beta` and `active`.

app/commands/site_updates/process_new_exercise_update.rb
```diff
@@ -0,0 +1,40 @@
+module SiteUpdates
+  class ProcessNewExerciseUpdate
+    include Mandate
+
+    initialize_with :exercise
+
+    def call
+      if exercise.wip?
```
ErikSchierboom: What happens in the unlikely case that a new exercise is created with the `deprecated` status?

iHiD:
@ErikSchierboom Thanks. Better?

test/commands/site_updates/process_new_exercise_update_test.rb
iHiD: Thanks. I reckon loads of flakey tests happen because of this bug.

ErikSchierboom (approved):
Nice!

PR #2018, opened 2021-10-02
Make timeouts explicit

### LINKED ISSUE #6006: Make timeouts explicit
I've mentored a student who keeps running into a situation where they can pass tests on their machine, but they get the "An error occurred while running your tests. This might mean that there was an issue in our infrastructure, or it might mean that you have something in your code that's causing our systems to break."

We believe that it might be a timeout, as the code takes a long time to run, but I see some talk in other issues that if that were the case, a timeout message would get called. We haven't seen a timeout message.

To try and duplicate the issue, I made a (poor) solution to one of the exercises that takes a long time to compute, but does pass on my machine: https://exercism.org/mentoring/external_requests/dc00fcd942444a43b0d551a407dc661b.

This has happened with 3 or 4 passing solutions for this student. He's gone and done research on algorithms and stuff to try and get his code faster, but I don't think it's the intention of the exercise authors that students should have to do that for "beginner" level exercises.

At the very least, it seems like the "timeout" messages would be helpful, or if it's another resource constraint too.

iHiD:
Rather than showing a general error message, explicitely say if the tests timed out.

Screenshot 2021-10-02 at 20 57 29

---

Fixes https://github.com/exercism/exercism/issues/6006

### DIFF
--- a/app/javascript/components/editor/TestRunSummary.tsx
+++ b/app/javascript/components/editor/TestRunSummary.tsx
@@ -161,7 +161,7 @@ const TestRunSummaryContent = ({
           <p>
             Your tests timed out. This might mean that there was an issue in our
             infrastructure, but more likely it suggests that your code is
-            running slowly. Could you have an infinite loop or similar?
+            running slowly. Is there an infinite loop or something similar?
           </p>
           <p>
             Please check your code, and if nothing seems to be wrong, try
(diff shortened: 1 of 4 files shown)


app/javascript/components/editor/TestRunSummary.tsx
```diff
@@ -160,8 +160,8 @@ const TestRunSummaryContent = ({
         <div className="ops-error">
           <p>
             Your tests timed out. This might mean that there was an issue in our
-            infrastructure, or it might mean that you have some infinite loop in
-            your code.
+            infrastructure, but more likely it suggests that your code is
+            running slowly. Could you have an infinite loop or similar?
```
kotp:
```suggestion
            running slowly. Is there an infinite loop or something similar?
```

kotp: We might add "or something similar"

iHiD:
Thanks @kotp :)

PR #2046, opened 2021-10-06
Clear iteration.published_at and solution.completed_at in Iteration::Destroy when needed

### LINKED ISSUE #5909: [Exercises] Deleting the Only Iteration should unpublish a solution
Working with an exercise that has been both completed and published, but has only one iteration.

1.  Visit the exercise page, and click the [iteration] section/tab.
2.  Click the [...] in the Analyzer/Tests area and select [Delete Iteration]

The following displays:

image

3. Select [Delete Iteration]

The following displays:
image

🔴   -expected a warning about the solution no longer being published (_there is no iteration **to** publish_).
🔴   -expected a warning about the exercise no longer being completed/answered (_there is no iteration to base those states on_).

4. Return to the exercise overview.

The following displays:
image

🔴  -expected to not have a "completed" message.
🔴  -expected to not have a "Your published solution" message/section

5.  Click on the "published solution" box
The following displays:
image

🔴  -expected a message about there being no published solution, or better yet -- this page would not exist.

From here, the state feels more confused.  If I now return to the exercise and "open in editor", the deleted iteration shows back up.  If I then "submit", the result is now two iterations, not one.  While this makes sense from an app perspective (_after all, there are really two iterations ... we just deleted one of them_), it feels a bit weird from a user perspective.

image

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/5909

### DIFF
--- a/app/commands/iteration/destroy.rb
+++ b/app/commands/iteration/destroy.rb
@@ -6,7 +6,8 @@ class Iteration
 
     def call
       iteration.update!(deleted_at: Time.current)
-      solution.update!(published_iteration_id: nil) if solution.published_iteration_id == iteration.id
+      solution.update!(published_iteration_id: nil, published_at: nil) if solution.published_iteration_id == iteration.id
+      solution.update!(completed_at: nil) if solution.iterations.not_deleted.empty?
     end
 
     private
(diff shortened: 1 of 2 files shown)


app/commands/iteration/destroy.rb
```diff
@@ -6,7 +6,8 @@ class Destroy
 
     def call
       iteration.update!(deleted_at: Time.current)
-      solution.update!(published_iteration_id: nil) if solution.published_iteration_id == iteration.id
+      solution.update!(published_iteration_id: nil, published_at: nil) if solution.published_iteration_id == iteration.id
+      solution.update!(completed_at: nil) if solution.iterations.not_deleted.empty?
```
iHiD: Rather than `empty?` you might like to do `exists?` which is quicker and change to `unless`

PR #2053, opened 2021-10-07
Fix typos in `publish-exercise` modal

fernandogmo:
`its publish` -> `it's published`

Also fixes inconsistency between `.haml` and `.tsx` files for this modal.


iHiD:
Thanks :)

PR #2061, opened 2021-10-12
Calculate median time for tracks and exercises

### LINKED ISSUE #15: Add median wait times back in
https://github.com/exercism/v3-beta/issues/129

ErikSchierboom:
Closes https://github.com/exercism/project-management/issues/15


app/commands/exercise/update_median_wait_times.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+    end
+
+    private
+    def median_wait_time(exercise)
+      wait_times = Mentor::Discussion.
+        joins(:request).
+        select('TIMESTAMPDIFF(SECOND, request.created_at, mentor_discussions.created_at) AS wait_time').
+        where('request.exercise_id': exercise.id).
+        map(&:wait_time)
```
ErikSchierboom, line 17: This executes the following query:

```sql
SELECT TIMESTAMPDIFF(SECOND, request.created_at, mentor_discussions.created_at) AS wait_time FROM `mentor_discussions` INNER JOIN `mentor_requests` `request` ON `request`.`id` = `mentor_discussions`.`request_id` WHERE `request`.`exercise_id` = 300
```

This query uses two keys and is thus quite fast.

app/commands/track/update_median_wait_times.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    private
+    def median_wait_time(track)
+      wait_times = Mentor::Discussion.
+        joins(:request).
+        joins(:exercise).
+        where('exercise.track_id': track.id).
+        select('TIMESTAMPDIFF(SECOND, mentor_requests.created_at, mentor_discussions.created_at) AS wait_time').
+        map(&:wait_time)
```
ErikSchierboom, line 18: This executes the following query:

```sql
SELECT TIMESTAMPDIFF(SECOND, mentor_requests.created_at, mentor_discussions.created_at) AS wait_time
FROM `mentor_discussions`
INNER JOIN `mentor_requests` ON `mentor_requests`.`id` = `mentor_discussions`.`request_id`
INNER JOIN `solutions` ON `solutions`.`id` = `mentor_discussions`.`solution_id`
INNER JOIN `exercises` `exercise` ON `exercise`.`id` = `solutions`.`exercise_id`
WHERE `exercise`.`track_id` = 3
```

This query uses only PRIMARY keys as is thus quite fast.

app/commands/exercise/update_median_wait_times.rb
ErikSchierboom: The v2 code had a separate class for the shared bit on joining requests with discussions, but the queries are now subtly different so I've opted not to do that. I also don't think it is a real problem due to the good test coverage I've added in this PR and I'd rather optimize this query than deduplicate the code.

app/javascript/components/common/MedianWaitTime.tsx
```diff
@@ -0,0 +1,29 @@
...
+import {
+  durationFromSeconds,
+  durationTimeElementFromSeconds,
+} from '../../utils/time'
+
+type MedianWaitTimeProps = {
+  seconds?: number
+}
+
+export function MedianWaitTime({ seconds }: MedianWaitTimeProps) {
```
ErikSchierboom, line 11: I've created a new component to render the median wait time, as we are displaying it in two locations and might want to use it in more.

app/javascript/components/common/MedianWaitTime.tsx
```diff
@@ -0,0 +1,29 @@
...
+  if (seconds === undefined || seconds === null) {
+    return null
+  }
+
+  const medianWaitTime = durationFromSeconds(seconds)
+
+  return (
+    <div>
+      Recent median waiting time: ~
+      <time
```
ErikSchierboom, line 21: I'm rendering the waiting time as a `<time>` element to help with A11y.

app/javascript/components/common/MedianWaitTime.tsx
```diff
@@ -0,0 +1,29 @@
...
+        dateTime={durationTimeElementFromSeconds(seconds)}
```
ErikSchierboom, line 22: It turned out the `<time>` element support a time duration, which I've used here.

config/sidekiq-schedule.yml
```diff
@@ -48,3 +48,9 @@ sync_tracks:
   cron: "0 1 * * 3"
   class: "SyncTracksJob"
   queue: dribble
+
+# 1am Thursday
+update_median_wait_times:
+  cron: "0 1 * * 4"
```
ErikSchierboom, line 59: Is once a week good enough, or do we want to run this more often?

PR #2073, opened 2021-10-19
Show multiple tests in tests tab

### LINKED ISSUE #6033: Website does not show multiple test files
In the Go track we have some exercises that contain two test files, one with the actual test code and one for the test data table (cases_test.go).

Since students had issues solving some  exercises without seeing the cases file on the website, we added it to the files.test key in the config.json of the website. However, the cases file did not show up on the website.

We were wondering whether the website actually supports multiple test files. As a workaround we added the cases file to files.editor instead as it was done in the Python track.

If multiple test files are indeed not supported currently, I want to say that at least for the Go track we are planning to remove the cases files anyway in the future so we wouldn't need that feature anymore.

In any case, if listing multiple files under files.test will not work for some time to come, it might make sense that configlet throws an error that says to add additional files to files.editor instead.

If multiple test files are supported in the website already, then this issue can be closed as the bug was probably just related to the update process issue tracked in #6010.

kntsoriano:
### UI Changes:

An exercise with multiple test files:

### Latest update 
Test tabs in mentoring discussion/representation discussion - preview

Screenshot 2022-12-15 at 17 38 11

### Before:

Screenshot 2022-12-14 at 17 45 02

### After:

Screenshot 2022-12-15 at 11 36 50

Screenshot 2022-12-14 at 17 45 50

An exercise with a single test file:

Screenshot 2022-12-14 at 17 51 40

---

closes https://github.com/exercism/exercism/issues/6033

@iHiD The HTML rendered is correct now, but the CSS for the tests panel only accommodates one file. I also updated the mentor discussion to show multiple tests.


app/models/git/exercise.rb
```diff
@@ -131,7 +131,7 @@ def example_files
       {}
     end
 
-    def tests
+    def test_files
```
ErikSchierboom, line 144: This was the root of all the above changes, as we also have `exemplar_files`, `solution_files`, etc. as methods in this class.

app/javascript/components/editor/Props.tsx
```diff
@@ -34,7 +34,7 @@ type EditorPanels = {
     exampleFiles: File[]
   }
   tests?: {
-    tests: readonly TestFile[]
+    testFiles: readonly TestFile[]
```
ErikSchierboom: `tests.testFiles` also makes a bit more sense than `tests.tests`.

iHiD:
Screenshot 2022-02-24 at 13 54 04

Screenshot 2022-02-24 at 13 54 17

I see this in the editor. Are there supposed to be tabs here for the tests? Or does that still need to be implemented?  Feels like there's HTML structure etc missing.

ErikSchierboom:
> I see this in the editor. Are there supposed to be tabs here for the tests? Or does that still need to be implemented? Feels like there's HTML structure etc missing.

I think the tabs might be missing. I'm not sure if Karlo planned on adding them, or whether he was unsure what the desired HTML was.

kntsoriano:
The HTML is correct for this branch, but we still need the CSS I think.

app/css/components/iteration-pane.css
```diff
@@ -32,3 +32,27 @@
...
+        margin-bottom: -1px;
+        @apply border-b-1;
+
+        .selected {
+            @apply bg-backgroundColorA;
+            @apply text-textColor2 font-medium;
+            border-bottom: 1px solid transparent;
+        }
+    }
+}
```
iHiD: This should be `c-...` or nested, not global.

ErikSchierboom:
Looks really good:

image

And still works as expected for single test files:

image

iHiD:
@ErikSchierboom Could you do a final check and merge pls?

iHiD (approved):

PR #2074, opened 2021-10-20
Upgrade to codemirror 0.19.0

kntsoriano:
ttps://github.com/exercism/exercism/issues/6046

- Update elixir package to be compatible with 0.19.0
- Use julia-lang package

The only problem with this is that https://github.com/exercism/codemirror-lang-wren isn't compatible with codemirror 0.19.0. We should bump that first before merging this.


package.json
```diff
@@ -33,7 +33,7 @@
     "ace-builds": "^1.4.12",
     "actioncable": "^5.2.4-3",
     "chart.js": "^3.1.0",
-    "codemirror-lang-elixir": "^1.0.2",
+    "codemirror-lang-elixir": "https://github.com/sachinraja/codemirror-lang-elixir",
```
kntsoriano, line 36: The package on npm isn't updated to the latest version. I pointed the package to github instead.

joshgoebel:
> The only problem with this is that https://github.com/exercism/codemirror-lang-wren isn't compatible with codemirror 0.19.0. 

Is what is involved in this mentioned anywhere?  I tried to see what was involved in "Update elixir package to be compatible with 0.19.0" but nothing popped when skimming this PR.

kntsoriano:
@joshgoebel I'm not sure what changed between codemirror 0.18 and 0.19. You could try upgrading the wren package and see what happens. Hopefully, it does not entail too much.

joshgoebel:
@exercism/codemirror-lang-wren@0.7.2 is pinned against 0.19 code mirror now.

kntsoriano:
@joshgoebel https://github.com/codemirror/language/blob/cd7f7e66fa51ddbce96cf9396b1b6127d0ca4c94/CHANGELOG.md#0190-2021-08-11

`LezerLanguage` was updated to `LRLanguage`. Can you make the changes to the wren highlighter?

joshgoebel:
As in: https://github.com/exercism/codemirror-lang-wren/commit/d512d2a0cbc9156c861e6860d7f8379bea38f838

Is this really a blocker for your though as I already bumped versions and all my tests are green... I presume on the CodeMirror side that (for compatibility) LanguageLezer and LRLezer just point to the same object/class, no?

Let me know if that PR looks good and if you really need a whole new NPM release.

kntsoriano:
@joshgoebel Thanks! That works well :)

@iHiD @ErikSchierboom This should be ready.

joshgoebel:
I could have pushed another actual release. I was just curious how the naming  was blocking you.

iHiD:
Thanks :)

PR #2080, opened 2021-10-21
Add IDs to document headings

### LINKED ISSUE #5984: Add IDs to docs headings
Currently, headings on exercism.org/docs pages are rendered without IDs. This makes linking to them impossible. It would be great if we could link to these specific headings.

ErikSchierboom:
**Ready to Merge**

---

Closes https://github.com/exercism/exercism/issues/5984

### DIFF
--- a/app/commands/markdown/parse.rb
+++ b/app/commands/markdown/parse.rb
@@ -8,11 +8,12 @@ class Markdown::Parse
     new(*args, **kwargs).()
   end
 
-  def initialize(text, nofollow_links: false, strip_h1: true, lower_heading_levels_by: 1)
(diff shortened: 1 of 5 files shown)


iHiD:
Oooh, nice. I think we should prefix (or maybe suffix) them all though, so they don't clash with "real" ids. Let's choose (and reserve for this) a prefix? we could use `#d-...` for `doc-`

ErikSchierboom:
What about using `h-` for heading?

iHiD:
Sure :) Tests are red.

ErikSchierboom:
Fixed

PR #2084, opened 2021-10-22
Enable lines of code counter

ErikSchierboom:
- Allow calculating lines of code
- Automatically calculate LoC when new iteration is created

Depends on https://github.com/exercism/lines-of-code-counter/pull/1 being merged and deployed


app/commands/solution/publish_iteration.rb
```diff
@@ -0,0 +1,23 @@
...
+    private
+    memoize
+    def iteration
+      solution.iterations.find_by(idx: iteration_idx)
+    end
+
+    def num_loc
+      return solution.iterations.last.num_loc if iteration_idx.nil?
+
+      iteration.num_loc
```
iHiD: The idx might be present but invalid. And the `iterations.last` might be deleted so we need to use `latest_iteration` (which might also be nil, thus the `&`). Could you add tests accordingly pls?

```suggestion
      iteration ? iteration.num_loc : solution.latest_iteration&.num_loc
```

ErikSchierboom: Done. I also changed the logic in the `GenerateIterationSnippetJob` to use the latest iteration instead of the last: https://github.com/exercism/website/pull/2084/files#diff-578db803a141a22b493d6bc4dec4f050741e90e57b9923dc8b959f17bf8a28e7R40. Is that correct? If so, I'll also update the snippet generator job, from which I copied the existing code.

test/jobs/calculate_lines_of_code_job_test.rb
```diff
@@ -0,0 +1,138 @@
...
+    assert_equal 0, iteration.reload.num_loc
+    assert_equal 0, iteration.solution.reload.num_loc
+  end
+
+  test "solution is updated if iteration is latest" do
+    submission = create :submission
+    create :submission_file, submission: submission, content: "Some source code"
+    create :iteration, solution: submission.solution
+    latest_iteration = create :iteration, submission: submission
+    create :iteration, solution: submission.solution, deleted_at: Time.current # Last iteration
```
ErikSchierboom, line 47: @iHiD I added this iteration for the last vs latest functionality.

app/commands/solution/publish_iteration.rb
iHiD: Yes, that's correct. Good catch :)

ErikSchierboom: Updated

db/migrate/20211028110123_change_solutions_num_loc_to_null.rb
```diff
@@ -0,0 +1,7 @@
+class ChangeSolutionsNumLocToNull < ActiveRecord::Migration[6.1]
+  def change
+    add_column :solutions, :num_loc, :integer, null: true
+
+    Solution.where(num_loc: 0).update(num_loc: nil)
```
iHiD:
```suggestion
```

I'll run this manually on bastion in batches. Otherwise it'll lock the whole table for ages in the migration.

iHiD:
Green! 🦜

PR #2093, opened 2021-10-25
Fix median wait times

### LINKED ISSUE #6060: "Recent median waiting time: ~0 seconds"
When I try to request mentoring in for example Elixir or Go, I'm seeing this statistic that's definitely wrong - "Recent median waiting time: ~0 seconds". @szabgab reports the same issue with Python so I assume this happens for all tracks.

Screen Shot 2021-10-24 at 16 38 14

iHiD:
@ErikSchierboom This needs tests adding, and a similar thing for the exercises.

---

Fixes https://github.com/exercism/exercism/issues/6060

### DIFF
--- a/app/commands/track/update_median_wait_times.rb
+++ b/app/commands/track/update_median_wait_times.rb
@@ -15,7 +15,9 @@ class Track
         joins(:exercise).
         where('exercise.track_id': track.id).
         select('TIMESTAMPDIFF(SECOND, mentor_requests.created_at, mentor_discussions.created_at) AS wait_time').
-        map(&:wait_time)
+        where('mentor_discussions.created_at > ?', Time.current - 4.weeks).
+        map(&:wait_time).
+        reject(&:zero?)
 
       calculate_median(wait_times)
     end
--- a/app/commands/track/update_median_wait_times.rb
+++ b/app/commands/track/update_median_wait_times.rb
@@ -14,10 +14,10 @@ class Track
         joins(:request).
         joins(:exercise).
         where('exercise.track_id': track.id).
-        select('TIMESTAMPDIFF(SECOND, mentor_requests.created_at, mentor_discussions.created_at) AS wait_time').
         where('mentor_discussions.created_at > ?', Time.current - 4.weeks).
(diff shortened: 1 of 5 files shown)


app/commands/track/update_median_wait_times.rb
```diff
@@ -15,7 +15,9 @@ def median_wait_time(track)
         joins(:exercise).
         where('exercise.track_id': track.id).
         select('TIMESTAMPDIFF(SECOND, mentor_requests.created_at, mentor_discussions.created_at) AS wait_time').
-        map(&:wait_time)
+        where('mentor_discussions.created_at > ?', Time.current - 4.weeks).
```
iHiD, line 17: Only include things that have started been discussed in the last 4 weeks (this removes all the old data which is largely zeros)

app/commands/track/update_median_wait_times.rb
```diff
@@ -15,7 +15,9 @@ def median_wait_time(track)
...
+        map(&:wait_time).
+        reject(&:zero?)
```
iHiD: Add a final guard against any weird things.

iHiD: I think this should actually be reject anything under 5s, because when we created private mentors discussions, it immediately creates a request and a discussion with it, but they could potentially fall a couple of seconds apart if our servers are running slow etc.

app/commands/track/update_median_wait_times.rb
ErikSchierboom: Right. That makes sense. I hadn't anticipated the old data being zeros.

app/commands/track/update_median_wait_times.rb
ErikSchierboom: Sure. I'll update.

db/seeds.rb
```diff
@@ -97,7 +97,8 @@
       blurb: git_track.blurb,
       tags: git_track.tags,
       repo_url: repo_url,
-      synced_to_git_sha: first_commit.oid
+      synced_to_git_sha: first_commit.oid,
+      active: false
```
ErikSchierboom: Was this change intended to go in this PR?

iHiD: Nope. I'll let you remove it :)

PR #2095, opened 2021-10-25
Add mailshot for v3 launch

iHiD:
@ErikSchierboom Can you check this looks sane pls. I'll run a script to actually batch send the things, by calling `User::Mailshot::Send.(..., :v3_launch)` for each user we want to send too. We can discuss priority order tomorrow.

### DIFF
--- a/app/views/mailers/mailshots_mailer/v3_launch.text.erb
+++ b/app/views/mailers/mailshots_mailer/v3_launch.text.erb
@@ -1,0 +1,20 @@
+Hi <%= @user.handle %>,
+
+Two years ago we set out to fix the major problems with Exercism, especially the frustrations of losing momentum waiting for mentors. Solving this was a complex endeavour, and involved rethinking the whole flow of using the platform, redesigning our mentoring paradigm, and creating a whole new set of content to teach language concepts, not just to practice them.
+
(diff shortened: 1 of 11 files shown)


app/views/mailers/mailshots_mailer/v3_launch.text.erb
```diff
@@ -0,0 +1,20 @@
+Hi <%= @user.handle %>,
+
+Two years ago we set out to fix the major problems with Exercism, especially the frustrations of losing momentum waiting for mentors. Solving this was a complex endeavour, and involved rethinking the whole flow of using the platform, redesigning our mentoring paradigm, and creating a whole new set of content to teach language concepts, not just to practice them.
+
+Last month we finally launched the new platform, rewritten from the ground up, with new Learning Mode, a totally redesigned mentoring experience, an online editor, and much more. Since launch, our average mentor wait-time has dropped from a few days to three hours, and in many tracks is now under an hour!
```
ErikSchierboom, line 5:
```suggestion
Last month we finally launched the new platform, rewritten from the ground up, with a new Learning Mode, a totally redesigned mentoring experience, an online editor, and much more. Since launch, our average mentor wait-time has dropped from a few days to three hours, and in many tracks is now under an hour!
```

ErikSchierboom (approved):

PR #2100, opened 2021-10-26
Improve healthcheck to use DB

iHiD:


app/controllers/pages_controller.rb
```diff
@@ -27,6 +27,13 @@ def index
   end
 
   def health_check
-    render json: { ruok: true }
+    user = User.find(User::SYSTEM_USER_ID)
```
ErikSchierboom, line 30: Ah, so this now only succeeds when the database is ready and migrations have run. Nice.

iHiD: Yep, exactly! :)

PR #2104, opened 2021-10-27
Hide server version

luilver:
#### Description

Showing the server version is a security flaw:

image

Server token off directive avoids to expose it as per described here: http://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens


iHiD:
@dougal A second pair of eyes pls?

joshgoebel:
Do we have a commercial ngninx subscription? Setting a custom value would be fun.

joshgoebel (approved):
I agree this is a good security practice.

iHiD:
> Do we have a commercial ngninx subscription?

No :)

dougal (approved):
Change looks good.

Agree, is good security practice.

PR #2123, opened 2021-10-30
Fix: Let's play fairly

kotp:


iHiD:
It's supposed to be "pay" :)

PR #2130, opened 2021-11-01
Speed things up

iHiD:
This PR speeds up the slowest requests by about 2s, the average 95th percentile request by 50ms. I also added another missing index in production on the auth-tokens, which has sped up all API requests by about 200ms.

I've already run the migrations in production FYI.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -257,10 +257,7 @@ class User < ApplicationRecord
     dismissed_introducers.where(slug: slug).exists?
   end
 
-  # TODO: Remove if not used by launch
-  # def favorited_by?(mentor)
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -257,10 +257,7 @@ def introducer_dismissed?(slug)
...
 
-  # TODO: Remove if not used by launch
-  # def favorited_by?(mentor)
-  #   relationship = Mentor::StudentRelationship.find_by(student: self, mentor: mentor)
-
-  #   relationship ? relationship.favorited? : false
-  # end
+  def send_devise_notification(notification, *args)
+    devise_mailer.send(notification, self, *args).deliver_later
+  end
```
iHiD, line 262: This uses background jobs for email sending.

PR #2135, opened 2021-11-02
Add median wait time to track select

### LINKED ISSUE #2101: Revert #1796
We now have median wait times calculated (!! 🦜) so we can undo #1796. Could you do that pls, @kntsoriano? Thanks!

The correct column is tracks.median_wait_time - we should probably rename throughout to be consistent 🙂

kntsoriano:
Closes https://github.com/exercism/website/issues/2101.

Screen Shot 2021-11-02 at 2 13 22 PM

This needs fixing CSS.


ErikSchierboom (approved):

test/system/flows/mentor_changes_mentored_tracks_test.rb
```diff
@@ -19,8 +19,8 @@ class MentorChangesMentoredTracksTest < ApplicationSystemTestCase
         visit mentoring_queue_path
         within(".mentor-queue-filtering") { click_on "Ruby" }
         click_on "Change the tracks you mentor"
-        find("label.track", text: "Ruby").click
-        find("label.track", text: "C#").click
+        find("label.track", text: "Ruby\nRecent median waiting time: ~16 minutes").click
+        find("label.track", text: "C#\nRecent median waiting time: ~33 minutes").click
```
iHiD: `Recent median waiting time` needs to change to `Avg. wait time` as it won't fit responsively here if it's that long. I don't mind it changing in `common` to this too.

iHiD:
Thanks for sorting this! One comment to fix pls :)

kntsoriano:
@iHiD This should be ready now.

test/system/flows/mentor_changes_mentored_tracks_test.rb
habere-et-dispertire: What is the limit for fitting responsively? Do we have room here for `Average wait time`?
(Fewer abbreviations means less confusion, particularly to non-English speakers unfamiliar with this abbreviation.)

iHiD: I'm not sure. If you add an issue referencing this to exercism/exercism we'll take a look in due course (otherwise I'll forget a comment in an old PR here)

Issue #200, opened 2020-10-20
Fix eslint issues


ErikSchierboom:
In https://github.com/exercism/v3-website/pull/174 support for ESLint was added. That PR doesn't check/update the existing code, which is something that we should be doing.

zacnomore:
~~It looks like there are no linting errors currently, can this be closed or are you looking for something like `lint-staged`?~~
```
./bin/script website lint
Inspecting 1304 files
1304 files inspected, no offenses detected
```

EDIT: The lint script in the dev-env repo doesn't run all of the linters and formatters, I think it's just rubocop. There are eslint issues

zacnomore:
I see now husky is configured to lint staged files already, I would suggest this be closed.

PR #2147, opened 2021-11-05
Eslint - Add scoping to switch statement declarations

zacnomore:
To help address #200 

This adds blocks to each case to ensure scope is intuitive
More details: https://eslint.org/docs/rules/no-case-declarations

### DIFF
--- a/app/javascript/components/misc/CodeMirror/languageCompartment.ts
+++ b/app/javascript/components/misc/CodeMirror/languageCompartment.ts
@@ -7,177 +7,227 @@ export const loadLanguageCompartment = async (
   language: string
 ): Promise<Extension> => {
   switch (language) {
-    case 'cpp':
+    case 'cpp': {
       const { cpp } = await import('@codemirror/lang-cpp')
       return compartment.of(cpp())
-    case 'java':
+    }
+    case 'java': {
       const { java } = await import('@codemirror/lang-java')
       return compartment.of(java())
-    case 'cfml':
+    }
+    case 'cfml': {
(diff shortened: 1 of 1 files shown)


iHiD:
@SleeplessByte @ErikSchierboom @kntsoriano Could one of you take a look pls.

app/javascript/components/misc/CodeMirror/languageCompartment.ts
```diff
@@ -7,177 +7,227 @@ export const loadLanguageCompartment = async (
...
       return compartment.of(java())
-    case 'cfml':
+    }
+    case 'cfml': {
       const { javascript: cfml } = await import('@codemirror/lang-javascript')
       return compartment.of(cfml())
+    }
     case 'javascript':
-    case 'typescript':
+    case 'typescript': {
```
ErikSchierboom, line 23: The documentation's "When not to use" section says:

> If you depend on fall through behavior and want access to bindings introduced in the case block.

Should we copy the code to not have fall through behavior then?

SleeplessByte: No, this works.

```typescript
switch (language) {
  case 'javascript':
  case 'typescript': { 
    // runs on both js and ts
    console.log(language);
  } 
}

// => 'javascript' is logged
```

SleeplessByte (approved):

kntsoriano (approved):

ErikSchierboom (approved):
Thanks!

PR #2149, opened 2021-11-05
Use opensearch for community solutions

ErikSchierboom:
- Working on opensearch
- Allow indexing of solutions
- Add elasticsearch-dsl gem
- Working on opensearch
- Revert
- Add indexing tests
- Remove elasticsearch-dsl
- Use opensearch to search for community solutions
- Use opensearch for community solutions
- Fix rubocop


app/commands/solution/index.rb
```diff
@@ -0,0 +1,65 @@
+class Solution::Index
```
iHiD:
```suggestion
class Solution::SyncToSearchIndex
```

As `index` can be a verb or a noun, I initially read this as being a "thing", not a command, so I think I'd rather rename to be more explicit.

app/commands/solution/index.rb
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    client.index(index: 'solutions', type: 'solution', body: body)
+  end
+
+  private
+  memoize
+  def published_iteration
+    solution.published_iterations.last
+  end
```
iHiD: Seeing this made me realise that when we add a script to bulk-add these, we're going to need to do lots of `includes` to avoid n+1s. Do you have a bulk-add script yet?

app/commands/solution/mark_as_out_of_date_in_index.rb
```diff
@@ -0,0 +1,32 @@
+class Solution::MarkAsOutOfDateInIndex
```
iHiD: I wonder if this should be `Exercism::MarkSolutionsAsOutOfDateInIndex` as it's acting on an exercise, not a solution. I could see it either way. Thoughts?

app/commands/solution/mark_as_out_of_date_in_index.rb
```diff
@@ -0,0 +1,32 @@
...
+  def call
+    client.update_by_query(index: 'solutions', body: {
+      script: {
+        source: 'ctx._source.out_of_date = false'
+      },
+      query: {
+        bool: {
+          must: [
+            { term: { 'exercise_id': exercise.id } },
+            { term: { 'out_of_date': true } }
```
iHiD: Does this not only change things that are `out_of_date = true` and change them to `out_of_date` = false? Did you mean to call the field "up_to_date" or is the logic wrong, or am I misreading it?

app/commands/solution/search_community_solutions.rb
```diff
@@ -17,25 +17,64 @@ def initialize(exercise, page: nil, per: nil, criteria: nil)
     end
 
     def call
-      @solutions = exercise.solutions.published
+      results = client.search(index: 'solutions', body: search_body)
 
-      filter_criteria!
-      sort!
+      solution_ids = results["hits"]["hits"].map { |hit| hit["_source"]["id"] }
+      solutions = Solution.where(id: solution_ids).
```
iHiD: Before this line, you need to deal with the fact that `solution_ids` might be `empty`, because the query below will break if there are no ids.

I'd probably do `solutions = solution_ids.present? ? .. : []` or something.

app/commands/solution/search_community_solutions.rb
```diff
@@ -17,25 +17,64 @@ def initialize(exercise, page: nil, per: nil, criteria: nil)
...
+        includes(:exercise, :track).
+        order(Arel.sql("FIND_IN_SET(id, '#{solution_ids.join(',')}')")).
+        to_a
 
-      @solutions.page(page).per(per)
+      total_count = results["hits"]["total"]["value"].to_i
+      Kaminari.paginate_array(solutions, total_count: total_count).
+        page(page).per(per)
```
iHiD, line 31: Just for niceness...

```suggestion
      Kaminari.paginate_array(
        solutions, 
        total_count: results["hits"]["total"]["value"].to_i
      ).page(page).per(per)
```

app/commands/solution/search_community_solutions.rb
```diff
@@ -17,25 +17,64 @@ def initialize(exercise, page: nil, per: nil, criteria: nil)
...
+        query: search_query,
+        sort: search_sort,
 
-      @solutions = @solutions.joins(:user).where("users.handle LIKE ?", "%#{criteria}%")
+        # Only return the solution IDs, not the entire document, to improve performance
+        _source: [:id],
+
+        # Paging information
+        from: (page - 1) * per,
+        size: per
```
iHiD, line 50: Just checked above to make sure these weren't ever possibly nil and noticed you could remove the `....present? && ` checks from L14 and L15 too if you fancied it.

app/commands/solution/search_community_solutions.rb
```diff
@@ -17,25 +17,64 @@ def initialize(exercise, page: nil, per: nil, criteria: nil)
...
     end
 
-    def sort!
-      @solutions = @solutions.order(num_stars: :desc, id: :desc)
+    def search_query
+      {
+        bool: {
+          must: [
+            { term: { 'exercise.id': exercise.id } },
+            { exists: { field: :published_at } },
```
iHiD: No idea which is quicker, but you could also check for `status` being `"published"` here, rather than the existence of the field.

app/commands/solution/search_user_solutions.rb
```diff
@@ -21,60 +21,67 @@ def initialize(user, criteria: nil, track_slug: nil, status: nil, mentoring_stat
...
-      filter_criteria!
-      filter_track!
-      filter_status!
-      filter_mentoring_status!
-      sort!
+      results = client.search(index: 'solutions', body: search_body)
 
-      @solutions.page(page).per(per)
+      solution_ids = results["hits"]["hits"].map { |hit| hit["_source"]["id"] }
+      solutions = Solution.where(id: solution_ids).
```
iHiD: Same comments as above about empty results.

app/jobs/mark_solutions_as_out_of_date_in_index_job.rb
```diff
@@ -0,0 +1,7 @@
+class MarkSolutionsAsOutOfDateInIndexJob < ApplicationJob
```
iHiD, line 1: Ha - you've basically used the naming I suggested here already :)

app/serializers/serialize_community_solutions.rb
```diff
@@ -4,7 +4,7 @@ class SerializeCommunitySolutions
   initialize_with :solutions
 
   def call
-    solutions.includes(:exercise, :track).map do |solution|
+    solutions.map do |solution|
```
iHiD: Can you do two things here:
1. Add a comment here saying that `includes` occur in the searching upstream as the kaminari array is manually put together, with a warning that if this is used outside of that then, `includes(:exercise, :track)` should be added. (You could also check for the type of `solutions` and if it's an `ActiveRecord::Relation` add the includes, but the comment is probably fine for now).
2. Check that this serializer isn't called anywhere else. If it is then you'll need to do the actual work in (1) not add the comment.

iHiD (approved):
Looks really good :)

app/commands/solution/index.rb
ErikSchierboom: Should be just `Solution.find_each {|s| Solution::Index.(s)}`. And then with the right `includes`

app/commands/solution/mark_as_out_of_date_in_index.rb
ErikSchierboom: Yeah, I considered that. I'm perfectly happy with that.

app/commands/solution/mark_as_out_of_date_in_index.rb
ErikSchierboom: I think the logic is inverted.

app/commands/solution/search_community_solutions.rb
```diff
@@ -17,25 +17,68 @@ def initialize(exercise, page: nil, per: nil, criteria: nil)
...
+          includes(:exercise, :track).
+          order(Arel.sql("FIND_IN_SET(id, '#{solution_ids.join(',')}')")).
+          to_a : []
 
-      @solutions.page(page).per(per)
+      total_count = results["hits"]["total"]["value"].to_i
+      Kaminari.paginate_array(solutions, total_count: total_count).
+        page(page).per(per)
+    rescue StandardError
+      Fallback.(exercise, page, per, criteria)
```
iHiD:
```suggestion
    rescue StandardError => e
      Bugsnag.notify(e)
      Fallback.(exercise, page, per, criteria)
```

app/commands/solution/search_user_solutions.rb
```diff
@@ -21,59 +21,129 @@ def initialize(user, criteria: nil, track_slug: nil, status: nil, mentoring_stat
...
+        Solution.where(id: solution_ids).
+          includes(:exercise, :track).
+          order(Arel.sql("FIND_IN_SET(id, '#{solution_ids.join(',')}')")).
+          to_a : []
+
+      total_count = results["hits"]["total"]["value"].to_i
+      Kaminari.paginate_array(solutions, total_count: total_count).
+        page(page).per(per)
+    rescue StandardError
+      Fallback.(user, page: page, per: per, track_slug: track_slug, status: status, mentoring_status: mentoring_status,
```
iHiD:
```suggestion
    rescue StandardError => e
      Bugsnag.notify(e)
      Fallback.(user, page: page, per: per, track_slug: track_slug, status: status, mentoring_status: mentoring_status,
```

app/commands/solution/search_user_solutions.rb
```diff
@@ -21,59 +21,129 @@ def initialize(user, criteria: nil, track_slug: nil, status: nil, mentoring_stat
...
+      def initialize(user, criteria: nil, track_slug: nil, status: nil, mentoring_status: nil, page: nil, per: nil, order: nil)
+        @user = user
+        @criteria = criteria
+        @track_slug = track_slug
+        @status = status
+        @mentoring_status = mentoring_status
+        @page = page
+        @per = per
+        @order = order
+      end
```
iHiD: I'd be pretty tempted to get rid of the kwargs and the attr_readers and just use initialize_with here, as its an internal class, rather than something more "public".

app/commands/solution/sync_to_search_index.rb
```diff
@@ -0,0 +1,55 @@
...
+      },
+      published_iteration: published_iteration ? {
+        tests_passed: published_iteration.submission.tests_passed?,
+        code: published_iteration.submission.files.map(&:content) || []
+      } : nil,
+      latest_iteration: latest_iteration ? {
+        tests_passed: latest_iteration.submission.tests_passed?,
+        code: latest_iteration.submission.files.map(&:content) || []
+      } : nil
+    }
```
iHiD: I'd move this into a method called `body`

iHiD (commented):
Just some small comments - looking good! :)

test/test_helper.rb
```diff
@@ -232,6 +233,27 @@ def download_s3_file(bucket, key)
       key: key
     ).body.read
   end
+
+  ###################
+  # OpenSearch Helpers #
+  ###################
+  def reset_opensearch!
```
ErikSchierboom, line 241: I've added some helper methods. Let me know if this is the right approach

PR #2154, opened 2021-11-09
Fallback to DB Search

iHiD:
If ElasticSearch fails (e.g. during us deploying things), fall back to normal search. We don't need to keep all the fallback options to match the ES ones, but just something so it's not totally wrong. As a follow-on step we can add a status that says we're using fallback and display that to a user.


ErikSchierboom:
Merged.

PR #2160, opened 2021-11-12
Fix order of they/them pronouns

andrewda:
Fixes the order of `they/them/their` pronouns to be equivalent to `he/him/his` and `she/her/her`.


iHiD:
Good catch! Thanks :)

PR #2176, opened 2021-11-17
Split into three deployed services

ErikSchierboom:
- Fix typos
- Split webservice into three services

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -155,7 +155,7 @@ jobs:
       # but then chained from action to action for the others.
       - name: Download task definition
         run: |
-          aws ecs describe-task-definition --task-definition ${{ env.webservers_task_definition_name }} --query taskDefinition > task-definition.json
+          aws ecs describe-task-definition --task-definition ${{ env.webservers_task_definition_name }} --query taskdefinition > task-definition.json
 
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -144,76 +153,169 @@ jobs:
...
+      - name: Update website task definition for puma container
+        id: update_definition_for_website_puma
+        uses: aws-actions/amazon-ecs-render-task-definition@v1
+        with:
+          task-definition: website-task-definition.json
+          container-name: puma
+          image: "${{ steps.build_rails.outputs.image }}"
+
+      - name: Update website task definition for nginx container
+        id: update_definition_for_website_nginx
```
iHiD, line 150: I believe this needs to be `${{ steps.update_definition_for_website_puma.outputs.task-definition }}`

The `website-task-definition.json` is immutable, so each step has to daisy-chain onto the previous one.

.github/workflows/deploy.yml
```diff
@@ -23,9 +30,7 @@ env:
   s3_bucket: "exercism-v3-assets"
 
 on:
-  push:
-    branches:
-      - main
+  - push
```
iHiD, line 33: We need to revert this before merging.

PR #2178, opened 2021-11-22
Don't sync untaught prerequisites

### LINKED ISSUE #1827: Some exercises are unreachable due to requiring unteached concepts
In the gitter there was someone wondering about exercises beeing locked, despite the fact that they fullfil all listed requirements when manipulating the URL to get on the exercises page.

They explicitely mentioned the `leap` exercise, which according to the `config.json` requires 3 concepts:

https://github.com/exercism/csharp/blob/main/config.json#L441

It seems as if `if-statements` and `numbers` are taught by the `cars-assemble` concept exercise, but I can not find any concept exercise teaching `math-operators`.

The word `math-operator` appears roughly 20 times within the configuration. So it is likely that there are more exercises affected.

It is also likely that there are more untaught concepts, that block students from doing some of the exercises not directly teaching concepts.

ErikSchierboom:
Closes https://github.com/exercism/csharp/issues/1827


app/commands/git/sync_concept_exercise.rb
```diff
@@ -88,6 +88,11 @@ def exercise_config
       head_git_track.concept_exercises[exercise_index]
     end
 
+    memoize
+    def exercise_config_taught_prerequisites
```
iHiD:
```suggestion
    def exercise_config_prerequisites
```

I don't think we need to have these as "taught" they're effectively just the "valid" data.

PR #2183, opened 2021-11-25
Search docs

ErikSchierboom:
- Enqueue sync job when document changes
- Parameterize opensearch wait
- Sync document to index
- Parameterize opensearch
- Searching docs
- Searching docs
- Add API method

### DIFF
--- a/app/assemblers/assemble_docs.rb
+++ b/app/assemblers/assemble_docs.rb
@@ -1,0 +1,19 @@
+class AssembleDocs
+  include Mandate
+
+  initialize_with :params
+
+  def call
+    SerializePaginatedCollection.(docs, serializer: SerializeDocs)
+  end
+
+  memoize
+  def docs
+    Document::SearchDocs.(
+      criteria: params[:criteria],
+      track_slug: params[:track_slug],
+      page: params[:page],
+      per: params[:per_page]
+    )
+  end
+end
--- a/app/commands/document/create_search_index_document.rb
+++ b/app/commands/document/create_search_index_document.rb
@@ -1,0 +1,20 @@
+class Document::CreateSearchIndexDocument
+  include Mandate
+
+  initialize_with :doc
+
+  def call
+    {
+      id: doc.id,
+      title: doc.title,
+      blurb: doc.blurb,
+      content_html: doc.content_html,
+      updated_at: doc.updated_at,
+      track: doc.track.nil? ? nil : {
+        id: doc.track.id,
+        slug: doc.track.slug,
+        title: doc.track.title
+      }
+    }
+  end
+end
--- a/app/commands/document/create_search_index_document.rb
+++ b/app/commands/document/create_search_index_document.rb
@@ -8,7 +8,7 @@ class Document::CreateSearchIndexDocument
       id: doc.id,
       title: doc.title,
       blurb: doc.blurb,
-      content_html: doc.content_html,
+      markdown: doc.markdown,
       updated_at: doc.updated_at,
       track: doc.track.nil? ? nil : {
         id: doc.track.id,
--- a/app/commands/document/create_search_index_document.rb
+++ b/app/commands/document/create_search_index_document.rb
@@ -9,12 +9,7 @@ class Document::CreateSearchIndexDocument
       title: doc.title,
       blurb: doc.blurb,
       markdown: doc.markdown,
-      updated_at: doc.updated_at,
-      track: doc.track.nil? ? nil : {
-        id: doc.track.id,
-        slug: doc.track.slug,
-        title: doc.track.title
-      }
+      track: doc.track.nil? ? nil : { slug: doc.track.slug }
     }
   end
 end
--- a/app/commands/document/search_docs.rb
+++ b/app/commands/document/search_docs.rb
@@ -40,7 +40,7 @@ class Document
     def search_body
       {
         query: search_query,
-        sort: search_sort,
+        sort: :_score,
 
         # Only return the solution IDs, not the entire document, to improve performance
         _source: [:id],
@@ -55,26 +55,16 @@ class Document
       {
         bool: {
           must: [
-            track_slug.blank? ? nil : {
-              terms: {
-                'track.slug': [track_slug].flatten
-              }
-            },
+            track_slug.blank? ? nil : { terms: { 'track.slug': [track_slug].flatten } },
             criteria.blank? ? nil : {
-              multi_match: {
-                query: criteria,
+              query_string: {
+                query: criteria.split(' ').map { |c| "*#{c}*" }.join(' AND '),
                 fields: ['title^5', 'blurb^3', 'markdown']
               }
             }
           ].compact
         }
       }
-    end
-
-    def search_sort
-      sort = [{ updated_at: { order: :desc, unmapped_type: "date" } }]
-      sort.prepend(:_score) if criteria.present?
-      sort
     end
 
     class Fallback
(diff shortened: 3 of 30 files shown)


app/assemblers/assemble_docs.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def call
+    SerializePaginatedCollection.(docs, serializer: SerializeDocs)
+  end
+
+  memoize
+  def docs
+    Document::SearchDocs.(
+      criteria: params[:criteria],
+      track_slug: params[:track_slug],
```
ErikSchierboom, line 14: For now, I've envisioned that people could either search on criteria (plaintext) or filter on a track. Whether we need the latter, 🤷

app/commands/document/create_search_index_document.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  initialize_with :doc
+
+  def call
+    {
+      id: doc.id,
+      title: doc.title,
+      blurb: doc.blurb,
+      markdown: doc.markdown,
+      updated_at: doc.updated_at,
```
ErikSchierboom: I originally used this for sorting, but seeing how people are always search for text I think, we could remove this. Unless you want to order by the updated at date if the score of documents is equal.

app/commands/document/search_docs.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def search_query
+      {
+        bool: {
+          must: [
+            track_slug.blank? ? nil : { terms: { 'track.slug': [track_slug].flatten } },
+            criteria.blank? ? nil : {
+              query_string: {
+                query: criteria.split(' ').map { |c| "*#{c}*" }.join(' AND '),
+                fields: ['title^5', 'blurb^3', 'markdown']
```
ErikSchierboom, line 62: The `^N` bits are the _boost_ modifiers. So in this case, a title match is 5 times more important than a markdown match.

app/commands/document/search_docs.rb
```diff
@@ -0,0 +1,106 @@
...
+      end
+
+      private
+      attr_reader :docs
+
+      def filter_criteria!
+        return if criteria.blank?
+
+        criteria.strip.split(" ").each do |crit|
+          @docs = @docs.where("title LIKE ? OR blurb LIKE ?", "%#{crit}%", "%#{crit}%")
```
ErikSchierboom, line 91: We can't search in the Markdown field, as that is not in the DB

app/commands/document/sync_all_to_search_index.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+
+      Exercism.opensearch_client.bulk(body: body)
+    end
+  end
+
+  private
+  attr_reader :since
+
+  BATCH_SIZE = 1000
```
ErikSchierboom, line 31: We're unlikely to hit this anytime soon, but I figured I'd keep it in for safeguarding.

app/commands/solution/search_user_solutions.rb
```diff
@@ -65,9 +65,12 @@ def search_query
...
             mentoring_status.blank? ? nil : { terms: { mentoring_status: [mentoring_status].flatten } },
-            criteria.blank? ? nil : { query_string: { query: criteria.split(' ').map do |c|
-                                                               "*#{c}*"
-                                                             end.join(' AND '), fields: ['exercise.title', 'track.title'] } }
+            criteria.blank? ? nil : {
+              query_string: {
+                query: criteria.split(' ').map { |c| "*#{c}*" }.join(' AND '),
+                fields: ['exercise.title', 'track.title']
+              }
+            }
```
ErikSchierboom, line 73: Just to make this a bit more readable.

app/controllers/api/docs_controller.rb
```diff
@@ -0,0 +1,10 @@
+module API
+  class DocsController < BaseController
+    skip_before_action :authenticate_user!
+    before_action :authenticate_user
```
ErikSchierboom, line 4: Do we need this?

app/serializers/serialize_doc.rb
```diff
@@ -0,0 +1,22 @@
+class SerializeDoc
+  include Mandate
+
+  initialize_with :doc
+
+  def call
```
ErikSchierboom, line 6: I was unsure which data we want to display on the search page, so I've just gathered some data. Let me know if you'd like more (or less) data.

test/commands/document/search_docs_test.rb
```diff
@@ -0,0 +1,189 @@
...
+    matching_markdown.stubs(:markdown).returns('Install')
+    Document::SyncToSearchIndex.(matching_title)
+    Document::SyncToSearchIndex.(matching_blurb)
+    Document::SyncToSearchIndex.(matching_markdown)
+
+    wait_for_opensearch_to_be_synced(Document::OPENSEARCH_INDEX)
+
+    # We've setup the document to have the "Install" text in different properties (title/blurb/markdown)
+    # to verify that the right boosting is applied
+    assert_equal [matching_title, matching_blurb, matching_markdown], Document::SearchDocs.(criteria: "install")
```
ErikSchierboom, line 73: This is an important assertion as it verifies that the field boosting works.

app/commands/document/create_search_index_document.rb
iHiD: I think sorting should be based on the ES match score, not on anything else.

ErikSchierboom: Perfect. I've removed the field.

app/commands/document/search_docs.rb
```diff
@@ -0,0 +1,106 @@
...
+      }
+    end
+
+    def search_query
+      {
+        bool: {
+          must: [
+            track_slug.blank? ? nil : { terms: { 'track.slug': [track_slug].flatten } },
+            criteria.blank? ? nil : {
+              query_string: {
```
ErikSchierboom, line 60: @iHiD The `query_string` option allows us to do wildcard searches (`*XYZ*`). While flexible, this is less performant than the default query search, which searches words. I don't think this is an issue for the documents, but it is something to consider to the solution searching.

test/commands/document/search_docs_test.rb
```diff
@@ -0,0 +1,185 @@
...
+    # We need to manually re-sync as markdown isn't a DB column and the indexing
+    # is started automatically when a Document record is saved (before the stub is setup)
+    non_track_doc.stubs(:markdown).returns('# Response')
+    ruby_doc_1.stubs(:markdown).returns('Instructions to install')
+    ruby_doc_2.stubs(:markdown).returns('## Guides')
+    elixir_doc.stubs(:markdown).returns('Great links')
+    Document::SyncToSearchIndex.(non_track_doc)
+    Document::SyncToSearchIndex.(ruby_doc_1)
+    Document::SyncToSearchIndex.(ruby_doc_2)
+    Document::SyncToSearchIndex.(elixir_doc)
```
ErikSchierboom, line 39: Hold off on merging, I can make this slightly more elegant.

app/commands/document/search_docs.rb
```diff
@@ -0,0 +1,106 @@
...
+        from: (page - 1) * per,
+        size: per
+      }
+    end
+
+    def search_query
+      {
+        bool: {
+          must: [
+            track_slug.blank? ? nil : { terms: { 'track.slug': [track_slug].flatten } },
```
iHiD, line 58: Why `flatten`?

app/controllers/api/docs_controller.rb
iHiD: Yes, as docs are viewable by non-logged-in users.

app/serializers/serialize_doc.rb
iHiD: Yep - this looks good I think.

app/commands/document/search_docs.rb
ErikSchierboom: Because track slug can be both an array and a string.

test/commands/document/search_docs_test.rb
ErikSchierboom: Nevermind, my plan did not work out :(

Issue #2141, opened 2021-11-04
List of issues closed by searchable solutions


iHiD:
Thought it useful to keep track of these:

- https://github.com/exercism/exercism/issues/6085
- https://github.com/exercism/exercism/issues/5963
- https://github.com/exercism/exercism/issues/6071
- https://github.com/exercism/exercism/issues/5616
- https://github.com/exercism/exercism/issues/5354

PR #2189, opened 2021-11-26
Filtering community solutions

### LINKED ISSUE #5616: Sorting Community Solutions
It would be good if the community solutions could be sorted by 
1. Date
2.  Stars
3.  Comments.

The v2 website already had this feature.

### LINKED ISSUE #5963: Try to make stars of community solutions more objective
IMO currently there are a lot of exercises where the number of stars are biased towards old solutions and don't really show "the best" solutions. This is probably not only an issue with exercism, but probably with all systems that have some kind of rating system and later sort according to those ratings. This leads to a situation where people look at the first few solutions and give some of them even more stars but probably rarely look at the bottom (or even all the) solutions to find the best ones.

I'm not sure how one could solve this situation for good, but it probably has to involve randomness in one form or another. Maybe replace the second and third of the shown solutions with random onces from all solutions. Or even a new "rating" feature where the user is shown random pairs of solutions and is asked which of them they like best (and maybe give a star to that one).

### LINKED ISSUE #6071: Sorting my published solutions seems a bit off
When I try to sort my published solutions by newest or oldest first, it seems to "almost work". Most of the solutions seems ordered correctly according to the date they display, but a few of them are out of order. 

I'm attaching a screenshot of one of them being out of order (parallel letter frequency). I added to the screenshots the dates that I see when I actually look at the solution. In some cases that date is different, but not in a way that would match the broken sort order.

Screen Shot 2021-10-28 at 09 04 12

This issue was originally mentioned on Slack by @NatiAris

### LINKED ISSUE #6085: Filter to list exercises that have been updated
The https://exercism.org/tracks/python/exercises page shows small triangles on the exercises that have been updated. It would be nice if it was easy to filter them according to this. e.g. by another circle at the top with the number of updated exercises.

ErikSchierboom:
TODO:

- [x] Update when published iteration changes.
- [x] Work out whether we need to update on student git-head update or not
- [x] Maybe just update on solution#after_save as most of the time it will be a no-op and that's safest?

Closes https://github.com/exercism/exercism/issues/6071
Closes https://github.com/exercism/exercism/issues/6085
Closes https://github.com/exercism/exercism/issues/5616
Closes https://github.com/exercism/exercism/issues/5963

See https://github.com/exercism/website/issues/2141


app/assemblers/assemble_exercise_community_solutions_list.rb
```diff
@@ -22,7 +22,10 @@ def solutions
     Solution::SearchCommunitySolutions.(
       exercise,
       page: params[:page],
-      criteria: params[:criteria]
+      criteria: params[:criteria],
+      tests_status: params[:tests_status],
+      mentoring_status: params[:mentoring_status],
+      up_to_date: params[:up_to_date]
```
iHiD: You could probably do:

```ruby
**params.slice(:page, :criteria, :tests_status, :mentoring_status, up_to_date)
```

app/commands/solution/search_community_solutions.rb
```diff
@@ -10,11 +10,15 @@ def self.default_per
       DEFAULT_PER
     end
 
-    def initialize(exercise, page: nil, per: nil, criteria: nil)
+    def initialize(exercise, page: nil, per: nil, order: nil, criteria: nil, tests_status: nil, mentoring_status: nil, up_to_date: nil)
```
iHiD, line 13: I'd probably break this over multiple lines at this stage, with some logical grouping. 

Probably exercise,page,per,order, then the rest on the next line

app/commands/solution/search_community_solutions.rb
```diff
@@ -68,28 +72,84 @@ def search_query
           must: [
             { term: { 'exercise.id': exercise.id } },
             { term: { status: 'published' } },
-            @criteria.blank? ? nil : { wildcard: { 'user.handle': "*#{criteria}*" } }
+            @up_to_date.nil? ? nil : { term: { 'out_of_date': !@up_to_date } },
```
iHiD: This is sort of a four-way boolean. We have:
1. show up to date only
2. show not up to date
3. show everything
4. Nothing passed in (so use default - probably 3)

Thoughts on how best to handle this?

ErikSchierboom: I'd argue it's a three way boolean, which option 3 and 4 being the same. I'm not a big fan of boolean arguments myself, but I could not come up with a good name that allowed me to pass in keywords. Thoughts?

app/assemblers/assemble_exercise_community_solutions_list.rb
```diff
@@ -21,8 +21,7 @@ def call
   def solutions
     Solution::SearchCommunitySolutions.(
       exercise,
-      page: params[:page],
-      criteria: params[:criteria]
+      **params.slice(:page, :criteria, :tests_status, :mentoring_status, :up_to_date)
```
iHiD: Maybe?
```suggestion
      **params.slice(self.class.keys)
```

ErikSchierboom: `**params.slice(*self.class.keys)` works

app/helpers/react_components/track/exercise_community_solutions_list.rb
```diff
@@ -11,13 +11,18 @@ def to_s
...
       end
 
       def data
-        AssembleExerciseCommunitySolutionsList.(exercise, params)
+        AssembleExerciseCommunitySolutionsList.(exercise, search_params)
+      end
+
+      memoize
+      def search_params
+        params.permit(*AssembleExerciseCommunitySolutionsList.keys)
```
iHiD: The only challenge with this is that `permit` doesn't exist for `Hash` so this will **only** work if params is Paramaters.

ErikSchierboom: Should I be using `slice` then?

app/assemblers/assemble_exercise_community_solutions_list.rb
```diff
@@ -22,7 +22,11 @@ def solutions
     Solution::SearchCommunitySolutions.(
       exercise,
       page: params[:page],
-      criteria: params[:criteria]
+      order: params[:order],
+      criteria: params[:criteria],
+      sync_status: params[:up_to_date] == "true" ? :up_to_date : nil,
```
iHiD: Is this how we check for booleans in other places? I'd have thought Rails would magically do this for us.

app/commands/solution/search_community_solutions.rb
```diff
@@ -10,11 +10,18 @@ def self.default_per
...
+                   mentoring_status: nil, sync_status: nil)
       @exercise = exercise
       @page = page.present? && page.to_i.positive? ? page.to_i : DEFAULT_PAGE # rubocop:disable Style/ConditionalAssignment
       @per = per.present? && per.to_i.positive? ? per.to_i : self.class.default_per # rubocop:disable Style/ConditionalAssignment
+      @order = order
       @criteria = criteria
+      @tests_status = tests_status
+      @head_tests_status = head_tests_status
+      @mentoring_status = mentoring_status
+      @sync_status = sync_status
```
iHiD, line 24: I suggest symbolizing these here, so you know they're symbols elsewhere and can remove the other `to_sym` throughout the command.
```suggestion
      @sync_status = sync_status&.to_sym
```

app/commands/solution/search_community_solutions.rb
```diff
@@ -73,17 +81,34 @@ def search_query
...
             { term: { status: 'published' } },
-            @criteria.blank? ? nil : { wildcard: { 'user.handle': "*#{criteria}*" } }
+            @sync_status.nil? ? nil : { term: { 'out_of_date': @sync_status.to_sym == :out_of_date } },
+            @mentoring_status.blank? ? nil : { terms: { 'mentoring_status': to_terms(@mentoring_status) } },
+            @tests_status.blank? ? nil : { terms: { 'published_iteration.tests_status': to_terms(@tests_status) } },
+            @head_tests_status.blank? ? nil : { terms: { 'published_iteration.head_tests_status': to_terms(@head_tests_status) } },
+            @criteria.blank? ? nil : {
+              query_string: {
+                query: criteria.split(' ').map { |c| "*#{c}*" }.join(' AND '),
+                fields: ['user.handle', 'published_iteration.code']
```
iHiD, line 91: I think the code should probably be a seperate field in the UI. So maybe remove that from here for now?

app/controllers/api/community_solutions_controller.rb
```diff
@@ -15,7 +15,10 @@ def use_exercise
     end
 
     def search_params
-      params.permit(AssembleExerciseCommunitySolutionsList.keys)
+      params.permit(AssembleExerciseCommunitySolutionsList.keys).tap do |p|
+        # Filter on passed head tests by default
+        p[:passed_head_tests] = "true" if p[:passed_head_tests].nil?
+      end
```
iHiD: What's the there a reason to have this here rather than having the default in the assembler?

app/helpers/react_components/track/exercise_community_solutions_list.rb
```diff
@@ -11,13 +11,21 @@ def to_s
...
-        AssembleExerciseCommunitySolutionsList.(exercise, params)
+        AssembleExerciseCommunitySolutionsList.(exercise, search_params)
+      end
+
+      memoize
+      def search_params
+        params.permit(*AssembleExerciseCommunitySolutionsList.keys).tap do |params|
+          # Set passed_head_tests to true by default
+          params[:passed_head_tests] = true if params[:passed_head_tests].blank?
+        end
```
iHiD, line 30: Feels like this needs to be dried up with the bit above I commented on.

db/migrate/20211201132830_add_published_iteration_head_tests_status_to_solutions.rb
```diff
@@ -0,0 +1,5 @@
+class AddPublishedIterationHeadTestsStatusToSolutions < ActiveRecord::Migration[6.1]
+  def change
+    add_column :solutions, :published_iteration_head_tests_status, :integer, null: false, default: 0
+  end
+end
```
iHiD: Want to rebase this onto `main` which means you can remove this?

app/helpers/react_components/track/exercise_community_solutions_list.rb
ErikSchierboom: This is to have the passed head tests option be enabled by default in React.

masahx:
hi, 
I didn't find topic but this seems close. Not sure if I miss something but community solutions don't have horizontal scroll? For example in this exercise: Java/RemoteControlCompetition - there are 4 solution files (I even have 5), you can't scroll horizontally
to see them all:

https://exercism.org/tracks/java/exercises/remote-control-competition/solutions
for example: https://exercism.org/tracks/java/exercises/remote-control-competition/solutions/PixelSam123

(there is hor. scroll when you look at your solution but not as community solution)
sorry maybe I miss something, topic or issue

usually you have no more than 2 files

PR #2193, opened 2021-11-30
Add timeout to elasticsearch

ErikSchierboom:


app/commands/document/search_docs.rb
```diff
@@ -17,7 +17,10 @@ def initialize(criteria: nil, track_slug: nil, page: nil, per: nil)
     end
 
     def call
-      results = Exercism.opensearch_client.search(index: Document::OPENSEARCH_INDEX, body: search_body)
+      results = Exercism.opensearch_client.search(index: Document::OPENSEARCH_INDEX,
+        body: search_body,
+        timeout: TIMEOUT,
+        allow_partial_search_results: false)
```
iHiD:
```suggestion
      results = Exercism.opensearch_client.search(
        index: Document::OPENSEARCH_INDEX,
        body: search_body,
        timeout: TIMEOUT,
        allow_partial_search_results: false
      )
```

iHiD: And same in other places.

PR #2194, opened 2021-11-30
Return iterations in reverse idx order for nudge

### LINKED ISSUE #6137: Select the last iteration by default when publishing only 1 iteration
When on “Publish your code and share your knowledge” screen, if I select “Yes, I’d like to share my solution with the community.” and “Single iteration”, “Iteration 1" is always selected:

Screenshot 2021-11-30 at 13 51 58

Even when there are multiple iterations:
Screenshot 2021-11-30 at 13 52 04

For me, personally, in 99% of the cases, I’d want to publish the last iteration. I think it would also be similar for other users.
So, if the select box would have the latest iteration selected by default, it would be 1 less click for me

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/6137

### DIFF
--- a/app/helpers/react_components/student/nudge.rb
+++ b/app/helpers/react_components/student/nudge.rb
@@ -13,7 +13,7 @@ module ReactComponents
           discussions: discussions,
           request: request,
           exercise_type: exercise_type,
-          iterations: solution.iterations.map { |iteration| SerializeIteration.(iteration) },
+          iterations: solution.iterations.order(idx: :desc).map { |iteration| SerializeIteration.(iteration) },
           links: links
         })
       end


app/helpers/react_components/student/nudge.rb
```diff
@@ -13,7 +13,7 @@ def to_s
           discussions: discussions,
           request: request,
           exercise_type: exercise_type,
-          iterations: solution.iterations.map { |iteration| SerializeIteration.(iteration) },
+          iterations: solution.iterations.order(idx: :desc).map { |iteration| SerializeIteration.(iteration) },
```
ErikSchierboom, line 16: This matches the behavior for the complete button on the right hand side. See https://github.com/exercism/website/blob/main/app/helpers/react_components/student/complete_exercise_button.rb#L11

iHiD: Nice :)

PR #2197, opened 2021-11-30
Add Nudge for mentoring

iHiD:
This will run daily and send an email to everyone who has submitted a non-hello-world practice solution that's not been mentored. People only ever get one notification.

Note: This needs an entry adding the schedule - but I'll run it manually the first time, so I've not done that for now.

### DIFF
--- a/app/jobs/nudge_users_to_request_mentoring_job.rb
+++ b/app/jobs/nudge_users_to_request_mentoring_job.rb
@@ -1,0 +1,43 @@
+class NudgeUsersToRequestMentoringJob < ApplicationJob
+  queue_as :notifications
+
+  def perform
+    # Don't change this without actively benchmarking it on bastion.
+    # Things that logically feel like they should make it faster often don't.
+    # At time of writing it takes 4.8s.
+    users = User.where(
+      id: Iteration.joins(solution: :exercise).
+        # Iterated since v3 launch
+        where('iterations.created_at >= ?', Date.new(2021, 9, 1)).
+        # Iterated over a day ago (give time to request manually)
+        where('iterations.created_at < ?', Time.current - 1.day).
+
+        # Don't include Concept Exercises as we don't nudge on those
+        where('solutions.type': 'PracticeSolution').
+
+        # Don't include hello-world
+        where.not('exercises.slug': 'hello-world').
+
+        # Not for users who have mentor requests
+        where.not('solutions.user_id': Mentor::Request.select(:student_id)).
+
+        # Not for users who have already had the notification
+        where.not('solutions.user_id': User::Notifications::NudgeToRequestMentoringNotification.select(:user_id)).
+
+        # Just the user ids
+        select('solutions.user_id')
+    )
+
+    users.find_each do |user|
+      track = PracticeSolution.status_iterated.where(user: user).
+        joins(:exercise).where.not('exercises.slug': 'hello-world').
+        last.track
+
+      User::Notification::Create.(
+        user,
+        :nudge_to_request_mentoring,
+        track: track
+      )
+    end
+  end
+end
--- a/app/views/mailers/notifications_mailer/nudge_to_request_mentoring.html.haml
+++ b/app/views/mailers/notifications_mailer/nudge_to_request_mentoring.html.haml
@@ -1,0 +1,10 @@
+%p Hi #{@user.handle},
+%p You've completed a few exercises (nice work!) but not yet requested mentoring on any of your solutions.
+%p
+  Receiving feedback from our mentors is an amazing way of discovering your knowledge gaps and levelling up your skills.
+  It's one of Exercism's most valuable features, and like all of Exercism, it's absolutely free!
+%p
+  Choose a solution that you feel has the most room for improvement and request a mentor to take a look at it.
+  They should get back to you within a day or so with some ideas!
+
+= email_button_to "Choose a solution →", @notification.url
--- a/app/views/mailers/notifications_mailer/nudge_to_request_mentoring.text.haml
+++ b/app/views/mailers/notifications_mailer/nudge_to_request_mentoring.text.haml
@@ -4,7 +4,7 @@ You've completed a few exercises (nice work!) but not yet requested mentoring on
 \
 Receiving feedback from our mentors is an amazing way of discovering your knowledge gaps and levelling up your skills. It's one of Exercism's most valuable features, and like all of Exercism, it's absolutely free!
 \
-Choose a solution that you feel has the most room for improvement and request a mentor to take a look at it.  They should get back to you within a day or so with some ideas!
+Choose a solution that you feel has the most room for improvement and request a mentor to take a look at it. They should get back to you within a day or so with some ideas!
 \
-Choose a solution: #{@notification.url}
+Choose a solution at: #{@notification.url}
 
--- a/app/views/mailers/notifications_mailer/nudge_to_request_mentoring.text.haml
+++ b/app/views/mailers/notifications_mailer/nudge_to_request_mentoring.text.haml
@@ -2,7 +2,7 @@ Hi #{@user.handle},
 \
 You've completed a few exercises (nice work!) but not yet requested mentoring on any of your solutions.
 \
-Receiving feedback from our mentors is an amazing way of discovering your knowledge gaps and levelling up your skills.  It's one of Exercism's most valuable features, and like all of Exercism, it's absolutely free!
+Receiving feedback from our mentors is an amazing way of discovering your knowledge gaps and levelling up your skills. It's one of Exercism's most valuable features, and like all of Exercism, it's absolutely free!
 \
 Choose a solution that you feel has the most room for improvement and request a mentor to take a look at it.  They should get back to you within a day or so with some ideas!
 \
--- a/app/views/mailers/notifications_mailer/nudge_to_request_mentoring.text.haml
+++ b/app/views/mailers/notifications_mailer/nudge_to_request_mentoring.text.haml
@@ -1,0 +1,10 @@
+Hi #{@user.handle},
+\
+You've completed a few exercises (nice work!) but not yet requested mentoring on any of your solutions.
+\
+Receiving feedback from our mentors is an amazing way of discovering your knowledge gaps and levelling up your skills.  It's one of Exercism's most valuable features, and like all of Exercism, it's absolutely free!
+\
+Choose a solution that you feel has the most room for improvement and request a mentor to take a look at it.  They should get back to you within a day or so with some ideas!
+\
+Choose a solution: #{@notification.url}
+
--- a/config/locales/notifications/en.yml
+++ b/config/locales/notifications/en.yml
@@ -25,3 +25,8 @@ en:
     joined_exercism:
       1: >
         Welcome to Exercism! Explore your dashboard to learn more.
+
+    nudge_to_request_mentoring:
+      1: >
+        <strong>It's time to get mentored!</strong>
+        Learning with our mentors is an amazing way to level up your knowledge, and it's 100% free. Choose one of your solutions to begin.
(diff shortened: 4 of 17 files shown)


db/migrate/20211130133210_add_index_for_nudge_job.rb
```diff
@@ -0,0 +1,13 @@
+class AddIndexForNudgeJob < ActiveRecord::Migration[6.1]
+  def change
+    add_index :solutions, [:last_iterated_at, :user_id], name: "fast_1"
+    add_index :solutions, [:user_id, :last_iterated_at], name: "fast_2"
+    add_index :solutions, [:type, :user_id, :exercise_id, :last_iterated_at], name: "fast_3"
+    add_index :solutions, [:exercise_id, :type, :user_id, :last_iterated_at], name: "fast_4"
+    add_index :user_notifications, [:type], name: "fast_5"
+    add_index :user_notifications, [:user_id, :type], name: "fast_7"
```
iHiD: These need removing before I merge.

db/migrate/20211130133210_add_index_for_nudge_job.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    # Definites
+    add_index :user_notifications, [:type, :user_id], name: "fast_6"
```
iHiD: This needs the algorthm/lock thing adding and the name removing.

app/jobs/nudge_users_to_request_mentoring_job.rb
```diff
@@ -0,0 +1,43 @@
...
+  def perform
+    # Don't change this without actively benchmarking it on bastion.
+    # Things that logically feel like they should make it faster often don't.
+    # At time of writing it takes 4.8s.
+    users = User.where(
+      id: Iteration.joins(solution: :exercise).
+        # Iterated since v3 launch
+        where('iterations.created_at >= ?', Date.new(2021, 9, 1)).
+        # Iterated over a day ago (give time to request manually)
+        where('iterations.created_at < ?', Time.current - 1.day).
```
ErikSchierboom, line 13: Good idea

app/jobs/nudge_users_to_request_mentoring_job.rb
```diff
@@ -0,0 +1,43 @@
+class NudgeUsersToRequestMentoringJob < ApplicationJob
+  queue_as :notifications
+
+  def perform
+    # Don't change this without actively benchmarking it on bastion.
+    # Things that logically feel like they should make it faster often don't.
+    # At time of writing it takes 4.8s.
+    users = User.where(
```
ErikSchierboom, line 8: With all the comment, this query is strangely pleasing to read :)

app/views/mailers/notifications_mailer/nudge_to_request_mentoring.html.haml
```diff
@@ -0,0 +1,10 @@
+%p Hi #{@user.handle},
+%p You've completed a few exercises (nice work!) but not yet requested mentoring on any of your solutions.
```
ErikSchierboom, line 2: At first I thought: "a few exercise", but then I realized that students will all have done `hello-world` :) (although you can't request mentoring on that)

app/views/mailers/notifications_mailer/nudge_to_request_mentoring.html.haml
```diff
@@ -0,0 +1,10 @@
...
+%p
+  Receiving feedback from our mentors is an amazing way of discovering your knowledge gaps and levelling up your skills.
```
ErikSchierboom, line 4: Is it `levelling` in US English or `leveling`? I've done a quick search and found:

https://en.wikipedia.org/wiki/Levelling#:~:text=Levelling%20or%20leveling%20(American%20English,points%20relative%20to%20a%20datum.&text=It%20is%20also%20known%20as%20spirit%20levelling%20and%20differential%20levelling.
https://ell.stackexchange.com/questions/146181/in-us-english-which-is-the-appropriate-spelling-leveled-a-look-or-levelled-a-l

It _looks_ like it should be `leveling`, but 🤷

app/views/mailers/notifications_mailer/nudge_to_request_mentoring.text.haml
```diff
@@ -0,0 +1,10 @@
+Hi #{@user.handle},
+\
+You've completed a few exercises (nice work!) but not yet requested mentoring on any of your solutions.
+\
+Receiving feedback from our mentors is an amazing way of discovering your knowledge gaps and levelling up your skills.  It's one of Exercism's most valuable features, and like all of Exercism, it's absolutely free!
```
ErikSchierboom:
```suggestion
Receiving feedback from our mentors is an amazing way of discovering your knowledge gaps and levelling up your skills. It's one of Exercism's most valuable features, and like all of Exercism, it's absolutely free!
```

app/views/mailers/notifications_mailer/nudge_to_request_mentoring.text.haml
```diff
@@ -0,0 +1,10 @@
...
+\
+Choose a solution that you feel has the most room for improvement and request a mentor to take a look at it.  They should get back to you within a day or so with some ideas!
```
ErikSchierboom:
```suggestion
Choose a solution that you feel has the most room for improvement and request a mentor to take a look at it. They should get back to you within a day or so with some ideas!
```

app/views/mailers/notifications_mailer/nudge_to_request_mentoring.text.haml
```diff
@@ -0,0 +1,10 @@
...
+\
+Choose a solution: #{@notification.url}
```
ErikSchierboom: Maybe:

```suggestion
Choose a solution at: #{@notification.url}
```

db/migrate/20211130133210_add_index_for_nudge_job.rb
```diff
@@ -0,0 +1,13 @@
+class AddIndexForNudgeJob < ActiveRecord::Migration[6.1]
+  def change
+    add_index :solutions, [:last_iterated_at, :user_id], name: "fast_1"
+    add_index :solutions, [:user_id, :last_iterated_at], name: "fast_2"
+    add_index :solutions, [:type, :user_id, :exercise_id, :last_iterated_at], name: "fast_3"
+    add_index :solutions, [:exercise_id, :type, :user_id, :last_iterated_at], name: "fast_4"
```
ErikSchierboom: Will this temporarily lock the table? I'm not sure how mysql does this.

test/jobs/nudge_users_to_request_mentoring_job_test.rb
```diff
@@ -0,0 +1,38 @@
...
+    # Set things up for the correct user
+    ruby = create :track, slug: :ruby
+    javascript = create :track, slug: :javascript
+    user = create :user
+    create :iteration, created_at: 2.days.ago, solution: create(:practice_solution, track: javascript, user: user, status: :iterated)
+    create :iteration, created_at: 2.days.ago, solution: create(:practice_solution, track: ruby, user: user, status: :iterated)
+
+    # Too recent
+    create :iteration, created_at: 5.minutes.ago, solution: create(:practice_solution)
+
```
ErikSchierboom, line 14: Should there be a call to `NudgeUsersToRequestMentoringJob.perform_now` after this to verify that the notification is not sent?

```ruby
User::Notification::Create.expects(:call).never
NudgeUsersToRequestMentoringJob.perform_now
```

If correct, you probably want to do this for the other exception cases below too.

config/locales/notifications/en.yml
```diff
@@ -25,3 +25,8 @@ en:
     joined_exercism:
       1: >
         Welcome to Exercism! Explore your dashboard to learn more.
+
+    nudge_to_request_mentoring:
+      1: >
+        <strong>It's time to get mentored!</strong>
+        Learning with our mentors is an amazing way to level up your knowledge, and it's 100% free. Choose one of your solutions to begin.
```
ErikSchierboom, line 32: Maybe be slightly more specific on what we expect?

```suggestion
        Learning with our mentors is an amazing way to level up your knowledge, and it's 100% free. Choose one of your solutions to request mentoring on.
```

app/jobs/nudge_users_to_request_mentoring_job.rb
```diff
@@ -0,0 +1,43 @@
...
+        where.not('solutions.user_id': User::Notifications::NudgeToRequestMentoringNotification.select(:user_id)).
+
+        # Just the user ids
+        select('solutions.user_id')
+    )
+
+    users.find_each do |user|
+      track = PracticeSolution.status_iterated.where(user: user).
+        joins(:exercise).where.not('exercises.slug': 'hello-world').
+        last.track
```
ErikSchierboom, line 34: Does `.last` by default return the latest track? I'm asking because there is no explicit sorting here.

ErikSchierboom (approved):

app/jobs/nudge_users_to_request_mentoring_job.rb
iHiD: It will return the last practice solution, which will be default ordered by the primary key, so it'll be the last submitted.

app/views/mailers/notifications_mailer/nudge_to_request_mentoring.html.haml
iHiD: Yes - I wondered if we should hold off and only do it when someone has done three exercises, but it made the query too heavy :)

db/migrate/20211130133210_add_index_for_nudge_job.rb
iHiD: Not if you use the algorithm+lock commands as per https://github.com/exercism/website/pull/2197/files#diff-e85712804207d693a50b13df15695b213427e6f51c2eab47a534762de03a3230R3 - Mysql now supports non-blocking table modifications if those are specified.

test/jobs/nudge_users_to_request_mentoring_job_test.rb
iHiD: Because L35 specifies uses `expects` on `User::Notification::Create`, if only allows calls with the exact params. If that gets called with any of the other users, it explodes. So this is actually tested, but I'll add a comment below to make that clear.

test/jobs/nudge_users_to_request_mentoring_job_test.rb
```diff
@@ -0,0 +1,38 @@
...
+    s = create(:practice_solution, user: u)
+    create :iteration, created_at: 2.days.ago, solution: s
+    create :mentor_request, solution: s
+
+    # Already had notification
+    u = create :user
+    create :iteration, created_at: 2.days.ago, solution: create(:practice_solution, user: u)
+    create :nudge_to_request_mentoring_notification, user: u, track: ruby
+
+    User::Notification::Create.expects(:call).with(user, :nudge_to_request_mentoring, track: ruby)
```
iHiD, line 36:
```suggestion
    # Only expect call with the initial correct user. This will raise if called with any other user.
    User::Notification::Create.expects(:call).with(user, :nudge_to_request_mentoring, track: ruby)
```

config/locales/notifications/en.yml
iHiD: I'm going to leave this as it is, purely so it doesn't roll over onto another line!

app/views/mailers/notifications_mailer/nudge_to_request_mentoring.html.haml
iHiD: I'm going to leave this one I think. One `l` just looks too wrong to me!!

db/migrate/20211130133210_add_index_for_nudge_job.rb
ErikSchierboom: Cool

test/jobs/nudge_users_to_request_mentoring_job_test.rb
ErikSchierboom: Ah okay. Thanks

PR #2198, opened 2021-11-30
Add ability to have different test runs for different git shas

iHiD:
This will allow us to store test runs against both a user's stored sha and the main up to date exercise sha.


app/commands/submission/test_run/init.rb
```diff
@@ -15,15 +19,22 @@ def call
...
+          submission.tests_queued! if in_band?
         end
       end
 
+      private
+      attr_reader :submission, :git_sha
+
+      def in_band?
+        @in_band
+      end
```
iHiD: Looking forward to making this a new one-line method in Ruby 3!

db/migrate/20211130210947_add_git_sha_to_submission_test_runs.rb
```diff
@@ -0,0 +1,13 @@
+class AddGitShaToSubmissionTestRuns < ActiveRecord::Migration[6.1]
+  def up
+    execute "ALTER TABLE `submission_test_runs` ADD `git_sha` varchar(50) NULL, ALGORITHM=INPLACE, LOCK=NONE"
+
+    # Run the following to update data:
+    connection.execute("SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED")
+    Submission::TestRun.joins(submission: :solution).update_all("submission_test_runs.git_sha = solutions.git_sha")
```
iHiD: These need commenting out, then running on Bastion.

app/models/solution.rb
```diff
@@ -116,6 +116,12 @@ def published_iterations
     iterations.not_deleted
   end
 
+  def up_to_date_tests_status
+    return if published_iterations.blank?
+
+    published_iterations.last.submission.test_runs.find_by(git_sha: exercise.git_sha)&.tests_status
+  end
```
iHiD: This will be the method that we use in the search index. We need to work out how we know to trigger the update to the search index though.

app/commands/submission/test_run/init.rb
```diff
@@ -3,7 +3,11 @@ class TestRun
     class Init
       include Mandate
 
-      initialize_with :submission
+      def initialize(submission, in_band: true, git_sha: nil)
```
iHiD: I'm not 100% sure about overloading this command or making another new one.

app/commands/exercise/queue_solution_head_test_runs.rb
```diff
@@ -0,0 +1,27 @@
+class Exercise::QueueSolutionHeadTestRuns
```
iHiD, line 1: This class batch queues all the new tests runs when an exercise is updated.

app/commands/git/generate_hash_for_important_exercise_files.rb
```diff
@@ -2,7 +2,10 @@ module Git
   class GenerateHashForImportantExerciseFiles
```
iHiD, line 2: This has changed to allow us to calculate the hash for any point in time.

app/commands/submission/test_run/init.rb
```diff
@@ -3,7 +3,11 @@ class TestRun
     class Init
```
iHiD, line 3: This can now take a custom git_sha to run against, and also has a status for `head_run` which we use both here, and when the job is completed to know whether we're dealing with a normal run or a run against git head. 

I'm open to a better name than `head_run` if you have an opinion!

app/commands/submission/test_run/process.rb
```diff
@@ -3,17 +3,16 @@ class TestRun
     class Process
```
iHiD, line 3: This now pivots on where it was a head run or not as to how some of the behaviours work

db/migrate/20211130210947_add_git_sha_to_submission_test_runs.rb
```diff
@@ -0,0 +1,52 @@
+class AddGitShaToSubmissionTestRuns < ActiveRecord::Migration[6.1]
+  def up
+    execute "ALTER TABLE `submissions` ADD `git_important_files_hash` varchar(50) NULL, ALGORITHM=INPLACE, LOCK=NONE"
+
+    execute "ALTER TABLE `submission_test_runs` ADD `git_important_files_hash` varchar(50) NULL, ALGORITHM=INPLACE, LOCK=NONE"
+    execute "ALTER TABLE `submission_test_runs` ADD `git_sha` varchar(50) NULL, ALGORITHM=INPLACE, LOCK=NONE"
+
+    return
```
iHiD: This is deliberately here to stop the bit below running.

app/commands/submission/test_run/init.rb
iHiD: Having thought about this more, I think I'll switch to having another type column here, rather than `head_run` as a flag. Maybe `test_run_type` with `[submission, published_solution_git_head]` or something. Thoughts welcome.

ErikSchierboom: I prefer the type column approach. That allows for future enhancements and is more explicit I feel.

app/commands/submission/test_run/process.rb
```diff
@@ -72,30 +57,73 @@ def handle_error!
...
           submission.send("tests_#{status}!")
         end
       end
 
+      def update_solution_status!(status)
+        submission.solution.with_lock do
+          return unless submission.exercise.git_sha == git_sha
+
+          submission.solution.send("published_iteration_head_tests_status_#{status}!")
+        end
```
ErikSchierboom: Is the lock here necessary? It looks like the check inside (`submission.exercise.git_sha == git_sha`) is unrelated to the locked object (`submission.solution`).

app/jobs/queue_solution_head_test_runs_job.rb
```diff
@@ -0,0 +1,7 @@
+class QueueSolutionHeadTestRunsJob < ApplicationJob
+  queue_as :default
```
ErikSchierboom, line 2: Were you planning to create a separate queue for this?

app/models/exercise.rb
```diff
@@ -63,7 +63,10 @@ def self.for(track_slug, exercise_slug)
   end
 
   after_update do
-    MarkSolutionsAsOutOfDateInIndexJob.perform_later(self) if saved_changes.include?(:git_important_files_hash)
+    if saved_changes.include?(:git_important_files_hash)
+      MarkSolutionsAsOutOfDateInIndexJob.perform_later(self)
+      QueueSolutionHeadTestRunsJob.perform_later(self)
+    end
```
ErikSchierboom, line 69: Nice

app/jobs/queue_solution_head_test_runs_job.rb
iHiD: Hmmm, not considered it. I copied the existing one (`MarkSolutionsAsOutOfDateInIndexJob`)

app/commands/submission/test_run/process.rb
iHiD: There's a race here where the exercise's git sha could change, but I guess if that happens another job will get scheduled, so it can probably be safely removed.

PR #2210, opened 2021-12-02
Parallellize non-system tests

ErikSchierboom:


.github/workflows/tests.yml
```diff
@@ -79,9 +79,32 @@ jobs:
...
+      matrix: ${{ steps.test-files.outputs.matrix }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - id: test-files
+        env:
+          FILES_PER_BATCH: 60
+        run: |
+          tests=$(find test -name *_test.rb -not -path system | xargs -n ${{ env.FILES_PER_BATCH }} | xargs -I {} echo '"{}"' | tr '\n' ',')
```
iHiD: From looking at the logs, system isn't getting excluded properly, which is why this is slow.

iHiD: (e.g. https://github.com/exercism/website/runs/4395693256?check_suite_focus=true)

ErikSchierboom: Ah, feck. It worked locally. I'll figure it out

ErikSchierboom: Should be fixed now

PR #2238, opened 2021-12-07
Allow sorting user solutions by most starred

ErikSchierboom:
Allow sorting user solutions by most starred


iHiD:
@ErikSchierboom This is red.

ErikSchierboom:
@iHiD This is now failing because the head tests status is set to queued in the tests. I'll have to figure out how to fix it (or you can :))

PR #2248, opened 2021-12-08
Filter journey solutions

ErikSchierboom:
- Move journey filters to submenu
- Remove unused code
- Allow selecting sync status
- Allow passing in tests_status and head_tests_status filters
- Allow filtering on tests_status, head_tests_status and sync_status
- Allow filtering on tests_status, head_tests_status and sync_status


app/commands/solution/search_user_solutions.rb
```diff
@@ -70,6 +76,9 @@ def search_query
             track_slug.blank? ? nil : { terms: { 'track.slug': [track_slug].flatten } },
             status.blank? ? nil : { terms: { status: [status].flatten } },
             mentoring_status.blank? ? nil : { terms: { mentoring_status: [mentoring_status].flatten } },
+            sync_status.nil? ? nil : { term: { 'out_of_date': sync_status == :out_of_date } },
+            tests_status.blank? ? nil : { terms: { 'published_iteration.tests_status': to_terms(tests_status) } },
+            head_tests_status.blank? ? nil : { terms: { 'published_iteration.head_tests_status': to_terms(head_tests_status) } },
```
ErikSchierboom, line 81: I currently have copied what we use for the searching of the community solutions. However, there we were only showing published solutions. Do we want to use the published iteration on the journey solutions page too? AFAICT, the tests status wasn't actually shown on the journey page, so this might not be bad.

app/javascript/components/journey/solutions-list/HeadTestsStatusSelect.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import { SingleSelect } from '../../common'
+
+export type HeadTestsStatus =
+  | undefined
+  | 'passed'
+  | 'failed'
+  | 'errored'
+  | 'exceptioned'
```
ErikSchierboom, line 9: Do we want to have individual states, or just a `passed`/`failed` option?

And what do we do with the `not_queued`, `queued` and `cancelled` states?

app/javascript/components/journey/solutions-list/MentoringStatusSelect.tsx
```diff
@@ -1,12 +1,17 @@
 import React from 'react'
 import { SingleSelect } from '../../common'
 
-type Status = undefined | 'none' | 'requested' | 'in_progress' | 'finished'
+export type MentoringStatus =
```
ErikSchierboom, line 4: Same question as above.

app/commands/solution/search_user_solutions.rb
iHiD: Good question. I think we should show the double tick on the card if it's got an iteration solution and passing HEAD and the general tests-status. With regards to this specific bit, I think what you have here is fine.

app/javascript/components/journey/solutions-list/HeadTestsStatusSelect.tsx
iHiD: For now, I'm happy with these individual states I think.

PR #2257, opened 2021-12-10
DRY up logic slightly

ErikSchierboom:


app/models/git/exercise.rb
```diff
@@ -162,13 +162,15 @@ def tooling_files
       end
     end
 
-    # This includes meta files
+    # This includes meta files, but excludes docs files
     memoize
     def tooling_filepaths
-      filepaths
+      filepaths.reject do |filepath|
+        filepath.starts_with?(".docs")
```
iHiD: Worth making this `.docs/`?

iHiD:
This looks good. Worth quickly double checking with `#maintainers` that none of them are using `.docs` for anything?

app/models/git/exercise.rb
ErikSchierboom: Absolutely.

app/commands/submission/representation/init.rb
```diff
@@ -28,7 +28,7 @@ def call
 
       def exercise_filepaths
         exercise_repo.tooling_filepaths.select do |filepath|
-          filepath.starts_with?(".meta")
+          filepath.starts_with?(".meta/")
```
ErikSchierboom: This is the only real exception, where we currently don't output the test/editor files for the representer.

app/commands/submission/test_run/init.rb
```diff
@@ -43,15 +43,8 @@ def call
...
-        exercise_repo.tooling_filepaths.select do |filepath|
-          # Skip non-functional files
-          next false if filepath.starts_with?(".docs")
-          next false if filepath == "README.md"
-
-          # Skip submitted files
-          next false if submission.valid_filepaths.include?(filepath)
-
-          true
+        exercise_repo.tooling_filepaths.reject do |filepath|
```
ErikSchierboom: I've tried moving this functionality to submission, as it is the same here and for the analyzer, but didn't do that as the `exercise_repo` used here is not necessarily the same as the submission exercise's git repo (due to the test run being allowed to pass in the `git_sha`).

test/commands/submission/analysis/init_test.rb
```diff
@@ -22,7 +22,36 @@ class Submission::Analysis::InitTest < ActiveSupport::TestCase
...
         exercise_git_dir: exercise_repo.dir,
-        exercise_filepaths: [".meta/config.json", ".meta/design.md", ".meta/exemplar.rb"]
+        exercise_filepaths: [".meta/config.json", ".meta/design.md", ".meta/exemplar.rb", "log_line_parser_test.rb"]
+      }
+    )
+    Submission::Analysis::Init.(submission)
+    assert_equal 'queued', submission.analysis_status
+  end
+
+  test "calls to publish_message for exercise with editor files" do
```
ErikSchierboom, line 33: This test case verifies that the original issue (editor files not being present)

iHiD:
@ErikSchierboom Any risks in merging this?

ErikSchierboom:
@iHiD This is technically a breaking change, as tooling no longer receives the `.docs` files. However, I asked in the #maintainers channel if people used files in the `.docs` folder in their tooling, and they told me they didn't. I also find it unlikely they do.

What this _does_ do is to give tooling access to editor files, which is something that does not happen at the moment.

iHiD:
Happy for this to go down, but think #2348 might be worth merging into this first :)

ErikSchierboom:
@iHiD Merged. Those changes look good.

PR #2259, opened 2021-12-10
Refactor everything to be less racey

iHiD:
The TL;DR: on this is that we should now be able to run `Solution::QueueHeadTestRun` for every solution and it will do one of:
- Update the status if there's a head test run.
- Do nothing if the solution isn't published or there's no test runner.
- Queue a new head test run.

This is called whenever a solution is updated (in the after_save) so whenever a new iteration is published etc, this will be called. 99% of the time it will be a no-op but sometimes it'll do stuff.

### DIFF
--- a/test/commands/submission/test_run/process_test.rb
+++ b/test/commands/submission/test_run/process_test.rb
@@ -142,27 +142,19 @@ class Submission::TestRun::ProcessTest < ActiveSupport::TestCase
   end
 
   test "changes solution not submission for solution run" do
-    submission = create :submission
+    # Set the exercise and test run to be one sha, and the solution to be a different one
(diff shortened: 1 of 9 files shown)


test/commands/submission/test_run/process_test.rb
```diff
@@ -131,8 +131,15 @@ class Submission::TestRun::ProcessTest < ActiveSupport::TestCase
   end
 
   test "does not broadcast for solution run" do
+    # see changes solution not submission for solution run for explaination of this setup
```
ErikSchierboom:
```suggestion
    # see changes solution not submission for solution run for explanation of this setup
```

ErikSchierboom (approved):

PR #2260, opened 2021-12-10
Set num_loc correctly when publishing a solution

ErikSchierboom:
This fixes the issue of an incorrect num_loc if a solution had multiple iterations and was then published using a specific iteration.

### DIFF
--- a/app/commands/solution/publish.rb
+++ b/app/commands/solution/publish.rb
@@ -13,7 +13,8 @@ class Solution
         ActiveRecord::Base.transaction do
           solution.update(
             published_at: Time.current,
-            published_iteration_id: published_iteration_id
+            published_iteration: published_iteration,
+            num_loc: num_loc
           )
         end
       end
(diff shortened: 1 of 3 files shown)


iHiD:
So one question with this is what happens when someone has set to publish **all** iterations, and then publish a new iteration. Will this get updated?

ErikSchierboom:
I think that's when https://github.com/exercism/website/blob/main/app/commands/solution/publish_iteration.rb is called (which already had the right code). I'll double check.

ErikSchierboom:
Ah no, the path is different. The num_loc gets updated in https://github.com/exercism/website/blob/main/app/jobs/calculate_lines_of_code_job.rb#L31

iHiD:
Should we DRY this up with https://github.com/exercism/website/blob/0ea896d8a3941068bbabf5ae71e1665eb0bf1a6b/app/commands/solution/publish_iteration.rb#L2 ?

ErikSchierboom:
I considered doing that. The only difference is that Publish also sets the published_at date. Thoughts on how to elegantly handle that?

iHiD:
@ErikSchierboom I've DRY'd it up. Can you check you're happy pls.

ErikSchierboom (commented):
Yes, nice!

PR #2280, opened 2021-12-14
Add support for running the head-tests on non-published solutions

iHiD:
Currently we only re-run the HEAD tests on published iterations. This adds support for adding on the latest iterations.

Following merging:
- [ ] Run all solutions through again (using `Solution::QueueHeadTestRun`)
- [ ] Work out how to store this in OpenSearch
- [ ] Reprocess everything in OpenSearch
- [ ] Update user solution searching

### DIFF
--- a/app/commands/solution/create_search_index_document.rb
+++ b/app/commands/solution/create_search_index_document.rb
@@ -36,6 +36,7 @@ class Solution::CreateSearchIndexDocument
       } : nil,
       latest_iteration: latest_iteration ? {
         tests_status: latest_iteration.tests_status,
+        head_tests_status: solution.latest_iteration_head_tests_status,
         code: latest_iteration.submission.files.map(&:content) || []
       } : nil
     }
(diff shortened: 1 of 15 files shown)


ErikSchierboom (approved):
Looks good!

>  Work out how to store this in OpenSearch

This PR adds a field to store this in opensearch, right? Or are you referring to the option to maybe also have a top-level field for ease of searching?

>  Reprocess everything in OpenSearch

Won't this happen automatically once the background processing is started?

iHiD:
> This PR adds a field to store this in opensearch, right? Or are you referring to the option to maybe also have a top-level field for ease of searching?

Yeah - basically wondering which of those to do. This PR took the approach I guessed was correct but I'm not sure if you agree.

> Won't this happen automatically once the background processing is started?

Yeah - but we'll need to redo it for all existing records.

ErikSchierboom:
> Yeah - basically wondering which of those to do. This PR took the approach I guessed was correct but I'm not sure if you agree.

It depends on how we'll use the new field. If we don't want to ever access the field individually, then it might make sense as a top-level property, but the current implementation makes it very symmetric with the published iteration to I think this is fine.

ErikSchierboom:
Where do we want to use the new field?

PR #2308, opened 2022-01-04
Don't link to concept pages if no concepts

### LINKED ISSUE #478: Unable to view concepts page
Hi team,

Apologies if this is not the right place to reach out.

I started the Julia track yesterday and after completing the first exercise I was told to visit the [concepts page](https://exercism.org/tracks/julia/concepts) but when I open that page I do not see any content (refer below image).
Screenshot from 2021-12-21 20-41-58

I have tried multiple browsers, different computers, and different operating systems to no avail.

Am I missing something?

Kind regards,
Mitchell O'Sullivan

ErikSchierboom:
Closes https://github.com/exercism/julia/issues/478

@iHiD Do we want to redirect from the concept page for tracks without a course? As in: if someone just goes to that page directly.

### DIFF
--- a/app/views/tracks/exercises/show/_action_box_join_exercism.html.haml
+++ b/app/views/tracks/exercises/show/_action_box_join_exercism.html.haml
@@ -10,7 +10,7 @@
     Sign up to Exercism to learn and master
     = link_to track.title, track_path(track)
     with
-    - if user_track.num_concepts.positive?
+    - if @track.course? && user_track.num_concepts.positive?
       #{link_to "#{user_track.num_concepts} #{'concept'.pluralize(user_track.num_concepts)}", track_concepts_path(track)},
     #{link_to "#{user_track.num_exercises} #{'exercise'.pluralize(user_track.num_exercises)}", track_exercises_path(track)},
     and real human mentoring,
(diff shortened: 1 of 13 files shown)


app/views/tracks/exercises/show/_action_box_join_exercism.html.haml
```diff
@@ -10,7 +10,7 @@
     Sign up to Exercism to learn and master
     = link_to track.title, track_path(track)
     with
-    - if user_track.num_concepts.positive?
+    - if @track.course? && user_track.num_concepts.positive?
```
iHiD: For all of these in this PR, I think we can just do `@track.course?` and not bother with `user_track.num_concepts.positive?`. `configlet` shouldn't let tracks be courses without concepts, etc.

iHiD:
Nice! One overarching comment.

Re the redirect, yes, that'd be nice. Redirect from `/concepts` to the track dashboard.

app/views/tracks/exercises/show/_action_box_join_exercism.html.haml
ErikSchierboom: > configlet shouldn't let tracks be courses without concepts, etc.

This isn't yet the case, but I can fix that easily enough.

ErikSchierboom:
Fun little date-specific bug I think: https://github.com/exercism/website/runs/4702460594?check_suite_focus=true

PR #2313, opened 2022-01-07
Skip creating github issue in development

ErikSchierboom:
While testing the Rails 7 branch, any failures created issues on GH. This PR aims to prevent that from happening.

### DIFF
--- a/app/commands/github/issue/open.rb
+++ b/app/commands/github/issue/open.rb
@@ -6,6 +6,8 @@ module Github
       initialize_with :repo, :title, :body
 
       def call
+        return if Rails.env.development?
+
         if missing?
           create_issue
         elsif closed?


app/commands/github/issue/open.rb
```diff
@@ -6,6 +6,8 @@ class Open
       initialize_with :repo, :title, :body
 
       def call
+        return if Rails.env.development?
```
iHiD, line 9: I feel like we want this to be `return unless Rails.env.production?` but then I imagine that will break the tests?

ErikSchierboom: It will

iHiD:
So I'm not against this, but it feels wrong to me that the dev-env has keys that can create issues. I guess that's because it needs those keys to be able to do syncing of PRs etc? I wonder if there's a more robust way of doing this that's more permissions-based.

iHiD (approved):

PR #2331, opened 2022-01-13
Add timeouts to CI builds

SaschaMann:
Right now the builds are all stuck for some reason, clogging up the queues. They normal appear to be taking ~2-3 minutes, so a timeout of 30 minutes should be more than enough to prevent this issue in the future.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,6 +14,7 @@ jobs:
     if: github.repository == 'exercism/website'
 
     runs-on: ubuntu-latest
+    timeout-minutes: 30
 
     steps:
       - uses: actions/checkout@v2
@@ -225,6 +226,7 @@ jobs:
     if: github.repository == 'exercism/website'
 
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     outputs:
       matrix: ${{ steps.test-files.outputs.matrix }}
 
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
> They normal appear to be taking ~2-3 minutes

I think only the rubocop tests run under 2 minutes. All others take more time, with the Ruby tests and system tests taking around 10 minutes. See https://github.com/exercism/website/actions/runs/1657384569

That said, 30 minutes _should_ be fine.

ErikSchierboom (commented):
Could you also add the timeouts to the system tests?

SaschaMann:
> I think only the rubocop tests run under 2 minutes. All others take more time, with the Ruby tests and system tests taking around 10 minutes. See https://github.com/exercism/website/actions/runs/1657384569
>
> That said, 30 minutes should be fine.

Oops, then I misread the overview. Bumping to 60 minutes or so would still be fine, I think. Up to you :)

ErikSchierboom:
I think 30 minutes is fine. Should be done in that time.

ErikSchierboom (approved):
LGTM @iHiD?

ErikSchierboom (approved):

iHiD:
Great. Thanks @SaschaMann :)

PR #2342, opened 2022-01-25
Enable esbuild splitting

iHiD:
This:
- Enables splitting in esbuild
- Adds `type: module` to allow Chrome to still read the files
- Overrides propshaft server to not try and server digested versions of esbuild-created files.

If it works:
- You should see chunk-xxx.js and FooterForm-xxx.js and such things in the network traffic.
- The donations footer is probably the best place to see if this is working properly or not

### DIFF
--- a/app/javascript/esbuild.js
+++ b/app/javascript/esbuild.js
@@ -17,6 +17,8 @@ function build() {
       ],
       bundle: true,
       sourcemap: true,
+      format: 'esm',
+      splitting: true,
       minify: process.env.NODE_ENV === 'production',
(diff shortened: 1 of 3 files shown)


app/javascript/esbuild.js
```diff
@@ -17,6 +17,8 @@ function build() {
       ],
       bundle: true,
       sourcemap: true,
+      format: 'esm',
```
ErikSchierboom, line 20: Support looks good: https://caniuse.com/es6-module

ErikSchierboom (approved):

PR #2374, opened 2022-02-08
Fix publish data not being cleared when removing last non-deleted iteration

### LINKED ISSUE #6231: Some exercises don't count as completed while they are completed
In my C# track, I have 0 available exercises while 105 are still locked. I completed all my available exercises:

It seems that some completed exercises don't count as such. If I take a look at the _locked_ **Accumulate** exercise, it needs the completion of 2 exercises to be unlocked: **Tracks on Tracks on Tracks** and **Log Analysis**. 
If I compare the main page of these two exercises, I got the following:

While **Log Analysis** displays that it is completed, it is not the case for **Tracks on Tracks on Tracks**. Maybe that's part of the problem?

@ErikSchierboom maybe you have an idea of what's causing the trouble here?

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/6231

### DIFF
--- a/app/commands/iteration/destroy.rb
+++ b/app/commands/iteration/destroy.rb
@@ -7,11 +7,8 @@ class Iteration
     def call
       iteration.update!(deleted_at: Time.current)
 
-      if !solution.iterations.not_deleted.exists?
-        Solution::Uncomplete.(solution)
-      elsif solution.published_iteration_id == iteration.id
-        Solution::Unpublish.(solution)
-      end
+      Solution::Uncomplete.(solution) unless solution.iterations.not_deleted.exists?
+      Solution::Unpublish.(solution) if solution.published_iteration_id == iteration.id
     end
 
     private
--- a/app/commands/iteration/destroy.rb
+++ b/app/commands/iteration/destroy.rb
@@ -7,14 +7,15 @@ class Iteration
     def call
       iteration.update!(deleted_at: Time.current)
 
-      # Unpublish the solution if this was the published iteration
-      solution.update!(published_iteration_id: nil, published_at: nil) if solution.published_iteration_id == iteration.id
-
-      # Uncomplete the solution if there are now no iterations
-      solution.update!(completed_at: nil) unless solution.iterations.not_deleted.exists?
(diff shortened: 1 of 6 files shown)


app/commands/iteration/destroy.rb
```diff
@@ -7,14 +7,15 @@ class Destroy
     def call
       iteration.update!(deleted_at: Time.current)
 
-      # Unpublish the solution if this was the published iteration
-      solution.update!(published_iteration_id: nil, published_at: nil) if solution.published_iteration_id == iteration.id
-
-      # Uncomplete the solution if there are now no iterations
-      solution.update!(completed_at: nil) unless solution.iterations.not_deleted.exists?
+      if !solution.iterations.not_deleted.exists?
```
iHiD, line 10:
```suggestion
      unless solution.iterations.not_deleted.exists?
```

app/commands/solution/uncomplete.rb
```diff
@@ -0,0 +1,12 @@
+class Solution
+  class Uncomplete
+    include Mandate
+
+    initialize_with :solution
+
+    def call
+      solution.update!(completed_at: nil)
+      Solution::Unpublish.(solution)
```
iHiD: Maybe wrap this in a Transaction?

solution.transaction do
...
end

app/commands/iteration/destroy.rb
```diff
@@ -7,14 +7,15 @@ class Destroy
...
+        Solution::Uncomplete.(solution)
+      elsif solution.published_iteration_id == iteration.id
+        Solution::Unpublish.(solution)
+      end
```
iHiD: How about this to make it more readable. The second bit might run twice, but will just be a no-op the second time so it's not really painful.

```suggestion
      Solution::Uncomplete.(solution) unless solution.iterations.not_deleted.exists?
      Solution::Unpublish.(solution) if solution.published_iteration_id == iteration.id
```

ErikSchierboom:
> The second bit might run twice, but will just be a no-op the second time so it's not really painful.

I had that originally but reverted for the above-mentioned duplicate run. But if it is no big deal, that is much cleaner.

PR #2375, opened 2022-02-08
Deploy all tasks

ErikSchierboom:


.github/workflows/deploy.yml
```diff
@@ -9,13 +9,13 @@ env:
   website_service_name: website-staging
```
iHiD: This needs changing too.

ErikSchierboom:
```suggestion
  website_service_name: website
```

PR #2376, opened 2022-02-08
One final gems bump

iHiD:


Gemfile
```diff
@@ -81,7 +81,7 @@ end
 group :development do
   gem 'haml_lint', require: false
   gem 'solargraph'
-  gem 'rubocop', require: false
+  gem 'rubocop', '1.25.0', require: false
```
ErikSchierboom, line 84: Should this be `1.25.1`?

```suggestion
  gem 'rubocop', '1.25.1', require: false
```

See https://github.com/rubocop/rubocop/releases/tag/v1.25.1

PR #2380, opened 2022-02-09
Log API usage

ErikSchierboom:


app/controllers/api/base_controller.rb
```diff
@@ -7,6 +7,7 @@
 module API
   class BaseController < ApplicationController
     skip_before_action :verify_authenticity_token
+    after_action :log_usage!
```
ErikSchierboom, line 10: This is to ensure that the user has been authenticated.

app/controllers/api/base_controller.rb
```diff
@@ -90,5 +91,13 @@ def render_error(status, type, data = {})
         }.merge(data)
       }, status: status
     end
+
+    def log_usage!
+      Thread.new do
+        log_time = Time.current.utc.strftime('%Y-%m-%d %H:%M')
```
ErikSchierboom, line 99: This leaves out the seconds, effectively grouping by minute

app/controllers/api/base_controller.rb
```diff
@@ -90,5 +91,13 @@ def render_error(status, type, data = {})
...
+        Exercism.redis_tooling_client.incr("#{Exercism.env}:request:url:#{request.original_fullpath}:#{request.method}:#{log_time}")
```
ErikSchierboom: I've only logged the path, not the URL, as I _think_ we're not really interested in the host name, are we?

test/controllers/api/base_controller_test.rb
```diff
@@ -0,0 +1,85 @@
...
+      setup_user(user_1)
+      get api_docs_path, headers: @headers, as: :json
+      get api_docs_path, headers: @headers, as: :json
+      get api_docs_path, headers: @headers, as: :json
+      sign_out user_1
+
+      setup_user(user_2)
+      get api_docs_path, headers: @headers, as: :json
+
+      sleep(0.1)
```
ErikSchierboom, line 14: I tried stubbing `Thread.new`, but that lead to weird Redis errors caused by (I assume) them also using `Thread.new`.

app/controllers/api/base_controller.rb
```diff
@@ -90,5 +91,14 @@ def render_error(status, type, data = {})
         }.merge(data)
       }, status: status
     end
+
+    def log_usage!
+      Thread.new do
+        log_time = Time.current.utc.strftime('%Y-%m-%d %H:%M')
+        redis = Redis.new(url: "#{Exercism.config.tooling_redis_url}/5")
+        redis.incr("#{Exercism.env}:request:url:#{request.path}:#{request.method}:#{log_time}")
+        redis.incr("#{Exercism.env}:request:user:#{current_user.handle}:#{log_time}") if user_signed_in?
```
iHiD: - Merge these two lines together.
- Move `if user_signed_in?` to `return unless user_signed_in?` at the start of the method.
- `request.path` probably needs escaping
- We should probably use `id` not `handle` for `current_user` (safer and smaller).

app/controllers/api/base_controller.rb
```diff
@@ -90,5 +91,17 @@ def render_error(status, type, data = {})
...
+      return unless user_signed_in?
+
+      Thread.new do
+        log_time = Time.current.utc.strftime('%Y-%m-%d %H:%M')
+        path_hash = Digest::SHA1.hexdigest(request.path)
+        redis = Redis.new(url: "#{Exercism.config.tooling_redis_url}/5")
+        redis.incr("#{Exercism.env}:request:#{path_hash}:#{request.method}:#{current_user.id}:#{log_time}")
+        redis.set("#{Exercism.env}:url:#{path_hash}", request.path)
+      end
+    end
```
iHiD, line 107: Can you throw a `rescue` before the final end of this method, just in case there's anything that blows up pls (I can't see why there would be, but I always like to do this for temporary code)

PR #2382, opened 2022-02-09
Allow preloading JS from other domain

ErikSchierboom:
Without this change, this appeared in the console log of the browser:

> A preload for 'https://d24y9kuxp2d7l2.cloudfront.net/assets/core-8e98e125fb7acfe200ab5d1acf16a3dd2216e4fe.js' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.

After this change, that message disappears.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin for more information

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -10,12 +10,12 @@
     = stylesheet_link_tag "internal", "data-turbo-track": "reload"
     = stylesheet_link_tag "website", "data-turbo-track": "reload"
 
-    = javascript_include_tag('core', type: :module, 'data-turbo-track': 'reload', 'data-turbo-eval': false)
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.haml
```diff
@@ -10,12 +10,12 @@
     = stylesheet_link_tag "internal", "data-turbo-track": "reload"
     = stylesheet_link_tag "website", "data-turbo-track": "reload"
 
-    = javascript_include_tag('core', type: :module, 'data-turbo-track': 'reload', 'data-turbo-eval': false)
+    = javascript_include_tag('core', type: :module, crossorigin: :anonymous, 'data-turbo-track': 'reload', 'data-turbo-eval': false)
```
ErikSchierboom, line 13: Using `anonymous` mode means that the CORS request won't be sending any user credentials. See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin

PR #2398, opened 2022-02-10
Don't html escape single quote in title

ErikSchierboom:


iHiD:
I presume this works? (rather you you just trusting what I said? :))

ErikSchierboom:
It does.

iHiD:
Excellent :)

PR #2403, opened 2022-02-10
Fix issues with rerunning tests after a sync

### LINKED ISSUE #6229: Analyzer spins forever, sometimes
According to the UI there is a single message (probably: yay exemplary), but the message doesn't show up.

image

### LINKED ISSUE #6236: When updating outdated exercise with test changes, cannot re-run tests without making a change
Might be intended, but is confusing ;)

iHiD:
Fixes https://github.com/exercism/exercism/issues/6236
Fixes https://github.com/exercism/exercism/issues/6229

This does two things:
1. It fixes a bug where the wrong git_sha was being passed into the test runner.
2. This hardens the logic around updating an exercise. We now also update the most recent iteration's submission (ie the last meaningful submission the person did) to reflect the new git_sha, and we rerun the tests on that code. We **don't** as of yet rerun the analyzer/representer. Currently we (mentally) expect submission.git_sha to be immutable and so this breaks the (mental) model of how representers reference submissions, which will take some deeper thinking.

### DIFF
--- a/app/commands/solution/update_to_latest_exercise_version.rb
+++ b/app/commands/solution/update_to_latest_exercise_version.rb
@@ -18,7 +18,10 @@ class Solution
       return unless submission
 
       submission.update!(git_sha: solution.git_sha, git_slug: solution.git_slug)
-      Submission::TestRun::Init.(submission, type: :solution)
+
+      # We run this in submission mode (the default) so as to set the last
+      # iteration to look like it's reprocessing in the UI.
+      Submission::TestRun::Init.(submission, run_in_background: true)
     end
   end
 end
--- a/app/commands/solution/update_to_latest_exercise_version.rb
+++ b/app/commands/solution/update_to_latest_exercise_version.rb
@@ -1,0 +1,24 @@
+class Solution
+  class UpdateToLatestExerciseVersion
+    include Mandate
+
+    initialize_with :solution
+
+    def call
+      solution.sync_git!
+      update_latest_iteration!
+    end
+
+    # This updates the submission of the latest iteration
+    # by setting its git_sha and then rerunning the tests
+    # in solution mode, which leads the status of the solution
+    # being set to the result of this latest run.
+    def update_latest_iteration!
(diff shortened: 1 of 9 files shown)


app/commands/tooling_job/create.rb
```diff
@@ -22,7 +22,7 @@ def call
           submission_efs_root: submission.uuid,
           submission_filepaths: submission.valid_filepaths,
           exercise_git_repo: solution.track.slug,
-          exercise_git_sha: solution.git_sha,
+          exercise_git_sha: git_sha,
```
iHiD, line 25: This is the first bug mentioned in the description.

app/commands/solution/update_to_latest_exercise_version.rb
```diff
@@ -0,0 +1,27 @@
...
+    initialize_with :solution
+
+    def call
+      solution.sync_git!
+      update_latest_iteration!
+    end
+
+    # This updates the submission of the latest iteration
+    # by setting its git_sha and then rerunning the tests
+    # in solution mode, which leads the status of the solution
```
ErikSchierboom:
```suggestion
    # in submission mode, which leads the status of the solution
```

test/commands/solution/update_to_latest_exercise_version_test.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    Solution::UpdateToLatestExerciseVersion.(solution)
+    [old_submission, new_submission, deleted_submission].each(&:reload)
+
+    assert_equal "foo", old_submission.git_sha
+    assert_equal "bar", old_submission.git_slug
+    assert_equal solution.exercise.git_sha, new_submission.git_sha
+    assert_equal solution.exercise.slug, new_submission.git_slug
+    assert_equal "foo", deleted_submission.git_sha
+    assert_equal "bar", deleted_submission.git_slug
```
ErikSchierboom, line 40: Nice test

ErikSchierboom (approved):

PR #2411, opened 2022-02-11
Fix notification asset path duplicate slashes

ErikSchierboom:
image

Once this is merged, the cache needs to be cleared for the involved records

### DIFF
--- a/app/models/blog_post.rb
+++ b/app/models/blog_post.rb
@@ -33,8 +33,8 @@ class BlogPost < ApplicationRecord
   def image_url
     attributes['image_url'].presence ||
       [
-        Rails.application.config.action_controller.asset_host,
-        compute_asset_path("graphics/blog-placeholder-article.svg")
+        Rails.application.config.action_controller.asset_host&.delete_suffix('/'),
+        compute_asset_path("graphics/blog-placeholder-article.svg")&.delete_prefix('/')
       ].compact.join('/')
   end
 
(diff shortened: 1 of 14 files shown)


app/models/user/notification.rb
```diff
@@ -68,8 +68,8 @@ def non_cacheable_rendering_data
 
   def image_url
     [
-      Rails.application.config.action_controller.asset_host,
-      compute_asset_path(image_path)
+      Rails.application.config.action_controller.asset_host.delete_suffix('/'),
+      compute_asset_path(image_path).delete_prefix('/')
```
ErikSchierboom: Another option would be to do a gsub on the result. Let me know what you prefer

iHiD:
Looks good. Looks like there's a failing test.

app/models/blog_post.rb
```diff
@@ -32,10 +32,7 @@ def to_param
 
   def image_url
     attributes['image_url'].presence ||
-      [
-        Rails.application.config.action_controller.asset_host,
-        compute_asset_path("graphics/blog-placeholder-article.svg")
-      ].compact.join('/')
+      "#{Rails.application.config.action_controller.asset_host}#{compute_asset_path('graphics/blog-placeholder-article.svg')}"
```
ErikSchierboom, line 35: The new version should be okay, because `compute_asset_path` return a path that starts with a slash and `Rails.application.config.action_controller.asset_host` is either a host name without a slash at the end in production, or an absolute path (`/assets`) for all other environments.

PR #2431, opened 2022-02-16
Prevent duplicate key execeptions when re-seeding

ErikSchierboom:


iHiD:
I've not sure I get the purpose of this. Normally I do `rake db:reset db:seed` which wipes the db and reseeds it. Are you wanting to be able to run `db:seeds` multiple times?

ErikSchierboom:
I sometimes have data in the database that I want to keep, some state I want to preserve. But it turned out that somewhere the seeding was wrong in my DB, which is why I wanted to re-run the seeding. I'm fine with us closing if we never want to allow re-running the seed, but it also doesn't hurt I feel.

iHiD (approved):

iHiD:
Happy for you to merge it 🙂

PR #2438, opened 2022-02-18
Restrict access to inactive tracks to maintainers

ErikSchierboom:
- Fix invalid 404 in unauthorized tests
- Gracefully handle track without online_editor
- Gracefully handle user_track without exercises
- Gracefully handle track without exercises
- Render 404 when viewing inactive track using non-maintainer user
- Seed users with maintainer role
- Restrict inactive track usage more

### DIFF
--- a/app/controllers/api/tracks_controller.rb
+++ b/app/controllers/api/tracks_controller.rb
@@ -23,7 +23,7 @@ module API
         return render_404(:track_not_found, fallback_url: tracks_url)
       end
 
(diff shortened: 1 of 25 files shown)


ErikSchierboom:
@iHiD There are one or two places in the API where tracks/exercises are retrieved too. Do we also want to update those?

ErikSchierboom:
We probably want to give some visual indication the track is currently inactive, to prevent maintainers from forgetting that.

PR #2457, opened 2022-02-23
Link hints to tasks via ID

### LINKED ISSUE #6105: Hints should be matched by index, not by title
Currently, the code links a task to its hints via its title. This should be done via its index.

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/6105﻿

### DIFF
--- a/app/serializers/serialize_exercise_assignment.rb
+++ b/app/serializers/serialize_exercise_assignment.rb
@@ -32,7 +32,7 @@ class SerializeExerciseAssignment
   end
 
   def general_hints
-    hints["general"].to_a
+    hints[0].to_a
   end
 
   memoize
@@ -49,8 +49,8 @@ class SerializeExerciseAssignment
       next unless header.header_level == 2
       next unless list.type == :list
 
-      heading = parse_title(header).downcase
-      hints[heading] = list.each.map { |list_item| Markdown::Parse.(list_item.each.first.to_commonmark) }
+      task_id = parse_task_id(header)
+      hints[task_id] = list.each.map { |list_item| Markdown::Parse.(list_item.each.first.to_commonmark) }
     end
   end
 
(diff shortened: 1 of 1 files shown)


app/serializers/serialize_exercise_assignment.rb
```diff
@@ -87,9 +88,17 @@ def hints_doc
...
+    # "## General" becomes "General"
     header.to_plaintext.strip.gsub(/^(^\d+)\.\s*(.*)/, '\2')
   end
+
+  def parse_task_id(header)
+    # Parse the task's ID, which is the starting number.
+    # "## 2. Do stuff" becomes 2
+    # "## General" becomes 0
+    header.to_plaintext.strip.gsub(/^(^\d+)\.\s*(.*)/, '\1').to_i
+  end
```
ErikSchierboom, line 103: I've opted for some duplication, but I could also be converting these into a single method returning a hash 🤷

app/serializers/serialize_exercise_assignment.rb
```diff
@@ -32,7 +32,7 @@ def overview
   end
 
   def general_hints
-    hints["general"].to_a
+    hints[0].to_a
```
iHiD, line 35: Is `general` always first? Is that in the spec?

app/serializers/serialize_exercise_assignment.rb
iHiD: I'm fine with either, although you **could** use the funky Ruby 3 hash-key-pattern matching stuff here probably.

app/serializers/serialize_exercise_assignment.rb
ErikSchierboom: No, but the first hint starts at `1` and if there is no number there (meaning: "General") the `to_i` call resolves to `0` :)

iHiD: #rubyskillz

iHiD (approved):
I'll just trust you on this one!

ErikSchierboom:
Well, the tests all pass and they look to be quite extensive.

PR #2458, opened 2022-02-23
Correctly determine when a track is new

### LINKED ISSUE #5835: [Tracks] Almost all tracks have the `New` badge
From what I have observed in the Tracks section, the tracks which have 0 concepts have a `New` badge even if they were present in v2 and some tracks which are newly created and have concepts don't show a `New` badge

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/5835

### DIFF
--- a/app/serializers/serialize_track.rb
+++ b/app/serializers/serialize_track.rb
@@ -18,10 +18,7 @@ class SerializeTrack
       icon_url: track.icon_url,
       tags: map_tags(track.tags),
       last_touched_at: user_track.last_touched_at&.iso8601,
-
-      # TODO: Set these
-      is_new: true,
-
+      is_new: track.created_at > Time.current - 6.months,
       links: {
         self: Exercism::Routes.track_url(track),
         exercises: Exercism::Routes.track_exercises_url(track),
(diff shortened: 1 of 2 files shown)


app/models/track.rb
```diff
@@ -92,6 +92,10 @@ def average_test_duration
     git.average_test_duration + INFRASTRUCTURE_DURATION_S
   end
 
+  def new?
```
ErikSchierboom: Do you think it makes sense to add this to the `Track` model, or should this just reside in `SerializeTrack`?

app/models/track.rb
```diff
@@ -92,6 +92,10 @@ def average_test_duration
...
+    created_at > Time.zone.now - 6.months
```
ErikSchierboom: I think we said 6 months, didn't we?

app/models/track.rb
iHiD: I think do it in SerializeTrack. If we ever then use it somewhere else, we can move it here, but I think keeping the models small is good.

app/serializers/serialize_track.rb
```diff
@@ -18,10 +18,7 @@ def call
       icon_url: track.icon_url,
       tags: map_tags(track.tags),
       last_touched_at: user_track.last_touched_at&.iso8601,
-
-      # TODO: Set these
-      is_new: true,
-
+      is_new: track.created_at > Time.zone.now - 6.months,
```
iHiD: I normally use `Time.current` here instead of `Time.zone.now` - probably worth changing this to be consistent.

PR #2472, opened 2022-02-28
Fix account deletion issue

iHiD:
@nobbz When this is deployed, the user will be able to delete their account.

### DIFF
--- a/app/commands/user/reset_account.rb
+++ b/app/commands/user/reset_account.rb
@@ -22,15 +22,25 @@ class User
     end
 
     def reset_tracks!
+      # If someone has left a user track then we won't reset
+      # and things will break. So we need to temporarily recreate
+      # the user tracks.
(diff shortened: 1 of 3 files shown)


app/commands/user/reset_account.rb
```diff
@@ -22,15 +22,25 @@ def call
     end
 
     def reset_tracks!
+      # If someone has left a user track then we won't reset
+      # and things will break. So we need to temporarily recreate
+      # the user tracks.
```
ErikSchierboom: Isn't this only an issue when the user has submitted some solutions?

```suggestion
      # If someone has submitted solutions then left a user track,
      # then we won't reset those solutions and things will break.
      # So we need to temporarily recreate the user tracks here.
```

ErikSchierboom (approved):

PR #2490, opened 2022-03-04
Tasks should be optional for v3 runners

iHiD:
See https://exercism-team.slack.com/archives/GC3K95MRR/p1646410077309919 (cc @ErikSchierboom)

### DIFF
--- a/app/javascript/components/editor/TestRunOutput.tsx
+++ b/app/javascript/components/editor/TestRunOutput.tsx
@@ -8,13 +8,10 @@ export const TestRunOutput = ({
 }: {
   testRun: TestRun
 }): JSX.Element => {
-  const testsWithoutTasks = testRun.tests.filter(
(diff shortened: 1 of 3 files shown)


app/javascript/components/editor/TestRunOutput.tsx
```diff
@@ -8,7 +8,12 @@ export const TestRunOutput = ({
 }: {
   testRun: TestRun
 }): JSX.Element => {
-  if (testRun.version >= 3 && testRun.tasks.length > 0) {
+  const hasTasks =
+    testRun.version >= 3 &&
+    testRun.tasks.length > 0 &&
+    testRun.tests.filter((t) => t.taskId == null || t.taskId == undefined) == 0
```
ErikSchierboom, line 14: Funny thing: this is testing whether an empty array is equal to `0`, which funnily enough, it is in JS! I'll refactor next week.

PR #2497, opened 2022-03-09
Simplify task ID code

ErikSchierboom:
- Fix typing error
- Simplify task ID test

### DIFF
--- a/app/javascript/components/editor/TestRunOutput.tsx
+++ b/app/javascript/components/editor/TestRunOutput.tsx
@@ -11,7 +11,7 @@ export const TestRunOutput = ({
   const hasTasks =
     testRun.version >= 3 &&
     testRun.tasks.length > 0 &&
-    testRun.tests.filter((t) => t.taskId == null || t.taskId == undefined) == 0
+    testRun.tests.every((t) => t.taskId !== null && t.taskId !== undefined)
 
   if (hasTasks) {
     return (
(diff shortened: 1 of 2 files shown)


app/javascript/components/editor/TestRunSummary.tsx
```diff
@@ -70,7 +70,7 @@ const TestRunSummaryStatus = ({
   children,
   percentagePassing,
 }: React.PropsWithChildren<{
-  status: string
+  statusClass: string
```
ErikSchierboom, line 73: This was a typing error

app/javascript/components/editor/TestRunOutput.tsx
```diff
@@ -11,7 +11,7 @@ export const TestRunOutput = ({
   const hasTasks =
     testRun.version >= 3 &&
     testRun.tasks.length > 0 &&
-    testRun.tests.filter((t) => t.taskId == null || t.taskId == undefined) == 0
+    testRun.tests.every((t) => t.taskId)
```
iHiD: Just to be sure, can taskId validly be 0 (ie do tracks every index from 0)? Just because 0 is falsey in JS.

ErikSchierboom: Output `0` as the task ID would be an error in the test runner, but I've made it more explicit

iHiD:
@ErikSchierboom Any chance you could fix the flakey test pls? It's a feburary bug.

PR #2499, opened 2022-03-09
Fix time_ago_in_words tests when previous month is Feb

ErikSchierboom:
The problem was that for the month of March, `Time.current - 1.month` would return `28d` instead of `1m`.

### DIFF
--- a/test/helpers/time_helper_test.rb
+++ b/test/helpers/time_helper_test.rb
@@ -2,6 +2,7 @@ require "test_helper"
 
 class TimeHelperTest < ActionView::TestCase
   test "time_ago_in_words" do
+    travel_to(Date.new(2022, 6, 3))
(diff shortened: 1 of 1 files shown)


test/helpers/time_helper_test.rb
```diff
@@ -14,6 +15,7 @@ class TimeHelperTest < ActionView::TestCase
   end
```
iHiD, line 15: Does `travel_to` need a block to unravel itself at the end of the test, or is it locally scoped?

ErikSchierboom: According to [the docs](https://api.rubyonrails.org/v6.0.3/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel_to)

> The stubs are automatically removed at the end of the test.

PR #2501, opened 2022-03-09
Allow awarding arbitrary reputation

ErikSchierboom:


app/models/user/reputation_period.rb
```diff
@@ -1,7 +1,7 @@
 class User::ReputationPeriod < ApplicationRecord
   enum period: { forever: 0, year: 1, month: 2, week: 3 }
   enum about: { everything: 0, track: 1 }, _prefix: true
-  enum category: { any: 0, building: 1, maintaining: 2, authoring: 3, mentoring: 4 }, _suffix: true
+  enum category: { any: 0, building: 1, maintaining: 2, authoring: 3, mentoring: 4, other: 5 }, _suffix: true
```
ErikSchierboom: I've added an `other` category. Do we want that or something else?

app/models/user/reputation_tokens/arbitrary_token.rb
```diff
@@ -0,0 +1,22 @@
+class User::ReputationTokens::ArbitraryToken < User::ReputationToken
+  params :arbitrary_value, :arbitrary_reason
```
ErikSchierboom, line 2: As said, these have to be prefixed with `arbitrary_` due to conflicts with existing `value` and `reason` methods.

app/models/user/reputation_tokens/arbitrary_token.rb
```diff
@@ -0,0 +1,22 @@
...
+    SecureRandom.uuid
+  end
+
+  def i18n_params
+    { reason: }
+  end
+
+  protected
+  def determine_value
+    arbitrary_value
```
ErikSchierboom, line 20: Overridden to not use the default logic for determining the value

ErikSchierboom (commented):
The main question is whether this token needs to show up anywhere in the contributions.

app/models/user/reputation_period.rb
iHiD: I don't think we need this at all.

app/models/user/reputation_tokens/arbitrary_token.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
```
iHiD: Want to go all posh Ruby 3?
```suggestion
  def determine_value = arbitrary_value
```

ErikSchierboom: 😮

ErikSchierboom: But: yes!

app/models/user/reputation_period.rb
ErikSchierboom: Something errored without this. I'll take a deeper dive

iHiD: Interesting. Thanks. This bit is quite complex in terms of what is going on, so shout if you want me to look with you.

iHiD:
> The main question is whether this token needs to show up anywhere in the contributions.

It should appear on the journey page and the public profile. Let me know if you want to bounce how to make that happen.

ErikSchierboom:
For a demonstration, see: https://www.loom.com/share/589263106279458c9813ac76d46af05c

app/models/user/reputation_period.rb
ErikSchierboom: This is due to the `MarkForNewToken` class not being able to deal with the category. I've changed the code to exclude the `:misc` category, but want to double check if that is what we want. It depends on whether or not we want to show misc reputation on the contributors page:

image

app/commands/user/reputation_period/mark_for_new_token.rb
```diff
@@ -10,7 +10,7 @@ def call
       rows = %i[forever year month week].flat_map do |period|
         # Make rows for :any and the category as long as the category isn't publishing
         categories = [:any]
-        categories << token.category unless token.category.to_sym == :publishing
+        categories << token.category unless %i[publishing misc].include?(token.category.to_sym)
```
iHiD, line 13: Maybe update the comment two lines up too?

app/commands/user/reputation_period/mark_for_new_token.rb
```diff
@@ -10,7 +10,7 @@ def call
       rows = %i[forever year month week].flat_map do |period|
         # Make rows for :any and the category as long as the category isn't publishing
```
ErikSchierboom:
```suggestion
        # Make rows for :any and the category as long as the category isn't publishing or misc
```

PR #2503, opened 2022-03-10
Use POST when connecting to GitHub from settings

### LINKED ISSUE #6151: can't integrate with github
I can't integrate my existing account with GitHub.

# Step to reproduce

after logged in,

* go to settings
* under GitHub Integration -> click Connect with GitHub
* get return 500 error

ErikSchierboom:
The GitHub authentication button on the settings page was still using a link, whereas on other pages a (POST) form with a button was used.

@iHiD Is there an easy to test if this fixes it? ﻿

Closes https://github.com/exercism/exercism/issues/6151

### DIFF
--- a/app/views/settings/show.html.haml
+++ b/app/views/settings/show.html.haml
@@ -41,8 +41,10 @@
               reputation
               = icon "external-link", "(opens in a new tab)"
             on Exercism when your submit pull requests on GitHub, please connect your accounts using the button below.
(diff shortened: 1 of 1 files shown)


iHiD:
No, it's hard to test. Easiest thing would be to get someone without a bazillion rep to test it manually. We disconnect them, then they hit the button, then we see if they're reconnected.

PR #2504, opened 2022-03-10
Allow skip sending email when awarding badges

ErikSchierboom:
What is the best way to test whether or not an email actually gets sent? I think Rails will not send emails by default in development.

### DIFF
--- a/app/commands/user/acquired_badge/create.rb
+++ b/app/commands/user/acquired_badge/create.rb
@@ -3,7 +3,11 @@ class User
     class Create
       include Mandate
 
-      initialize_with :user, :slug
+      def initialize(user, slug, skip_email: false)
+        @user = user
+        @slug = slug
+        @skip_email = skip_email
+      end
 
       def call
         # Check to see if it exists already before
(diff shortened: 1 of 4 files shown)


app/commands/user/acquired_badge/create.rb
```diff
@@ -3,7 +3,11 @@ class AcquiredBadge
     class Create
       include Mandate
 
-      initialize_with :user, :slug
+      def initialize(user, slug, skip_email: false)
+        @user = user
+        @slug = slug
+        @skip_email = skip_email
+      end
```
iHiD, line 10: We really need to support `initialize_with` in mandate for kwargs!

ErikSchierboom: We do!

app/commands/user/acquired_badge/create.rb
```diff
@@ -3,7 +3,11 @@ class AcquiredBadge
     class Create
       include Mandate
 
-      initialize_with :user, :slug
+      def initialize(user, slug, skip_email: false)
```
iHiD:
```suggestion
      def initialize(user, slug, send_email: true)
```

iHiD (approved):

PR #2517, opened 2022-03-17
Dispatch repo_update event when appends file or org-wide-files config changes

ErikSchierboom:
Recently, we added support for a `.github/org-wide-files-config.toml` file to specify some org-wide-files syncing options.
Currently, we only use this to allow tracks to enable running `configlet fmt` in their `configlet` workflow: https://github.com/exercism/csharp/blob/main/.github/org-wide-files-config.toml
This PR now also submits the (renamed) `repo_update` event (used to be `appends_update`, but we're renaming that) when the above-mentioned config file changes. 
This is needed as otherwise changing that config file would not do anything unless somehow a new org-wide-files run starts.

Once this is merged and deployed, https://github.com/exercism/org-wide-files/pull/186 should be merged.

CC @SaschaMann

### DIFF
--- a/app/commands/github/dispatch_event_to_org_wide_files_repo.rb
+++ b/app/commands/github/dispatch_event_to_org_wide_files_repo.rb
@@ -21,7 +21,7 @@ module Github
       }.to_json
     end
 
-    EVENT_TYPES = %i[appends_update].freeze
+    EVENT_TYPES = %i[repo_update].freeze
     private_constant :EVENT_TYPES
   end
 end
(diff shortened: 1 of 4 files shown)


iHiD:
Two thoughts:
1. Do we have a race between merging this and merging equivs on the tracks?
2. Should we rerun a sync on all tracks in case we've dropped any hooks once this is all done?

@ErikSchierboom I'm approving and you can merge when you feel it's appropriate. 

Thanks to you and @SaschaMann for all your hard work on this.

iHiD (approved):

ErikSchierboom:
@iHiD There would be a race condition yes. I've come up with a better solution: I'm modifying the org-wide-files workflow to temporarily support both types: https://github.com/exercism/org-wide-files/pull/186/files @SaschaMann that would work, wouldn't it?

PR #2532, opened 2022-03-23
Shorten badge descriptions

ErikSchierboom:
Do we need to manually update the database?

### DIFF
--- a/app/models/badges/all_your_base_badge.rb
+++ b/app/models/badges/all_your_base_badge.rb
@@ -3,7 +3,7 @@ module Badges
     seed "All your base are belong to us",
       :rare,
       'all-your-base',
-      'Awarded for completing the "All Your Base" exercise'
+      'Completed the "All Your Base" exercise'
 
     def award_to?(user)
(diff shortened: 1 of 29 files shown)


iHiD:
> Do we need to manually update the database?

We need to reset the cached data for the existing badges.

I think you want `UserAcquiredBadge.update_all(cached_rendering_data: "{}")` but definitely run this locally to check first, then also run online just for you (`UserAcquiredBadge.where(user_id: xxx).update_all(cached_rendering_data: "{}")`)

You want to use `update_all` to do it via SQL. If you do `update` it will try to reset the cahced data and that will fail on bastion.

ErikSchierboom:
This is the script to update the badges:

```ruby
Badge.find_each do |badge|
    seed_data = badge.class.instance_variable_get("@seed_data")

    badge.update(
    name: seed_data[:name],
    rarity: seed_data[:rarity],
    icon: seed_data[:icon],
    description: seed_data[:description]
)
end
```

ErikSchierboom:
The script just ran successfully

PR #2535, opened 2022-03-24
Check tooling status before initializing job

ErikSchierboom:
We recently noticed that the Elm analyzer has run and added comments to a solution, even though the `status.analyzer` field was still disabled.
This is both wrong (an analyzer that is being built would already be running) and wasteful.
I've also done the same for the representer and for the one case where the test runner check was not applied.

### DIFF
--- a/app/models/git/track.rb
+++ b/app/models/git/track.rb
@@ -51,22 +51,22 @@ module Git
 
     memoize
     def has_concept_exercises?
-      !!config[:status][:concept_exercises]
+      !!status[:concept_exercises]
     end
 
     memoize
     def has_test_runner?
-      !!config[:status][:test_runner]
+      !!status[:test_runner]
     end
 
     memoize
     def has_representer?
-      config[:status][:representer]
+      !!status[:representer]
     end
 
     memoize
     def has_analyzer?
-      config[:status][:analyzer]
+      !!status[:analyzer]
     end
 
     memoize
@@ -152,6 +152,11 @@ module Git
     memoize
     def test_runner
       config[:test_runner] || {}
+    end
+
+    memoize
+    def status
+      config[:status] || {}
     end
   end
 end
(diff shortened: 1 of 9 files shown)


app/models/git/track.rb
```diff
@@ -51,22 +51,22 @@ def key_features
 
     memoize
     def has_concept_exercises?
-      !!config[:status][:concept_exercises]
+      !!status[:concept_exercises]
```
ErikSchierboom, line 54: I've changed this to use a (memoized) `status` method to gracefully handle tracks that don't have a `status` property.

test/controllers/api/profiles/solutions_controller_test.rb
```diff
@@ -63,7 +63,7 @@ class API::Profiles::SolutionsControllerTest < API::BaseTestCase
     setup_user
 
     profile_user = create(:user_profile).user
-    5.times { create :practice_solution, :published, user: profile_user }
+    5.times { |i| create :practice_solution, :published, user: profile_user, num_stars: i }
```
ErikSchierboom, line 66: This test was failing on my machine, and it had to do with the sorting. All solutions had the same number of stars (which is the sorted field), but somehow the returned and the expected value were different. This assigns them a unique number of stars.

iHiD (approved):

db/migrate/20220324090941_add_tooling_status_columns_to_tracks.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    add_column :tracks, :has_representer, :boolean, null: false, default: false
+    add_column :tracks, :has_analyzer, :boolean, null: false, default: false
+
+    Track.find_each do |track|
+      track.update(
+         has_representer: track.git.has_representer?,
+         has_analyzer: track.git.has_analyzer?,
+      )
+    end
```
iHiD:
```suggestion
```

Run that bit on bastion instead.

PR #2542, opened 2022-03-25
Allow reputation from issue creation

ErikSchierboom:
One open issue is what do to with deleting a reputation token. So far, reputation tokens are never deleted AFAICT, but here it could be the case that a label is applied and then removed. This was not an issue for PRs, as they had a base reputation awarded. I think there are two solutions:

1. Removing the reputation token
2. Setting the reputation token's value to 0

### DIFF
--- a/app/commands/user/reputation_token/award_for_issue.rb
+++ b/app/commands/user/reputation_token/award_for_issue.rb
@@ -7,22 +7,14 @@ class User
 
       def call
         return unless label_change?
-        return unless reputation_level
         return unless author
         return unless author.maintainer? || author.admin?
 
-        token = User::ReputationToken::Create.(
-          author,
-          :issue_author,
-          level: reputation_level,
-          repo: params[:repo],
-          issue_node_id: params[:node_id],
-          issue_number: params[:number],
-          issue_title: params[:title],
-          external_url: params[:html_url],
-          opened_at: params[:opened_at]
-        )
-        token&.update!(level: reputation_level)
+        if reputation_level
+          create_token!
+        else
+          delete_token!
+        end
       end
 
       private
@@ -41,6 +33,29 @@ class User
       memoize
       def author
         User.find_by(github_username: params[:opened_by_username])
+      end
+
+      def create_token!
+        User::ReputationToken::Create.(
+          author,
+          :issue_author,
+          level: reputation_level,
+          repo: params[:repo],
+          issue_node_id: params[:node_id],
+          issue_number: params[:number],
+          issue_title: params[:title],
+          external_url: params[:html_url],
+          opened_at: params[:opened_at]
+        )&.tap do |token|
+          token.update!(level: reputation_level)
+        end
+      end
+
+      def delete_token!
+        User::ReputationTokens::IssueAuthorToken.where(user: author, external_url: params[:html_url]).first&.tap do |token|
+          token.destroy!
+          User::ReputationPeriod::MarkForToken.(token)
+        end
       end
     end
   end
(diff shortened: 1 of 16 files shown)


app/commands/user/reputation_token/award_for_issue.rb
```diff
@@ -0,0 +1,47 @@
...
+    class AwardForIssue
+      include Mandate
+
+      initialize_with :params
+
+      def call
+        return unless label_change?
+        return unless reputation_level
+        return unless author
+        return unless author.maintainer? || author.admin?
```
ErikSchierboom, line 11: I could change this to:

```suggestion
        return unless author&.maintainer? || author&.admin?
```

but I was unsure whether this is idiomatic

iHiD: Yeah - this is fine.

app/commands/user/reputation_token/award_for_issue.rb
```diff
@@ -0,0 +1,62 @@
...
+          issue_title: params[:title],
+          external_url: params[:html_url],
+          opened_at: params[:opened_at]
+        )&.tap do |token|
+          token.update!(level: reputation_level)
+        end
+      end
+
+      def delete_token!
+        User::ReputationTokens::IssueAuthorToken.where(user: author, external_url: params[:html_url]).first&.tap do |token|
```
ErikSchierboom: Ideally, I'd be using the uniqueness key here to find the token, but for that I need to create the token first. Any thoughts here? There is no index on `external_url` either, but it might not be that bad with the filter already using the type of the token and the user.

app/models/user/reputation_token.rb
```diff
@@ -32,8 +32,9 @@ class User::ReputationToken < ApplicationRecord
     self.value = self.determine_value
   end
 
-  after_save_commit do
-    summing_sql = Arel.sql("(#{user.reputation_tokens.select('SUM(value)').to_sql})")
+  after_commit do
```
ErikSchierboom, line 35: This needed to change to `after_commit` as otherwise the rep was not recalculated after deleting the token.

app/commands/user/reputation_token/award_for_issue.rb
iHiD: You can use `find_by` instead of `where...first`

iHiD: I think it'll be fine with the user/type filter. Run a query in a rails c and check how long it takes and if it's slow send me the query to check?

iHiD: I'm ok with this not being a command for now, but if we get to the point we're deleting other tokens, it should be DRY'd into that as it's a two-step thing.

iHiD (approved):
I'd probably check the query in rails c before merging. But other than that, this looks solid :)

app/commands/user/reputation_token/award_for_issue.rb
ErikSchierboom: I'm fine with adding it as a command.

ErikSchierboom: I've updated

iHiD (approved):
Looks solid!

ErikSchierboom:
@iHiD The query is reasonably fast (89ms):

```shell
irb(main):006:0> User::ReputationToken.find_by(user: user, external_url: external_url)
D, [2022-03-31T14:06:49.629017 #229509] DEBUG -- :   User::ReputationToken Load (88.7ms)  SELECT `user_reputation_tokens`.* FROM `user_reputation_tokens` WHERE `user_reputation_tokens`.`user_id` = 23560 AND `user_reputation_tokens`.`external_url` = 'https://github.com/exercism/github-actions/pull/57' LIMIT 1
```

This result is for me, and I'm likely the one with the most reputation tokens (by far), so should be faster for other users.

PR #2545, opened 2022-03-25
Show correct time for initial mentor request post

### LINKED ISSUE #6198: Bug: The timestamp of the student's starting point comment is being overridden.
I've noticed that, in the mentor discussion section, whenever you start a mentoring discussion and leave your first comment, the timestamp of the student's starting point comment gets overridden with your comment timestamp.

My gut feeling tells me this has something to do with `shortFromNow(post.updatedAt)` or something from that like.

In the following recording, see how the timestamp of the student's comment  goes from `5d ago` to `now` as soon as I post my comment:

https://user-images.githubusercontent.com/87103727/149981771-79857d05-b266-48de-8d3b-90094abb8466.mov

Hope this helps! Thanks!

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/6198﻿

### DIFF
--- a/app/models/mentor/request_comment.rb
+++ b/app/models/mentor/request_comment.rb
@@ -30,7 +30,7 @@ class Mentor::RequestComment
       by_student: true,
       content_markdown: request.comment_markdown,
       content_html: request.comment_html,
-      updated_at: request.updated_at,
+      updated_at: request.created_at,
       request: request
     )
   end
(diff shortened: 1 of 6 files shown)


app/models/mentor/request_comment.rb
```diff
@@ -30,7 +30,7 @@ def self.from(request)
       by_student: true,
       content_markdown: request.comment_markdown,
       content_html: request.comment_html,
-      updated_at: request.updated_at,
+      updated_at: request.created_at,
```
ErikSchierboom, line 33: The problem was that once a mentor request is accepted, its status column is updated which causes the `updated_at` column to be updated too. I've replaced this with the `created_at` column. Note that the returned field needs to be `updated_at` because this is basically faking an actual request comment, and for those we _do_ want the updated date,

iHiD (approved):

PR #2575, opened 2022-04-05
Don't redirect to /site.webmanifest after GitHub registration

### LINKED ISSUE #6314: Bug report: redirected to site.manifest after going through GitHub Oauth flow
IIRC what I did was:

1. Was viewing the list of tracks https://exercism.org/tracks
2. Selected a track (I selected Racket)
3. Saw the login-with-github-or-create-an-account-page
4. Clicked GitHub
5. Clicked whatever on the GitHub page that gives permission to exercism for me to login
6. When redirected to exercism, wound up looking at: https://exercism.org/site.webmanifest

As soon as I just went to a new tab and went to exercism.org, I saw the main site and was already authenticated; so the auth worked, but something about the redirect probably did something it wasn't supposed to.

Just wanted you all to know!

PS, just visiting the site after not having seen it for many years and it looks AMAZING, you all are awesome & thank you!

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/6314

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -72,7 +72,8 @@ class ApplicationController < ActionController::Base
 
   private
   def storable_location?
-    request.get? && is_navigational_format? && !devise_controller? && !request.xhr?
+    request.get? && is_navigational_format? && !devise_controller? && !request.xhr? &&
+      request.fullpath != '/site.webmanifest'
   end
 
   def after_sign_in_path_for(resource_or_scope)
--- a/test/system/flows/user_registration_test.rb
+++ b/test/system/flows/user_registration_test.rb
@@ -114,6 +114,34 @@ module Flows
       OmniAuth.config.test_mode = false
     end
 
+    test "user registers via Github, onboards, and is not redirected to /site.webmanifest" do
+      OmniAuth.config.test_mode = true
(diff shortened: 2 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -72,7 +72,8 @@ def self.disable_site_header!(*args)
 
   private
   def storable_location?
-    request.get? && is_navigational_format? && !devise_controller? && !request.xhr?
+    request.get? && is_navigational_format? && !devise_controller? && !request.xhr? &&
+      request.fullpath != '/site.webmanifest'
```
ErikSchierboom, line 76: For now, I've just kept it simple with just the `request.fullpath != '/site.webmanifest'` condition. If we want to add more urls later, we can refactor.

test/system/flows/user_registration_test.rb
```diff
@@ -114,6 +114,34 @@ class UserRegistrationTest < ApplicationSystemTestCase
       OmniAuth.config.test_mode = false
     end
 
+    test "user registers via Github, onboards, and is not redirected to /site.webmanifest" do
```
ErikSchierboom, line 117: This test fails with the old code but passes with the new code.

iHiD:
Brilliant. Thanks.

PR #2582, opened 2022-04-07
Add track completion badges

ErikSchierboom:
- Add badge for track completion
- Keep track of concept exercise progress
- Add conceptual badge

What would the best way be to award this to existing users? Iterating over their `.user_tracks`?


app/images/icons/completer.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="384" height="384" viewBox="0 0 384 384" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ErikSchierboom: Another badge to create :)

app/models/badges/completer_badge.rb
```diff
@@ -0,0 +1,18 @@
+module Badges
+  class CompleterBadge < Badge
```
ErikSchierboom, line 2: I named the badge "Completer" to correspond to the terminology we use on the track page (Completed % exercises)

app/models/badges/completer_badge.rb
```diff
@@ -0,0 +1,18 @@
...
+    seed "Completer",
+      :ultimate,
+      :completer,
+      "Completed all exercises in a track"
+
+    def award_to?(user)
+      user.user_tracks.joins(:track).
+        where('tracks.active').
```
ErikSchierboom, line 10: This is a safeguard to not award it to tracks that are not yet finished.

app/images/icons/conceptual.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="384" height="384" viewBox="0 0 384 384" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ErikSchierboom: And another

app/models/badges/completer_badge.rb
```diff
@@ -0,0 +1,18 @@
+module Badges
+  class CompleterBadge < Badge
+    seed "Completer",
+      :ultimate,
```
ErikSchierboom, line 4: I made them ultimate. I've not yet done the bastion check to see how many users would get this badge.

test/commands/iteration/create_test.rb
```diff
@@ -199,7 +199,7 @@ class Iteration::CreateTest < ActiveSupport::TestCase
       submission = create :submission, solution: solution
       Iteration::Create.(solution, submission)
 
-      assert_equal Badges::DieUnendlicheGeschichteBadge, user.reload.badges.first.class
+      assert_includes user.reload.badges.map(&:class), Badges::DieUnendlicheGeschichteBadge
```
ErikSchierboom, line 202: I've changed the assertion slightly to make it clear that we don't care for the order of the badges. 
This was, we could refactor the order of the award jobs being run without the test breaking.

app/models/badges/completer_badge.rb
```diff
@@ -0,0 +1,18 @@
...
+      :completer,
+      "Completed all exercises in a track"
+
+    def award_to?(user)
+      user.user_tracks.joins(:track).
+        where('tracks.active').
+        any? { |ut| ut.completed_percentage == 100 }
```
iHiD, line 11: Is `completed_percentage` a DB column?

iHiD: Ah. ignore me.

app/models/badges/conceptual_badge.rb
```diff
@@ -0,0 +1,18 @@
...
+    seed "Conceptual",
+      :ultimate,
+      :conceptual,
+      "Completed all learning exercises in a track"
+
+    def award_to?(user)
+      user.user_tracks.joins(:track).
+        where('tracks.active').
+        any? { |ut| ut.completed_concept_exercises_percentage == 100 }
+    end
```
iHiD, line 12: This needs to guard to only look at tracks that have concepts.

iHiD:
> What would the best way be to award this to existing users? Iterating over their .user_tracks?

Yeah - probably select all users that have completed >19 exercises (or whatever the lowest number of exercises in any track is) and then run queue jobs for all of them.

ErikSchierboom:
@iHiD I've assigned you to help you remember to create the icons.

iHiD (approved):

iHiD (approved):
Looks good!

ErikSchierboom:
@iHiD these are the commands that I'll run to award the badges. Could you sanity check? I've also made them send out the email when they're awarded, is that okay?

```ruby
Track.active.each do |track|
    num_exercises = UserTrack::External.new(track).exercises.count
    
    completer_user_ids =
        Arel.sql(
             Solution
                 .joins(:exercise)
                 .where(status: %i[completed published])
                 .where('exercises.track_id = ?', track.id)
                 .group('user_id')
                 .select('user_id')
                 .having("COUNT(*) >= #{num_exercises}")
                 .distinct
                 .to_sql)

     User.where("id IN (#{completer_user_ids})").find_each do |user|
        AwardBadgeJob.perform_later(user, :completer)
    end
end
```

and 

```ruby
Track.active.where(course: true).each do |track|
    num_concept_exercises = UserTrack::External.new(track).concept_exercises.count
    
    conceptual_user_ids =
        Arel.sql(
             Solution
                 .joins(:exercise)
                 .where(status: %i[completed published])
                 .where('exercises.track_id = ?', track.id)
                 .where('exercises.type = ?', 'ConceptExercise')
                 .group('user_id')
                 .select('user_id')
                 .having("COUNT(*) >= #{num_concept_exercises}")
                 .distinct
                 .to_sql)

     User.where("id IN (#{conceptual_user_ids})").find_each do |user|
        AwardBadgeJob.perform_later(user, :conceptual)
    end
end
```

iHiD:
1. I'd use `count` not `size`
2. Do we need to look at `status: :active` counts for concept exercises, not just raw counts?

Other than that, looks solid to me! :)

ErikSchierboom:
1. Updated
2. We could, but then we'd also need to account for `beta` and `deprecated` exercises that the user had started. I think that the current code will match some more people than being strict, but the actual badge awarding will take care of that I think.

iHiD:
OK. Good :)

PR #2587, opened 2022-04-08
Use x:rep label instead of x:size for reputation

ErikSchierboom:
Use `x:rep` label to determine reputation for issues and pull requests.
For backwards compatibility, we'll still support the `x:size` label for pull requests.

PR #2588, opened 2022-04-08
Update gems

iHiD:
This should deal with all the dependabot PRs we have.


ErikSchierboom (approved):
Looks good.

One thing to consider is to update the `mysql2` gem, which helped solved a M1 MacBook issue for me (see https://github.com/exercism/website/pull/2587/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fR12).

PR #2616, opened 2022-04-14
Support any markdown within hints

ErikSchierboom:
This PR adds support for any Markdown content to be used within hints.
As such, it also fixes the issue that nested list items would not be rendered.

Currently, we require the hints to all be list items, which is a bit limiting in terms of how the hints can be displayed visually.
For example, the admonitions that we support via `~~~exercism/note` and its siblings can't be used, but that could be really useful.
In some cases, people have also expressed a desire to work with paragraph in the notes, as that might provide for a better reading experience.

This PR aims to fix this by supporting _any_ Markdown content within the hints.
The main thing that needs to be done is to update the styling to work well with the different types of Markdown.
There is a bit commented code in app/serializers/serialize_exercise_assignment.rb:86 that can be used to force the hints contents to be what is specified there, which is really helpful when testing.

### DIFF
--- a/app/serializers/serialize_exercise_assignment.rb
+++ b/app/serializers/serialize_exercise_assignment.rb
@@ -83,6 +83,7 @@ class SerializeExerciseAssignment
 
   memoize
   def hints_doc
+    # TODO: remove
     hints = <<~HINTS.strip
       # Hints
 
--- a/app/serializers/serialize_exercise_assignment.rb
+++ b/app/serializers/serialize_exercise_assignment.rb
@@ -42,16 +42,14 @@ class SerializeExerciseAssignment
     #  ## Header
     #  - foo
     #  - bar
-    hints_doc.each_cons(2).each_with_object({}) do |(header, list), hints|
-      # TODO: Add an issue to the relevant track via a async job
-      # if any of these are invalid
-      next unless header.type == :header
-      next unless header.header_level == 2
-      next unless list.type == :list
 
-      task_id = parse_task_id(header)
-      hints[task_id] = list.each.map { |list_item| Markdown::Parse.(list_item.each.first.to_commonmark) }
-    end
+    hints_doc.
+      each.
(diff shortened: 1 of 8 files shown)


test/serializers/serialize_exercise_assignment_test.rb
```diff
@@ -7,8 +7,11 @@ class SerializeExerciseAssignmentTest < ActiveSupport::TestCase
     serialized = SerializeExerciseAssignment.(solution)
 
     expected = [
-      "<p>The <a href=\"http://ruby-for-beginners.rubymonstas.org/built_in_classes/strings.html\" target=\"_blank\" rel=\"noopener\">rubymostas strings guide</a> has a nice\nintroduction to Ruby strings.</p>\n", # rubocop:disable Layout/LineLength
-      "<p>The <code>String</code> object has many useful <a href=\"https://ruby-doc.org/core-2.7.0/String.html\" target=\"_blank\" rel=\"noopener\">built-in methods</a>.</p>\n" # rubocop:disable Layout/LineLength
+      <<~HINT
```
ErikSchierboom, line 10: I've converted the expected string values to HEREDOC where appropriate, which improves readability IMHO.

app/serializers/serialize_exercise_assignment.rb
```diff
@@ -42,16 +42,26 @@ def hints
...
-      next unless header.type == :header
-      next unless header.header_level == 2
-      next unless list.type == :list
 
-      task_id = parse_task_id(header)
-      hints[task_id] = list.each.map { |list_item| Markdown::Parse.(list_item.each.first.to_commonmark) }
-    end
+    hints_doc.
+      each.
+      # Create chunks of nodes whenever the node types is a level 2 heading
```
ErikSchierboom, line 48: As the code is not obvious, I've added a comment to explain its working. Let me know if this comment is clear enough.

iHiD (approved):

PR #2618, opened 2022-04-15
Support passing parameters to badges

ErikSchierboom:
This PR takes a stab at implementing the TODO described at https://github.com/exercism/website/blob/main/app/commands/solution/complete.rb#L16-L18

At its core, the TODO is about optimizing the badge awarding, where we don't do (expensive) queries if we don't need to.
One example being that we only need to check whether the `whatever` should be awarded if the completed exercise is `bob`.
When completing any other exercise, this check should just be skipped.

Implementation-wise, we still want the `AwardBadgeJob` call to not contain any logic.
My proposed solution is to have a "context" that can be passed to the awarding of badges.
This context that the form of keywords, which are passed onto the badge's `award_to?` method.

I'll leave some comments on the source code to help clarify things.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,8 +29,7 @@ gem 'rest-client' # http
 gem 'octokit' # GitHub
 
 # General-purpose gems
-# gem 'mandate', '1.0.0.beta1'
-gem 'mandate', path: '../../mandate'
+gem 'mandate', '2.0.0'
 gem 'kaminari'
 gem 'oj'
 
@@ -55,7 +54,7 @@ gem 'devise', '~> 4.7'
 
 # Omniauth
 gem 'omniauth-github'
-gem "omniauth-rails_csrf_protection"
+gem 'omniauth-rails_csrf_protection'
 
 # Payments
 gem 'stripe'
--- a/Gemfile
+++ b/Gemfile
@@ -35,7 +35,8 @@ gem 'kaminari'
 gem 'oj'
 
 # Setup dependencies
-gem 'exercism-config', '>= 0.83.0' # gem 'exercism-config', path: '../exercism_config'
+gem 'exercism-config', '>= 0.83.0'
+# gem 'exercism-config', path: '../exercism_config'
 
 # Model-level dependencies
 gem 'image_processing', '~> 1.2'
@@ -54,7 +55,7 @@ gem 'devise', '~> 4.7'
 
 # Omniauth
 gem 'omniauth-github'
-gem 'omniauth-rails_csrf_protection'
+gem "omniauth-rails_csrf_protection"
 
 # Payments
 gem 'stripe'
--- a/Gemfile
+++ b/Gemfile
@@ -29,13 +29,13 @@ gem 'rest-client' # http
 gem 'octokit' # GitHub
 
 # General-purpose gems
-gem 'mandate', '1.0.0.beta1'
+# gem 'mandate', '1.0.0.beta1'
+gem 'mandate', path: '../../mandate'
 gem 'kaminari'
 gem 'oj'
 
 # Setup dependencies
-gem 'exercism-config', '>= 0.83.0'
-# gem 'exercism-config', path: '../exercism_config'
+gem 'exercism-config', '>= 0.83.0' # gem 'exercism-config', path: '../exercism_config'
 
 # Model-level dependencies
 gem 'image_processing', '~> 1.2'
@@ -54,7 +54,7 @@ gem 'devise', '~> 4.7'
 
 # Omniauth
 gem 'omniauth-github'
-gem "omniauth-rails_csrf_protection"
+gem 'omniauth-rails_csrf_protection'
 
 # Payments
 gem 'stripe'
--- a/app/commands/solution/complete.rb
+++ b/app/commands/solution/complete.rb
@@ -14,10 +14,10 @@ class Solution
       end
 
       %i[anybody_there all_your_base whatever lackadaisical].each do |badge|
-        AwardBadgeJob.perform_later(user, badge, exercise_slug: exercise.slug)
+        AwardBadgeJob.perform_later(user, badge, context: exercise)
       end
 
-      AwardBadgeJob.perform_later(user, :conceptual, exercise_type: exercise.type)
+      AwardBadgeJob.perform_later(user, :conceptual, context: exercise)
       AwardBadgeJob.perform_later(user, :completer)
 
       record_activity!
--- a/app/commands/solution/complete.rb
+++ b/app/commands/solution/complete.rb
@@ -14,12 +14,11 @@ class Solution
       end
 
       %i[anybody_there all_your_base whatever lackadaisical].each do |badge|
-        AwardBadgeJob.perform_later(user, badge, exercise: exercise.slug)
+        AwardBadgeJob.perform_later(user, badge, exercise_slug: exercise.slug)
       end
 
-      %i[completer conceptual].each do |badge|
-        AwardBadgeJob.perform_later(user, badge)
-      end
+      AwardBadgeJob.perform_later(user, :conceptual, exercise_type: exercise.type)
+      AwardBadgeJob.perform_later(user, :completer)
 
       record_activity!
     end
--- a/app/commands/solution/complete.rb
+++ b/app/commands/solution/complete.rb
@@ -13,12 +13,14 @@ class Solution
         solution.update!(completed_at: Time.current)
       end
 
-      # TODO: Think about how we can add a guard here for hello-world so
-      # we only go through the checks in the job if the exercise is hello-world.
-      # But at the job level, not at this level. So passing some args etc?
-      %i[anybody_there all_your_base whatever lackadaisical completer conceptual].each do |badge|
+      %i[anybody_there all_your_base whatever lackadaisical].each do |badge|
+        AwardBadgeJob.perform_later(user, badge, exercise: exercise.slug)
+      end
+
+      %i[completer conceptual].each do |badge|
         AwardBadgeJob.perform_later(user, badge)
       end
+
       record_activity!
     end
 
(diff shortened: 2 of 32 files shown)


Gemfile
```diff
@@ -29,13 +29,13 @@ gem 'rest-client' # http
 gem 'octokit' # GitHub
 
 # General-purpose gems
-gem 'mandate', '1.0.0.beta1'
+# gem 'mandate', '1.0.0.beta1'
+gem 'mandate', path: '../../mandate'
```
ErikSchierboom: This PR requires the https://github.com/iHiD/mandate/pull/15 functionality in Mandate, as otherwise kwargs are not handled correctly.
Of course, we should remove this path if we agree that it is a sensible PR to merge into Mandate.

app/commands/solution/complete.rb
```diff
@@ -17,7 +17,7 @@ def call
       # we only go through the checks in the job if the exercise is hello-world.
       # But at the job level, not at this level. So passing some args etc?
       %i[anybody_there all_your_base whatever lackadaisical].each do |badge|
-        AwardBadgeJob.perform_later(user, badge)
+        AwardBadgeJob.perform_later(user, badge, exercise: exercise.slug)
```
ErikSchierboom: The magic here is the passing of the `exercise` keyword. 
All the badges that are to be awarded here are specific to an exercise, so with just the passing of the exercise slug they have enough information to short-circuit their behavior.

app/models/badges/all_your_base_badge.rb
```diff
@@ -5,7 +5,9 @@ class AllYourBaseBadge < Badge
       'all-your-base',
       'Completed the "All Your Base" exercise'
 
-    def award_to?(user)
+    def award_to?(user, exercise:)
+      return false unless exercise == 'all-your-base'
```
ErikSchierboom: The main downside of the current setup is that it requires the `exercise` keyword argument to be passed to the `AwardBadgeJob.perform_later` method call, which is not something that is obvious. We _could_ consider making the `exercise` keyword have a `nil` value by default, but I opted for not doing that 🤷 
Let me know what you think (pending this decision, I've not yet updated the unit tests, so they'll be failing).

iHiD:
```ruby
def self.worth_queuing?(exercise: nil)
  exercise == 'all-your-base'
end
```

iHiD:
```ruby
class AwardBadgeJob
  self.perform_later(user, badge_slug, send_email: true, **context)
    # Where do you convert badge_slug to Badge object?
    badge = badge_slug.camelize.constantize
    return unless badge.worth_queuing?(**context)

    super(user, badge_slug, send_email: true, **context)
  end
end
```

app/models/badge.rb
```diff
@@ -48,7 +48,12 @@ def send_email_on_acquisition?
   # notifications when they are created
   def notification_key; end
 
-  def award_to?(_user, **_context)
+  # Stub to allow badges to short-circuit queueing
+  def self.worth_queuing?(**_context)
```
iHiD, line 52:
```suggestion
  def self.worth_queuing?(...)
```

Neater?

app/models/badge.rb
```diff
@@ -48,7 +48,12 @@ def send_email_on_acquisition?
...
+    true
+  end
+
+  def award_to?(_user)
```
iHiD, line 56:
```suggestion
  def award_to?(...)
```

?

app/models/badges/all_your_base_badge.rb
```diff
@@ -5,9 +5,11 @@ class AllYourBaseBadge < Badge
       'all-your-base',
       'Completed the "All Your Base" exercise'
 
-    def award_to?(user, exercise:)
-      return false unless exercise == 'all-your-base'
+    def self.worth_queuing?(context)
```
iHiD:
```suggestion
    def self.worth_queuing?(exercise:)
```

I decided I liked this in the end. I actually think its much safer than using hash params.

app/models/badge.rb
ErikSchierboom: Ehm, I sort of like `_user` as it gives the reader a better indication of the interface of the method (which arguments can be expected to be passed to it)

app/models/badge.rb
ErikSchierboom: Ehm, I sort of like `**_context` as it gives the reader a better indication of the interface of the method (which arguments can be expected to be passed to it)

app/models/badges/all_your_base_badge.rb
ErikSchierboom: Great. I'll update

app/commands/solution/complete.rb
```diff
@@ -13,12 +13,14 @@ def call
...
 
-      # TODO: Think about how we can add a guard here for hello-world so
-      # we only go through the checks in the job if the exercise is hello-world.
-      # But at the job level, not at this level. So passing some args etc?
-      %i[anybody_there all_your_base whatever lackadaisical completer conceptual].each do |badge|
+      %i[anybody_there all_your_base whatever lackadaisical].each do |badge|
+        AwardBadgeJob.perform_later(user, badge, exercise: exercise.slug)
+      end
+
+      %i[completer conceptual].each do |badge|
```
ErikSchierboom: These badges don't take an `exercise` keyword argument, as they're not exercise-specific

app/jobs/award_badge_job.rb
```diff
@@ -3,7 +3,17 @@ class AwardBadgeJob < ApplicationJob
...
+
+    super(user, badge_slug, send_email:)
+  end
+
   def perform(user, badge_slug, send_email: true)
     User::AcquiredBadge::Create.(user, badge_slug, send_email:)
   end
+
+  def self.badge(badge_slug)
+    "Badges::#{badge_slug.to_s.camelize}Badge".constantize
```
ErikSchierboom: So Ruby-ish!

app/jobs/award_badge_job.rb
```diff
@@ -16,4 +16,10 @@ def perform(user, badge_slug, send_email: true)
   def self.badge(badge_slug)
     "Badges::#{badge_slug.to_s.camelize}Badge".constantize
   end
+
+  def self.worth_queuing_context(context)
+    return {} if context.nil?
+
+    { context.class.base_class.name.demodulize.underscore => context }.symbolize_keys
```
ErikSchierboom: I need the `base_class` name otherwise I would never get `reputation_token` as the keyword, as it would be `code_merge_token` for one case, but `exercise_author_token` for another. 

The `demodulize` is used to remove any namespace (e.g. `User::ReputationToken` becomes `ReputationToken`).

app/models/badges/new_years_resolution_badge.rb
```diff
@@ -5,6 +5,10 @@ class NewYearsResolutionBadge < Badge
       'new-years-resolution',
       'Submitted a solution on January 1st'
 
+    def self.worth_queuing?
+      Time.now.utc.yday == 1
+    end
```
iHiD, line 10: I think this should take the solution.

app/jobs/award_badge_job.rb
```diff
@@ -3,7 +3,23 @@ class AwardBadgeJob < ApplicationJob
 
   discard_on BadgeCriteriaNotFulfilledError
 
+  def self.perform_later(user, badge_slug, send_email: true, context: nil)
+    return unless badge(badge_slug).worth_queuing?(**worth_queuing_context(context))
```
iHiD: If context is `blank?`, we shouldn't run this line.

app/jobs/award_badge_job.rb
```diff
@@ -3,7 +3,23 @@ class AwardBadgeJob < ApplicationJob
...
   def perform(user, badge_slug, send_email: true)
     User::AcquiredBadge::Create.(user, badge_slug, send_email:)
   end
+
+  def self.badge(badge_slug)
+    "Badges::#{badge_slug.to_s.camelize}Badge".constantize
+  end
+
+  def self.worth_queuing_context(context)
+    return {} if context.nil?
```
ErikSchierboom: Move to L7

app/jobs/award_badge_job.rb
```diff
@@ -3,7 +3,21 @@ class AwardBadgeJob < ApplicationJob
 
   discard_on BadgeCriteriaNotFulfilledError
 
+  def self.perform_later(user, badge_slug, send_email: true, context: nil)
+    return unless self.worth_queuing?(badge_slug, context:)
```
ErikSchierboom, line 7: I've refactored slightly

app/jobs/award_badge_job.rb
```diff
@@ -3,7 +3,21 @@ class AwardBadgeJob < ApplicationJob
...
+
+    super(user, badge_slug, send_email:)
+  end
+
   def perform(user, badge_slug, send_email: true)
     User::AcquiredBadge::Create.(user, badge_slug, send_email:)
   end
+
+  def self.worth_queuing?(badge_slug, context:)
+    return true if context.nil?
```
iHiD:
```suggestion
    return true unless context.present?
```

iHiD: `content: {}`

iHiD (approved):

app/jobs/award_badge_job.rb
```diff
@@ -3,7 +3,21 @@ class AwardBadgeJob < ApplicationJob
...
+
   def perform(user, badge_slug, send_email: true)
     User::AcquiredBadge::Create.(user, badge_slug, send_email:)
   end
+
+  def self.worth_queuing?(badge_slug, context:)
+    return true if context.blank?
+
+    context_key = context.class.base_class.name.demodulize.underscore
+    badge = "Badges::#{badge_slug.to_s.camelize}Badge".constantize
```
SleeplessByte, line 20: You probably want to whitelist this, for the off-chance that `badge_slug` is malicious. It's hard to break, but I cannot guarantee you cannot.

If you don't want this to error if it doesn't exist at all, use [safe_constantize](https://apidock.com/rails/ActiveSupport/Inflector/safe_constantize) to make it return nil instead.

PR #2621, opened 2022-04-19
Submission analysis status takes into account has_representer and has_analyzer

### LINKED ISSUE #6325: Excessive time to analyze submissions for Java
Hello! Code analysis seems to be stuck for any exercise that has been submitted on the Java tack in the last few days. Each submission passes all the tests, but remains "Processing" seemingly indefinitely. 

processing

ErikSchierboom:
The pending analysis status did not account for the fact that a `not_queued` status for the representation or analysis could simply be a result of the track not having a representer and/or analyzer.

Closes https://github.com/exercism/exercism/issues/6325

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -68,6 +68,10 @@ class Submission < ApplicationRecord
     return false if (representation_exceptioned? || representation_cancelled?) &&
                     (analysis_exceptioned? || analysis_cancelled?)
     return false if representation_not_queued? && analysis_not_queued?
+
+    # TODO: materialize the track's has_representer and has_analyzer status to the
+    # submission when it is created, and then use those here instead of the track's
+    # status fields
     return true if !representation_generated? && !analysis_completed? && track.has_representer? && track.has_analyzer?
     return false if has_automated_feedback?
     return true if (representation_queued? || representation_not_queued?) && track.has_representer?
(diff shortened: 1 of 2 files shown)


app/models/submission.rb
```diff
@@ -68,11 +68,10 @@ def automated_feedback_pending?
...
                     (analysis_exceptioned? || analysis_cancelled?)
     return false if representation_not_queued? && analysis_not_queued?
-    return true if !representation_generated? && !analysis_completed?
-
+    return true if !representation_generated? && !analysis_completed? && track.has_representer? && track.has_analyzer?
     return false if has_automated_feedback?
-    return true if representation_queued? || representation_not_queued?
-    return true if analysis_queued? || analysis_not_queued?
+    return true if (representation_queued? || representation_not_queued?) && track.has_representer?
+    return true if (analysis_queued? || analysis_not_queued?) && track.has_analyzer?
```
ErikSchierboom, line 78: The main downside of using `track.has_representer?` and `track.has_analyzer?` is that if these values ever switch from `false` to `true`, the UI will think an analysis is being done in the background, whereas in fact it won't.

Some possible solutions are:

1. We detect changes to `track.has_analyzer` and `track.has_representer` and automatically queue the corresponding jobs 
2. We store the `has_representer` and `has_analyzer` status on the submission
3. We add an extra `skipped` value to the `representation_status` and `analysis_status` field

iHiD (approved):

PR #2628, opened 2022-04-26
Support non-editor filepaths to be important filepaths

### LINKED ISSUE #6332: When/How are exercises updated to new versions?
(I'm a maintainer on the Cpp track)

We had an issue with the online editor for 2 exercises, which required a change to a build file to fix (https://github.com/exercism/cpp/pull/495). Some users are still seeing issues even though others seem to be able to use it just fine (https://github.com/exercism/cpp-test-runner/issues/48)

What I _think_ is happening is that new users have no issues solving it, however if a user has already submitted a solution there doesn't seem to be a way to update to the latest version. My account has submitted one of those exercises but not the other so I can see both of those states on the website, and I confirmed with a test account that the exercises work fine for a new user.

I don't really understand how the process of updating a users exercise version actually works and I couldn't find a description in the `exercism/docs` repo, so I'm not sure what's going on. There used to be a way to update the version on the website but the option in the editor to "Revert to exercise start" doesn't seem to update the build file.

My wild guess is that existing exercises aren't updated to a new commit unless a file listed in the `.meta/config.json` file is updated, but newcomers to the exercise get the latest version. The build file that we needed to change isn't listed in the `config.json`.

Q1: Is there still a way to force update a users exercise to the latest version?
Q2: How can we make sure that updates to the build files cause an update to the user's existing solutions?

ErikSchierboom:
This PR adds support for allowing files that are not displayed in the editor to be accounted for when determining whether an exercise needs an update. 

Closes https://github.com/exercism/exercism/issues/6332


app/models/git/exercise.rb
```diff
@@ -195,7 +200,8 @@ def important_filepaths
         introduction_append_filepath,
         hints_filepath,
         *test_filepaths,
-        *editor_filepaths
+        *editor_filepaths,
+        *invalidator_filepaths
```
ErikSchierboom, line 204: Including the invalidator filepaths in this list causes them to be included in the important files hash

test/models/git/exercise_test.rb
```diff
@@ -18,6 +18,48 @@ class ExerciseTest < ActiveSupport::TestCase
...
+      expected = [".meta/example.rb"]
+      assert_equal(expected, exercise.example_filepaths)
+    end
+
+    test "exemplar_filepaths" do
+      exercise = Git::Exercise.new(:booleans, "concept", "HEAD",
+        repo_url: TestHelpers.git_repo_url("track-with-exercises"))
+      expected = [".meta/exemplar.rb"]
+      assert_equal(expected, exercise.exemplar_filepaths)
+    end
```
ErikSchierboom, line 61: Weirdly enough, these methods weren't tested yet, so I've added them

iHiD (approved):
Let's go!

PR #2634, opened 2022-04-28
Allow admins to see discussion

ErikSchierboom:


app/controllers/mentoring/discussions_controller.rb
```diff
@@ -17,6 +17,6 @@ def use_mentor_discussion
     @mentor_discussion = Mentor::Discussion.find_by!(uuid: params[:id])
 
     # TODO: (Required) Do we want an unauthorised page here?
-    redirect_to mentoring_path unless @mentor_discussion.mentor_id == current_user.id
+    redirect_to mentoring_path unless @mentor_discussion.viewable_by_mentor?(current_user)
```
ErikSchierboom, line 20: I've created a new `viewable_by_mentor?` method on the `Discussion` class, as I could not re-use the `viewable_by?` method as that would also return `true` if the current user was the student. I'm not entirely happy with the new method's name though. Let me know what you think.

iHiD (approved):

app/controllers/api/mentoring/discussion_posts_controller.rb
```diff
@@ -58,7 +58,7 @@ def destroy
     def use_mentor_discussion
       @discussion = Mentor::Discussion.find_by(uuid: params[:discussion_uuid])
       return render_404(:mentor_discussion_not_found) unless @discussion
-      return render_403(:mentor_discussion_not_accessible) unless @discussion.mentor_id == current_user.id
+      return render_403(:mentor_discussion_not_accessible) unless @discussion.viewable_by_mentor?(current_user)
```
ErikSchierboom, line 61: This ensure that admins can see the mentor discussion's posts

app/models/submission.rb
```diff
@@ -97,6 +97,9 @@ def viewable_by?(user)
     # A user can always see their own stuff
     return true if solution.user == user
 
+    # Admins can always see submissions
+    return true if user&.admin?
```
ErikSchierboom, line 101: `&.` is used as we're testing with `nil` values for the user.

test/system/flows/admin/admin_views_mentor_discussion_test.rb
```diff
@@ -0,0 +1,34 @@
+require "application_system_test_case"
+require_relative "../../../support/capybara_helpers"
+
+module Flows
+  module Admin
+    class AdminViewsMentorDiscussionTest < ApplicationSystemTestCase
+      include CapybaraHelpers
+
+      test "admin views mentor discussion but is not student nor mentor" do
```
ErikSchierboom, line 9: I've added a system test to verify that admins can view mentor discussions. I should probably also add this for viewing exercises at some point.

For now, I've put this test in a separate `admin` directory (we already have `mentor` and `student` directories). Let me know if this is fine or if you'd like to see it moved.

iHiD (approved):

PR #2639, opened 2022-05-03
Use safe_constantize to mitigate attack vector for dynamic badge slug

ErikSchierboom:


app/jobs/award_badge_job.rb
```diff
@@ -17,7 +17,7 @@ def self.worth_queuing?(badge_slug, context:)
     return true if context.blank?
 
     context_key = context.class.base_class.name.demodulize.underscore
-    badge = "Badges::#{badge_slug.to_s.camelize}Badge".constantize
+    badge = "Badges::#{badge_slug.to_s.camelize}Badge".safe_constantize
```
ErikSchierboom, line 20: @SleeplessByte I've opted for the `safe_constantize` option, as the `badge.worth_queuing?` call on the next line will then fail with a nil reference error, which we'll automatically log.

iHiD (approved):

iHiD:
So I'm clear, do we have an example of what this protects us against?

ErikSchierboom:
> So I'm clear, do we have an example of what this protects us against? 

@SleeplessByte ?

SleeplessByte:
I'd need to see if I can craft an attack. I have no proof at the moment that it's possible, but whatever you do, whenever you call `constantize` it is important that you whitelist (which this PR does not do). A next refactor may remove `Badge::` from the string, and then it becomes "dangerous".

ErikSchierboom:
It sounds like it requires us to merge code before there is an attack vector? Is that correct?

SleeplessByte:
I think so, yes. Which is good!

PR #2641, opened 2022-05-04
Update placeholder to mention searching by track name is allowed

ErikSchierboom:


app/javascript/components/mentoring/queue/ExerciseFilterList.tsx
```diff
@@ -153,7 +150,7 @@ const Component = ({ exercises, value, setValue }: Props): JSX.Element => {
             value={searchQuery}
             onChange={handleSearchBarChange}
             className="--search"
-            placeholder="Search by Exercise name"
+            placeholder="Search by exercise or track name"
```
iHiD: I'm presuming so, but is this true for this mentoring component too?

ErikSchierboom: It is not, I tested on the wrong page. Doh

iHiD (approved):
Let's go!

PR #2642, opened 2022-05-04
Return 404 for missing docs

### LINKED ISSUE #6341: Docs with a slug that doesn't exist result in 500 instead of 404
Correct (anything but docs): https://exercism.org/tracks/surely-this-does-not-exist

Incorrect: https://exercism.org/docs/xvcbcvb

@ErikSchierboom 

Related: #5680

ErikSchierboom:
This PR ensures that a 404 is rendered when a document is not found, instead of a 500 (which would 
also result in a bugsnag entry I believe).
Whilst implementing this, I found that unit tests were a bit lacking, so I added some.

- Rename tests for docs controller
- Add tests for docs section
- Add tests for docs tracks section
- Add tests for docs track index section
- Add tests for docs track doc section
- Add 404 tests to docs controller tests
- Simplify code a bit

Closes https://github.com/exercism/exercism/issues/6341

### DIFF
--- a/app/controllers/docs_controller.rb
+++ b/app/controllers/docs_controller.rb
@@ -14,7 +14,7 @@ class DocsController < ApplicationController
   end
 
   def show
-    @doc = Document.where(section: @section).find(params[:slug])
+    @doc = @nav_docs.find(params[:slug])
     render action: :show
   rescue ActiveRecord::RecordNotFound
     render_404
@@ -25,7 +25,7 @@ class DocsController < ApplicationController
   def track_index; end
 
   def track_show
-    @doc = Document.where(track: @track).find(params[:slug])
+    @doc = @nav_docs.find(params[:slug])
     @section = :tracks
     render action: :show
   rescue ActiveRecord::RecordNotFound
--- a/app/controllers/docs_controller.rb
+++ b/app/controllers/docs_controller.rb
@@ -9,15 +9,11 @@ class DocsController < ApplicationController
   def section
     @doc = Document.find_by!(section: @section, slug: "APEX")
     render action: :show
-  rescue ActiveRecord::RecordNotFound
-    render_404
   end
 
   def show
     @doc = @nav_docs.find(params[:slug])
     render action: :show
-  rescue ActiveRecord::RecordNotFound
-    render_404
   end
 
   def tracks; end
@@ -28,8 +24,6 @@ class DocsController < ApplicationController
     @doc = @nav_docs.find(params[:slug])
     @section = :tracks
     render action: :show
-  rescue ActiveRecord::RecordNotFound
-    render_404
   end
 
   private
@@ -43,7 +37,5 @@ class DocsController < ApplicationController
     @nav_docs = Document.where(track_id: @track.id)
 
     render_404 unless @track.accessible_by?(current_user)
-  rescue ActiveRecord::RecordNotFound
-    render_404
   end
 end
--- a/app/controllers/docs_controller.rb
+++ b/app/controllers/docs_controller.rb
@@ -7,13 +7,17 @@ class DocsController < ApplicationController
   def index; end
 
   def section
-    @doc = Document.find_by(section: @section, slug: "APEX")
+    @doc = Document.find_by!(section: @section, slug: "APEX")
     render action: :show
+  rescue ActiveRecord::RecordNotFound
+    render_404
   end
 
   def show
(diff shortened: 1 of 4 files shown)


app/controllers/docs_controller.rb
```diff
@@ -7,23 +7,29 @@ class DocsController < ApplicationController
...
-    @doc = Document.find_by(section: @section, slug: "APEX")
+    @doc = Document.find_by!(section: @section, slug: "APEX")
     render action: :show
+  rescue ActiveRecord::RecordNotFound
+    render_404
   end
 
   def show
-    @doc = Document.where(section: @section).find(params[:slug])
+    @doc = @nav_docs.find(params[:slug])
```
ErikSchierboom, line 15: `@nav_docs` has the same code as `Document.where(section: @section)` and is set in `use_section`, which is automatically called.

app/controllers/docs_controller.rb
```diff
@@ -7,23 +7,29 @@ class DocsController < ApplicationController
...
+    render_404
   end
 
   def tracks; end
 
   def track_index; end
 
   def track_show
-    @doc = Document.where(track: @track).find(params[:slug])
+    @doc = @nav_docs.find(params[:slug])
```
ErikSchierboom, line 24: Same as above, but with `use_track` instead of `use_section`.

app/controllers/docs_controller.rb
```diff
@@ -7,23 +7,29 @@ class DocsController < ApplicationController
   def index; end
 
   def section
-    @doc = Document.find_by(section: @section, slug: "APEX")
+    @doc = Document.find_by!(section: @section, slug: "APEX")
```
ErikSchierboom, line 10: I've changed to `find_by!` to raise an error when the document is not found.

app/controllers/docs_controller.rb
```diff
@@ -7,23 +7,29 @@ class DocsController < ApplicationController
...
     render action: :show
+  rescue ActiveRecord::RecordNotFound
+    render_404
```
iHiD: Out of interest, do you need this? I'd have guessed Rails would gracefully handle the `ActiveRecord::RecordNotFound` to show a 404.

app/controllers/docs_controller.rb
```diff
@@ -7,23 +7,29 @@ class DocsController < ApplicationController
...
   end
 
   def show
-    @doc = Document.where(section: @section).find(params[:slug])
+    @doc = @nav_docs.find(params[:slug])
     render action: :show
+  rescue ActiveRecord::RecordNotFound
+    render_404
```
iHiD: Same question.

app/controllers/docs_controller.rb
ErikSchierboom: You are correct. I've checked this by changing `config.consider_all_requests_local = true` to `config.consider_all_requests_local = false` in `config/environments/development.rb`.

The only "downside" is that the tests will change from

```ruby
get docs_section_path(:mentoring)
assert_rendered_404
```

to:

```ruby
assert_raises ActiveRecord::RecordNotFound do
  get docs_section_path(:mentoring)
end
```

I like the former a bit better.

iHiD: So, I think it's definitely not idiomatic to be catching ActiveRecord::RecordNotFound and reraising as 404s, so we shouldn't do it.

I agree about the tests though. I'm surprised the first doesn't work tbh. I'd have thought it would have done reflected the production output more. But I think we probably just have to live with the uglier tests.

ErikSchierboom: > I agree about the tests though. I'm surprised the first doesn't work tbh.

Well, that's because `environments/test.rb` has the same define:

```ruby
config.consider_all_requests_local = true
```

Do we want that?

iHiD: Yeah, I imagine we should leave it as it is. But I just figured Rails would be smarter, despite that :)

config/environments/development.rb
```diff
@@ -19,7 +19,7 @@
   config.eager_load = false
 
   # Show full error reports.
-  config.consider_all_requests_local = true
+  config.consider_all_requests_local = false
```
iHiD: Revert this.

ErikSchierboom:
```suggestion
  config.consider_all_requests_local = true
```

iHiD (approved):

PR #2643, opened 2022-05-04
Use correct exercise count in metadata

### LINKED ISSUE #5711: Embed shows wrong number of exercises
If I post https://exercism.org/tracks/julia in a context where embeds are enabled, such as Slack, it shows a different number of exercises than on the track page itself:

grafik

The website states 56.

I'm guessing this might be caused by concept exercises being disabled that are incorrectly counted for the embed.

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/5711

I considered adding tests, but I was unsure how much value there would be in doing that.

### DIFF
--- a/app/helpers/meta_tags_helper.rb
+++ b/app/helpers/meta_tags_helper.rb
@@ -17,10 +17,10 @@ module MetaTagsHelper
     content_for?(:meta_url) ? content_for(:meta_url) : request.original_url.gsub(%r{/$}, "")
   end
 
-  def track_meta_tags(track)
-    content_for :meta_title, "#{track.title} on Exercism"
+  def track_meta_tags(user_track)
+    content_for :meta_title, "#{user_track.track.title} on Exercism"
     content_for :meta_description,
-      "Get fluent in #{track.title} by solving #{track.num_exercises} exercises. And then level up with mentoring from our world-class team." # rubocop:disable Layout/LineLength
+      "Get fluent in #{user_track.track.title} by solving #{user_track.num_exercises} exercises. And then level up with mentoring from our world-class team." # rubocop:disable Layout/LineLength
   end
 
   def concept_meta_tags(concept, user_track)
--- a/app/helpers/meta_tags_helper.rb
+++ b/app/helpers/meta_tags_helper.rb
@@ -18,20 +18,20 @@ module MetaTagsHelper
   end
 
   def track_meta_tags(user_track)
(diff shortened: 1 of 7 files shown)


app/views/tracks/about.html.haml
```diff
@@ -1,4 +1,4 @@
-- track_meta_tags(@track)
+- track_meta_tags(@user_track)
```
iHiD, line 1: I'd have thought this would break if you haven't joined the track?

ErikSchierboom: No, because then `@user_track` will be `UserTrack::External`. See https://github.com/exercism/website/blob/main/app/controllers/tracks_controller.rb#L51 for  the call to create the user track, and https://github.com/exercism/website/blob/0f0a8ba30660cdfcf17b66e9bb6060821e3ceeb8/app/models/user_track.rb#L38 for its implementation. The user track will only ever be `nil` if the track can't be found, but that is already covered in https://github.com/exercism/website/blob/main/app/controllers/tracks_controller.rb#L50

ErikSchierboom: Oh, and we're already using the user track to display the exercises on that page: https://github.com/exercism/website/blob/main/app/views/tracks/about.html.haml#L17

iHiD: Ah nice!

app/helpers/meta_tags_helper.rb
```diff
@@ -17,10 +17,10 @@ def meta_url
     content_for?(:meta_url) ? content_for(:meta_url) : request.original_url.gsub(%r{/$}, "")
   end
 
-  def track_meta_tags(track)
-    content_for :meta_title, "#{track.title} on Exercism"
+  def track_meta_tags(user_track)
+    content_for :meta_title, "#{user_track.track.title} on Exercism"
     content_for :meta_description,
-      "Get fluent in #{track.title} by solving #{track.num_exercises} exercises. And then level up with mentoring from our world-class team." # rubocop:disable Layout/LineLength
+      "Get fluent in #{user_track.track.title} by solving #{user_track.num_exercises} exercises. And then level up with mentoring from our world-class team." # rubocop:disable Layout/LineLength
```
iHiD: Bonus points for doing adding a `track_title` method to `user_track` rather than daisy-chaining the methods. You can use delegate with `prefix` IIRC.

iHiD (approved):
Approved. I'll leave the daisy-chaining tweak up to you :)

PR #2649, opened 2022-05-05
Set page size to 24 for solutions

### LINKED ISSUE #6339: Feature Request: please make pagination a multiple of your view
Hi,

a picture is worth a thousand words:

FRQgBWnXEAArmKk

that is the pagination of my profile -> solutions. I'm not sure if this is a valid concern on all different type of screens/resolutions, but it would make sense for pagination to be a multiple of the solutions "view". In this case 24 vs 25 would result in an evenly distributed number of "cards".

Thank you :)

ErikSchierboom:
Setting the page size to 24 allows the displayed items to be aligned in a grid of size 1, 2, 3, 4, 6, 12 (or 24).

Closes https://github.com/exercism/exercism/issues/6339

### DIFF
--- a/app/commands/solution/search_community_solutions.rb
+++ b/app/commands/solution/search_community_solutions.rb
@@ -3,7 +3,7 @@ class Solution
     include Mandate
 
     DEFAULT_PAGE = 1
-    DEFAULT_PER = 30
+    DEFAULT_PER = 24
     MAX_ROWS = 10_000
 
     def self.default_per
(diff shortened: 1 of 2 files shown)


app/commands/solution/search_community_solutions.rb
```diff
@@ -3,7 +3,7 @@ class SearchCommunitySolutions
     include Mandate
 
     DEFAULT_PAGE = 1
-    DEFAULT_PER = 30
+    DEFAULT_PER = 24
```
ErikSchierboom, line 6: I've aligned this value (community solutions) with that of the user solutions

iHiD (approved):

PR #2657, opened 2022-05-12
Concept map shows concept exercises first

### LINKED ISSUE #5760: Learning exercise isn't always first in the concept box in the syllabus
When I look at C# logged out, I see this: the learning exercises (light bulb) aren't always the first ones listed for a concept. In Elixir, they always are. I am not aware of anything special that I have done in the Elixir config to make it happen. I thought that's what the platform does by default.

As a learner looking at the syllabus, I _should_ be guided into the learning exercise first because the other exercises in the concept should be locked anyway.

Screen Shot 2021-09-04 at 08 46 37

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/5760

### DIFF
--- a/app/commands/user_track/generate_exercise_status_mapping.rb
+++ b/app/commands/user_track/generate_exercise_status_mapping.rb
@@ -48,7 +48,7 @@ class UserTrack
             type: user_track.exercise_type(slug),
             position: user_track.exercise_position(slug)
           }
-        end.compact.sort_by { |h| h[:position] } # rubocop:disable Style/MultilineBlockChain
+        end.compact.sort_by { |h| [h[:type], h[:position]] } # rubocop:disable Style/MultilineBlockChain
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/commands/user_track/generate_exercise_status_mapping.rb
```diff
@@ -48,7 +48,7 @@ def mapping
             type: user_track.exercise_type(slug),
             position: user_track.exercise_position(slug)
           }
-        end.compact.sort_by { |h| h[:position] } # rubocop:disable Style/MultilineBlockChain
+        end.compact.sort_by { |h| [h[:type], h[:position]] } # rubocop:disable Style/MultilineBlockChain
```
ErikSchierboom, line 51: The `position` field is unique _within an exercise type_. This means that we can have a concept exercise with position 7 and a practice exercise with position 7. The issue occurs when we have a concept exercise which position is greater than the practice exercise's position.

An alternative solution would be to have the position be contiguous between concept exercises and practice exercises.

iHiD (approved):
Nice!

PR #2660, opened 2022-05-17
Add new content role

iHiD:
We will email existing applicants about this change once this is merged.

### DIFF
--- a/app/views/about/hiring_content.html.haml
+++ b/app/views/about/hiring_content.html.haml
@@ -159,18 +159,18 @@
           A link to a short video introducing yourself in more detail, telling us why you're interested joining Exercism, and giving us some more information on your experience and why you're right for the role.
           Don't worry about the quality of the video itself! Try and keep it around 5-10 minutes long. We use #{link_to 'Loom', 'https://www.loom.com'} for easy recording and sharing.
         %li
-          A sample article on a programming topic you have written that embeds the values and style we list above.
-          %em.text-italic (We will not use the article on Exercism unless we hire you)
+          A sample article on a programming topic you have written that embeds the values and style we list above
+          %em.text-italic (we will not use the article on Exercism unless we hire you).
           Everyone applying for this job will be required to write a sample article on one of the topics listed below.
           However, if you want to link us to a post you've previously written on any programming topic for this stage, that is fine.
           If you are writing something from scratch, please use one of the following topics:
-          %ul
+          %ul.mt-4
             %li Types of numbers (e.g. integers, big ints, tiny ints, floating points, decimals)
             %li Bit manipulation
             %li Exceptions
             %li Testing Randomness
 
-      %p.text-p-xlarge.mb-16
+      %p.text-p-xlarge.mb-8
         The application process will run as follows:
 
       %ul.text-p-xlarge.mb-16
--- a/app/views/about/hiring_content.html.haml
+++ b/app/views/about/hiring_content.html.haml
@@ -160,6 +160,7 @@
           Don't worry about the quality of the video itself! Try and keep it around 5-10 minutes long. We use #{link_to 'Loom', 'https://www.loom.com'} for easy recording and sharing.
         %li
           A sample article on a programming topic you have written that embeds the values and style we list above.
+          %em.text-italic (We will not use the article on Exercism unless we hire you)
           Everyone applying for this job will be required to write a sample article on one of the topics listed below.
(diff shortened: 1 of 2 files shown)


kytrinyx (approved):
This looks excellent.  We might want to mention that none of the articles they write as part of the application process will actually be used on the site.

app/views/about/hiring_content.html.haml
```diff
@@ -53,39 +54,48 @@
...
-        Discussing and exploring those ideas is a key piece of the role, and something it's important you'll enjoy.
-        At Exercism, we place a high value on
-        %strong pragmatism
-        \- the ability to balance finding the best answer with the one that makes the path forward easiest.
-        To succeed in this role it's important you value that pragmatism too and will enjoy digging into people's opinions, balancing them with your own, and finding pragmatic routes forward.
+        Central to this role is seeing content in the same way our founders do.
+        We want to create content that embodies a sense of playfulness, fun, and approachability.
+        We want our content to feel conversational, maybe even chatty - maybe more like a blog post than a formal teaching article.
+        Articles will ideally include a lot of visuals.
+        Everything we do is centred around making programming approachable, empowering, and never making anyone feel stupid.
```
ErikSchierboom, line 77: I'm still not a fan of the "stupid" part, but I also don't have a good way to phrase it differently.

app/views/about/hiring_content.html.haml
```diff
@@ -142,12 +152,22 @@
...
-      %ul.text-p-xlarge.mb-16
-        %li Subject of "Educational Content Factotum"
+      %ul.text-p-xlarge.mb-24
+        %li Subject of "Educational Content Creator"
         %li A couple of paragraphs introducing yourself and providing some brief background.
         %li
           A link to a short video introducing yourself in more detail, telling us why you're interested joining Exercism, and giving us some more information on your experience and why you're right for the role.
           Don't worry about the quality of the video itself! Try and keep it around 5-10 minutes long. We use #{link_to 'Loom', 'https://www.loom.com'} for easy recording and sharing.
+        %li
+          A sample article on a programing topic you have written that embeds the values and style we list above.
```
ErikSchierboom:
```suggestion
          A sample article on a programming topic you have written that embeds the values and style we list above.
```

ErikSchierboom (approved):

app/views/about/hiring_content.html.haml
lunohodov, line 77:
```suggestion
        Everything we do is centred around making programming approachable, empowering, and never making anyone feel lost.
```

Would this make sense?

kytrinyx: Someone suggested "lost" instead of stupid, but now I don't see the comment anymore. That could work (though I'll admit that I often feel stupid, not lost, when reading programming books).

ErikSchierboom: I'd be fine with merging this as in, and then potentially reconsidering later.

iHiD: I strongly feel that "stupid" is the correct word. I often feel stupid reading articles, especially if they're maths heavy or the author presumes I know things I don't. I really appreciate content that goes out of its way to not make me feel like that.

I have treid to add the first bit of the sentence to add the "positive" side as well as the "negative" side. Happy to revisit in the future, but I feel the vibe is right.

iHiD: Thanks for the suggestion @lunohodov. There's a longer chat about this above.

app/views/about/hiring_content.html.haml
```diff
@@ -142,12 +152,22 @@
...
-      %ul.text-p-xlarge.mb-16
-        %li Subject of "Educational Content Factotum"
+      %ul.text-p-xlarge.mb-24
+        %li Subject of "Educational Content Creator"
         %li A couple of paragraphs introducing yourself and providing some brief background.
         %li
           A link to a short video introducing yourself in more detail, telling us why you're interested joining Exercism, and giving us some more information on your experience and why you're right for the role.
           Don't worry about the quality of the video itself! Try and keep it around 5-10 minutes long. We use #{link_to 'Loom', 'https://www.loom.com'} for easy recording and sharing.
+        %li
+          A sample article on a programming topic you have written that embeds the values and style we list above.
```
iHiD:
```suggestion
          A sample article on a programming topic you have written that embeds the values and style we list above.
          %em.text-italic (We will not use the article on Exercism unless we hire you)
```

PR #2663, opened 2022-05-18
Ensure that the HOME and END keys make the cursor go to the beginning/end of the same line on which the cursor is placed.

### LINKED ISSUE #5875: Comment Editing: the end of the line
When the END key is pressed while in the Comment box the cursor goes to the end of the paragraph instead of the end of the line. This is unwelcome when editing in the middle of a paragraph.

ErikSchierboom:
https://www.loom.com/share/1aec870385494046b1503ee5f87fadc9

Closes https://github.com/exercism/exercism/issues/5875


iHiD (approved):
Nice :)

PR #2670, opened 2022-05-20
Increment num_comments on solution when a new comment is added

### LINKED ISSUE #6281: Published solution says 0 comments despite there being comments
In all places where I can see my solution it reports as 0 comments given...
https://exercism.org/profiles/SleeplessByte/solutions
https://exercism.org/tracks/javascript/exercises/poetry-club-door-policy/solutions
image

...despite there being comments.
https://exercism.org/tracks/javascript/exercises/poetry-club-door-policy/solutions/SleeplessByte
image

ErikSchierboom:
This PR fixes the `num_comments` field not having the correct count (it is always `0`).

We could update the existing solutions with the following query:

```ruby
  Solution.update_all("num_comments = (SELECT COUNT(*) FROM solution_comments WHERE solution_comments.solution_id=solutions.id)")
end
```

Closes https://github.com/exercism/exercism/issues/6281

### DIFF
--- a/app/models/solution/comment.rb
+++ b/app/models/solution/comment.rb
@@ -1,5 +1,5 @@
 class Solution::Comment < ApplicationRecord
-  belongs_to :solution
+  belongs_to :solution, counter_cache: :num_comments
   belongs_to :author, class_name: "User", foreign_key: :user_id, inverse_of: :solution_comments
 
   validates :content_markdown, presence: true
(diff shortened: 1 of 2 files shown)


app/models/solution/comment.rb
```diff
@@ -1,5 +1,5 @@
 class Solution::Comment < ApplicationRecord
-  belongs_to :solution
+  belongs_to :solution, counter_cache: :num_comments
```
ErikSchierboom, line 2: This was the culprit

iHiD:
Yep - merge and then update!

iHiD (approved):

PR #2673, opened 2022-05-25
Fix cc-by-nc-sa doc link

### LINKED ISSUE #6374: cc-sa-4 license URL is broken
After sign up, there's a page to accept the COC, licensing and privacy policies. On this page, the link to cc-sa-4 is broken, it doesn't actually open the license. Perhaps because the license is at https://exercism.org/docs/using/licenses/cc-by-nc-sa, but the link goes to https://exercism.org/licences/cc-sa-4.

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/6374

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -479,8 +479,8 @@ Rails.application.routes.draw do
 
   # Licences
   %w[licence license].each do |spelling|
-    get "#{spelling}s/cc-sa-4", to: redirect("docs/using/licenses/cc-by-nc-sa-4"), as: "cc_sa_4_#{spelling}"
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -479,8 +479,8 @@
 
   # Licences
   %w[licence license].each do |spelling|
-    get "#{spelling}s/cc-sa-4", to: redirect("docs/using/licenses/cc-by-nc-sa-4"), as: "cc_sa_4_#{spelling}"
-    get "#{spelling}s/cc-by-nc-sa-4", to: redirect("docs/using/licenses/cc-by-nc-sa-4"), as: "cc_by_nc_sa_4_#{spelling}"
```
ErikSchierboom, line 483: The docs file does not have the `-4` suffix. Neither does the `AGPL 3` link (it's just `agpl`), so I fixed it here instead of changing the filename in the docs.

iHiD (approved):
Nice. Thanks.

PR #2674, opened 2022-05-25
Use average test runner time in test results

### LINKED ISSUE #5650: Testing code to work should use average test runner time
When viewing the tests for a practice exercise that was submitted for v3, I get the following screen:

image

Maybe we should be using the test runner average number of seconds as specified in the `config.json`.

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/5650


app/controllers/api/solutions/submission_test_runs_controller.rb
```diff
@@ -10,9 +10,12 @@ def show
...
           test_run: test_run,
-          test_runner_status: {
-            exercise: submission.exercise.has_test_runner,
-            track: submission.track.has_test_runner
+          test_runner: {
+            average_test_duration: submission.track.average_test_duration,
+            status: {
+              exercise: submission.exercise.has_test_runner,
+              track: submission.track.has_test_runner
+            }
```
ErikSchierboom, line 18: I now also need to pass in the average test duration. I could have added it as just a new field, but I felt that encapsulating it in a new, top-level `test_runner` field made more sense.

app/helpers/react_components/editor.rb
```diff
@@ -27,7 +27,9 @@ def to_s
               highlightjs_language: track.highlightjs_language
             } : nil,
             results: {
-              average_test_duration: track.average_test_duration
+              test_runner: {
+                average_test_duration: track.average_test_duration
+              }
```
ErikSchierboom, line 32: This is to align with what is done for submissions above.

app/javascript/components/editor/TestRunSummary.tsx
```diff
@@ -1,25 +1,23 @@
...
-  averageTestDuration,
   showSuccessBox,
 }: {
   testRun: TestRun
-  testRunnerStatus?: TestRunnerStatus
+  testRunner: TestRunner
   onSubmit?: () => void
   isSubmitDisabled?: boolean
   onCancel?: () => void
-  averageTestDuration?: number
```
ErikSchierboom, line 22: It said here that the average test duration could be null or undefined, whereas in fact it could not be. For tracks that don't have it set explicitly in their `config.json`, a default value is used.

app/javascript/components/editor/TestRunSummary.tsx
```diff
@@ -60,7 +58,10 @@ export const TestRunSummary = ({
     <div className="automated-feedback-pending">
       <GraphicalIcon icon="spinner" />
       <h3>We&apos;re testing your code to check it works</h3>
-      <p>This usually takes 5-20 seconds.</p>
+      <p>
+        This usually takes {testRunner.averageTestDuration}-
+        {testRunner.averageTestDuration * 4} seconds.
```
ErikSchierboom, line 63: This was the original requested change :)

app/javascript/components/editor/TestRunSummary.tsx
```diff
@@ -256,18 +257,18 @@ const TestRunSummaryContent = ({
...
-                className="bar"
-                style={{ animationDuration: `${averageTestDuration}s` }}
-              />
-            ) : null}
+            <div
+              className="bar"
+              style={{
+                animationDuration: `${testRunner.averageTestDuration}s`,
+              }}
+            />
```
ErikSchierboom, line 265: As the test duration is always guaranteed to be set, we can simplify this

app/javascript/components/editor/TestRunSummary.tsx
```diff
@@ -256,18 +257,18 @@ const TestRunSummaryContent = ({
...
           </div>
           <p>
             <strong>Running tests...</strong>
-            {averageTestDuration !== undefined ? (
-              <span>Estimated running time ~ {averageTestDuration}s</span>
-            ) : null}
+            <span>
+              Estimated running time ~ {testRunner.averageTestDuration}s
+            </span>
```
ErikSchierboom, line 271: As the test duration is always guaranteed to be set, we can simplify this

app/models/track.rb
```diff
@@ -90,7 +90,7 @@ def highlightjs_language
   end
 
   def average_test_duration
-    git.average_test_duration + INFRASTRUCTURE_DURATION_S
+    git.average_test_duration.round + INFRASTRUCTURE_DURATION_S
```
ErikSchierboom, line 93: I've opted for rounding this value, as showing a test runner to run for `5.6` seconds gives a false impression of precision I feel.

iHiD (approved):
Looks great. Nice refactor!

PR #2690, opened 2022-05-31
Prevent race conditions

ErikSchierboom:
Some models used the `after_save` and `after_update` callbacks, but these could lead to race conditions as the code called
within those methods could read an outdated version of the model as there was no guarantee the changes would have been committed.

This was e.g. problematic when dealing with a solution's published head tests status, which would not be updated due to the published
solution's published_iteration_id not having yet been updated.

See https://github.com/exercism/exercism/issues/5810

### DIFF
--- a/app/models/submission/file.rb
+++ b/app/models/submission/file.rb
@@ -13,7 +13,12 @@ class Submission::File < ApplicationRecord
   # any content that has been passed in via the
   # attr_writer. We only want to do this on create
   # and never want to update or change this content.
-  after_create_commit do
+  #
+  # Note that we don't use after_create_commit as
+  # we want this to be done as quickly as possible
+  # and all data that is used is present (no risk of
+  # a race condition).
+  after_create do
     if @content
       Exercism.s3_client.put_object(
         bucket: s3_bucket,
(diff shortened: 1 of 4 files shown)


app/models/submission/file.rb
```diff
@@ -13,7 +13,7 @@ class Submission::File < ApplicationRecord
   # any content that has been passed in via the
   # attr_writer. We only want to do this on create
   # and never want to update or change this content.
-  after_create do
+  after_create_commit do
```
ErikSchierboom: I think we don't need it here, as the code inside seems to only use the current model's values. I still changed it to `after_create_commit` to prevent any possible future mistakes. Let me know what you think.

iHiD: I think its probably worth reverting this one, but adding a comment to explain why it's reverted.

We want this to run as quickly/soon as possible, and so `after_save` is the better choice. As no jobs are queued in here, and if something fails we don't mind the extra thing in s3/efs, it's safe to do.

ErikSchierboom: Okay, sounds good

ErikSchierboom: Updated

iHiD (approved):

PR #2691, opened 2022-06-01
Fix iteration publishing not showing all iterations

### LINKED ISSUE #6303: Cannot update published iteration to different iteration
image

When updating a single-published iteration to a new one, you only see that single published one in the drop down.
After selecting "all", you can select _all_ the iterations in the dropdown.

### LINKED ISSUE #6385: "Change published iterations" doesn't show all iterations
"Change published iterations" only shows the first iteration. The workaround is to unpublish and publish again.

ErikSchierboom:
This PR fixes two things:

1. There was a bug where the publishing on the community solutions page only showed a single iteration if the solution on published a single iteration
2. There was a bug the you could publish a deleted iteration

Closes https://github.com/exercism/exercism/issues/6303
Closes https://github.com/exercism/exercism/issues/6385

### DIFF
--- a/app/controllers/tracks/community_solutions_controller.rb
+++ b/app/controllers/tracks/community_solutions_controller.rb
@@ -22,6 +22,7 @@ class Tracks::CommunitySolutionsController < ApplicationController
 
     @author = @solution.user
     @comments = @solution.comments
+    @own_solution = @author == @current_user
 
     # TODO: (Required) Real algorithm here
     @other_solutions = @exercise.solutions.published.where.not(id: @solution.id).limit(3)
--- a/app/helpers/react_components/common/solution_view.rb
+++ b/app/helpers/react_components/common/solution_view.rb
@@ -5,16 +5,24 @@ module ReactComponents
 
       def to_s
         super("common-solution-view", {
-          iterations: solution.published_iterations.map { |iteration| SerializeIteration.(iteration) },
+          iterations: iterations.map { |iteration| SerializeIteration.(iteration) },
           language: solution.track.highlightjs_language,
           indent_size: solution.track.indent_size,
           out_of_date: solution.out_of_date?,
           published_iteration_idx: solution.published_iteration.try(:idx),
+          published_iteration_idxs: solution.published_iterations.pluck(:idx),
           links: {
             change_iteration: change_iteration_link,
             unpublish: unpublish_link
           }
         })
+      end
+
+      private
+      def iterations
+        return solution.iterations.not_deleted if solution.user == current_user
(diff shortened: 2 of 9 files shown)


app/controllers/tracks/community_solutions_controller.rb
```diff
@@ -22,6 +22,7 @@ def show
 
     @author = @solution.user
     @comments = @solution.comments
+    @own_solution = @author == @current_user
```
ErikSchierboom, line 25: I'm using this in the HAML template later

app/helpers/react_components/common/solution_view.rb
```diff
@@ -5,18 +5,26 @@ class SolutionView < ReactComponent
...
           }
         })
       end
 
+      private
+      def iterations
+        return solution.iterations.not_deleted if solution.user == current_user
+
+        solution.published_iterations
+      end
```
ErikSchierboom, line 26: In order to be able to allow the user to change their published iteration to any non-deleted iteration, we need to return the non-deleted iterations if you're viewing your own solution.
Then to be able to only show the published solutions (which is the current behavior), I've needed to introduce the `published_iteration_idxs` field.

app/helpers/react_components/student/published_solution.rb
```diff
@@ -9,7 +9,7 @@ def to_s
           {
             solution: SerializeCommunitySolution.(solution),
             published_iteration_idx: solution.published_iteration.try(:idx),
-            iterations: solution.iterations.order(idx: :desc).map { |iteration| SerializeIteration.(iteration) },
+            iterations: solution.iterations.not_deleted.order(idx: :desc).map { |iteration| SerializeIteration.(iteration) },
```
ErikSchierboom, line 12: This fixes a bug where deleted iterations could be published.

app/views/tracks/community_solutions/show.html.haml
```diff
@@ -19,7 +19,7 @@
         .seperator /
         = link_to "Community Solutions", track_exercise_solutions_path(@track, @exercise)
         .seperator /
-        .exercise-title #{@author.handle}'s solution
+        .exercise-title #{@own_solution ? 'Your' : "#{@author.handle}'s"} solution
```
ErikSchierboom, line 22: It has always bothered me that my own community solution was visually not different from another user's community solution, except for some extra options. What do you think of this change?

iHiD: Yep!

iHiD (approved):

PR #2704, opened 2022-06-09
Exclude current discussion from previous

### LINKED ISSUE #5750: The prior sessions mentored is perhaps off by one?

ErikSchierboom:
- Allow excluding of discussions when retrieving
- Exclude current discussion from previous discussions
- Add system test for current discussion exclusion

Closes https://github.com/exercism/exercism/issues/5750

### DIFF
--- a/app/commands/mentor/discussion/retrieve.rb
+++ b/app/commands/mentor/discussion/retrieve.rb
@@ -16,7 +16,8 @@ module Mentor
                      student_handle: nil,
                      criteria: nil, order: nil,
                      track_slug: nil,
-                     sorted: true, paginated: true)
+                     sorted: true, paginated: true,
+                     exclude: nil)
 
         # This will be a code-level exception rather than a user-level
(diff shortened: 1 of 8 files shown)


iHiD (approved):
This looks great at a glance :)

kotp:
Is this in play now? 

Because I am seeing this, which has more than 2 previous (if previous is the marker for counting) and it is more than off by one.
incorrect_number_for_previous_discussions

kotp:
To clarify what I mean by "if previous" is the marker for counting, is that I got to this view from Boutique Inventory, and there is one future, and 4 previous discussions.  So something is not right.

kotp:
Sorry, I got there from Simple Calculator.  Boutique was simply highlighted due to a mouse cursor position.  So it appears to be correct.  It could be nice if it were highlighting the current exercise, though, so that you can visually see prior and later discussions.

PR #2718, opened 2022-06-16
Only show concepts preview if track has course

ErikSchierboom:


test/system/flows/user_views_track_test.rb
```diff
@@ -0,0 +1,64 @@
+require "application_system_test_case"
+require_relative "../../support/capybara_helpers"
+
+module Flows
+  class UserViewsTrackTest < ApplicationSystemTestCase
+    include CapybaraHelpers
+
+    test "shows about information" do
```
ErikSchierboom, line 8: I used this occasion to also test the showing of the about information and key features (in a subsequent test)

iHiD (approved):

PR #2727, opened 2022-06-21
Ignores private discussions for available mentor slots

### LINKED ISSUE #6093: Private mentoring use the regular mentoring slots - but only when one is available
As far as I could understand private mentoring requests do NOT normally do not use mentoring slots and thus are not limited. However, as reported by one of my students, if the student has open slots or if a used mentoring slot is freed by ending the discussion then the ongoing private mentoring occupies the now free mentoring slot.

This feels to be a bit inconsistent.

I wonder if either of this behavior is considered to be a bug or if this is the intention.

ErikSchierboom:
Closes https://github.com/exercism/exercism/issues/6093


app/controllers/mentoring/external_requests_controller.rb
```diff
@@ -15,21 +15,7 @@ def accept
...
-      )
-      request.fulfilled!
-      Mentor::Discussion.create!(
-        mentor: current_user,
-        request: request,
-        awaiting_student_since: Time.current
-      )
-    end
-
+    discussion = Mentor::Request::AcceptExternal.(current_user, @solution)
```
ErikSchierboom, line 18: I've taken the opportunity to move this functionality into its own command. Let me know what you think about the name.

app/models/user_track/mentoring_slots.rb
```diff
@@ -20,7 +20,8 @@ def num_available_mentoring_slots
 
   memoize
   def num_used_mentoring_slots
-    active_mentoring_discussions.size + pending_mentoring_requests.size
+    active_mentoring_discussions.size - active_external_mentoring_discussions.size +
```
ErikSchierboom: This is the core change: ignoring any active, external mentoring discussions

db/migrate/20220621081536_add_external_to_mentor_request_and_discussion.rb
```diff
@@ -0,0 +1,9 @@
+class AddExternalToMentorRequestAndDiscussion < ActiveRecord::Migration[7.0]
+  def change
+    add_column :mentor_requests, :external, :boolean, default: false, null: false, if_not_exists: true
+    add_column :mentor_discussions, :external, :boolean, default: false, null: false, if_not_exists: true
```
ErikSchierboom: So far, I've chosen to add the `external` property to both the request and discussion, although technically I only need it on discussions for this PR. I'm happy to remove it from requests if you think that would be better.

db/migrate/20220621081536_add_external_to_mentor_request_and_discussion.rb
```diff
@@ -0,0 +1,9 @@
...
+
+    Mentor::Request.where(comment_markdown: "This is a private review session").update_all(external: true)
+    Mentor::Discussion.joins(:request).where('request.external': true).update_all(external: true)
```
ErikSchierboom: Looking at the existing data, I think the only way to extract whether something was accepted via an external request was to check the fixed comment markdown value. In theory, someone could have used the exact same comment and we'd flag it as private here, but that may not be very likely.

iHiD: Don't run these in the migration, run them manually afterwards on bastion. 

If they blow up for any reason (which often happens to me!) then you'll have a migration in a half-run state and it's a real pain to get out of (involves multiple deploys of commenting and uncommenting - or fudging on bastion).

So I'd wrap them in `unless Rails.env.production?` so we still get them locally, but not when we deploy.

app/models/user_track/mentoring_slots.rb
```diff
@@ -20,7 +20,8 @@ def num_available_mentoring_slots
...
+      pending_mentoring_requests.size
```
iHiD:
```suggestion
    active_mentoring_discussions.where.not(:external).count + pending_mentoring_requests.count
```

iHiD: Better to just not include the external ones via SQL, I think.

iHiD (changes requested):
This is great pending my suggestions! :)

iHiD (approved):
Great! When you run it in production, you'll want to set the isolation level to read_committed to avoid locking. Message me if you're unsure how.

ErikSchierboom:
@iHiD I've ran the query using `READ_COMMITTED`.

PR #2743, opened 2022-06-28
Fix flaky badge test

ErikSchierboom:
- Add badge test for ordered_by_rarity scope
- Fix flake test by explicitly ordering by badge id

### DIFF
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -133,7 +133,8 @@ class UserTest < ActiveSupport::TestCase
     create :user_acquired_badge, user: user, badge: legendary_badge_2, revealed: true
     create :user_acquired_badge, user: user, badge: ultimate_badge_1, revealed: true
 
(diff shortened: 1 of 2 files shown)


test/models/user_test.rb
```diff
@@ -133,7 +133,8 @@ class UserTest < ActiveSupport::TestCase
     create :user_acquired_badge, user: user, badge: legendary_badge_2, revealed: true
     create :user_acquired_badge, user: user, badge: ultimate_badge_1, revealed: true
 
-    assert_equal [legendary_badge_1, legendary_badge_2, ultimate_badge_1, rare_badge_1, common_badge_1], user.featured_badges
+    assert_equal [legendary_badge_2, legendary_badge_1, ultimate_badge_1, rare_badge_1, common_badge_2],
```
ErikSchierboom, line 136: This test was flaky in that the order was sometimes different of badges with the same rarity. This fixes that.

This is similar to what https://github.com/exercism/website/blob/main/app/controllers/dashboard_controller.rb#L5 does BTW.

iHiD (approved):

PR #2756, opened 2022-07-12
Add request throttling

ErikSchierboom:


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,27 @@
+require 'http_authentication_token'
+
+# Enable the Rack Attack cache store in test mode,
+# as otherwise we can't verify limits being exceeded
+Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new if Rails.env.test?
+
+Rack::Attack.throttled_response_retry_after_header = true
```
ErikSchierboom, line 3: I've enabled this which will add a `retry-after` header for throttled requests.

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,23 @@
+require 'http_authentication_token'
+
+Rack::Attack.throttled_response_retry_after_header = true
```
ErikSchierboom, line 3: Enabling this ensures that a `Retry-After` header is sent back when a request is throttled. This can be used by callers to wait for X seconds before making a new request.

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,23 @@
...
+  next 30 if req.patch? && req.path =~ %r{^/api/v2/mentoring/testimonials/[^/]+/reveal}
+  next 20 if req.patch? && req.path =~ %r{^/api/v2/reputation/[^/]+/mark_as_seen}
+  next 8 if req.patch? && req.path == '/api/v2/settings/communication_preferences'
+  next 8 if req.patch? && req.path == '/api/v2/settings/sudo_update'
+
+  5
+end
+
+Rack::Attack.throttle("API - POST/PATCH/PUT/DELETE", limit: api_non_get_limit_proc, period: 1.minute) do |req|
+  next unless req.post? || req.patch? || req.put? || req.delete?
```
ErikSchierboom, line 18: I considered doing `next unless req.get?`, but as we can also have `req.options?` I figured this is likely safer.

lib/http_authentication_token.rb
```diff
@@ -0,0 +1,13 @@
+module HttpAuthenticationToken
+  def self.from_header(header)
+    token = header.to_s.match(/^Token (.*)/) { |m| m[1] }
+    return nil unless token
+
+    token.split(",").find do |kv|
+      key, value = kv.strip.split(/="?/)
+      return value.chomp('"').gsub(/\\"/, '"') if key == 'token'
+    end
```
ErikSchierboom, line 9: Cleaned this up a bit from the original source code

test/controllers/api/base_test_case.rb
```diff
@@ -7,7 +7,7 @@ def self.guard_incorrect_token!(path, args: 0, method: :get)
         url = send(path, *Array.new(args, 'a')) # rubocop:disable Lint/RedundantSplatExpansion
         send(method, url, as: :json)
 
-        assert_response 401
+        assert_response :unauthorized
```
ErikSchierboom, line 10: I noticed that the existing code used both `assert_response 200` and `assert_response :success`. I looked into using the keyword variant a bit, and quite liked it. Let me know if you like the change or would like me to revert.

test/test_helper.rb
```diff
@@ -132,13 +134,9 @@ class ActiveSupport::TestCase
...
   def setup
-    # Clear out elasticsearch
     reset_opensearch!
-
-    # Clear out redis
-    redis = Exercism.redis_tooling_client
-    keys = redis.keys("#{Exercism.env}:*")
-    redis.del(*keys) if keys.present?
+    reset_redis!
+    reset_rack_attack!
```
ErikSchierboom, line 137: Minor refactoring

test/initializers/rack_attack_test.rb
```diff
@@ -0,0 +1,114 @@
...
+      team: {
+        name: 'reviewers'
+      },
+      organization: {
+        login: 'exercism'
+      }
+    }
+
+    50.times do
+      post webhooks_membership_updates_path, headers: headers(payload), as: :json, params: payload
```
ErikSchierboom, line 88: This was the only non-GET method I could quickly find that did not require authentication

ErikSchierboom:
Do we want to show a message in the UI if a rate limit is hit? It is quite unlikely to occur for regular users and should only really happen for scripts.

test/controllers/api/base_test_case.rb
iHiD: This is good! :)

test/controllers/api/base_test_case.rb
```diff
@@ -24,5 +28,11 @@ def setup_user(user = nil)
       auth_token = create :user_auth_token, user: @current_user
       @headers = { 'Authorization' => "Token token=#{auth_token.token}" }
     end
+
+    def enable_rate_limiting
+      Rack::Attack.enabled = true
+      yield
+      Rack::Attack.enabled = true
```
iHiD:
```suggestion
    ensure
      Rack::Attack.enabled = false
```

iHiD: Two changes:
- `ensure`
- `false` not `true`

test/test_helper.rb
```diff
@@ -30,6 +30,8 @@
   c.stubbing_non_public_method = :prevent
 end
 
+Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
```
iHiD: We should probably set this in the exercism config gem.

iHiD: Does this get set in dev or production anywhere currently?

iHiD:
Love how thorough this is, @ErikSchierboom. Really brilliant job 💙

iHiD:
> Do we want to show a message in the UI if a rate limit is hit? It is quite unlikely to occur for regular users and should only really happen for scripts.

I think not for now, no.

test/test_helper.rb
ErikSchierboom: It defaults to `Rails.cache`. I don't yet set this anywhere.

test/controllers/api/base_test_case.rb
ErikSchierboom: This was old code and has subsequently been removed.

iHiD (approved):
Let's do this!

PR #2757, opened 2022-07-13
Restrict creating profile to 5 reputation

ErikSchierboom:
- Move profile creation into command
- Render 404 when trying to view profile without the correct permissions
- Working on rendering locked window
- Render locked page
- Update styling of locked profile page
- Show locked profile information
- Updated styling


app/views/profiles/intro/_locked.html.haml
```diff
@@ -0,0 +1,17 @@
+.border-1.border-borderColor6.shadow-smZ1.rounded-8.p-32.mb-32
```
ErikSchierboom: I have no idea if extracting this bit of markup to a separate view makes sense. Let me know.

test/controllers/profiles_controller_test.rb
```diff
@@ -58,21 +58,30 @@ class ProfilesControllerTest < ActionDispatch::IntegrationTest
...
-    user = create :user
+    user = create :user, reputation: 10
     create :user_profile, user: user
     sign_in!(user)
 
     get new_profile_url
     assert_redirected_to profile_path(user)
   end
 
+  test "new: 404s if the user hasn't unlocked creating a profile" do
```
ErikSchierboom: Once styling has been added, I'll add some more tests

app/commands/user/profile/create.rb
```diff
@@ -0,0 +1,13 @@
+class User::Profile
+  class Create
+    include Mandate
+
+    initialize_with :user
+
+    def call
+      raise ProfileCriteriaNotFulfilledError unless user.profile_unlocked?
+
+      User::Profile.new(user:)
```
iHiD: This should actually create (as per app/controllers/api/profiles_controller.rb:L5).

app/controllers/profiles_controller.rb
```diff
@@ -52,12 +52,17 @@ def tooltip
...
+
+    @profile_min_reputation = User::Profile::MIN_REPUTATION
+    @profile_unlocked = current_user&.profile_unlocked?
   end
 
   def new
     return redirect_to profile_path(current_user) if current_user&.profile
 
-    @profile = User::Profile.new
+    @profile = User::Profile::Create.(current_user)
```
iHiD: As this is a GET controller, it shouldn't call a mutating (create here) method. Instead we should add a guard with the `return redirect_to` above, such as `return redirect_to action: :intro unless user.profile_unlocked?`

app/models/user.rb
```diff
@@ -261,4 +261,6 @@ def introducer_dismissed?(slug)
   def send_devise_notification(notification, *args)
     devise_mailer.send(notification, self, *args).deliver_later
   end
+
+  def profile_unlocked? = reputation >= User::Profile::MIN_REPUTATION
```
iHiD:
```suggestion
  def may_create_profile? = reputation >= User::Profile::MIN_REPUTATION
```

When I read "unlocked" I initially interpreted it as the profile having been made, so I think renaming it like this is wiser.

app/models/user/profile.rb
```diff
@@ -5,6 +5,8 @@ class User::Profile < ApplicationRecord
 
   delegate :to_param, to: :user
 
+  MIN_REPUTATION = 5
```
iHiD: Pop this right at the top of the class definition pls (before the extend).

app/controllers/profiles_controller.rb
```diff
@@ -52,10 +52,14 @@ def tooltip
...
   def intro
     return redirect_to profile_path(current_user) if current_user&.profile
+
+    @profile_min_reputation = User::Profile::MIN_REPUTATION
+    @may_create_profile = current_user&.may_create_profile?
   end
 
   def new
     return redirect_to profile_path(current_user) if current_user&.profile
+    return redirect_to action: :intro unless current_user&.may_create_profile?
```
iHiD: Neither of these lines need the `&`, nor does the one on L57 as we already check that we have a current user via L2.

iHiD (approved):
One small comment, then get it merged! :)

PR #2769, opened 2022-07-15
Add country code to metrics

ErikSchierboom:
There's still stuff to do, like deciding upon the reverse geocoding service we want to use.


app/controllers/application_controller.rb
```diff
@@ -53,6 +53,8 @@ def ensure_not_mentor!
     redirect_to mentoring_inbox_path
   end
 
+  def country_code = request.location&.country_code.presence
```
iHiD: We need to ensure this is instant and foolproof if going here. That means the database needs to be local to this docker image really, and the lookup in like 1ms.

The alternative, which feels safer to me, is to pass the IP address here into the `Metric::Queue` method, then do the lookup as part of the `Metric::Create` call

.github/workflows/deploy.yml
```diff
@@ -72,9 +72,10 @@ jobs:
           ECR_REPOSITORY: ${{ env.ecr_repository_rails }}
           IMAGE_NAME: "${{ steps.login_to_ecr.outputs.registry}}/${{ env.ecr_repository_rails }}:${{ github.sha }}"
           DOCKERFILE: ${{ env.dockerfile_rails }}
+          GEOIP_LICENSE_KEY: ${{ secrets.GEOIP_LICENSE_KEY }}
```
ErikSchierboom, line 75: I have a key that we can use

app/commands/metric/create.rb
```diff
@@ -6,10 +6,13 @@ class Metric::Create
...
-    klass.new(occurred_at:, params:).tap do |metric|
+    klass.new(occurred_at:, country_code:, params: params.except(:remote_ip)).tap do |metric|
       metric.save!
     rescue ActiveRecord::RecordNotUnique
       return klass.find_by!(uniqueness_key: metric.uniqueness_key)
     end
   end
+
+  private
+  def country_code = Geocoder.search(params[:remote_ip]).first&.country_code
```
ErikSchierboom: This is where the country code is determined from the remote IP address. This method is always called from within a job, so it should be safe to do so without blocking anything

config/initializers/geocoder.rb
```diff
@@ -0,0 +1,14 @@
+# rubocop:disable Naming/VariableNumber
+Geocoder.configure(
+  ip_lookup: :geoip2,
+  geoip2: {
+    file: ENV.fetch('GEOIP_FILE', File.join('/usr', 'share', 'GeoIP', 'GeoLite2-Country.mmdb'))
```
ErikSchierboom, line 7: I've used the default location that maxmind's `geoipupdate` tool uses.

docker/rails.Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM ruby:3.1.0-bullseye
 
+ARG GEOIP_LICENSE_KEY
```
ErikSchierboom, line 3: `ARG` means that you can pass something in to be used as an environment variable within the Dockerfile, but it'll only exist while building the Docker image; it is _not_ baked into the actual Docker image.

config/initializers/geocoder.rb
ErikSchierboom: I'm using the country database as that fits or need and is also a smaller database than the other databases (which have more information).

app/commands/mentor/discussion/finish_by_student.rb
```diff
@@ -91,7 +91,8 @@ def award_badge!
       end
 
       def log_metric!
-        Metric::Queue.(:finish_mentoring, discussion.finished_at, discussion:, track:, user: student)
+        Metric::Queue.(:finish_mentoring, discussion.finished_at,
+          discussion:, track:, user: student, remote_ip: Exercism.request_context[:remote_ip])
```
iHiD: Any reason not to move this into Metric::Queue itself?

config/initializers/exercism_config.rb
```diff
@@ -8,8 +8,12 @@
 
 Exercism.config.hcaptcha_endpoint = "https://hcaptcha.com"
 
-# TODO: Move this upstream
 module Exercism
+  # We'll store the request context for easy access from commands
+  # without having to pass it all the way down from the controller
+  cattr_accessor :request_context
```
iHiD, line 14: Nice comment :)

config/initializers/geocoder.rb
iHiD: Maybe add these as comments to this file as they're useful for future reference.

docker/rails.Dockerfile
```diff
@@ -9,6 +10,12 @@ RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
     curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
     apt-get install -y cmake make nodejs yarn graphicsmagick libvips42
 
+WORKDIR /usr/share/GeoIP
+
+RUN curl "https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-Country&license_key=${GEOIP_LICENSE_KEY}&suffix=tar.gz" --output geolite2-country.tar.gz && \
+    tar -xvf geolite2-country.tar.gz --strip-components=1 --wildcards '*/GeoLite2-Country.mmdb' && \
+    rm geolite2-country.tar.gz
```
iHiD, line 18: The only downside with this being here is that it'll never get update, whereas if it's further down it'll get updated every time we deploy. The size of the db should probably determine this. If it's like 10mb, let's move it down. If it's huge, let's leave it here. Thoughts?

docker/rails.Dockerfile
iHiD: Oooh, that's nice.

docker/rails.Dockerfile
ErikSchierboom: The size is 5.3MB. I'll move it down then.

app/commands/mentor/discussion/finish_by_student.rb
ErikSchierboom: Yes, as there are a couple of metrics (the Github-related ones) that should have `remote_ip` set to `nil`. If I were to remove it here, the logic to determine whether or not has to move elsewhere. There are a couple of options:

1. Have `Metric::Queue` determine whether or not the `remote_ip` should be added. This feels like a violation of the separation of concerns, so I don't really like this.
2. Another would be to add a `remote_ip` method to the `Metric` base class that subclasses can override (which the GitHub ones would do), and then use that when creating the metric. 

I'll try the second approach.

.github/workflows/deploy.yml
```diff
@@ -72,9 +72,10 @@ jobs:
...
         run: |
           # Build a docker container and push it to ECR
-          docker build -t $IMAGE_NAME -f $DOCKERFILE .
+          docker build -t $IMAGE_NAME -f $DOCKERFILE --build-arg GEOIP_LICENSE_KEY --build-arg GEOIP_CACHE_BUSTER=$(date -d "`date +%Y%m01`" +%Y-%m-%d) .
```
ErikSchierboom, line 78: Instead of moving the GeoIP DB download command down in the Dockerfile, I've found (what I think is) a nicer solution: adding a cache buster build argument. What I'm doing is to passing in a build argument named `GEOIP_CACHE_BUSTER` which values is `YYYY-MM-01`, so basically the current year and month (with a fixed day). This means that this value changes every month, but that it stays the same within the same month. Once the value changes, it'll cause the cache to be invalidated and the DB is downloaded again.

app/commands/mentor/discussion/finish_by_student.rb
ErikSchierboom: Check https://github.com/exercism/website/pull/2769/commits/9e058ef0a5cf6babd6fcdf6e9cdad53861f78b57

iHiD:
This can be merged. The error is fixed by https://github.com/exercism/website/pull/2803

iHiD (approved):
Let's go!

PR #2771, opened 2022-07-15
Revert "Add request throttling"

iHiD:
Reverts exercism/website#2756

Based on a couple of things (e.g. me seeing editor errors, and Josh seeing mentor errors) I'm reverting this for now


config/initializers/rack_attack.rb
```diff
@@ -1,23 +0,0 @@
...
-  next 4 if req.post? && req.path =~ %r{^/api/v2/solutions/[^/]+/iterations}
-  next 12 if req.post? && req.path =~ %r{^/api/v2/solutions/[^/]+/submissions}
-  next 30 if req.post? && req.path =~ %r{^/api/v2/markdown/parse}
-  next 30 if req.patch? && req.path =~ %r{^/api/v2/mentoring/testimonials/[^/]+/reveal}
-  next 20 if req.patch? && req.path =~ %r{^/api/v2/reputation/[^/]+/mark_as_seen}
-  next 8 if req.patch? && req.path == '/api/v2/settings/communication_preferences'
-  next 8 if req.patch? && req.path == '/api/v2/settings/sudo_update'
-
-  5
-end
```
iHiD, line 15: @ErikSchierboom I'm wondering if these are grouping correctly. So say I hit the `5` at the end, does that then block me for the rest of the minute from all the rest? And if I do 4 submissions, but then one `markdown parse` do I then get blocked. ie when I'm blocked, am I blocked from *everything* regardless of the type of the new thing. If so, then us missing something might be problematic.

iHiD: See https://exercism-team.slack.com/archives/CAQP7JL3T/p1657908532249979

iHiD: (Update: My issue is something else - not rate limiting)

ErikSchierboom: I have an idea. I think what is missing is the request URL and method in the return value of throttle. So something like:

"#{req.path}/#{req.action}/#{token || req.ip}"

iHiD: Ah! Yes. That makes sense.

PR #2776, opened 2022-07-19
Support multiple cohorts

ErikSchierboom:
- Parameterize cohorts
- Add partial for cohorts


app/controllers/cohorts_controller.rb
```diff
@@ -1,17 +1,32 @@
...
   end
+
+  private
+  def use_cohort
+    @cohort = COHORTS[params[:id].to_sym]
+    render_404 if @cohort.blank?
+
+    user_track = UserTrack.for(@current_user, @cohort.track_slug)
+    @num_concepts = user_track.num_concepts
+    @num_exercises = user_track.num_exercises
```
ErikSchierboom: This data was used in the template, but was always empty.

app/models/cohort_membership.rb
```diff
@@ -1,7 +1,15 @@
 class CohortMembership < ApplicationRecord
   belongs_to :user
+  belongs_to :cohort
+
+  enum status: {
+    on_waiting_list: 0,
+    enrolled: 1
+  }
```
ErikSchierboom, line 10: I've opted for an enum as I quite like their verbosity (and the automatically generated scopes).

db/migrate/20220719131059_create_cohorts.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.string :slug, null: false
+      t.string :name, null: false
+      t.column :capacity, :int, null: false, default: 0
+      t.datetime :begins_at, null: false
+      t.datetime :ends_at, null: false
+      t.belongs_to :track, null: false, foreign_key: true
+
+      t.timestamps
+
+      t.index :slug, unique: true
```
ErikSchierboom: I assume that we're only making the slug unique? That _would_ mean that having another gohort requires a new slug.

db/migrate/20220719135115_add_cohort_to_cohort_memberships.rb
```diff
@@ -0,0 +1,18 @@
...
+      c.name = 'Go-Hort'
+      c.capacity = 100
+      c.track = Track.find_by(slug: 'go')
+      c.begins_at = Time.utc(2022, 07, 01)
+      c.ends_at = Time.utc(2022, 07, 30)
+    end
+
+    CohortMembership.update(cohort: Cohort.find_by(slug: :cohort_slug))
+
+    change_column_null :cohort_memberships, :cohort_id, false
```
ErikSchierboom: I know running database queries inside a migration is not a nice thing to do. In this case, I did it because I can only make the `cohort_id` column not-null if it has a value, which means that it needs to point to something.

I've also got to make it be a foreign key.

app/commands/cohort/join.rb
```diff
@@ -0,0 +1,15 @@
+class Cohort::Join
+  include Mandate
+
+  initialize_with :user, :cohort, :introduction
+
+  def call
+    # TODO: make this safe against race issues
+    user.cohort_memberships.create!(cohort:, status:, introduction:)
```
iHiD: I have a preference for:
```suggestion
    User::CohortMembership.create!(user: ,cohort:, status:, introduction:)
```

app/commands/cohort/join.rb
```diff
@@ -0,0 +1,15 @@
...
+  rescue ActiveRecord::RecordNotUnique
+    # This is fine
```
iHiD: You might want to lookup + return the correct membership back from here so that this always returns a membership

app/controllers/cohorts_controller.rb
```diff
@@ -1,17 +1,24 @@
...
-    end
+    Cohort::Join.(current_user, @cohort, params[:introduction])
 
     redirect_to action: :show, anchor: "register"
   end
+
+  private
+  def use_cohort
+    @cohort = ::Cohort.find_by(slug: params[:id])
+    render_404 if @cohort.blank?
```
iHiD: This can be reduced to:

```suggestion
    @cohort = ::Cohort.find_by!(slug: params[:id])
```

app/models/cohort.rb
```diff
@@ -0,0 +1,8 @@
+class Cohort < ApplicationRecord
+  belongs_to :track
+
+  has_many :cohort_memberships, dependent: :destroy
+  has_many :memberships, through: :cohort_memberships, source: :cohort
```
iHiD: This is wrong.

I think you mean
```
has_many :members, through: :cohort_memberships, source: :users
```

(memberships -> members, cohort -> user)

db/migrate/20220719131059_create_cohorts.rb
iHiD: Yes, this is correct.

I'm pretty sure you can inline `index: { unique: true }` onto L4 btw.

db/migrate/20220719131059_create_cohorts.rb
```diff
@@ -0,0 +1,16 @@
+class CreateCohorts < ActiveRecord::Migration[7.0]
+  def change
+    create_table :cohorts do |t|
+      t.string :slug, null: false
+      t.string :name, null: false
+      t.column :capacity, :int, null: false, default: 0
+      t.datetime :begins_at, null: false
+      t.datetime :ends_at, null: false
+      t.belongs_to :track, null: false, foreign_key: true
```
iHiD: I tend to always put relations at the top of a table definition with a blank line below them.

iHiD:
Looking good.

app/models/cohort.rb
ErikSchierboom: I do. Thanks

app/commands/cohort/join.rb
```diff
@@ -0,0 +1,27 @@
...
+  def call
+    time = Time.now.utc.to_formatted_s(:db)
+
+    # rubocop:disable Layout/LineLength
+    id = CohortMembership.connection.insert(%{
+      INSERT INTO cohort_memberships (user_id, cohort_id, status, introduction, created_at, updated_at)
+      SELECT #{user.id}, #{cohort.id}, IF(COUNT(*)>=#{cohort.capacity},#{ON_WAITING_LIST_STATUS},#{ENROLLED_STATUS}), "#{introduction}", "#{time}", "#{time}"
+      FROM cohort_memberships
+      WHERE cohort_id = #{cohort.id}
+    })
```
ErikSchierboom: I needed to use manual SQL here to ensure that there isn't a race condition regarding the status of the new cohort membership. There is a test to verify that this works well (the test failed for the old, `.create` code).

app/views/cohorts/show/_form.html.haml
```diff
@@ -0,0 +1,44 @@
...
+    - elsif membership
+      - if membership.enrolled?
+        %h3.text-h1.text-center.mb-16 You're all set!
+        %p.text-p-xlarge.mb-16.text-center.mx-auto{ style: "max-width:800px;" }
+          Congratulations - you're all set!
+          One of our team will email you shortly to welcome you to the #{cohort.name} and tell you how to get set up!
+      - else
+        %h3.text-h1.text-center.mb-16 You're on the waiting list
+        %p.text-p-xlarge.mb-16.text-center.mx-auto{ style: "max-width:800px;" }
+          Sorry - the #{cohort.name} filled up before you got here! You're currently on the waiting list. If a space becomes available we'll let you know. Otherwise we'll message you about the next #{cohort.name} later in the year.
```
ErikSchierboom: One thing that we lost with this PR is the bit about the position in the waiting list. I could add that back if we want to, but then I'd need to store the enrollment position with the cohort membership 🤷

db/migrate/20220719131059_create_cohorts.rb
```diff
@@ -0,0 +1,15 @@
+class CreateCohorts < ActiveRecord::Migration[7.0]
+  def change
+    create_table :cohorts do |t|
+      t.belongs_to :track, null: false, foreign_key: true
+
+      t.string :slug, null: false, index: { unique: true }
+      t.string :name, null: false
+      t.column :capacity, :int, null: false, default: 0
+      t.datetime :begins_at, null: false
+      t.datetime :ends_at, null: false
```
ErikSchierboom, line 10: We don't yet do anything with these fields. I assume we want to show a special page if a cohort has not been opened or has ended?

db/migrate/20220720075226_remove_user_id_index_on_cohort_memberships.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveUserIdIndexOnCohortMemberships < ActiveRecord::Migration[7.0]
+  def change
+    remove_foreign_key :cohort_memberships, :users, column: :user_id
+    remove_index :cohort_memberships, :user_id, if_not_exists: true
+  end
+end
```
ErikSchierboom, line 6: I had some issues with getting the migrations working when I combined things. It basically prevented me from doing rollbacks, which was a bit unfortunate. Once we're happy with the PR, I could try to merge the migrations into one if desired.

test/commands/cohort/join_test.rb
```diff
@@ -0,0 +1,60 @@
...
+        user = create :user
+        Cohort::Join.(user, cohort, 'Hi')
+      end
+    end
+    threads.map(&:join)
+
+    assert_equal member_count, CohortMembership.count
+    assert_equal cohort.capacity, CohortMembership.enrolled.count
+    assert_equal member_count - cohort.capacity, CohortMembership.on_waiting_list.count
+  end
```
ErikSchierboom, line 59: This test verifies that even with concurrent joining, the number of enrolled students does not exceed the capacity.

app/controllers/cohorts_controller.rb
```diff
@@ -1,17 +1,25 @@
 class CohortsController < ApplicationController
   skip_before_action :authenticate_user!
+  before_action :use_cohort
 
   def show
-    @membership = current_user.cohort_memberships.first if user_signed_in?
+    @membership = CohortMembership.find_by(user: current_user, cohort: @cohort) if user_signed_in?
```
iHiD:
```suggestion
    @membership = current_user.cohort_memberships.find_by(cohort: @cohort) if user_signed_in?
```

I prefer this version when looking up 😁

iHiD:
Reminder for this that we discussed finding the waiting list position and still showing that.

ErikSchierboom:
The waiting list position functionality has been restored: https://github.com/exercism/website/pull/2776/commits/dc30a3bba46639fdddca68706797a9f8187fb03b

iHiD (approved):

PR #2791, opened 2022-07-26
Add cache_column

iHiD:
This knocks 250ms off all the community solution pages for popular exercises.

### DIFF
--- a/app/jobs/cache_num_published_solutions_on_exercise_job.rb
+++ b/app/jobs/cache_num_published_solutions_on_exercise_job.rb
@@ -1,0 +1,9 @@
+class CacheNumPublishedSolutionsOnExerciseJob < ApplicationJob
+  def perform(exercise)
+    sql = Arel.sql(exercise.solutions.published.select("COUNT(*)").to_sql)
+
+    ActiveRecord::Base.transaction(isolation: Exercism::READ_COMMITTED) do
+      Exercise.where(id: exercise.id).update_all("num_published_solutions = (#{sql})")
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


ErikSchierboom (commented):
I think you'll also need to update this in https://github.com/exercism/website/blob/main/app/commands/solution/publish_iteration.rb, which is called in https://github.com/exercism/website/blob/main/app/controllers/api/solutions_controller.rb#L119 without using `Solution::Publish`.

iHiD:
> I think you'll also need to update this in...

I don't think I do as that code doesn't change the quantity of published solutions, unless I'm misunderstanding.

iHiD:
Happy for you to merge this if you're happy too.

ErikSchierboom (approved):

ErikSchierboom:
Maybe also update: https://github.com/exercism/website/blob/main/app/controllers/tracks/community_solutions_controller.rb#L11?

iHiD:
Could you do this pls?

app/jobs/cache_num_published_solutions_on_exercise_job.rb
```diff
@@ -0,0 +1,9 @@
+class CacheNumPublishedSolutionsOnExerciseJob < ApplicationJob
+  def perform(exercise)
```
ErikSchierboom, line 4: Maybe be explicit about the queue being used? We do that in other jobs too.

```suggestion
  queue_as :default

  def perform(exercise)
```

iHiD: Good shout. Thanks.

PR #2792, opened 2022-07-26
Increase tolerance for community solution searching

iHiD:
When this value is hit, the actual SQL takes over 20s (!!) to execute, so I'd rather be more tolerant on the search index.

### DIFF
--- a/app/commands/solution/search_community_solutions.rb
+++ b/app/commands/solution/search_community_solutions.rb
@@ -108,7 +108,7 @@ class Solution
       [value].flatten
     end
 
-    TIMEOUT = '100ms'.freeze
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
Is searching for user solutions also this slow when searching via the DB? If so, you could consider also increasing the timeout here: https://github.com/exercism/website/blob/main/app/commands/solution/search_user_solutions.rb#L110

iHiD:
Good shout. I need to check this in `skylight`. I'd like to get this merged regardless, then come back to it. If you're happy with it, could you merge it in the morning pls?

ErikSchierboom (approved):

PR #2793, opened 2022-07-26
Add MandateJob

iHiD:
This means that:
```ruby
class SendDonationPaymentEmailJob < ApplicationJob
  queue_as :notifications

  def perform(payment)
    Donations::Payment::SendEmail.(payment)
  end
end
```

Can be reduced to 
```ruby
Donations::Payment::SendEmail.defer(payment)
```

---

## TODO

- [x] Move `queue_as :notifications` into the command and tap into it somehow.
- [x] Handle `wait` options
- [ ] Check whether there are other params we want.


iHiD:
@ErikSchierboom Could you review this pls. If you're happy, could you move any other things you want into this format too pls? And then we can merge it. Upstreaming it is a little complex, so I'm happy to leave this here for now.

ErikSchierboom:
Once this is merged, I want to do a follow-up PR to slightly refactor the github webhooks code, which can also use this new MandateJob feature. I'd like to do this separately though, as it'll involve some shuffling around of code and I don't want to clutter this PR.

iHiD:
Also, once this is merged I might need to spin up an old Sidekiq server to process any old jobs using the original format.

PR #2794, opened 2022-07-26
cohorts: fix typo in headers

ee7:
Fixes this:

https://exercism.org/cohorts/gohort-aug-22
https://exercism.org/cohorts/exhort-aug-22


iHiD:
Thanks :)

PR #2803, opened 2022-07-28
Remove use of threads

iHiD:
Rails isn't threadsafe without doing work (especially about ActiveRecord connection sharing), so this is resulting in sporadic errors.


ErikSchierboom (approved):

ErikSchierboom:
@iHiD I've added the comment

PR #2809, opened 2022-07-28
Fix metrics

iHiD:
By the time `Metric::Create` is called, it's in Sidekiq, which no-longer has access to the request as it was at the time of the user. This moves the extraction logic into `Queue` and then passing it explicitely into `Create`

### DIFF
--- a/app/commands/metric/create.rb
+++ b/app/commands/metric/create.rb
@@ -3,16 +3,29 @@ class Metric::Create
 
   queue_as :metrics
 
-  initialize_with :type, :occurred_at, :params
+  def initialize(type, occurred_at, params)
+    @type = type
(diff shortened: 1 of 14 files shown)


ErikSchierboom (approved):

app/commands/metric/create.rb
```diff
@@ -3,16 +3,29 @@ class Metric::Create
...
     end
   end
+
+  private
+  attr_reader :type, :occurred_at, :params, :request_context
+
+  memoize
+  def remote_ip
+    request_context[:remote_ip]
+  end
```
ErikSchierboom:
```suggestion
  def remote_ip = request_context[:remote_ip]
```

PR #2814, opened 2022-07-29
Fix git diffs

iHiD:
If files are added to one of the categories of files in config.json but they already exist in git, then the important_files_hash changes, but there are no changes in the actual contents of the files as generated by git, so we show a broken empty diff.

This fixes that by manually generating a diff from the start of time to the HEAD commit for any paths that are added to the files hash.

There's currently at least five things to consider:
- [x] We check for files twice if they're added to git and the config at the same time. We should add a guard in the block at line 32 rather than doing the diff at the end.
- [x] My brain is too tired to work out the consequences of **removing** files from the config.json, but we should consider it
- [x] There are no tests. This file itself seems to have no test coverage, and I don't have the energy to add any now.
- [x] `changed_files` is now too messy. We should split it into two methods (`changed_in_git` and `changed_in_config`) and then `changed_files` should just sum them or something.
- [x] `repo.send(:repo_dir)` needs DRYing up.

@ErikSchierboom If you fancy finishing this off, I'd appreciate it. But if you don't get chance, we can do it next week.


app/commands/git/generate_diff_between_exercise_versions.rb
```diff
@@ -22,7 +20,21 @@ def call
     def changed_files
       # This is a diff of two commits considering all files in the respective old and new directories
       raw_diff = `cd #{repo.send(:repo_dir)} && git diff #{old_sha} #{exercise.git_sha} -- #{old_git.dir} #{new_git.dir}` # rubocop:disable Layout/LineLength
-      ProcessDiff.(raw_diff)
+      changes = ProcessDiff.(raw_diff)
+
+      # We now need to check for any new files that have been added to they interesting_paths,
+      # but may have been added to git in a previous commit. In these cases we need to do a manual
+      # diff for each file between the very first commit in the repo, and this latest commit
+      return changes unless new_interesting_paths.present?
```
iHiD: The rubocop fix is for this line. It wants to change unless/present? to if/blank? but I think that just reads worse. I don't believe there's a global rule for this as if/unless have different semantic meanings to the reader. So I've turned the rule off.

test/models/metrics/complete_solution_metric_test.rb
```diff
@@ -14,7 +14,7 @@ class Metrics::CompleteSolutionTest < ActiveSupport::TestCase
       assert_equal occurred_at, metric.occurred_at
       assert_equal user, metric.user
       assert_equal track, metric.track
-      assert_equal 'US', metric.country_code
+      # assert_equal 'US', metric.country_code
```
iHiD: This and the one below need reverting.

app/commands/git/generate_diff_between_exercise_versions.rb
```diff
@@ -82,13 +102,11 @@ def call
       end
 
       def add_diff!(filepath, lines)
-        @diffs << Diff.new(filepath, lines.join) if filepath
+        @diffs << Diff.new(filepath, exercise.dir, lines.join) if filepath
```
iHiD:
```suggestion
        @diffs << Diff.new(filepath, exercise.git_dir, lines.join) if filepath
```

app/models/exercise.rb
```diff
@@ -50,7 +50,7 @@ def self.for(track_slug, exercise_slug)
     joins(:track).find_by('tracks.slug': track_slug, slug: exercise_slug)
   end
 
-  delegate :files_for_editor, :exemplar_files, :introduction, :instructions, :source, :source_url, to: :git
+  delegate :files_for_editor, :exemplar_files, :introduction, :instructions, :source, :source_url, :dir, to: :git
```
iHiD:
```suggestion
  delegate :files_for_editor, :exemplar_files, :introduction, :instructions, :source, :source_url, to: :git
  delegate :dir, to: :git, prefix: true
```

iHiD: I think I prefer the clarity of `exercise.git_dir` rather than `exercise.dir` here.

iHiD:
Couple of failing tests.

ErikSchierboom:
> Couple of failing tests.

@iHiD Those tests were flaky

PR #2838, opened 2022-08-09
Fix listen Gem

### LINKED ISSUE #636: Pegging CPU core on Apple M1 processor
Running Ruby 2.7.1 and Spring 2.1.0 (in a Rails 6.0.3.2 app) on an M1 MacBook Air under Rosetta (running the whole shell in x86 mode) I'm seeing spring pegging a whole CPU core (CPU usage of 99%).

Curious to know if others are seeing the same, and what the best way to debug this is!

iHiD:
Fixes https://github.com/rails/spring/issues/636#issuecomment-755404716

Goes from consistently being at 40% on every core to 0.3%...!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -84,7 +84,7 @@ end
 
 group :development do
   gem 'haml_lint', require: false
-  # gem 'solargraph'
+  gem 'solargraph'
   gem 'rubocop', require: false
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):
Yay!

PR #2839, opened 2022-08-09
Break out 'styles' into UI kit

iHiD:


dem4ron (approved):
Great, will be nicer to work with! I didn't notice any fractures in the UI on my machine.

PR #2843, opened 2022-08-10
Support Linux in bin/dev

ErikSchierboom:
With this PR, the bin/dev binary supports running on both MacOS and Linux.

We need this as `hivemind` is only supported on MacOS, whereas on Linux you'd use `overmind`.

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -1,3 +1,13 @@
 #!/usr/bin/env bash
 
-hivemind Procfile.dev
+case $(uname -s) in
+    Linux)
+        overmind start -f Procfile.dev
+        ;;
+    Darwin)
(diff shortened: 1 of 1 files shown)


iHiD:
I don't think this is correct actually. Hivemind and overmind are effectively the same thing, it's just that overmind has loads more config options (but they're made by the same dev and designed to be simple vs complex versions). So I don't think the split is mac/linux here, just which of those two tools you prefer.

ErikSchierboom:
Ah, I had some issues getting hivemind working on WSL and overmind was working correctly. I'll try and address that then.

iHiD:
I'm also happy to switch this to overmind in general.

PR #2847, opened 2022-08-10
Export solutions

ErikSchierboom:
This PR restores the functionality from v2 to export the files of the last 500 solutions, which can be very helpful with representers or analyzers.

CC @junedev


app/commands/exercise/export_solutions_to_zip_file.rb
```diff
@@ -0,0 +1,32 @@
...
+class Exercise::ExportSolutionsToZipFile
+  include Mandate
+
+  initialize_with :exercise
+
+  def call
+    Zip::File.open(zip_file_path, create: true) do |zip_file|
+      solutions.each.with_index do |solution, idx|
+        solution.latest_iteration.files.each do |iteration_file|
+          zip_file.get_output_stream("#{idx}/#{iteration_file.filename}") do |f|
```
ErikSchierboom: The v2 version of this code flattened all filenames (effectively removing any sub-directories). I've fixed this in this version.

app/commands/exercise/export_solutions_to_zip_file.rb
```diff
@@ -0,0 +1,32 @@
...
+            f.puts(iteration_file.content.force_encoding("utf-8"))
+          end
+        end
+      end
```
ErikSchierboom: The v2 version of this code wrote files to the file system to then zip the directory recursively. With this version, we'll directly write to the zip file without the use of any intermediate files or recursive directory zipping helper needed.

app/commands/exercise/export_solutions_to_zip_file.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  private
+  # We need to memoize this as there is randomness in the file path
+  memoize
+  def zip_file_path = Rails.root / "tmp" / "export_solutions_data" / "#{Time.now.to_i}-#{SecureRandom.uuid}.zip"
+
+  def solutions
+    exercise.solutions.includes(iterations: :files).
+      where(status: %i[iterated completed published]).
```
ErikSchierboom, line 35: I've omitted solutions that have the `:started` status.

app/controllers/api/solutions/export_controller.rb
```diff
@@ -0,0 +1,23 @@
+module API
+  module Solutions
+    class ExportController < BaseController
+      before_action :use_exercise
+
+      def index
+        return render_403(:solutions_export_not_accessible) unless current_user.maintainer? || current_user.admin?
```
ErikSchierboom: I've only made this functionality available for maintainers and admins. 

So far, I've not yet created a UI in the `/maintaining` part of the website.

config/routes/api.rb
```diff
@@ -81,6 +81,7 @@
           patch :start
         end
 
+        resources :export_solutions, only: [:index], controller: "solutions/export"
```
ErikSchierboom: I've considered various paths, but felt that `/api/v2/tracks/csharp/exercises/leap/export_solutions` was nice. Let me know what you think.

config/initializers/rack_attack.rb
```diff
@@ -1,15 +1,32 @@
...
+
+  memoize
+  def route_name
+    route_name = nil
+    Rails.application.routes.router.recognize(self) do |route|
+      route_name = route.name if route.name.present?
+    end
+    route_name
+  end
+end
```
ErikSchierboom: Now that I had two throttle blocks, I decided that it might be nice to extract this shared logic and monkey patch it.

config/initializers/rack_attack.rb
```diff
@@ -1,15 +1,32 @@
...
-  next 20 if req.patch? && req.path =~ %r{^/api/v2/reputation/[^/]+/mark_as_seen}
-  next 8 if req.patch? && req.path == '/api/v2/settings/communication_preferences'
-  next 8 if req.patch? && req.path == '/api/v2/settings/sudo_update'
+  next 4 if req.post? && req.route_name == 'api_solution_iterations'
+  next 12 if req.post? && req.route_name == 'api_solution_submissions'
+  next 30 if req.post? && req.route_name == 'api_parse_markdown'
+  next 30 if req.patch? && req.route_name == 'reveal_api_mentoring_testimonial'
+  next 20 if req.patch? && req.route_name == 'mark_as_seen_api_reputation'
+  next 8 if req.patch? && req.route_name == 'api_settings_communication_preferences'
+  next 8 if req.patch? && req.route_name == 'sudo_update_api_settings'
```
ErikSchierboom: I've changed this to do route name matching, which is less likely to be wrong than a regex.

app/commands/exercise/export_solutions_to_zip_file.rb
```diff
@@ -0,0 +1,28 @@
...
+
+class Exercise::ExportSolutionsToZipFile
+  include Mandate
+
+  initialize_with :exercise
+
+  def call
+    file_stream = Zip::OutputStream.write_buffer do |zip|
+      solutions.each.with_index do |solution, idx|
+        solution.latest_iteration.files.each do |iteration_file|
```
ErikSchierboom: I've used the v2 version of the code as inspiration, but I was wondering whether we might want to output the files like the representer/analyzer would receive them (which would include test files and such).

config/routes/api.rb
iHiD: Yeah, I'm good with this. But I'd make the controller match it (ie ExportSolutionsController).

app/controllers/api/solutions/export_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+        zip_file_stream = Exercise::ExportSolutionsToZipFile.(@exercise)
+
+        send_data zip_file_stream, type: 'application/zip', filename: 'solutions.zip'
+      end
+
+      private
+      def use_exercise
+        @track = Track.find_by(slug: params[:track_slug])
+        return render_track_not_found if @track.blank?
```
iHiD: I'd have a slight preference here to having a `use_track` method which is:

```ruby
def use_track
  @track = Track.find(params[:track_slug])
  render_track_not_found
end
```

and then the same for exercise

app/commands/exercise/export_solutions_to_zip_file.rb
iHiD: `solution.latest_iteration `

This should be the first iteration, not the latest one (we want to see what they look like pre-mentoring)

iHiD: With a maintainer hat on, I'd probably rather just get the students' files as those are the ones that I have to deal with. But if your maintainers hat (or other people) disagree, I don't have any strong preference. Do we actually pass test files etc to representers?

app/commands/exercise/export_solutions_to_zip_file.rb
```diff
@@ -0,0 +1,28 @@
...
+          zip.put_next_entry "#{idx}/#{iteration_file.filename}"
+          zip.print iteration_file.content.force_encoding("utf-8")
+        end
+      end
+    end
+
+    file_stream.rewind
```
iHiD, line 28: I always find it so weird that we have to manually do this!

app/commands/exercise/export_solutions_to_zip_file.rb
ErikSchierboom: > Do we actually pass test files etc to representers?

We do. All exercise files are passed to the tooling, except for the files within `.docs` (and the solution files overwrite any exercise files).

> But if your maintainers hat (or other people) disagree,

The reason why we do the above, is that some tracks might have non-solution files (e.g. project files) that are needed for the representer/analyzer to work well. E.g. the F# analyzer might want to analyze the project file for the order of file dependencies, which is not a solution file.

app/commands/exercise/export_solutions_to_zip_file.rb
ErikSchierboom: Yes!

config/routes/api.rb
```diff
@@ -81,6 +81,7 @@
           patch :start
         end
 
+        resources :export_solutions, only: [:index], controller: "export_solutions"
```
ErikSchierboom: @iHiD Is this the way you'd like the controller to be named?

app/commands/exercise/export_solutions_to_zip_file.rb
```diff
@@ -0,0 +1,38 @@
...
+        submission = solution.submissions.first
+
+        # Export the first iteration's files as that iteration won't have
+        # had any mentor/analyzer/representer comments applied to them
+        submission.files.each do |iteration_file|
+          zip.put_next_entry "#{idx}/#{iteration_file.filename}"
+          zip.print iteration_file.content.force_encoding("utf-8")
+        end
+
+        # Export the tooling files
```
ErikSchierboom:
```suggestion
        # Export the other (exercise) files
```

config/routes/api.rb
iHiD: Yes, which means you can leave the `controller` param out altogether on this line.

app/controllers/api/export_solutions_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      zip_file_stream = Exercise::ExportSolutionsToZipFile.(@exercise)
+
+      send_data zip_file_stream, type: 'application/zip', filename: 'solutions.zip'
+    end
+
+    private
+    def use_track
+      @track = Track.find_by(slug: params[:track_slug])
+      return render_track_not_found if @track.blank?
```
iHiD: You didn't like my suggestion of `find` and the rescue? (https://github.com/exercism/website/pull/2847#discussion_r943327471)

ErikSchierboom: I missed the rescue part. I'll update

app/controllers/api/export_solutions_controller.rb
```diff
@@ -18,13 +18,15 @@ def ensure_maintainer!
     end
 
     def use_track
-      @track = Track.find_by(slug: params[:track_slug])
-      render_track_not_found if @track.blank?
+      @track = Track.find_by!(slug: params[:track_slug])
```
iHiD:
```suggestion
      @track = Track.find(params[:track_slug])
```

iHiD: The `FriendlyId` gem means that we can just use `find` here.

app/controllers/api/export_solutions_controller.rb
```diff
@@ -18,13 +18,15 @@ def ensure_maintainer!
...
+    rescue ActiveRecord::RecordNotFound
+      render_track_not_found
     end
 
     def use_exercise
-      @exercise = @track.exercises.find_by(slug: params[:exercise_slug])
-      render_exercise_not_found if @exercise.blank?
+      @exercise = @track.exercises.find_by!(slug: params[:exercise_slug])
```
iHiD:
```suggestion
      @exercise = @track.exercises.find(params[:exercise_slug])
```

iHiD (approved):
I'll let you merge this on Tues in case rack attack causes issues.

PR #2853, opened 2022-08-14
Auto update solutions if latest tests pass

iHiD:
When the latest iteration of a solution passes the tests, automatically update that to the latest version to stop users having to click lots. This is configurable via a user preference in settings, which is on by default.

There are two proper atomic commits to review in this PR for the backend and the preference. Then there are a few code-review commits after that.

---

## UI Changes

Screenshot 2022-08-15 at 00 45 54


app/javascript/components/settings/UserPreferencesForm.tsx
```diff
@@ -0,0 +1,104 @@
...
+}
+
+const SuccessMessage = () => {
+  return (
+    <div className="status success">
+      <Icon icon="completed-check-circle" alt="Success" />
+      Your preferences have been updated
+    </div>
+  )
+}
```
iHiD, line 104: This is copied from communication preferences. I'd rather not hold up merging on this being perfect, but rather if there's anything that needs improving here, we can do this in a PR that updates both this and communication preferences.

app/models/user/preferences.rb
```diff
@@ -0,0 +1,11 @@
+class User::Preferences < ApplicationRecord
```
iHiD, line 1: I considered `Preferences` or `Settings` or `GeneralPreferences` and eventually went for this.

app/models/solution.rb
```diff
@@ -111,6 +111,7 @@ def update_latest_iteration_head_tests_status!(status)
 
     update_column(:latest_iteration_head_tests_status, status)
     Solution::SyncToSearchIndex.defer(self)
+    Solution::AutoUpdateToLatestExerciseVersion.(self)
```
iHiD, line 114: All the guarding is done in auto-update itself so there's none here. I considered deferring this, but as it's happening already in a background request and it's pretty lightweight, I decided against it.

db/migrate/20220814230425_create_user_preferences.rb
```diff
@@ -0,0 +1,15 @@
+class CreateUserPreferences < ActiveRecord::Migration[7.0]
+  def change
+    create_table :user_preferences do |t|
```
iHiD:
```suggestion
    create_table :user_preferences, if_not_exists: true do |t|
```

db/migrate/20220814230425_create_user_preferences.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.belongs_to :user, foreign_key: true, index: { unique: true }
+      t.boolean :auto_update_exercises, default: true, null: false
+
+      t.timestamps
+    end
+
+    # Run on Bastion in production
+    unless Rails.env.production?
+      User.find_each(&:create_preferences)
+    end
```
iHiD, line 13: We should run the create table **and** this block **before** deploying, and then the block again (with guards for `users.id>xxx`) once we're deployed. There will be a small window where we could break things for new users while that runs, so maybe we run it in a loop on bastion until we're fully deployed.

config/initializers/rack_attack.rb
```diff
@@ -8,6 +8,7 @@
   next 30 if req.post? && req.path =~ %r{^/api/v2/markdown/parse}
   next 30 if req.patch? && req.path =~ %r{^/api/v2/mentoring/testimonials/[^/]+/reveal}
   next 20 if req.patch? && req.path =~ %r{^/api/v2/reputation/[^/]+/mark_as_seen}
+  next 8 if req.patch? && req.path == '/api/v2/settings/user_preferences'
```
ErikSchierboom: This bit of the rack attack has been changed. Could you rebase and update?

test/factories/user/preferences.rb
```diff
@@ -1,5 +1,8 @@
 FactoryBot.define do
   factory :user_preferences, class: 'User::Preferences' do
-    user { create :user }
+    # Creating a user makes preferences, so this dance
+    # deletes the one that gets created by defualt and
```
ErikSchierboom:
```suggestion
    # deletes the one that gets created by default and
```

app/commands/solution/sync_latest_iteration_head_tests_status.rb
```diff
@@ -16,9 +16,8 @@ def call
       status = :errored
     end
 
-    return true if solution.latest_iteration_head_tests_status == status
-
-    solution.update_latest_iteration_head_tests_status!(status)
+    # Only update it if we need to.
+    solution.update_latest_iteration_head_tests_status!(status) unless solution.latest_iteration_head_tests_status == status
```
ErikSchierboom, line 20: I assume that the `unless` part was unnecessary as it would be a `noop` if Rails detected no change in the value? Or is that incorrect?

app/serializers/serialize_user_preferences.rb
```diff
@@ -0,0 +1,22 @@
+class SerializeUserPreferences
+  include Mandate
+
+  initialize_with :preferences
+
+  def call
+    {
+      automation: serialize(User::Preferences.automation_keys)
+    }
+  end
```
ErikSchierboom: Just maybe?

```suggestion
  def call = { automation: serialize(User::Preferences.automation_keys) }
```

config/locales/user_preferences/en.yml
```diff
@@ -0,0 +1,3 @@
+en:
+  user_preferences:
+    auto_update_exercises: "Automatically update my solutions to the latest versions of exercises if they pass the latest tests"
```
ErikSchierboom: I'm not entirely sure, but should this be: 

```suggestion
    auto_update_exercises: "Automatically update my solutions to the latest version of exercises if they pass the latest tests"
```

or maybe even:

```suggestion
    auto_update_exercises: "Automatically update my solutions to the latest version of exercises if they pass the latest version's tests"
```

db/schema.rb
```diff
@@ -845,6 +845,14 @@
     t.index ["uuid"], name: "index_user_notifications_on_uuid", unique: true
   end
 
+  create_table "user_preferences", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.bigint "user_id"
+    t.boolean "auto_update_exercises", default: true, null: false
```
ErikSchierboom, line 850: This defaulting to `true` might be something we want to mention in the newsletter.

ErikSchierboom (approved):
Nice

app/commands/solution/sync_latest_iteration_head_tests_status.rb
iHiD: This is actually the same as what the code was. I just didn't feel the premature return was the clearest way once I added other clauses here (which later got deleted). So this change is unrelated to this.

dem4ron (approved):

ErikSchierboom (commented):
Would it be possible to have a system test that verifies this is working correctly?

app/models/solution.rb
```diff
@@ -107,10 +107,10 @@ def update_published_iteration_head_tests_status!(status)
...
   def update_latest_iteration_head_tests_status!(status)
-    return if latest_iteration_head_tests_status == status.to_sym
-
-    update_column(:latest_iteration_head_tests_status, status)
-    Solution::SyncToSearchIndex.defer(self)
+    unless latest_iteration_head_tests_status == status.to_sym
+      update_column(:latest_iteration_head_tests_status, status)
+      Solution::SyncToSearchIndex.defer(self)
+    end
     Solution::AutoUpdateToLatestExerciseVersion.(self)
```
ErikSchierboom: I think I hit this problem when I tested things locally. I had already put `puts` statements around these lines :)

ErikSchierboom (approved):

app/models/solution.rb
iHiD: :D!

Nice! System test caught something 🦜

PR #2854, opened 2022-08-14
Add docker setup to bin/dev

iHiD:
I figure it's useful to have this. If the containers are already there the first two lines just error and continue.

I'd also love to get `setup_local_exercism_aws` in there too, but as that relies on waiting for docker, and it should open happen once, it's harder.

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -1,3 +1,5 @@
 #!/usr/bin/env bash
 
+docker run -dp 3042:8080 -p 3040:4566 -p 3041:4566 localstack/localstack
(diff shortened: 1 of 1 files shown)


dem4ron (commented):
awesome!

ErikSchierboom (approved):
Nice

ErikSchierboom:
> I'd also love to get setup_local_exercism_aws in there too, but as that relies on waiting for docker, and it should open happen once, it's harder.

Let's do that later. Would be really useful

PR #2856, opened 2022-08-16
Add missing zip gem

ErikSchierboom:
It looks like I had the `zip` gem installed locally, which is why it worked on my machine but it doesn't on the server.


iHiD (approved):
Nice work :)

PR #2857, opened 2022-08-16
Cache objects not just SQL

iHiD:
Speeds up basic requests by about 10%. I suspect for complex ones it'll speed it up even more but I don't know how to pass sessions around via CURL etc to test things while logged in :)


ErikSchierboom (approved):

ErikSchierboom:
It looks good.

PR #2858, opened 2022-08-17
Improve overall website speed by 2x by using Turbo Frames

iHiD:
@ErikSchierboom Give this branch a whirl and see if things feel faster to you.


app/controllers/concerns/body_class_concern.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  memoize
+  def body_class
+    classes = []
+    classes << "devise" if devise_controller?
+    classes << "namespace-#{namespace_name}"
+    classes << "controller-#{controller_name || 'none'}"
+    classes << "action-#{action_name}"
+    classes << "theme-light"
+    user_signed_in? ? classes << "user-signed_in" : classes << "user-signed_out"
```
ErikSchierboom: I assume rubocop doesn't let you do:

```suggestion
    classes << user_signed_in? ? "user-signed_in" : "user-signed_out"
```

app/helpers/layout_helper.rb
```diff
@@ -1,19 +1,8 @@
...
-    classes << "theme-light"
-    user_signed_in? ? classes << "user-signed_in" : classes << "user-signed_out"
-    classes.join(" ")
-  end
-
   def js_packs
     [
       ('application' unless landing_page?),
-      ('internal' if user_signed_in?),
+      ('internal' if controller.user_signed_in?),
```
ErikSchierboom: What does this change do?

app/helpers/react_components/react_component.rb
```diff
@@ -11,7 +11,7 @@ def to_s(id, data, fitted: false, wrapper_class_modifier: nil, style: nil)
         "",
         class: css_classes.join(" "),
         style:,
-        "data-react-#{id}": true,
+        "data-react-id": id,
```
ErikSchierboom, line 14: Makes much more sense this way

app/helpers/layout_helper.rb
iHiD: It got left in error. Removing.

dem4ron (approved):
Great!

app/controllers/application_controller.rb
```diff
@@ -58,24 +60,11 @@ def set_request_context
...
-
-  def disable_site_header!
-    @__render_site_header__ = false
-  end
-
-  def self.disable_site_header!(*args)
-    before_action :disable_site_header!, *args
+  private
+  def set_body_class_header
+    response.set_header("Exercism-Body-Class", body_class)
```
ErikSchierboom, line 65: Am I correct in assuming that this header is also sent for all API requests? If so, it might be a micro-optimization to not do that 🤷

app/controllers/concerns/body_class_concern.rb
```diff
@@ -0,0 +1,32 @@
...
+  extend ActiveSupport::Concern
+  extend Mandate::Memoize
+
+  memoize
+  def body_class
+    classes = []
+    classes << "devise" if devise_controller?
+    classes << "namespace-#{namespace_name}"
+    classes << "controller-#{controller_name || 'none'}"
+    classes << "action-#{action_name}"
```
ErikSchierboom, line 11: Question: it looks like `controller_name` can be empty. Can the action name be empty too?

app/javascript/utils/react-bootloader.jsx
```diff
@@ -24,27 +24,65 @@ Bugsnag.start({
...
     renderComponents(parentElement, mappings)
     renderTooltips(parentElement, mappings)
   }
 
+  // document.addEventListener('turbo:before-fetch-response', (e) => {
+  //   console.log(Date.now(), 'turbo:before-fetch-response')
+  // })
+  // document.addEventListener('turbo:render', (e) => {
+  //   console.log(Date.now(), 'turbo:render')
+  // })
```
ErikSchierboom, line 68: Did you purposefully leave these comments in the code? (same goes for other comments in this file)

app/javascript/utils/react-bootloader.jsx
```diff
@@ -53,55 +91,62 @@ const render = (elem, component) => {
...
-  for (const [name, generator] of Object.entries(mappings)) {
-    const selector = '[data-react-' + name + ']'
-    parentElement.querySelectorAll(selector).forEach((elem) => {
-      const data = JSON.parse(elem.dataset.reactData)
-      render(elem, generator(data, elem))
-    })
+  const elems = Array.from(
+    parentElement.getElementsByClassName('c-react-component')
+  )
+  for (let elem of elems) {
```
ErikSchierboom, line 117:
```suggestion
  for (const elem of parentElement.getElementsByClassName('c-react-component')) {
```

app/javascript/utils/react-bootloader.jsx
```diff
@@ -53,55 +91,62 @@ const render = (elem, component) => {
...
-      const component = generator(elem.dataset, elem)
-
-      const tooltipElem = document.createElement('span')
-      elem.insertAdjacentElement('afterend', tooltipElem)
-
-      render(
-        tooltipElem,
-        <ExercismTippy content={component} reference={elem} />
-      )
+      renderTooltip(mappings, elem)
```
ErikSchierboom: Instead of 

```js
.forEach((elem) => {
  renderTooltip(mappings, elem)
})
```

you can do:

```js
.forEach((elem) => renderTooltip(mappings, elem))
```

app/javascript/utils/react-bootloader.jsx
```diff
@@ -53,55 +91,62 @@ const render = (elem, component) => {
...
-      }, 1)
+      // console.log(Date.now(), 'rendered')
+      elem.classList.add('--hydrated')
     }
   )
 
   const unloadOnce = () => {
     ReactDOM.unmountComponentAtNode(elem)
     document.removeEventListener('turbo:before-render', unloadOnce)
   }
```
ErikSchierboom, line 102: This code is now no longer used. Was that intentional?

app/javascript/utils/react-bootloader.jsx
```diff
@@ -24,27 +24,65 @@ Bugsnag.start({
...
+  //   console.log(Date.now(), 'turbo:before-fetch-response')
+  // })
+  // document.addEventListener('turbo:render', (e) => {
+  //   console.log(Date.now(), 'turbo:render')
+  // })
+
   // This adds rendering for all future turbo clicks
-  document.addEventListener('turbo:load', () => {
-    console.log('Loading React from Turbo Load')
+  document.addEventListener('turbo:load', (e) => {
```
ErikSchierboom:
```suggestion
  document.addEventListener('turbo:load', () => {
```

app/javascript/utils/react-bootloader.jsx
```diff
@@ -24,27 +24,65 @@ Bugsnag.start({
...
+  }
+
+  document.querySelector(`#${styleElemId}`).textContent = style
+}
+document.addEventListener('turbo:before-fetch-request', () => {
+  setTurboStyle('* { cursor:wait !important }')
+})
+document.addEventListener('turbo:before-render', () => {
+  setTurboStyle('')
+})
```
ErikSchierboom, line 54: You could make this one-liners:

```suggestion
document.addEventListener('turbo:before-fetch-request', () => setTurboStyle('* { cursor:wait !important }'))
document.addEventListener('turbo:before-render', () => setTurboStyle(''))
```

app/javascript/utils/react-bootloader.jsx
```diff
@@ -24,27 +24,65 @@ Bugsnag.start({
...
+    'exercism-body-class'
+  )
+  document.querySelector('body').className = bodyClass
+})
+
+/********************************************************/
+/** Add a loading cursor when a turbo-frame is loading **/
+/********************************************************/
+const setTurboStyle = (style) => {
+  const styleElemId = 'turbo-style'
```
ErikSchierboom: Prevent the `styleElemId` from being allocated each time the `setTurboStyle` function is called

```suggestion
const styleElemId = 'turbo-style'
const setTurboStyle = (style) => {  
```

ErikSchierboom (approved):

app/javascript/utils/react-bootloader.jsx
dem4ron: This might be auto-formatted, but I think it is more readable in 3 lines than 1

app/controllers/concerns/body_class_concern.rb
iHiD: No, it can't. I have no idea under what scenario `controller_name` can be empty either, but this bit of code has followed me around for a decade so I'll trust some past version of myself on it!

app/javascript/utils/react-bootloader.jsx
iHiD:
```suggestion
  // As getElementsByClassName returns a live collection, it is recommended to use Array.from
  // when iterating through it, otherwise the number of elements may change mid-loop.
  const elems = Array.from(
    parentElement.getElementsByClassName('c-react-component')
  )
  for (let elem of elems) {
```

app/javascript/utils/react-bootloader.jsx
iHiD: I thought our linter would do this automatically for me 😢

app/controllers/application_controller.rb
iHiD: Good shout. Also done SPI controller. Done :+1:

app/javascript/utils/react-bootloader.jsx
iHiD: Yes, for now. They're a really useful set for benchmarking everything. Let's tidy when we're happy with the speed.

app/javascript/utils/react-bootloader.jsx
iHiD: Yeah as we now have things that persist (e.g. the ones in the header). We can probably make this more sophisticated, but I think it's ok to remove for now. If we see JS errors as a result, we can do the "more sophisticated" bit sooner, and add flags for whether things should unload or not, but I'm happy to try it without.

app/javascript/utils/react-bootloader.jsx
ErikSchierboom:
```suggestion
  const elems = Array.from(
    parentElement.getElementsByClassName('c-react-component')
  )
  for (const elem of elems) {
```

app/controllers/concerns/body_class_concern.rb
iHiD: Seems to work with brackets! 🎉

PR #2860, opened 2022-08-17
Remove n+1 from track page

iHiD:
I can't see any reason why we can't delegate to the db column here and avoid an n+1 lookup

### DIFF
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -11,7 +11,7 @@ class Exercise < ApplicationRecord
     deprecated: 3
   }
 
-  belongs_to :track, counter_cache: :num_exercises
+  belongs_to :track
 
   has_many :solutions, dependent: :destroy
   has_many :submissions, through: :solutions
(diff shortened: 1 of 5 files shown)


app/models/user_track/external.rb
```diff
@@ -10,7 +10,7 @@ def initialize(track)
       @track = track
     end
 
-    delegate :concepts, :num_concepts, :updated_at,
+    delegate :concepts, :num_exercises, :num_concepts, :updated_at,
```
ErikSchierboom, line 13: I think this might not work. 
Check the definition of `exercises`:

```ruby
def exercises
  enabled_exercises(track.exercises)
end
```

Whereas `num_exercises` contains the active _and_ inactive exercises IIRC.

iHiD: > Whereas num_exercises contains the active and inactive exercises IIRC

Interesting. I think we should change `num_exercises` to just be active + beta then. I can't see any point in us having just a total count.

ErikSchierboom: Agreed. That would be nice

app/models/exercise.rb
```diff
@@ -70,6 +70,10 @@ def self.for(track_slug, exercise_slug)
     end
   end
 
+  after_commit do
+    track.recache_num_exercises! if (saved_changes.keys & %w[id status]).present?
```
ErikSchierboom, line 74: Interesting. Your using the fact the `id` column changes if the record is deleted, right?

ErikSchierboom (approved):

app/models/exercise.rb
iHiD: Yup!

iHiD:
I've run the recache in production.

PR #2863, opened 2022-08-17
Perf tweaks

iHiD:
Three more performance updates.

- The first one should insulate us from any locking pain. 
- The second should speed up the query as less data will be used. 
- The third one knocks about 25ms off our power-users' requests.

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -20,7 +20,7 @@ class AboutController < ApplicationController
 
   private
   def use_num_individual_supporters
-    @num_individual_supporters = User::AcquiredBadge.joins(:user).includes(:user).
+    @num_individual_supporters = User::AcquiredBadge.includes(:user).
       where(badge_id: Badge.find_by_slug!("supporter")). # rubocop:disable Rails/DynamicFindBy
       count
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,7 +6,7 @@ class ApplicationController < ActionController::Base
   before_action :store_user_location!, if: :storable_location?
   before_action :authenticate_user!
   before_action :ensure_onboarded!
-  before_action :mark_notifications_as_read!
+  around_action :mark_notifications_as_read!
   before_action :set_request_context
 
   def process_action(*args)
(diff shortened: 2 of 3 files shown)


app/controllers/about_controller.rb
```diff
@@ -20,7 +20,7 @@ def supporter_gobridge
 
   private
   def use_num_individual_supporters
-    @num_individual_supporters = User::AcquiredBadge.joins(:user).includes(:user).
+    @num_individual_supporters = User::AcquiredBadge.includes(:user).
       where(badge_id: Badge.find_by_slug!("supporter")). # rubocop:disable Rails/DynamicFindBy
```
ErikSchierboom, line 24: Out of curiosity, why don't we do:

```suggestion
      where(badge_id: Badge.find_by!(slug: "supporter")).
```

app/controllers/application_controller.rb
```diff
@@ -30,16 +30,28 @@ def ensure_admin!
     redirect_to maintaining_root_path
   end
 
+  # We want to mark relevant notifications as read, but we don't
+  # care about doing this before the rest of the action is run, so we
+  # use a promise to kick it off async. However, we do want it to finish
+  # before we send the response (which loads notifications async) so we
+  # wait for the promise to finish before leaving this block.
```
ErikSchierboom, line 37: Neat

app/controllers/sitemaps_controller.rb
```diff
@@ -1,6 +1,12 @@
 class SitemapsController < ApplicationController
```
ErikSchierboom, line 1: How often are the sitemap pages accessed? I assume web crawlers check them, but is that often or fairly irregularly?

ErikSchierboom (approved):
Neat

app/controllers/about_controller.rb
iHiD: Good question! `find_by_slug` is a specific method on `Badge`.

Screenshot 2022-08-18 at 12 56 11

app/controllers/sitemaps_controller.rb
iHiD: Often. I imagine multiple times per day, and the sql query takes about 40secs.

ErikSchierboom: Okay. I was wondering whether it would be a good idea to cache, but a couple of times a day is not worth it I think.

app/controllers/about_controller.rb
ErikSchierboom: Ah, I was confused. Could we use `for_slug!` instead to reduce the ambiguity?

PR #2864, opened 2022-08-17
Reduce unnecessary joins when retrieving discussions

iHiD:
@glennj This **might** help if you're getting sporadic slow requests here, but it's hard to test.

### DIFF
--- a/app/controllers/api/mentoring/discussions_controller.rb
+++ b/app/controllers/api/mentoring/discussions_controller.rb
@@ -40,7 +40,7 @@ module API
         current_user,
         params[:status],
         sorted: false, paginated: false
-      ).joins(solution: [exercise: :track]).group(:track_id).count
+      ).joins(solution: {exercise: :track}).group(:track_id).count
 
       tracks = Track.where(id: track_counts.keys).order(:title)
       data = tracks.map do |track|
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):
LGTM! There was a build failure: https://github.com/exercism/website/runs/7887521652?check_suite_focus=true
This actually showed that less data is retrieved by default! I fixed it in https://github.com/exercism/website/pull/2864/commits/90b54f218f01ea50d410182d53c9056a7aa4445f

app/controllers/api/mentoring/discussions_controller.rb
```diff
@@ -40,7 +40,7 @@ def tracks
         current_user,
         params[:status],
         sorted: false, paginated: false
-      ).group(:track_id).count
+      ).joins(solution: [exercise: :track]).group(:track_id).count
```
iHiD:
```suggestion
      ).joins(solution: {exercise: :track}).group(:track_id).count
```

iHiD:
Nice - thanks for fixing! Updated the syntax slightly 🙂

app/controllers/api/mentoring/discussions_controller.rb
```diff
@@ -40,7 +40,7 @@ def tracks
         current_user,
         params[:status],
         sorted: false, paginated: false
-      ).group(:track_id).count
+      ).joins(solution: {exercise: :track}).group(:track_id).count
```
iHiD:
```suggestion
      ).joins(solution: { exercise: :track }).group(:track_id).count
```

PR #2868, opened 2022-08-18
Add automation page

ErikSchierboom:
Closes https://github.com/exercism/project-management/issues/29
Closes https://github.com/exercism/project-management/issues/42
Closes https://github.com/exercism/project-management/issues/24


app/assemblers/assemble_exercise_representations_with_feedback_list.rb
```diff
@@ -0,0 +1,36 @@
+class AssembleExerciseRepresentationsWithFeedbackList
+  include Mandate
+
+  initialize_with :user, :params
+
+  def self.keys = %i[page order criteria track_slug]
```
iHiD: I'd pop these above the initializer I think.

app/assemblers/assemble_exercise_representations_with_feedback_list.rb
```diff
@@ -0,0 +1,36 @@
+class AssembleExerciseRepresentationsWithFeedbackList
```
iHiD: I'd probably get rid of `List` on this and the second one.

app/commands/exercise/representation/update_num_submissions.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  initialize_with :representation
+
+  def call
+    # We're updating in a single query instead of two queries to avoid race-conditions
+    # and using read_committed to avoid deadlocks
+    ActiveRecord::Base.transaction(isolation: Exercism::READ_COMMITTED) do
+      Exercise::Representation.
+        where(id: representation.id).
+        update_all("num_submissions = (#{num_submissions_sql})")
```
iHiD, line 14: I have a **slight** bias here to this being `num_iterations`. As we only create representations for the submissions of representations, I feel like it's a clearer name as to the intention of the column (even though it's measuring submissions under the hood). But my bias is only very slight, so feel free to overrule it if you prefer `num_submissions`.

app/commands/exercise/representation/update_num_submissions.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+  end
+
+  private
+  def num_submissions_sql
+    Arel.sql(
+      Submission::Representation.
+        joins(submission: :solution).
+        where(submissions: { solutions: { exercise: representation.exercise } }).
```
iHiD: You can do this to avoid loading the exercise into memory.

```suggestion
        where(submissions: { solutions: { exercise_id: representation.exercise_id } }).
```

app/controllers/mentoring/automation_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Mentoring::AutomationController < ApplicationController
+  before_action :ensure_mentor!
```
iHiD: We'll want a new method for this. Let's call them `supermentors` (and we can rep-gate and num_solutions and average_rating that)

app/commands/exercise/representation/update_num_submissions.rb
ErikSchierboom: > As we only create representations for the submissions of representations

Did you mean to write:

> As we only create representations for the iterations of submissions

?

app/models/user.rb
```diff
@@ -4,6 +4,8 @@ class User < ApplicationRecord
 
   SYSTEM_USER_ID = 1
   GHOST_USER_ID = 720_036
+  SUPERMENTOR_MIN_REP = 500
+  SUPERMENTOR_MIN_SATISFACTION_PERCENTAGE = 95
```
ErikSchierboom: What would be nice place to store this information?

ErikSchierboom: I'm also unsure about the actual values. I just came up with these 🤷

app/commands/exercise/representation/update_num_submissions.rb
iHiD: I mean to write "for the submissions of iterations". Sorry.

app/assemblers/assemble_exercise_representations_with_feedback.rb
```diff
@@ -22,7 +22,7 @@ def representations
       mentor:,
       track:,
       status: :with_feedback,
-      criteria: params[:criteria],
+      criteria: params.fetch(:criteria, ''),
```
iHiD: Why is this needed?

ErikSchierboom: Same reason as https://github.com/exercism/website/pull/2872

iHiD: I don't understand 🙂 

This is passing it into a search method, whereas #2872 is changing what is rendered out (in the `query:` line). Unless I'm missing something, this won't change what React sees and is effectively a noop.

app/assemblers/assemble_exercise_representations_without_feedback.rb
```diff
@@ -22,7 +22,8 @@ def representations
       mentor:,
       track:,
       status: :without_feedback,
-      criteria: params[:criteria],
+      only_mentored_solutions: !!params[:only_mentored_solutions],
```
iHiD: Is this needed? Can we not rely on truthy/falsy? If it's needed, can you add a comment to the code explaining why pls.

app/assemblers/assemble_exercise_representations_without_feedback.rb
```diff
@@ -22,7 +22,8 @@ def representations
...
+      criteria: params.fetch(:criteria, ''),
```
iHiD: I don't think this is needed either. I think this can be `params[:criteria]` pending discussion on the other comment around this.

app/helpers/react_components/mentoring/representations/with_feedback.rb
```diff
@@ -0,0 +1,62 @@
...
+          { value: :recent, label: 'Sort by recent first' }
+        ].freeze
+        private_constant :SORT_OPTIONS
+
+        private
+        def representations_request
+          {
+            endpoint: Exercism::Routes.with_feedback_api_mentoring_representations_url,
+            query: {
+              criteria: params[:criteria],
```
iHiD: This is where I think you need the `fetch(:criteria, "")`

app/helpers/react_components/mentoring/representations/with_feedback.rb
```diff
@@ -0,0 +1,62 @@
...
+              sort_options: SORT_OPTIONS
+            }
+          )
+        end
+
+        SORT_OPTIONS = [
+          { value: :num_occurances, label: 'Sort by highest occurance' },
+          { value: :recent, label: 'Sort by recent first' }
+        ].freeze
+        private_constant :SORT_OPTIONS
```
iHiD: Move to the bottom of the file pls.

app/assemblers/assemble_representation_tracks_for_select.rb
```diff
@@ -0,0 +1,24 @@
+class AssembleRepresentationTracksForSelect
+  include Mandate
+
+  initialize_with :mentor, :status
```
iHiD: As there are only two statuses, I'd prefer a `with_feedback: ` here I think. Saves for typos with symbols. Unless you think there might ever be a third status?

app/assemblers/assemble_tracks_for_select.rb
```diff
@@ -1,19 +1,12 @@
 class AssembleTracksForSelect
   include Mandate
 
-  def initialize(tracks = default_tracks)
-    @tracks = tracks
-  end
+  initialize_with tracks: ::Track.active
```
iHiD: I'm not 100% sure if this is the same or not. I have a feeling that this will execute every time, even if `tracks` is passed in, whereas a method will only execute if `tracks:` is nil. But I'm only about 50% sure, so give it a check?

app/commands/exercise/representation/create_or_update.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      def call
+        representation = Exercise::Representation.create_or_find_by!(exercise:, ast_digest:) do |rep|
+          rep.source_submission = submission
+          rep.ast = ast
+          rep.mapping = mapping
+          rep.last_submitted_at = last_submitted_at
+        end
+
+        representation.update!(last_submitted_at:) unless representation.new_record?
```
iHiD: It should never be possible for `representation` to be a new record here, as the `create_or_find_by!` has a `!` at the end.

app/commands/exercise/representation/create_or_update.rb
```diff
@@ -0,0 +1,25 @@
...
+        Exercise::Representation::UpdateNumSubmissions.defer(representation)
```
iHiD, line 17: If we do want to allow unsaved records, this line needs guarding too.

app/commands/exercise/representation/search.rb
```diff
@@ -0,0 +1,79 @@
+class Exercise::Representation::Search
+  include Mandate
+
+  # Use class method rather than constant for easier stubbing during testing
+  def self.requests_per_page = 20
+
+  initialize_with criteria: nil, status: nil, mentor: nil, track: nil, order: :most_submissions,
```
iHiD: Can this ever be correctly called without a `status`? If not, it shouldn't be `nil` and I feel like it should probably be a boolean.

app/commands/exercise/representation/search.rb
```diff
@@ -0,0 +1,79 @@
...
+    return if track.blank?
+
+    @representations = @representations.for_track(track)
+  end
+
+  def filter_criteria!
+    return if criteria.blank?
+
+    @representations = @representations.where('exercises.title LIKE ?', "%#{criteria}%").
+      or(@representations.where('exercises.slug LIKE ?', "%#{criteria}%"))
```
iHiD: This will be much faster as two queries so we can use `exercises.id` key (or even just as a much simpler search column). Firstly get the exercises that match these things (with the context of the track selector), then use `exercises.id` in this query.

app/commands/exercise/representation/search.rb
```diff
@@ -0,0 +1,79 @@
...
+    @representations = @representations.mentored_by(mentor)
+  end
+
+  def sort!
+    case order
+    when :most_recent
+      @representations = @representations.order(last_submitted_at: :desc)
+    when :most_submissions
+      @representations = @representations.order(num_submissions: :desc)
+    end
```
iHiD, line 71: When considering indexes, we'll need two seperate indexes for these two, with these in the `order` position in the index.

app/javascript/components/common/Introducer.tsx
```diff
@@ -54,7 +56,7 @@ export const Introducer = ({
     .join(' ')
 
   return (
-    <div ref={ref} className={classNames}>
+    <div ref={ref} className={`${classNames} ${additionalClassNames}`}>
```
iHiD, line 59: I feel like maybe here the public key should be `className` and the concatenated key should be `baseClassNames`. That seems more consistent with having `className` as a public thing in general.

app/javascript/components/mentoring/automation/RepresentationList.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <Component {...props} />
+    </FetchingBoundary>
+  )
+}
+
+const Component = ({ resolvedData, latestData, page, setPage }: Props) => {
+  return (
+    <>
+      {resolvedData && resolvedData.results && (
+        <React.Fragment>
```
iHiD, line 46: Let's be consistent with either `<>` or `<React.Fragment>`. I prefer the first personally but I'm not fussy.

app/helpers/react_components/mentoring/representations/with_feedback.rb
```diff
@@ -0,0 +1,62 @@
...
+              page: params[:page]
+            }.compact,
+            options: {
+              initial_data: { representations: },
+              stale_time: 5000 # milliseconds
+            }
+          }
+        end
+
+        def representations = AssembleExerciseRepresentationsWithoutFeedback.(mentor, params)
```
iHiD: I feel that this should be `WithFeedback` not `WithoutFeedback`?

app/assemblers/assemble_exercise_representations_with_feedback.rb
ErikSchierboom: Ehm, feck me. You're right 🤦

app/assemblers/assemble_exercise_representations_without_feedback.rb
ErikSchierboom: > Can we not rely on truthy/falsy? 

We can. I'll update

app/javascript/components/mentoring/automation/RepresentationList.tsx
ErikSchierboom: I also prefer the first.

app/assemblers/assemble_representation_tracks_for_select.rb
ErikSchierboom: 🤷 I'm usually fairly against using boolean values, but it might be okay here.

app/helpers/view_components/mentor/header.rb
```diff
@@ -47,43 +47,60 @@ def stats
...
-
-          link_to(
-            Exercism::Routes.mentoring_testimonials_path,
-            class: tab_class(:testimonials)
-          ) do
-            graphical_icon(:testimonials) +
-              tag.span("Testimonials") +
-              tag.span(number_with_delimiter(num_testimonials), class: 'count')
-          end,
+      def tabs = [workspace_tab, queue_tab, testimonials_tab, automation_tab]
```
ErikSchierboom, line 50: The automation tab has logic in it to determine whether it is enabled or not. That led me to extracting the tabs into separate methods.

app/models/submission/representation.rb
```diff
@@ -9,6 +9,8 @@ def self.digest_ast(ast)
   end
 
   belongs_to :submission
+  belongs_to :mentor, optional: true, class_name: "User"
```
iHiD, line 12:
```suggestion
  belongs_to :mentored_by, optional: true, class_name: "User"
```

app/commands/user/reset_account.rb
```diff
@@ -46,6 +46,7 @@ def reset_mentoring!
       user.mentor_discussion_posts.update_all(user_id: User::GHOST_USER_ID)
       user.mentor_testimonials.update_all(mentor_id: User::GHOST_USER_ID)
       user.provided_testimonials.update_all(student_id: User::GHOST_USER_ID)
+      user.submission_representations.update_all(mentor_id: User::GHOST_USER_ID)
```
iHiD, line 49:
```suggestion
      user.submission_representations.update_all(mentored_by_id: User::GHOST_USER_ID)
```

app/commands/submission/representation/update_mentor.rb
```diff
@@ -0,0 +1,31 @@
+class Submission
+  class Representation
+    class UpdateMentor
```
iHiD, line 3:
```suggestion
    class UpdateMentoredBy
```

(etc)

PR #2871, opened 2022-08-19
Fix issue with non-iteration submissions and improve test performance

### LINKED ISSUE #6511: Export endpoint for testing the representer should return first iterations
I looked at 3 exercises so far. From the 500 solutions I got from the export endpoint, less then 200 compile and pass the tests. That makes the statistics rather "small".

Looking at those numbers, it feels to me the endpoint currently returns first submissions instead of first iterations (the successful ones coming from the CLI). I think it should return first iterations because this is what the representer will run on. Than, failing solutions can happen but should be rare.

(@ErikSchierboom here the issue as discussed)

iHiD:
- Fixes https://github.com/exercism/exercism/issues/6511
- Stops us from creating 500 solutions for the test (which took **ages**)

### DIFF
--- a/app/commands/exercise/export_solutions_to_zip_file.rb
+++ b/app/commands/exercise/export_solutions_to_zip_file.rb
@@ -10,7 +10,7 @@ class Exercise::ExportSolutionsToZipFile
   def call
     file_stream = Zip::OutputStream.write_buffer do |zip|
       solutions.each.with_index do |solution, idx|
-        submission = solution.iterations.first.submission
+        submission = solution.iterations.first&.submission
         next unless submission
 
         # Export the first iteration's files as that iteration won't have
--- a/app/commands/exercise/export_solutions_to_zip_file.rb
+++ b/app/commands/exercise/export_solutions_to_zip_file.rb
@@ -3,12 +3,15 @@ require 'zip'
 class Exercise::ExportSolutionsToZipFile
   include Mandate
 
+  def self.num_submissions = 500
+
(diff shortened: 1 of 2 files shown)


app/commands/exercise/export_solutions_to_zip_file.rb
```diff
@@ -3,12 +3,15 @@
...
+  def self.num_submissions = 500
+
   initialize_with :exercise
 
   def call
     file_stream = Zip::OutputStream.write_buffer do |zip|
       solutions.each.with_index do |solution, idx|
-        submission = solution.submissions.first
+        submission = solution.iterations.first.submission
+        next unless submission
```
ErikSchierboom, line 14: When would this condition be true? Just asking as this would cause the output to potentially not have 500 records. Which is fine, but might be odd.

ErikSchierboom (approved):

app/commands/exercise/export_solutions_to_zip_file.rb
iHiD: I was thinking if the first iteration is deleted. But we should probably change it to be `solution.iterations.first&.submission` for that case.

ErikSchierboom: It probably doesn't really matter much

iHiD: I think this will definitely fail without the guard for deleted iterations as that happens all the time. I'll update.

PR #2872, opened 2022-08-19
Don't cause track exercises data to become stale due to criteria being empty

### LINKED ISSUE #6505: When tracks/x/exercises loads, it immediately re-fetches the exercises
It shouldn't do this 🙂

ErikSchierboom:
The issue was not that there wasn't a stale time (there is by default), but that the criteria param is set to `nil` on the ruby side, but to an empty string on the React side (which is required for proper input handling)

Closes https://github.com/exercism/exercism/issues/6505

### DIFF
--- a/app/helpers/react_components/student/exercise_list.rb
+++ b/app/helpers/react_components/student/exercise_list.rb
@@ -12,18 +12,20 @@ module ReactComponents
 
       private
       def request
-        query = {
-          criteria: params[:criteria],
-          sideload: ["solutions"]
-        }.compact
-
         {
           endpoint: Exercism::Routes.api_track_exercises_path(track),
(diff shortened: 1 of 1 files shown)


iHiD:
Nice!!

iHiD (approved):

SleeplessByte:
You probably _also_ want to "fix" this on the client, as in make sure the _key_ is `query || ''`, so that it's the same cache key regardless of the nullness of the value.

ErikSchierboom:
> You probably also want to "fix" this on the client, as in make sure the key is query || '', so that it's the same cache key regardless of the nullness of the value.

It was already fixed on the hence, which is why there was a disconnect.

PR #2873, opened 2022-08-19
Upgrade to Mandate 2.2

ErikSchierboom:
- Bump Mandate to version 2.2
- Update Concept::Create
- Update ConceptExercise::Create
- Update PracticeExercise::Create
- Update ContributorTeam::Create
- Update ContributorTeam::Membership::CreateOrUpdate
- Update Github::Issue::CreateOrUpdate
- Update Github::PullRequest::CreateOrUpdate
- Update Github::Task::CreateOrUpdate
- Update Track::Create
- Update Metric::Queue
- Update User::ReputationToken::Create

### DIFF
--- a/app/assemblers/assemble_contributors.rb
+++ b/app/assemblers/assemble_contributors.rb
@@ -1,7 +1,7 @@
 class AssembleContributors
   include Mandate
 
-  initialize_with params: Mandate::KWARGS
+  initialize_with :params
 
   def call
     SerializePaginatedCollection.(
--- a/app/assemblers/assemble_contributors.rb
+++ b/app/assemblers/assemble_contributors.rb
@@ -1,7 +1,7 @@
 class AssembleContributors
   include Mandate
 
-  initialize_with :params
+  initialize_with params: Mandate::KWARGS
 
   def call
     SerializePaginatedCollection.(
--- a/app/serializers/serialize_tracks.rb
+++ b/app/serializers/serialize_tracks.rb
@@ -1,10 +1,7 @@
 class SerializeTracks
   include Mandate
 
-  def initialize(tracks, user = nil)
-    @tracks = tracks
-    @user = user
-  end
+  initialize_with :tracks, :user
 
   def call
     sorted_tracks.map do |track|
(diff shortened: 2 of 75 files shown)


iHiD:
@ErikSchierboom Rather than adding the `{}` everywhere, would changing https://github.com/exercism/website/blob/3cf9a7e4420f7169176ee24f9a44bf01b98fe209/app/commands/user/reputation_token/award_for_issue.rb#L6 to be `params: Mandate::KWARGS` work instead?

test/helpers/react_components/mentoring/inbox_test.rb
```diff
@@ -2,9 +2,9 @@
 
 class MentoringInboxTest < ReactComponentTestCase
   test "mentoring inbox rendered correctly" do
-    component = ReactComponents::Mentoring::Inbox.new(
+    component = ReactComponents::Mentoring::Inbox.new({
       criteria: "Ruby"
-    )
+    })
```
ErikSchierboom, line 7: I've left this to be using `{}`. I could change this, but I'd need to call `**@params` in a HAML template.

app/serializers/serialize_tracks.rb
```diff
@@ -1,10 +1,7 @@
 class SerializeTracks
   include Mandate
 
-  def initialize(tracks, user = nil)
-    @tracks = tracks
-    @user = user
-  end
+  initialize_with :tracks, :user
```
ErikSchierboom, line 4: I've converted `user` to be non-nil as it was always passed as an argument.

app/assemblers/assemble_contributors.rb
```diff
@@ -1,7 +1,9 @@
 class AssembleContributors
   include Mandate
 
-  initialize_with :params
+  initialize_with params: Mandate::KWARGS
+
+  def self.keys = %i[track_slug user_handle page category period]
```
ErikSchierboom: I needed to add this as otherwise I could not use `**` with Rails `params` in the calling controllers. This error was shown:

> ActionController::UnfilteredParameters: unable to convert unpermitted parameters to hash

PR #2903, opened 2022-08-23
Fix Scratchpad UI bug

### LINKED ISSUE #6519: Scratchpad Editor footer UI bug
the_bug

dem4ron:
Correct alignment of buttons in editor toolbar, better placement of Unsaved tag.

Closes [Issue #6519](https://github.com/exercism/exercism/issues/6519) 

### UI Change: 
Screenshot 2022-08-23 at 10 44 30

### DIFF
--- a/app/css/components/markdown-editor.css
+++ b/app/css/components/markdown-editor.css
@@ -2,6 +2,9 @@
 
 /* editor toolbar height = 49px */
 /* 30x30[px] button + 9[px] top-bot padding + 1[px] top border = 49px */
+:root {
+    --editor-toolbar-height: 49px;
+}
 
 .c-markdown-editor {
     @screen md {
@@ -112,8 +115,8 @@
         @apply absolute;
         bottom: 0;
         right: 0;
-        @apply p-8;
-        @apply flex items-center justify-end;
+        height: var(--editor-toolbar-height);
+        @apply flex items-center justify-end mr-8;
 
         & button {
             @apply ml-12;
@@ -122,8 +125,5 @@
     }
     & .editor-preview {
         @apply bg-backgroundColorA;
-    }
-    & .scratchpad-footer {
-        @apply py-[4.5px] px-8;
     }
 }
--- a/app/css/components/markdown-editor.css
+++ b/app/css/components/markdown-editor.css
@@ -1,5 +1,8 @@
 @import "../styles";
 
+/* editor toolbar height = 49px */
+/* 30x30[px] button + 9[px] top-bot padding + 1[px] top border = 49px */
+
 .c-markdown-editor {
     @screen md {
         min-width: 520px;
@@ -116,14 +119,11 @@
             @apply ml-12;
             animation: fadeIn 0.2s forwards;
         }
-
-        & .--warning {
-            @apply flex-grow;
-            @apply text-red font-medium;
-            @apply mr-12;
-        }
     }
     & .editor-preview {
         @apply bg-backgroundColorA;
+    }
+    & .scratchpad-footer {
+        @apply py-[4.5px] px-8;
     }
 }
--- a/app/css/components/markdown-editor.css
+++ b/app/css/components/markdown-editor.css
@@ -2,11 +2,9 @@
 
 /* editor toolbar height = 49px */
 /* 30x30[px] button + 9[px] top-bot padding + 1[px] top border = 49px */
-:root {
-    --editor-toolbar-height: 49px;
-}
 
 .c-markdown-editor {
+    --editor-toolbar-height: 49px;
     @screen md {
         min-width: 520px;
     }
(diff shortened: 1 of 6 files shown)


app/css/components/markdown-editor.css
```diff
@@ -116,14 +119,11 @@
...
-            @apply flex-grow;
-            @apply text-red font-medium;
-            @apply mr-12;
-        }
     }
     & .editor-preview {
         @apply bg-backgroundColorA;
     }
+    & .scratchpad-footer {
+        @apply py-[4.5px] px-8;
```
iHiD: Computer's can't render half pixels so this generally don't what you want. In this situation, it's better to set the height on the footer (set a variable above where your comment is) then do something like `flex flex-row items-center` and no vertical padding.

app/css/components/mentor/discussion.css
```diff
@@ -507,7 +507,7 @@
             & > .c-introducer {
                 @apply mb-20;
             }
-            & > .title {
+            & .title {
```
iHiD, line 510: Why this change? This will change any div with the class `.title` in the scratchpad to have this styling, rather than the one this was originally targetting. Is that what you intended?

iHiD: Looking at the JS, you might be better just inlining this CSS onto the `title` div now.

app/css/components/markdown-editor.css
dem4ron: Yes, that was one of my approaches, but then deleted the variable as I didn't use it for anything.

SleeplessByte: Computers _can_ render half pixels. It's implemented as sub-pixel rendering and work in most browsers. (But agree on the rest)

app/css/components/mentor/discussion.css
dem4ron: Yes, I moved the title a level deeper. because a level higher it spawned a one pixel invisible top something which only moved the icon 1px down. Unfortunately couldn't figure out why it did that. It still bugs me.. So I moved the title a level deeper, and that solved the icon shift. 

Then I will move the styles into inline classes.

app/css/components/markdown-editor.css
iHiD: My understanding is that most browsers (e.g. Chrome) round for paddings and margins and use sub-pixel rendering for borders, but even that is inconsistent. (e.g. this recent blog: https://medium.com/kajabi-ux/css-and-sub-pixel-rendering-the-case-of-the-clipped-border-4652c5a1b5ab)

Regardless, I'd rather stick to the range of paddings we have defined in Tailwind :)

app/css/components/mentor/discussion.css
iHiD: I'll check it out and take a look at the 1px thing :)

dem4ron: This is a what I used IIRC: (the effect is exactly the same)
```
       <div className="title justify-between">
        <div className="items-center py-[1px] flex">
          Your notes for <strong>{exercise.title}</strong> in
          <TrackIcon iconUrl={track.iconUrl} title={track.title} />
          <strong>{track.title}</strong>
        </div>

        {content === page.contentMarkdown ? null : <AlertTag>Unsaved</AlertTag>}
      </div>

```
you need to add the padding for the div under title, so the height matches the height of the AlertTag.

and you need to remove margin and alignment from AlertTag.tsx: 

```


import React from 'react'

export function AlertTag({
  children,
}: {
  children: React.ReactNode
}): JSX.Element {
  return (
    <div className="text-alert bg-veryLightRed text-13 rounded-8 font-semibold py-6 px-12 animate-fadeIn">
      {children}
    </div>
  )
}
```

app/css/components/markdown-editor.css
iHiD: @dem4ron Just this one to do, then we can merge this 🙂

dem4ron: ok!

dem4ron: Screenshot 2022-08-25 at 9 23 41

These are the results of the new commit on my machine. Checked it on Safari, Firefox, Chrome. Looks exactly the same.

app/css/components/markdown-editor.css
```diff
@@ -2,6 +2,9 @@
 
 /* editor toolbar height = 49px */
 /* 30x30[px] button + 9[px] top-bot padding + 1[px] top border = 49px */
+:root {
+    --editor-toolbar-height: 49px;
+}
```
iHiD: I think you don't need to store this in `:root` and can just scope it inside `.c-markdown-editor` below

iHiD:
Nice work dude 🙂

PR #2904, opened 2022-08-23
Use endless methods

ErikSchierboom:


iHiD (approved):

ErikSchierboom:
@iHiD I'll add some more endless methods to it. For now, I'm making this a draft. Low prio

iHiD:
Nice. Thoughts on grouping all endless methods together in class? Feels neater to me. Especially in a command.

ErikSchierboom:
> Nice. Thoughts on grouping all endless methods together in class? Feels neater to me. Especially in a command.

When they're next to each other, I already do that. But I'd rather not make this a rule, as I'd much prefer methods to be grouped by functionality, not by form.

iHiD (approved):

PR #2907, opened 2022-08-24
Add Impact page

iHiD:


app/commands/metric/create.rb
```diff
@@ -25,9 +28,18 @@ def call
...
+  def coordinates
+    geocoded&.coordinates.presence
+  end
+
+  memoize
+  def geocoded
     return if remote_ip.blank?
 
-    Geocoder.search(remote_ip).first&.country_code.presence
+    Geocoder.search(remote_ip).first
```
iHiD, line 44: @ErikSchierboom I've added an extra field to these metrics "coordinates"

app/controllers/about_controller.rb
```diff
@@ -18,6 +18,46 @@ def supporter_gobridge
...
+  #     '102.130.40.0',
+  #     '1.178.144.0',
+  #     '132.174.250.145'
+  #   ]
+  #   Metrics::SubmitSolutionMetric.last(6).map.with_index do |m, idx|
+  #     m.update(
+  #       coordinates: Geocoder.search(ips[idx]).first&.coordinates,
+  #       track: Track.order('rand()').first
+  #     )
+  #   end
```
iHiD, line 34: @ErikSchierboom You can run this to get some varied coordinates in your DB.

config/initializers/geocoder.rb
```diff
@@ -1,10 +1,14 @@
+db_file = Rails.env.production? ?
+  File.join('/usr', 'share', 'GeoIP', 'GeoLite2-City.mmdb') :
+  Rails.root.join('vendor', 'GeoLite2-City.mmdb')
```
iHiD, line 3: @ErikSchierboom Two changes: 
1. city not country DB
2. Local location has changed to vendor (so it can be consistent across OSes)

The dockerfile will need updating with the new db.

iHiD: (I've updated the Dockerfile too)

PR #2911, opened 2022-08-25
More metrics

iHiD:


app/commands/metric_period/search.rb
```diff
@@ -4,7 +4,7 @@ class MetricPeriod::Search
   PERIOD_TYPES = %i[month day minute].freeze
   METRIC_TYPES = %i[
     open_issue_metric open_pull_request_metric merge_pull_request_metric
-    submit_solution_metric complete_solution_metric publish_solution_metric
+    start_solution_metric complete_solution_metric publish_solution_metric
```
ErikSchierboom, line 7: Did you purposefully not include the two new metric typed here?

iHiD: No, bug. Next commit fixed it.

PR #2918, opened 2022-08-30
Improve performance of contributors page by a bazillion percent.

iHiD:
This changes the calculation of contextual data to use n queries (where n is number of users) rather than 1 nice big optimised query.

This is theoretically terrible but actually works out to be much more performant for two reasons:
1. MySQL's index performance ranges from 0.05s to 33s (!!) depending on how dispersed the indexes pages are. First website page load therefore takes 33s and subsequent ones are pretty instant. By reducing it down to n queries, we have a steady performance that's at the very low end of the range (~0.5s sum for all the queries)
2. We cache the values per user and invalidate the cache when a new reputation token is added, so although this is n queries in the worst case - we're never really actually there.

### DIFF
--- a/app/commands/user/reputation_token/calculate_contextual_data.rb
+++ b/app/commands/user/reputation_token/calculate_contextual_data.rb
@@ -11,7 +11,7 @@ class User::ReputationToken
     # By reducing it down to n queries, we have a steady performance that's at the very low
     # end of the range (~0.5s sum for all the queries)
     # 2. We cache the values per user and invalidate the cache when a new reputation token is
-    # added, so although this is n+1 in the worst case - we're never really actually there.
+    # added, so although this is n queries in the worst case - we're never really actually there.
     def initialize(user_ids, period: nil, earned_since: nil, track_id: nil, category: nil)
       @single_user = user_ids.is_a?(Integer)
       @user_ids = Array(user_ids)
--- a/app/commands/user/reputation_token/calculate_contextual_data.rb
+++ b/app/commands/user/reputation_token/calculate_contextual_data.rb
@@ -7,7 +7,7 @@ class User::ReputationToken
     # This does n queries (where n is number of users).
     # This is theoretically terrible but actually works out to be much more performant for two reasons:
     # 1. MySQL's index performance ranges from 0.05s to 33s (!!) depending on how dispersed the indexes pages
-    # are. First page load therefore takes 33s and subsequent ones are pretty instant. 
+    # are. First website page load therefore takes 33s and subsequent ones are pretty instant. 
     # By reducing it down to n queries, we have a steady performance that's at the very low
     # end of the range (~0.5s sum for all the queries)
     # 2. We cache the values per user and invalidate the cache when a new reputation token is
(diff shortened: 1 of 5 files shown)


app/commands/user/reputation_token/calculate_contextual_data.rb
```diff
@@ -4,6 +4,14 @@ class CalculateContextualData
     include ActionView::Helpers::TextHelper
     include ActionView::Helpers::NumberHelper
 
+    # This does n queries (where n is number of users).
+    # This is theoretically terrible but actually works out to be much more performant for two reasons:
+    # 1. MySQL's index performance ranges from 0.05s to 33s (!!) depending on how dispersed the indexes pages
+    # are. First page load therefore takes 33s and subsequent ones are pretty instant.
```
iHiD:
```suggestion
    # are. First website page load therefore takes 33s and subsequent ones are pretty instant. 
```

app/commands/user/reputation_token/calculate_contextual_data.rb
```diff
@@ -4,6 +4,14 @@ class CalculateContextualData
...
     include ActionView::Helpers::NumberHelper
 
+    # This does n queries (where n is number of users).
+    # This is theoretically terrible but actually works out to be much more performant for two reasons:
+    # 1. MySQL's index performance ranges from 0.05s to 33s (!!) depending on how dispersed the indexes pages
+    # are. First website page load therefore takes 33s and subsequent ones are pretty instant. 
+    # By reducing it down to n queries, we have a steady performance that's at the very low
+    # end of the range (~0.5s sum for all the queries)
+    # 2. We cache the values per user and invalidate the cache when a new reputation token is
+    # added, so although this is n+1 in the worst case - we're never really actually there.
```
iHiD:
```suggestion
    # added, so although this is n queries in the worst case - we're never really actually there.
```

app/models/user/reputation_token.rb
```diff
@@ -1,6 +1,10 @@
 class User::ReputationToken < ApplicationRecord
   include IsParamaterisedSTI
 
+  def self.cache_hash_for(user_id)
+    "users/#{user_id}/reputation"
+  end
```
ErikSchierboom:
```suggestion
  def self.cache_hash_for(user_id) = "users/#{user_id}/reputation"
```

:)

iHiD:
This needs to expire daily.

app/commands/user/reputation_token/calculate_contextual_data.rb
```diff
@@ -66,19 +72,37 @@ def format(value, thing, suffix = nil)
...
+    def with_cache(user)
+      user_key = User::ReputationToken.cache_hash_for(user)
+      value_key = ["contextual", earned_since, track_id, category].join("|")
+      redis = Exercism.redis_tooling_client
+      val = redis.hget(user_key, value_key)
+      return JSON.parse(val) if val
+
+      yield.tap do |val|
+        redis.hset(user_key, value_key, val.to_json)
+      end
```
ErikSchierboom, line 108: Neat

ErikSchierboom (approved):
Code looks good to me!

PR #2919, opened 2022-08-31
Add automation section to mentors dashboard

iHiD:
@ErikSchierboom @dem4ron I'm opening this so I can do a review 🙂 

## Run before deploy

```ruby
c = User.connection;0 

# Columns
c.add_column :exercise_representations, :num_submissions, :integer, limit: 3, null: true, if_not_exists: true
c.add_column :exercise_representations, :last_submitted_at, :datetime, null: false, default: -> { 'NOW(6)' }, if_not_exists: true
c.add_reference :exercise_representations, :track, null: true, foreign_key: true, if_not_exists: true
c.add_column :exercise_representations, :uuid, :string, null: true, index: { unique: true }, if_not_exists: true
c.add_reference :submission_representations, :mentored_by_id, null: true, foreign_key: { to_table: :users }, if_not_exists: true


# Indexes
c.add_index :exercise_representations, %i[exercise_id ast_digest], unique: false, if_not_exists: true
c.add_index :exercise_representations, :ast_digest
c.add_index :exercise_representations, %i[feedback_type track_id num_submissions], order: {feedback_type: :asc, track_id: :asc, num_submissions: :desc}, unique: false, name: 'index_exercise_representation_type_track_num_submissions', if_not_exists: true
c.add_index :exercise_representations, %i[feedback_type track_id last_submitted_at], order: {feedback_type: :asc, track_id: :asc, last_submitted_at: :desc}, unique: false, name: 'index_exercise_representation_type_track_last_submitted_at', if_not_exists: true
c.add_index :exercise_representations, %i[feedback_author_id track_id num_submissions], order: {feedback_author_id: :asc, track_id: :asc, num_submissions: :desc}, unique: false, name: 'index_exercise_representation_author_track_num_submissions', if_not_exists: true
c.add_index :exercise_representations, %i[feedback_author_id track_id last_submitted_at], order: {feedback_author_id: :asc, track_id: :asc, last_submitted_at: :desc}, unique: false, name: 'index_exercise_representation_author_track_last_submitted_at', if_not_exists: true
c.add_index :submission_representations, %i[submission_id ast_digest], unique: false, if_not_exists: true
c.add_index :submission_representations, :ast_digest

```

## Note these:
```
last_er_id = Exercise::Representation.last.id
# 837000

last_sr_id = Submission::Representation.last.id
# 1361000
```

## Run before deploy
```ruby
map = Exercise.pluck(:id, :track_id)
map.each do |exercise_id, track_id|
    Exercise::Representation.where(exercise_id: ).update_all(track_id: )
  end
end
```

## Run after deploy
```ruby
Exercise::Representation.where('id > ?', 223684).find_in_batches do |batch|
  batch.each do |representation|
      submission_representations = representation.submission_representations.pluck(:submission_id, :created_at)
      representation.update!(
          last_submitted_at: submission_representations.last.last,
          num_submissions: submission_representations.map(&:first).uniq.length
        )
    end
  end
  sleep(5)
end


map = Exercise.pluck(:id, :track_id)
map.each do |exercise_id, track_id|
    Exercise::Representation.where(track_id: nil, exercise_id: ).update_all(track_id: )
  end
end

Exercise::Representation.in_batches.each do |batch|
    batch.update_all('`uuid` = UUID()')
  end
end

# TODO: Incorporate Exercise.joins(:track).where('tracks.has_representer': true).pluck(:id).each do |exercise_id|

Iteration.where(id: Mentor::DiscussionPost.select(:iteration_id)).includes(:submission).find_each do |iteration|
  Submission::Representation::UpdateMentor.(iteration.submission)
end
```


app/assemblers/assemble_exercise_representations_with_feedback.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  initialize_with :mentor, :params
+
+  def call
+    SerializePaginatedCollection.(
+      representations,
+      serializer: SerializeExerciseRepresentations,
+      serializer_kwargs: { with_feedback: false },
+      meta: {
+        unscoped_total: Exercise::Representation.with_feedback.count
```
iHiD: This should be per user? Currently it's global.

app/assemblers/assemble_exercise_representations_without_feedback.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  initialize_with :mentor, :params
+
+  def call
+    SerializePaginatedCollection.(
+      representations,
+      serializer: SerializeExerciseRepresentations,
+      serializer_kwargs: { with_feedback: false },
+      meta: {
+        unscoped_total: Exercise::Representation.without_feedback.count
```
iHiD, line 13: I think this is going to be super slow.

We might want to cache this in redis using a pattern of:
- `cache-get`
- or if missing `db-get` -> `cache-set` (Same as in my other PR)
- `cache-set` in a ActiveJob everything a new ExerciseRepresentation is created

app/commands/exercise/representation/create_or_update.rb
```diff
@@ -0,0 +1,25 @@
+class Exercise
+  class Representation
+    class CreateOrUpdate
+      include Mandate
+
+      initialize_with :submission, :ast, :ast_digest, :mapping, :last_submitted_at
+
+      def call
+        representation = Exercise::Representation.create_or_find_by!(exercise:, ast_digest:) do |rep|
```
iHiD:
```suggestion
        representation = Exercise::Representation.find_create_or_find_by!(exercise:, ast_digest:) do |rep|
```

To avoid empty gaps in the `id` auto increment.

app/assemblers/assemble_exercise_representations_without_feedback.rb
iHiD: (That last bit could be part of https://github.com/exercism/website/pull/2919/files#diff-fe2dd7e4f217bfd3564d47d44f2909ad618f7590f116683124f939bdfc5beeb1R17)

app/commands/exercise/representation/find_example_submissions.rb
```diff
@@ -0,0 +1,44 @@
...
+      break example_submissions if submissions.last_page?
+      break example_submissions if page == MAX_PAGES_FETCHED
+
+      page += 1
+    end
+  end
+
+  private
+  def solution_files_hash(submission)
+    Digest::SHA1.hexdigest(submission.files.map(&:content).join)
```
iHiD: This is really expensive because we look up the files on s3.

```suggestion
    submission.files.map(&:digest).join
```

app/commands/exercise/representation/find_example_submissions.rb
```diff
@@ -0,0 +1,44 @@
+class Exercise::Representation::FindExampleSubmissions
+  include Mandate
+
+  initialize_with :representation
+
+  def call
+    example_submissions = [representation.source_submission]
+    example_submission_hashes = Set.new([solution_files_hash(representation.source_submission)])
```
iHiD: Alternatively we could have one hash of `hash: submission` then check for the `key?` below, then end the method by returning `values`. Rather than keeping two lists in play.

app/commands/exercise/representation/find_example_submissions.rb
```diff
@@ -0,0 +1,44 @@
...
+    page = 1
+
+    loop do
+      submissions = representation.
+        submission_representation_submissions.
+        where.not(id: representation.source_submission.id).
```
iHiD: For the sake of speed, I suggest removing this constraint in the SQL and moving it into the Ruby.

app/commands/exercise/representation/find_example_submissions.rb
```diff
@@ -0,0 +1,44 @@
...
+        order(created_at: :desc).
```
iHiD:
```suggestion
        order(id: :desc).
```

`id` (number) comparison is always faster than date comparison, and they're 1-1 for sorting intents.

app/commands/exercise/representation/find_example_submissions.rb
```diff
@@ -0,0 +1,44 @@
...
+        page(page).
+        per(NUM_EXAMPLES)
```
iHiD:
```suggestion
        per(NUM_EXAMPLES * 3)
```

Cheaper to return three times as many from one query than do three queries (presuming we think it'll take 3 queries to get the set). it could be 2x or 4x depending on the average.

app/commands/exercise/representation/find_example_submissions.rb
```diff
@@ -0,0 +1,44 @@
...
+
+        hash = solution_files_hash(submission)
+        next if example_submission_hashes.include?(hash)
+
+        example_submissions << submission
+        example_submission_hashes << hash
+      end
+
+      break example_submissions if submissions.last_page?
+      break example_submissions if page == MAX_PAGES_FETCHED
```
iHiD: I'd probably `return` not break here (so the same as L20)

app/commands/exercise/representation/search.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    @representations = @representations.for_track(track)
+  end
+
+  def filter_criteria!
+    return if criteria.blank?
+
+    exercise_ids = Exercise.where('title LIKE ?', "%#{criteria}%").
+      or(Exercise.where('slug LIKE ?', "%#{criteria}%")).
+      pluck(:id)
```
iHiD: I think we decided `select` was saner here, in case there are like 400 exercises that match `%a%` in which case mysql will blow up having that many things in an IN due to the sql being too long.

app/commands/mentor/discussion/finish_by_student.rb
```diff
@@ -90,6 +91,10 @@ def award_badge!
         AwardBadgeJob.perform_later(discussion.mentor, :mentor)
       end
 
+      def update_roles!
+        Mentor::UpdateRoles.defer(discussion.mentor)
```
iHiD: Nice.

iHiD: I feel like maybe I'd rather this was `User::UpdateMentorRoles`. I would have found that naming more obvious.

app/commands/exercise/representation/update_num_submissions.rb
```diff
@@ -0,0 +1,22 @@
...
+    ActiveRecord::Base.transaction(isolation: Exercism::READ_COMMITTED) do
+      Exercise::Representation.
+        where(id: representation.id).
+        update_all("num_submissions = (#{num_submissions_sql})")
+    end
+  end
+
+  private
+  def num_submissions_sql
+    Arel.sql(representation.submission_representations.select("COUNT(*)").to_sql)
```
iHiD: We will end up with multiple submission representations of the same submission as representers get rerun. We only want to count the cache of one per submission.

```suggestion
    Arel.sql(representation.submission_representations.select("COUNT(distinct(submission_id))").to_sql)
```

app/commands/submission/representation/update_mentor.rb
```diff
@@ -0,0 +1,31 @@
...
+      def first_mentor_comment
+        iteration.mentor_discussion_posts.
+          where.not(user_id: [student_id, User::GHOST_USER_ID]).
+          first
+      end
+
+      def student_id = solution.user_id
+    end
+  end
+end
```
iHiD, line 31: Nice.

app/commands/user/reset_account.rb
```diff
@@ -46,6 +46,7 @@ def reset_mentoring!
       user.mentor_discussion_posts.update_all(user_id: User::GHOST_USER_ID)
       user.mentor_testimonials.update_all(mentor_id: User::GHOST_USER_ID)
       user.provided_testimonials.update_all(student_id: User::GHOST_USER_ID)
+      user.submission_representations.update_all(mentor_id: User::GHOST_USER_ID)
```
iHiD: I think it's better to nullify them.

app/assemblers/assemble_exercise_representations_with_feedback.rb
iHiD: See comment below about caching. Less important on this one, but maybe still worth doing.

app/css/components/mentor/representer-row.css
```diff
@@ -0,0 +1,92 @@
...
+            @apply text-right;
+            width: 140px;
+        }
+        & time {
+            @apply ml-32;
+        }
+
+        & .--feedback-glimpse {
+            @apply text-right truncate mr-60 leading-150 text-textColor6;
+            width: 25rem;
```
iHiD: We use `px` not `rem`.

app/css/components/mentor/representer-row.css
```diff
@@ -0,0 +1,92 @@
...
+
+        & .action-icon {
+            @apply ml-36 text-textColor6;
+
+            height: 16px;
+            width: 16px;
+        }
+    }
+}
+.--most-popular {
```
iHiD, line 88: Generally nothing should be at global scope unless it's prefixed with `c-` (for component).

I think what you actually want here is:
```suggestion
  &.--most-popular {
```

which means "the current selector when it has this class".

(You'll then also need another bracket at the bottom)

app/css/pages/mentoring/automation.css
```diff
@@ -0,0 +1,69 @@
+@import "../../styles";
+
+.c-mentor-inbox {
```
iHiD, line 3: You're globally changing `c-mentor-inbox` here. You should either:
1. Make these changes in the component file (app/css/pages/mentoring/inbox.css) or 
2. If you want these only for automation, you should add a subselector here:

```.c-mentor-inbox.--automation {```

Then have the class for automation as class: `c-mentor inbox --automation`. 

3. Alternatively you could duplicate the c-mentor-inbox entirely here and change the class.

(1) or (2) are probably right depending on your purpose.

File: app/css/ui-kit/borders.css [hunk 18]
```diff
@@ -7,4 +7,12 @@
         background-origin: border-box;
         background-clip: padding-box, border-box;
     }
+    .border-gradient-primary {
+        border-style: solid;
+        border-color: transparent;
+        background-image: linear-gradient(#130b43, #130b43),
+            linear-gradient(#2200ff, #9e00ff);
+        background-origin: border-box;
+        background-clip: padding-box, border-box;
```
iHiD, line 16: Use an overload here instead of a whole new class.

```suggestion
    &.bg-primaryBtnBorder {
        background-image: linear-gradient(#130b43, #130b43),
            linear-gradient(#2200ff, #9e00ff);
    }
```

app/css/ui-kit/colors.css
```diff
@@ -240,6 +240,20 @@
...
+    }
+
+    .textColor6-filter {
+        filter: var(--textColor6-filter);
+    }
+
+    .white-filter {
+        filter: var(--white-filter);
+    }
+}
```
iHiD: Can you put these in `app/css/ui-kit/filters.css` pls, and switch the name so it's `filter-xxx`

app/css/ui-kit/inputs.css
```diff
@@ -20,6 +20,13 @@ textarea {
     }
 }
 
+/* maybe make this default for all inputs */
+.portable-input {
+    &::placeholder {
+        @apply text-textColor6 !important;
+    }
+}
+
```
iHiD, line 29: What does this do? (Any time there's an `!important` it's generally a smell!)

app/helpers/react_components/mentoring/representations/with_feedback.rb
```diff
@@ -0,0 +1,83 @@
...
+              initial_data: representations,
+              stale_time: 5000 # milliseconds
+            }
+          }
+        end
+
+        memoize
+        def representations_request_params
+          {
+            criteria: params.fetch(:criteria, ''),
```
iHiD, line 36: I still don't think we need the empty string thing here btw.

app/helpers/view_components/mentor/header.rb
```diff
@@ -47,43 +47,60 @@ def stats
...
+          return tag.div(
             class: "#{tab_class(:automation)} locked",
             'aria-label': "This tab is locked"
           ) do
             graphical_icon(:automation) +
-              tag.span("Automation")
+            tag.span("Automation")
           end
-        ]
+        end
```
iHiD, line 97: This should have a tooltip (as per figma) that explains why it's locked and how to unlock it. It's not blocking, but I'd like to ensure we eventually get that (if we don't have it already!)

app/images/icons/green-check.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="42" height="42" viewBox="0 0 42 42" fill="none" xmlns="http://www.w3.org/2000/svg">
+<circle cx="21" cy="21" r="21" fill="#59D2AE"/>
+<path d="M28.5 13.4993L18.4387 27.872C18.1694 28.259 17.7305 28.4929 17.2591 28.5007C16.7877 28.5085 16.3413 28.2892 16.0593 27.9113L13.5 24.4993" stroke="#130B43" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
+</svg>
```
iHiD, line 4: Use `app/images/icons/completed-check-circle.svg` instead of this pls.

app/images/icons/no-result-magnifier.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="48" height="48" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M37.7506 19.624C37.7514 26.9556 33.3354 33.5658 26.562 36.3719C19.7886 39.1779 11.9918 37.6273 6.80756 32.443C1.6233 27.2588 0.0726455 19.462 2.87872 12.6886C5.68479 5.91515 12.2949 1.49919 19.6266 1.5C24.4334 1.5 29.0433 3.40949 32.4422 6.8084C35.8411 10.2073 37.7506 14.8172 37.7506 19.624Z" stroke="#5C5589" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+<path d="M13.1367 13.136L25.8647 25.864" stroke="#5C5589" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+<path d="M13.1367 25.864L25.8647 13.136" stroke="#5C5589" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+<path d="M32.4434 32.4399L46.5014 46.4999" stroke="#5C5589" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
```
iHiD, line 5: All the colours in here (e.g. `#5C5589`) should be changed to `black`. All icons should be black and should have a `filter-xxx` applied when they're used.

app/javascript/components/common/SearchInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+type SearchInputProps = {
+  filter: string | undefined
+  setFilter: (filter: string) => void
+  placeholder: string
+  className?: string
+}
+
+const WRAPPER_CLASSNAMES =
+  'bg-unnamed15 text-textColor6 flex flex-row flex-grow rounded-[5px] border-1 border-transparent py-[11px] px-[21px] text-16 max-w-[420px] focus-within:focused-input hover:cursor-text'
```
iHiD, line 12: If there's a px padding of 21px, that's Tai making a mistake, so round it to 20px (I appreciate you doing it pixel perfect as a mindset though!)

For the py-11px, I'm presuming this is because the search box wants to be a fixed height. I'd rather specify the height than an arbitary vertical padding for that. 

```suggestion
  'bg-unnamed15 text-textColor6 flex flex-row flex-grow rounded-[5px] border-1 border-transparent py-[11px] px-20 text-16 max-w-[420px] focus-within:focused-input hover:cursor-text'
```

There are already a lot of other search boxes around the site. Are they different to this, or should they be standardised? If so, should we be reusing some of the CSS from them? (Genuine questions - I don't remember 🙂 )

app/javascript/components/common/SearchInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+        className={ICON_CLASSNAMES + (focused ? 'filter-none' : '')}
+        icon="search"
+      />
+      <input
+        onFocus={() => setFocused(true)}
+        onBlur={() => setFocused(false)}
+        ref={searchInputRef}
+        type="text"
+        className={INPUT_CLASSNAMES}
+        style={{ all: 'unset' }}
```
iHiD, line 45: What does `unset` do? Use browser defaults? If so, we might not want to do this. If something else then maybe it's great.

app/javascript/components/contributing/TasksList.tsx
```diff
@@ -63,18 +64,14 @@ export const TasksList = ({
...
-    request
-  )
+  const { status, resolvedData, latestData, isFetching, error } =
+    usePaginatedRequestQuery<PaginatedResult, Error | Response>(
+      ['contributing-tasks', request.endpoint, request.query],
+      request
+    )
   const track =
     tracks.find((t) => t.slug === request.query.trackSlug) || tracks[0]
+  useLogger('TRACK IN TRACKLIST', track)
```
iHiD, line 74: Is this debugging? Do we keep it in? Etc?

app/javascript/components/mentoring/Representation.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { CompleteRepresentationData } from '../types'
+import { LeftPane } from './representation/left-pane'
+import { RightPane } from './representation/right-pane'
+
+export function Representation({
+  data,
+}: {
+  data: CompleteRepresentationData
+}): JSX.Element {
+  console.log('data in representaiton:', data)
```
iHiD: Debugging

app/javascript/components/mentoring/automation/AutomationListElement.tsx
```diff
@@ -0,0 +1,68 @@
...
+          {fromNow(representation.lastSubmittedAt)}
+        </>
+      ),
+    }
+  }, [representation, withFeedback])
+
+  return (
+    <a
+      className="--representer"
+      href={`${withFeedback ? '' : 'automation/'}${representation.id}/edit`}
```
iHiD: Should this URL not be provided from upstream?

app/css/components/mentor/representer-row.css
iHiD: Actually, from reading code further, this should just be nested inside the row in the CSS, it's not a modifier.

app/javascript/components/mentoring/automation/RepresentationList.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function Component({
+  resolvedData,
+  latestData,
+  page,
+  setPage,
+  withFeedback,
+}: Props) {
+  return (
```
iHiD, line 42: How about you do 
```if(!resolvedData || !resolvedData.results) { return <></> }```

as a guard first, then the code below can be indented a couple of layers less.

File: app/javascript/components/mentoring/automation/TrackSelector.tsx [hunk 31]
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
+
+export function TrackSelector(): JSX.Element {
+  return <div></div>
+}
```
iHiD, line 5: Solid 😂

app/javascript/components/mentoring/automation/mock-data.tsx
```diff
@@ -0,0 +1,986 @@
+import { MentorDiscussion, MentoredTrack } from '../../types'
```
iHiD: Reminder to delete this (or move it outside of the production tree)

config/locales/api/en.yml
```diff
@@ -2,6 +2,8 @@ en:
   api:
     errors:
       not_onboarded: "You have not yet onboarded on the website"
+      no_supermentor: "You do not have supermentor permissions"
```
iHiD:
```suggestion
      not_supermentor: "You do not have supermentor permissions"
```

config/locales/api/en.yml
```diff
@@ -2,6 +2,8 @@ en:
...
+      no_maintainer: "You do not have maintainer permissions"
```
iHiD:
```suggestion
      not_maintainer: "You do not have maintainer permissions"
```

db/migrate/20220812093222_add_num_submissions_to_representations.rb
```diff
@@ -0,0 +1,5 @@
+class AddNumSubmissionsToRepresentations < ActiveRecord::Migration[7.0]
+  def change
+    add_column :exercise_representations, :num_submissions, :integer, null: false, default: 0
```
iHiD:
```suggestion
    add_column :exercise_representations, :num_submissions, :integer, null: false, default: 1
```

🤷

db/migrate/20220817075919_add_last_submitted_at_to_exercise_representations.rb
```diff
@@ -0,0 +1,11 @@
+class AddLastSubmittedAtToExerciseRepresentations < ActiveRecord::Migration[7.0]
+  def change
+    add_column :exercise_representations, :last_submitted_at, :datetime, null: false, default: Time.now, if_not_exists: true
+
+    unless Rails.env.production?
+      Exercise::Representation.find_each do |rep|
+        rep.update(last_submitted_at: rep.submission_representations.last.created_at)
+      end
```
iHiD, line 12: Can you put together a bit on the OP of this PR with all the things we need to run before we deploy pls.

db/migrate/20220823134038_add_ast_digest_index_to_exercise_representations.rb
```diff
@@ -0,0 +1,5 @@
+class AddAstDigestIndexToExerciseRepresentations < ActiveRecord::Migration[7.0]
+  def change
+    add_index :exercise_representations, %i[exercise_id ast_digest], unique: false
```
iHiD: Should this actually be `unique: true`? 🤔

db/migrate/20220824135136_add_mentor_to_submission_representations.rb
```diff
@@ -0,0 +1,12 @@
+class AddMentorToSubmissionRepresentations < ActiveRecord::Migration[7.0]
+  def change
+    add_reference :submission_representations, :mentor, null: true, foreign_key: { to_table: :users }, if_not_exists: true
```
iHiD, line 3: I'd still rather this was `mentored_by` rather than `mentor` and that changing throughout. I find it much clearer.

db/migrate/20220823134044_add_ast_digest_index_to_submission_representations.rb
```diff
@@ -0,0 +1,5 @@
+class AddAstDigestIndexToSubmissionRepresentations < ActiveRecord::Migration[7.0]
+  def change
+    add_index :submission_representations, %i[submission_id ast_digest], unique: false
```
iHiD:
```suggestion
    add_index :submission_representations, %i[submission_id ast_digest], unique: false, if_not_exists: true
```

db/migrate/20220823134038_add_ast_digest_index_to_exercise_representations.rb
iHiD:
```suggestion
    add_index :exercise_representations, %i[exercise_id ast_digest], unique: false, if_not_exists: true
```

iHiD:
```suggestion
    add_index :exercise_representations, %i[exercise_id ast_digest], unique: false, if_not_exists: true
```

db/migrate/20220812093222_add_num_submissions_to_representations.rb
iHiD:
```suggestion
    add_column :exercise_representations, :num_submissions, :integer, null: false, default: 0, if_not_exists: true
```

db/migrate/20220825134619_add_uuid_to_exercise_representations.rb
```diff
@@ -0,0 +1,16 @@
+class AddUuidToExerciseRepresentations < ActiveRecord::Migration[7.0]
+  def change
+    add_column :exercise_representations, :uuid, :string, null: true, index: { unique: true }
```
iHiD:
```suggestion
    add_column :exercise_representations, :uuid, :string, null: true, index: { unique: true }, if_not_exists: true
```

db/migrate/20220825134619_add_uuid_to_exercise_representations.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    # TODO: consider if we can run this in production
+    ActiveRecord::Base.transaction(isolation: Exercism::READ_COMMITTED) do
+      Exercise::Representation.find_each do |representation|
```
iHiD, line 7:
```suggestion
    Exercise::Representation.find_each do |representation|
      ActiveRecord::Base.transaction(isolation: Exercism::READ_COMMITTED) do
```

iHiD: Need a seperate transaction for each, else we'll lock the whole table for ages.

db/migrate/20220825134619_add_uuid_to_exercise_representations.rb
```diff
@@ -0,0 +1,16 @@
...
+        Exercise::Representation
+          .where(id: representation.id)
+          .update_all(uuid: SecureRandom.compact_uuid)
```
iHiD: This is not what you want :) This will set the uuid of all records to the same thing. Instead you want to use the Mysql UUID feature:
```suggestion
          .update_all("`uuid` = UUID()")
```

or if you really care about company uuids, you'd have to iterate every row (which means I think we don't care 😉)

db/migrate/20220831065513_add_track_to_exercise_representations.rb
```diff
@@ -0,0 +1,10 @@
+class AddTrackToExerciseRepresentations < ActiveRecord::Migration[7.0]
+  def change
+    add_reference :exercise_representations, :track, null: true, foreign_key: true, if_not_exists: true
+
+    unless Rails.env.production?
+      exercise_track_id_sql = Arel.sql(Exercise.where('id = exercise_id').select(:track_id).to_sql)
+      Exercise::Representation.update_all("track_id = (#{exercise_track_id_sql})")
+    end
```
iHiD, line 13: This should have the logic from the one above (I actually don't think the one above necessarily needs the looping - it was this one I intended it for!)

tailwind.config.js
```diff
@@ -315,8 +319,17 @@ module.exports = {
...
-  plugins: [],
+  variants: {
+    extend: {
+      borderColor: ['focus-within', 'focus', 'hover'],
+      backgroundColor: ['focus-within'],
+    },
+  },
+  plugins: [
+    function ({ addVariant }) {
+      addVariant('child', '& > *')
```
iHiD, line 336: Where is this used?

db/migrate/20220825134619_add_uuid_to_exercise_representations.rb
ErikSchierboom: Doh

ErikSchierboom: > This will set the uuid of all records to the same thing.

Are you saying that `SecureRandom.compact_uuid` returns the same thing when called in a tight loop? Otherwise, I think I'm just updating one representation.

```ruby
Exercise::Representation
          .where(id: representation.id)
          .update_all(uuid: SecureRandom.compact_uuid)
```

app/javascript/components/mentoring/automation/mock-data.tsx
dem4ron: yes, deleting this. :)

app/javascript/components/mentoring/automation/TrackSelector.tsx
dem4ron: the purest component :)

tailwind.config.js
dem4ron: in PreviewFooter.tsx and SubmittedAutomationModal.tsx.
The reason is the same: it leaves those components untouched, but does some minor styling on them. Eg. tabs or the reputation badge components, which I didn't want to modify.

dem4ron: and from Tailwind 3.1 you get this support out of the box without having to create a plugin for it. We use 3.0.7
https://tailwindcss.com/blog/tailwindcss-v3-1#arbitrary-values-but-for-variants

app/css/components/mentor/representer-row.css
dem4ron: I moved this outside, because I reused this tag in somewhere completely different place, but I'll find a better way of solving this.

app/helpers/view_components/mentor/header.rb
ErikSchierboom: I've created a tooltip, but the figma version is outdated as it uses reputation as the unlock mechanism, whereas we're actually using num_finished_mentoring_sessions >= 100 and satisfaction_percentage >= 95.

app/css/ui-kit/inputs.css
dem4ron: This is the only way the change took effect. I agree on the `!important` I don't like it using too :)

app/models/user/reputation_tokens/automation_feedback_author_token.rb
```diff
@@ -11,5 +11,5 @@ class User::ReputationTokens::AutomationFeedbackAuthorToken < User::ReputationTo
 
   def i18n_params =  { exercise_title: representation.exercise.title }
   def guard_params = "Representation##{representation.id}"
-  def internal_url = Exercism::Routes.edit_mentoring_automation_path(representation)
+  def internal_url = Exercism::Routes.edit_mentoring_automation_path(representation.uuid)
```
iHiD, line 14: The idomatic way to do this is to add a `to_param` method to the representation instead and leave these as they were (much more DRY).

```ruby
def to_param
  uuid
end
```

config/locales/mentoring/en.yml
```diff
@@ -0,0 +1,4 @@
+en:
+  mentoring:
+    rules:
+      global: "- Be kind"
```
ErikSchierboom, line 4: This needs some proper content.

ErikSchierboom (approved):

app/commands/exercise/representation/submit_feedback.rb
```diff
@@ -0,0 +1,30 @@
...
+  class Representation
+    class SubmitFeedback
+      include Mandate
+
+      initialize_with :mentor, :representation, :feedback_markdown, :feedback_type
+
+      def call
+        if representation.has_feedback?
+          representation.update(feedback_markdown:, feedback_type:, feedback_editor: mentor)
+          award_reputation_token!(:automation_feedback_editor) unless mentor_is_author?
```
ErikSchierboom, line 11: I figured we don't want to award more reputation if someone edits their own comments. Or do we?

app/helpers/react_components/mentoring/representations/with_feedback.rb
ErikSchierboom: We currently do, but we should be fixing it in typescript instead.

config/locales/mentoring/en.yml
ErikSchierboom: Maybe we could outsource this to our supermentors? 😉

app/commands/exercise/representation/submit_feedback.rb
iHiD: Agreed.

db/migrate/20220823134044_add_ast_digest_index_to_submission_representations.rb
```diff
@@ -0,0 +1,5 @@
+class AddAstDigestIndexToSubmissionRepresentations < ActiveRecord::Migration[7.0]
+  def change
+    add_index :submission_representations, %i[submission_id ast_digest], unique: false, if_not_exists: true
```
ErikSchierboom, line 3: @iHiD I think this index needs to be on `ast_digest` only.

```suggestion
    add_index :submission_representations, %i[ast_digest], unique: false, if_not_exists: true
```

I don't know if this can be changed or if it needs a new migration to remove and then add this?

app/css/components/search-input.css
```diff
@@ -0,0 +1,5 @@
+@layer base {
+    .focused-input {
+        @apply bg-white border-lightBlue shadow-inputSelected;
+    }
+}
```
iHiD, line 5: This class should have the same name as the css file.

If it's a component, it should be prefixed with `c-`. If it's a style (which I think it is), it should be in ui-kit

PR #2926, opened 2022-09-01
Optimise representations page

iHiD:
Readds migrations. See comments.

### DIFF
--- a/app/commands/exercise/representation/find_example_submissions.rb
+++ b/app/commands/exercise/representation/find_example_submissions.rb
@@ -10,10 +10,11 @@ class Exercise::Representation::FindExampleSubmissions
 
     loop do
       submissions = representation.
-        submission_representation_submissions.
+        submission_representations.
+        includes(:submission).
         order(id: :desc).
         page(page).
-        per(NUM_EXAMPLES * 2)
+        per(NUM_EXAMPLES * 2).map(&:submission)
 
       submissions.each do |submission|
         return example_submissions.values if example_submissions.size == NUM_EXAMPLES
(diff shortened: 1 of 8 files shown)


app/commands/exercise/representation/find_example_submissions.rb
```diff
@@ -10,10 +10,11 @@ def call
 
     loop do
       submissions = representation.
-        submission_representation_submissions.
+        submission_representations.
+        includes(:submission).
         order(id: :desc).
         page(page).
-        per(NUM_EXAMPLES * 2)
+        per(NUM_EXAMPLES * 2).map(&:submission)
```
iHiD, line 16: This is 2ms, not 20s, so a good 4 orders of magnitude faster. Got to love SQL! 🙂

app/helpers/view_components/mentor/header.rb
```diff
@@ -101,8 +101,8 @@ def automation_tab
           class: tab_class(:automation)
         ) do
           graphical_icon(:automation) +
-            tag.span("Automation") +
-            tag.span(number_with_delimiter(num_representations_without_feedback), class: 'count')
+            tag.span("Automation") #+
+          # tag.span(number_with_delimiter(num_representations_without_feedback), class: 'count')
```
iHiD, line 105: This count currently takes ages. Caching the individual track counts and summing them here would be the best bet. And we can then reuse those caches for the track counts as well.

app/commands/exercise/representation/find_example_submissions.rb
```diff
@@ -6,30 +6,28 @@ class Exercise::Representation::FindExampleSubmissions
   def call
     source_submission_hash = solution_files_hash(representation.source_submission)
     example_submissions = { source_submission_hash => representation.source_submission }
-    page = 1
```
iHiD, line 9: I've refactored all this a bit to use enumeration rather than a loop.

PR #2932, opened 2022-09-02
Add representer mentoring docs

ErikSchierboom:
- Add representer docs to tracks-with-exercises repo
- Support reading represents docs
- Add track- and exercise mentoring contents


app/helpers/react_components/mentoring/representations/representation.rb
```diff
@@ -10,7 +10,9 @@ def to_s
             examples: examples_data,
             mentor: mentor_data,
             rules: {
-              global_html: Markdown::Parse.(I18n.t("mentoring.rules.global"))
+              global_html: Markdown::Parse.(I18n.t("mentoring.rules.global")),
+              track_html: Markdown::Parse.(track.representer_mentoring),
+              exercise_html: exercise.mentoring_notes.content
```
ErikSchierboom: @iHiD I've added three files:

- One that we manage (global rules). Don't know if there is anything useful in there that our supermentors would appreciate, otherwise we can remove this
- The track-specific mentoring notes
- The exercise-specific mentoring notes

Will the representer mentoring document be specific to representation mentoring, or should it be a more generic track mentoring document?

iHiD (approved):

PR #2942, opened 2022-09-05
Automation-Locked tooltip

### LINKED ISSUE #6537: Link on info text for 'Automation' on 'Mentoring Inbox' page can't be clicked
When a logged in user hovers the UI pointer over the tab 'Automation on page 'Mentoring Inbox' (➙ https://exercism.org/mentoring/inbox), then the link in the displayed text can't be clicked, because the pop-up info disappears as soon as the pointer doesn't hover over the tab text.

The marked text in the screen shot _looks_ like a link, but it can't be reached.

Exercism-Mentoring-Automation

This happens in Safari, Firefox & Chrome.

dem4ron:
- Basic styling and interactivity of tooltip, remove a uselogger
- Tooltip improvements

Closes https://github.com/exercism/exercism/issues/6537

### DIFF
--- a/app/views/mentoring/automation/tooltip_locked.html.haml
+++ b/app/views/mentoring/automation/tooltip_locked.html.haml
@@ -1,17 +1,23 @@
 .heading.c-generic-tooltip.text-14.w-413
-  %h4.font-semibold.mx-24.mt-20.mb-4 What is automation?
 
-  -# %p.mx-24
-    = ReactComponent::
+  %p.mx-24
     Requires a satisfaction percentage of
     %strong= @min_satisfaction_percentage.to_s
     (current: #{@satisfaction_percentage})
 
-  -# %p.mx-24
+  .bg-veryLightBlue.flex.flex-row.rounded-8.mb-16.mx-24.some{style: 'height:16px;'}
+    .bg-gradient-to-r.from-gradientProgressBarA.to-gradientProgressBarB.rounded-8.some{style:'min-width:6px; height:100%; width:' +(@finished_mentoring_sessions+25).to_s+'%'}
+
+  %p.mx-24
     Requires
     %strong= @min_finished_mentoring_sessions.to_s
     finished mentoring sessions
     (current: #{@finished_mentoring_sessions})
+    
+  .bg-veryLightBlue.flex.flex-row.rounded-8.mb-16.mx-24.some{style: 'height:16px;'}
+    .bg-gradient-to-r.from-gradientProgressBarA.to-gradientProgressBarB.rounded-8.some{style:'min-width:6px; height:100%; width:' +(@finished_mentoring_sessions+25).to_s+'%'}
+
+  %h4.font-semibold.mx-24.mt-20.mb-4 What is automation?
 
   %p.mx-24.mb-4
     Automation allows you to submit standardised responses to code submissions that get shared as part of automated feedback on solutions.
--- a/app/views/mentoring/automation/tooltip_locked.html.haml
+++ b/app/views/mentoring/automation/tooltip_locked.html.haml
@@ -1,12 +1,13 @@
-.p.heading.c-generic-tooltip.text-14.some{style:'padding:16px 24px;width:413px;'}
+.heading.c-generic-tooltip.text-14{class:"!py-16 !px-24 w-[413px]"}
   %p.text-16.text-white.font-semibold.leading-170.mb-12
     Requires
-    %strong.font-bold= @min_finished_mentoring_sessions.to_s
-    finished mentoring sessions and a satisfaction rating of #{@min_satisfaction_percentage.to_s}% 
-    %br (current rating is #{@satisfaction_percentage}%)
-    
-  .bg-darkThemeBackgroundColor.flex.flex-row.rounded-8.mb-6.overflow-hidden.some{style: 'height:16px;'}
-    .bg-gradient-to-r.from-gradientProgressBarA.to-gradientProgressBarB.rounded-8.some{style:'min-width:6px; height:100%; width:' +(@finished_mentoring_sessions.to_s)+'%'}
+    %strong.font-bold= @min_finished_mentoring_sessions
+    finished mentoring sessions and a satisfaction rating of #{@min_satisfaction_percentage}% 
+    (current rating is #{@satisfaction_percentage}%)
+  .bg-darkThemeBackgroundColor.flex.flex-row.rounded-8.mb-6.overflow-hidden{class: 'h-[16px]'}
+    .bg-gradient-to-r.from-gradientProgressBarA.to-gradientProgressBarB.rounded-8.h-100{class:"min-w-[6px]", style: "width:"+@finished_mentoring_sessions.to_s+"%"}
+
+
   .mb-16 #{@finished_mentoring_sessions}/100
 
   %h4.font-semibold.mb-4 What is automation?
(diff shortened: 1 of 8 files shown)


dem4ron:
### UI Changes:

Screenshot 2022-09-05 at 15 49 40
Screenshot 2022-09-05 at 15 49 59

@iHiD, please review this :)

dem4ron:
After reviewing and removing the unnecessary margins, adjusting top padding according to the design
Screenshot 2022-09-05 at 15 58 23

app/views/mentoring/automation/tooltip_locked.html.haml
```diff
@@ -1,21 +1,20 @@
...
-  %h4.font-semibold.mx-24.mt-20.mb-4 What is automation?
-
-  -# %p.mx-24
-    = ReactComponent::
-    Requires a satisfaction percentage of
-    %strong= @min_satisfaction_percentage.to_s
-    (current: #{@satisfaction_percentage})
-
-  -# %p.mx-24
+.p.heading.c-generic-tooltip.text-14.some{style:'padding:16px 24px;width:413px;'}
```
iHiD:
```suggestion
.p.heading.c-generic-tooltip.text-14.some{style:'padding:16px 24px;width:413px;'}
```

1. What's `.p`? 
2. What's `.some`
3. Also use `py-16 px-24` rather than add styling.
4. use `class: "w-[413px]"` rather than `style`

app/views/mentoring/automation/tooltip_locked.html.haml
```diff
@@ -1,21 +1,20 @@
...
+  %p.text-16.text-white.font-semibold.leading-170.mb-12
     Requires
-    %strong= @min_finished_mentoring_sessions.to_s
-    finished mentoring sessions
-    (current: #{@finished_mentoring_sessions})
+    %strong.font-bold= @min_finished_mentoring_sessions.to_s
+    finished mentoring sessions and a satisfaction rating of #{@min_satisfaction_percentage.to_s}% 
+    %br (current rating is #{@satisfaction_percentage}%)
```
iHiD:
```suggestion
    %br 
    (current rating is #{@satisfaction_percentage}%)
```

`%br` is a self-closing tag, it doesn't have content.

That said, I don't think you want a `br` here at all really - they're normally a bad idea as they break differently in responsive-world.

app/views/mentoring/automation/tooltip_locked.html.haml
```diff
@@ -1,21 +1,20 @@
...
-    (current: #{@satisfaction_percentage})
-
-  -# %p.mx-24
+.p.heading.c-generic-tooltip.text-14.some{style:'padding:16px 24px;width:413px;'}
+  %p.text-16.text-white.font-semibold.leading-170.mb-12
     Requires
-    %strong= @min_finished_mentoring_sessions.to_s
-    finished mentoring sessions
-    (current: #{@finished_mentoring_sessions})
+    %strong.font-bold= @min_finished_mentoring_sessions.to_s
```
iHiD:
```suggestion
    %strong.font-bold= @min_finished_mentoring_sessions
```

No need for `to_s` here I don't think.

app/views/mentoring/automation/tooltip_locked.html.haml
```diff
@@ -1,21 +1,20 @@
...
+    finished mentoring sessions and a satisfaction rating of #{@min_satisfaction_percentage.to_s}% 
+    %br (current rating is #{@satisfaction_percentage}%)
+    
+  .bg-darkThemeBackgroundColor.flex.flex-row.rounded-8.mb-6.overflow-hidden.some{style: 'height:16px;'}
```
iHiD:
```suggestion
  .bg-darkThemeBackgroundColor.flex.flex-row.rounded-8.mb-6.overflow-hidden.some{ class: 'h-[16px]' }
```

app/views/mentoring/automation/tooltip_locked.html.haml
```diff
@@ -1,21 +1,20 @@
...
+    .bg-gradient-to-r.from-gradientProgressBarA.to-gradientProgressBarB.rounded-8.some{style:'min-width:6px; height:100%; width:' +(@finished_mentoring_sessions.to_s)+'%'}
```
iHiD: You have `h-100`, `min-w-[6px]` you can use as classes. Then the last thing width should correctly be a style as it includes a variable.

app/views/mentoring/automation/tooltip_locked.html.haml
dem4ron: Oops, `p` is a chunk I left behind, it was padding, I'll change it to `{class: "!py-16 !px-24"}`, as in other forms it doesn't overwrite the more !important padding in c-generic-tooltip, now seeing it the next day, it just makes no sense why I left it there 🥲 


About `.some` it is super embarrassing to admit, but I read the answer first here: 
https://stackoverflow.com/questions/8090010/haml-how-to-set-inline-style-for-element-in-haml
and immediately thought it was a pseudoselector-something in haml to insert inline styling..
but now that I read the question too, I see that this is only a random class.. sorry about this.

app/views/mentoring/automation/tooltip_locked.html.haml
dem4ron: yep, sorry didn't know the syntax here. 
But since tooltip has a fixed width, I don't think responsiveness has a big effect on this. (or should it?)
Shouldn't it stay?, as it organizes text better, otherwise
Screenshot 2022-09-06 at 10 04 17
it would be this 2 half lines, and it looked better to me in a new, clean line.

or should I modify the fixed width of the tooltip?

app/views/mentoring/automation/tooltip_locked.html.haml
dem4ron: OK (I removed and works well). 
and how does this work, isn't it a number? or when do I need to stringify it? only when I concatenate it with another string?

app/views/mentoring/automation/tooltip_locked.html.haml
dem4ron: I modified the width of the tooltip, whenever it is under `md` screen-width, it switches to `w-auto`

app/views/mentoring/automation/tooltip_locked.html.haml
```diff
@@ -1,8 +1,9 @@
-.heading.c-generic-tooltip.text-14{class:"!py-16 !px-24 w-[413px]"}
+.heading.c-generic-tooltip.text-14{class:"!py-16 !px-24 md:w-[413px] w-auto"}
```
iHiD: What is `.heading`?

dem4ron: It was present when I arrived, assumed, it was a parent element to the others, but it is nothing, so just removed it.

app/views/mentoring/automation/tooltip_locked.html.haml
iHiD: Cool cool!

app/views/mentoring/automation/tooltip_locked.html.haml
iHiD: OK, so I think the:
- The text should change to be "95% (currently 0%)"
- The width of the tooltip should change to fit that all onto two lines (so it might need an extra 20px or so wider).

app/views/mentoring/automation/tooltip_locked.html.haml
iHiD: Great question. So when something is rendered as a string in ruby, `to_s` is automatically called on it. So when something is interpolated (e.g. `width:#{num}%`}, num has `to_s` called on it. And when it's rendered here in haml, that happens too.

iHiD:
This is looking much better now. Couple of minor comments then we can merge it 🙂 I imagine it's been a fun learning exercise!

dem4ron:
### UI changes:
Screenshot 2022-09-07 at 12 37 46
Screenshot 2022-09-07 at 12 37 16 1

I went to use `max-width` only instead of `md:width` and `w-auto`. Looks better.

PR #2947, opened 2022-09-05
Reshape guidance in automation

iHiD:
@dem4ron This:
- Adds the scratchpad to the representation JSON
- Unifies it's format with that of the existing mentoring UI (by changing both). 

You'll need to make changes to the work you've already done for the automation UI and also change the mentoring UI. The tests will also need updating, which you might like to try and do, or ask me to do (or pair with me on).

You can build on this branch, or you can cherry pick the commit onto your branch (Go to your branch and do `git fetch && git cherry-pick a955b32a04739df2b8a8530fa169dc11030fc1c9`)

### DIFF
--- a/app/helpers/react_components/mentoring/representations/representation.rb
+++ b/app/helpers/react_components/mentoring/representations/representation.rb
@@ -53,6 +53,11 @@ module ReactComponents
               Exercism::Routes.with_feedback_mentoring_automation_index_path
         end
 
+        memoize
+        def scratchpad
+          ScratchpadPage.new(about: exercise)
+        end
+
         delegate :track, :exercise, to: :representation
       end
     end
--- a/app/helpers/react_components/mentoring/representations/representation.rb
+++ b/app/helpers/react_components/mentoring/representations/representation.rb
@@ -9,19 +9,15 @@ module ReactComponents
             representation: SerializeExerciseRepresentation.(representation),
             examples: examples_data,
             mentor: mentor_data,
+            mentor_solution:,
             guidance: {
+              representationsHtml: Markdown::Parse.(track.mentoring_representations).presence,
               exercise: exercise.mentoring_notes_content,
               track: track.mentoring_notes_content,
               exemplar_files: SerializeExemplarFiles.(exercise.exemplar_files),
               links: {
                 improve_notes: exercise.mentoring_notes_edit_url
               }
-            },
-            mentor_solution:,
-            information: {
-              representations: Markdown::Parse.(track.mentoring_representations).presence,
-              exercise: exercise.mentoring_notes_content,
-              track: track.mentoring_notes_content
             },
             scratchpad: {
               is_introducer_hidden: true,
--- a/app/helpers/react_components/mentoring/representations/representation.rb
+++ b/app/helpers/react_components/mentoring/representations/representation.rb
@@ -15,7 +15,7 @@ module ReactComponents
               track: track.mentoring_notes_content
             },
             scratchpad: {
-              is_introducer_hidden: current_user&.introducer_dismissed?("scratchpad"),
+              is_introducer_hidden: true,
               links: {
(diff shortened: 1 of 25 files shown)


iHiD:
(cc @ErikSchierboom)

app/javascript/components/mentoring/representation/right-pane/UtilityTabs.tsx
```diff
@@ -0,0 +1,61 @@
...
+      </Tab.Panel>
+      <Tab.Panel id="scratchpad" context={TabsContext}>
+        <Scratchpad
+          scratchpad={data.scratchpad}
+          track={data.representation.track}
+          exercise={data.representation.exercise}
+        />
+      </Tab.Panel>
+      <Tab.Panel id="guidance" context={TabsContext}>
+        <div>Guidance is coming soon!</div>
```
iHiD: 😁

iHiD: You should be able to copy anything you need from `app/helpers/react_components/mentoring/session.rb` into `app/helpers/react_components/mentoring/representations/representation.rb` to make this work.

app/helpers/react_components/mentoring/representations/representation.rb
```diff
@@ -9,10 +9,23 @@ def to_s
             representation: SerializeExerciseRepresentation.(representation),
             examples: examples_data,
             mentor: mentor_data,
+            mentor_solution:,
             guidance: {
-              representations_html: Markdown::Parse.(track.mentoring_representations).presence,
-              track_mentoring_notes_html: track.mentoring_notes_content.presence,
-              exercise_mentoring_notes_html: exercise.mentoring_notes_content.presence
+              representationsHtml: Markdown::Parse.(track.mentoring_representations).presence,
```
iHiD: This shouldn't be camel-cased, it should be underscored. It's down to JS to do that conversion (and there's a function we use in the bootstrap file for it). 

But I also don't think it needs the `html` suffix at all any more.

```suggestion
              representations: Markdown::Parse.(track.mentoring_representations).presence,
```

app/javascript/components/mentoring/Session.tsx
```diff
@@ -203,7 +203,7 @@ export const Session = (props: SessionProps): JSX.Element => {
                 </Tab.Panel>
                 <Tab.Panel id="guidance" context={TabsContext}>
                   <Guidance
-                    notes={notes}
+                    notes={guidance.exercise}
```
iHiD: This `Guidance` component should be updated to render both the exercise guidance and the track guidance (if present). So add an extra section to the accordion. The Track guidance section should go below the Exercise guidance section.

app/helpers/react_components/mentoring/representations/representation.rb
dem4ron: oh, true! yes, I readded the html suffix, because it sounded clearer what this is for. but I will remove it.

app/javascript/components/types.ts
```diff
@@ -464,12 +474,18 @@ export type CompleteRepresentationData = {
...
-    representationsHtml?: string
-    trackMentoringNotesHtml?: string
-    exerciseMentoringNotesHtml?: string
-  }
+  guidance: Guidance
+  scratchpad: Scratchpad
+}
+
+export type Guidance = {
+  representationsHtml: string
```
iHiD:
```suggestion
  representations: string
```

dem4ron (approved):

app/helpers/react_components/mentoring/representations/representation.rb
ErikSchierboom: I quite like the `_html` suffix, and we use it in several other places too. Not having it here makes it feel somewhat inconsistent to me.

PR #2958, opened 2022-09-15
Fix test summary overflow bug

dem4ron:
Will resolve this issue: https://github.com/exercism/exercism/issues/6544

## Parts of the problem:

Screenshot 2022-09-15 at 12 38 17

Screenshot 2022-09-15 at 15 47 54

### One possible solution:
Screenshot 2022-09-15 at 15 49 00

I think this one looks fine for the body of test summary.

#### Apply scroll to title too?

Screenshot 2022-09-15 at 16 03 29

But these have too many scrollbars. Maybe a responsive UI would be good here. But still what happens when there is a sentence_written_with_underscores_and_digested_as_one_word. Ellipsis is a solution

### DIFF
--- a/app/css/components/test-run.css
+++ b/app/css/components/test-run.css
@@ -200,6 +200,7 @@
             & .--summary-inner {
                 @apply flex items-center;
                 @apply py-10 px-24;
+                @apply overflow-y-scroll;
             }
 
             & .--status {
@@ -231,7 +232,7 @@
                     @apply font-bold text-darkGray uppercase leading-huge mb-8;
                 }
                 & pre {
-                    @apply bg-backgroundColorB py-16 px-20 whitespace-pre-wrap rounded-8;
+                    @apply bg-backgroundColorB py-16 px-20 whitespace-pre-wrap overflow-y-auto rounded-8;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


iHiD:
This page isn't supposed to be responsive yet, so I'm not too worried about the responsive case.

- **Long titles:** This is a pain. It's intended that tracks parse the test names to be prettier and easier to read, which fixes this in the best way. Which track was this?
- **Code run:** I'm happy with those boxes scrolling if they're too long. But we'd want to set `overflow:auto` not `scroll` else we add scrollbars to everything on windows even when they're not used.

dem4ron:
This is the Rust track.

iHiD:
So that needs updating here really: https://github.com/exercism/rust-test-runner/blob/main/src/output.rs#L21 

This is what Ruby does: https://github.com/exercism/ruby-test-runner/blob/e25627f465c31dbe4b22c331d715d69a901f71aa/lib/extract_test_metadata.rb#L45-L48 

Feel free to put a Rust PR in if you fancy a bit of Rust side-project (off the clock 🙂) - you could maybe ask if anyone fancies pairing in `#track-rust` (@bobahop might like to do it with you for example - although he's a little Rusted out atm I believe!)

PR #2968, opened 2022-09-20
Tweaks to impact page

iHiD:


app/javascript/components/impact/map.tsx
```diff
@@ -95,27 +95,27 @@ const MetricPointUserWithTooltip = ({
...
+        <div
+          ref={iconRef}
+          className="relative border-2 border-gradient rounded-circle translate-y-[-50%] translate-x-[-50%]"
+        >
+          <GraphicalIcon
+            icon="avatar-placeholder"
+            className="w-[32px] h-[32px]"
+          />
+        </div>
+      )
```
iHiD, line 112: @dem4ron `refs` don't work in functions apparently. Inlining this fixes that issue, and means the "New user signed up" stops appearing at the top-left of the page.

PR #2972, opened 2022-09-22
Add Donation modal

dem4ron:
Screenshot 2023-09-21 at 11 17 21


app/css/ui-kit/buttons.css
```diff
@@ -231,4 +231,10 @@
             @apply text-15;
         }
     }
+
+    .btn-casual {
+        @apply text-textColor1 font-semibold text-18;
+        @apply border-1 border-textColor1 bg-white rounded-8;
+        @apply shadow-xsZ1v3;
+    }
```
iHiD: We shouldn't include font-sizes, rounding, etc in button classes like this. The text-18 and rounded-8 etc should come from adding the `btn-l` to the element. I also think in this case, the actual button is `btn-default` or `btn-enhanced`. I think the only difference is the shadow?

app/css/ui-kit/filters.css
```diff
@@ -15,4 +15,9 @@
     .filter-lightGold {
         filter: var(--lightGold-filter);
     }
+
+    .filter-shadow-huge {
+        -webkit-filter: drop-shadow(0px 4px 128px rgba(79, 114, 205, 0.8));
+        filter: drop-shadow(0px 4px 128px rgba(79, 114, 205, 0.8));
```
iHiD: You can do this via tailwind, which removes the need for the extra webkit variant: https://tailwindcss.com/docs/drop-shadow#arbitrary-values

app/javascript/components/common/GraphicalIcon.tsx
```diff
@@ -5,8 +5,8 @@ export function GraphicalIcon({
   icon,
   className = '',
   category,
-  width = null,
-  height = null,
+  width = 0,
+  height = 0,
```
iHiD: This means something very different I think. It'll set the images to have 0px width/height if the thing isn't specified, rather than using the default of the image.

dem4ron:
@iHiD about modal a11y: 
It seems to be a bad idea to have the same ids multiple times, but these elements get replaced, and there is only one unique id present in the DOM, and I don't really know how it could be easily bypassed in [this structure](https://github.com/exercism/website/blob/2a40119d05184754a1a6e378f2a83c80238992a6/app/javascript/components/modals/student/FinishMentorDiscussionModal.tsx#L137-L146). What do you think about this? 
I think I'll summon @SleeplessByte :)

app/css/ui-kit/buttons.css
dem4ron: There are a couple differences between any of those and this one in the design. But `btn-enhanced` is the closest to this.

> We shouldn't include font-sizes, rounding, etc in button classes like this. The text-18 and rounded-8 etc should come from adding the btn-l to the element. 

Ok!

app/javascript/components/common/GraphicalIcon.tsx
dem4ron: You are right, took a proper look at it, and the correct default value is `undefined` if you want to get the default sizes. I updated the type signature on [GraphicalIcon.](https://github.com/exercism/website/blob/40e6f245bd3e35fa9c29d8e75034246936b1bf81/app/javascript/components/common/GraphicalIcon.tsx#L4-L10)

SleeplessByte:
- If the question is: should you ever have two or more elements with the same ID, the answer is *no*, which I think you also say (unique).
- If the question is: is it okay to _remove_ an element with an ID to later add an element with that same ID, even if the added element isn't the same as the removed one: that's fine. I think this is the case given your comment about there only being one instance per ID in the DOM at the same time.

app/javascript/components/modals/student/finish-mentor-discussion-modal/DonationStep.tsx
```diff
@@ -0,0 +1,100 @@
...
+//               "subscription": null
+//           }
+//       }
+//   },
+//   "userSignedIn": true,
+//   "links": {
+//       "settings": "http://local.exercism.io:3020/settings/donations",
+//       "donate": "http://local.exercism.io:3020/donate"
+//   }
+// }
```
ErikSchierboom: I think this should be removed

iHiD:
@dem4ron @ErikSchierboom Can we get this green pls (might just need a rebase). Then can one of you update the OP so I know what I need to do get it over the line. Thanks.

ErikSchierboom:
My latest commit should have fixed the failing test.
@dem4ron do you know what the status is on the client-side? I think the modal was not yet shown, was it?

app/css/ui-kit/filters.css
iHiD: @dem4ron

app/helpers/react_components/student/mentoring_session.rb
```diff
@@ -6,26 +6,48 @@ class MentoringSession < ReactComponent
...
+                }
+              },
+              # TODO: add correct endpoint and initial data
+              user_signed_in: user_signed_in?,
+              captcha_required: !current_user || current_user.captcha_required?,
+              recaptcha_site_key: ENV.fetch('RECAPTCHA_SITE_KEY', Exercism.secrets.recaptcha_site_key),
+              links: {
+                settings: Exercism::Routes.donations_settings_url,
+                donate: Exercism::Routes.donate_url
+              }
```
iHiD: This nesting is wrong. This shouldn't all be in `links`. It should be part of the main hash.

iHiD:
@dem4ron This just needs the "don't close when clicking the background" thing setting and then I think your bit is done.

iHiD:
And also just check my comments.

test/system/flows/student/mentor_finished_discussion_test.rb
```diff
@@ -48,9 +48,9 @@ class MentorFinishedDiscussionTest < ApplicationSystemTestCase
           click_on "It was good!"
           fill_in "Leave #{discussion.mentor.handle} a testimonial (optional)", with: "Good mentor!"
           click_on "Finish"
-          click_on "Back to the exercise"
+          click_on "Continue to exercise"
 
-          assert_text "Nice, it looks like you’re done here!"
+          assert_text "Want to try another mentor?"
```
iHiD: @dem4ron Why has this changed. If someone selects "It was good!", I don't think we should be taking them down the "request another mentor" path

app/helpers/react_components/student/mentoring_session.rb
```diff
@@ -6,26 +6,44 @@ class MentoringSession < ReactComponent
...
+          exercise: SerializeMentorSessionExercise.(exercise),
+          iterations:,
+          mentor: mentor_data,
+          track_objectives: user_track&.objectives.to_s,
+          out_of_date: solution.out_of_date?,
+          videos:,
+          donation: {
+            user_signed_in: user_signed_in?,
+            captcha_required: !current_user || current_user.captcha_required?,
+            recaptcha_site_key: ENV.fetch('RECAPTCHA_SITE_KEY', Exercism.secrets.recaptcha_site_key),
```
iHiD:
```suggestion
            recaptcha_site_key: Exercism.secrets.recaptcha_site_key,
```

app/helpers/react_components/student/mentoring_session.rb
```diff
@@ -6,26 +6,44 @@ class MentoringSession < ReactComponent
...
+          track: SerializeMentorSessionTrack.(track),
+          exercise: SerializeMentorSessionExercise.(exercise),
+          iterations:,
+          mentor: mentor_data,
+          track_objectives: user_track&.objectives.to_s,
+          out_of_date: solution.out_of_date?,
+          videos:,
+          donation: {
+            user_signed_in: user_signed_in?,
+            captcha_required: !current_user || current_user.captcha_required?,
```
iHiD:
```suggestion
            captcha_required: false
```

app/helpers/react_components/student/mentoring_session.rb
```diff
@@ -6,26 +6,44 @@ class MentoringSession < ReactComponent
...
+          discussion: discussion ? SerializeMentorDiscussionForStudent.(discussion) : nil,
+          track: SerializeMentorSessionTrack.(track),
+          exercise: SerializeMentorSessionExercise.(exercise),
+          iterations:,
+          mentor: mentor_data,
+          track_objectives: user_track&.objectives.to_s,
+          out_of_date: solution.out_of_date?,
+          videos:,
+          donation: {
+            user_signed_in: user_signed_in?,
```
iHiD:
```suggestion
            user_signed_in: true,
```

app/images/graphics/wizard-hat.svg
iHiD, line 1: Can we remove this now?

app/helpers/react_components/student/mentoring_session.rb
```diff
@@ -6,26 +6,44 @@ class MentoringSession < ReactComponent
...
+          exercise: SerializeMentorSessionExercise.(exercise),
+          iterations:,
+          mentor: mentor_data,
+          track_objectives: user_track&.objectives.to_s,
+          out_of_date: solution.out_of_date?,
+          videos:,
+          donation: {
+            user_signed_in: true,
+            captcha_required: false,
+            recaptcha_site_key: Exercism.secrets.recaptcha_site_key,
```
iHiD: Do we need this, @dem4ron as `captcha_required` is always false?

dem4ron: then probably we don't

app/images/graphics/wizard-hat.svg
dem4ron: yes, it isn't used elsewhere.

app/helpers/react_components/student/mentoring_session.rb
iHiD: OK. I'll leave you to remove it! :)

.github/labels.yml
```diff
@@ -175,7 +175,7 @@
 - name: "bot/bugsnag"
   description: ""
   color: "A32E92"
-  
+
```
ErikSchierboom: Why was this changed? This file is auto-generated so shouldn't be edited directly.

CODE_OF_CONDUCT.md
```diff
@@ -90,4 +90,4 @@ This policy was initially adopted from the Front-end London Slack community and
 A version history can be seen on [GitHub](https://github.com/exercism/website-copy/edit/main/pages/code_of_conduct.md).
 
 _This policy is a "living" document, and subject to refinement and expansion in the future.
-This policy applies to the Exercism website, the Exercism GitHub organization, any other Exercism-related communication channels (e.g.  Slack, Twitter, email) and any other Exercism entity or event._
+This policy applies to the Exercism website, the Exercism GitHub organization, any other Exercism-related communication channels (e.g. Slack, Twitter, email) and any other Exercism entity or event._
```
ErikSchierboom: This file is auto-synced from https://github.com/exercism/org-wide-files/blob/main/global-files/CODE_OF_CONDUCT.md, so any local changes will get overwritten at some point. Maybe revert?

app/helpers/react_components/donations/footer_form.rb
```diff
@@ -17,7 +17,7 @@ def to_s
             recaptcha_site_key: ENV.fetch('RECAPTCHA_SITE_KEY', Exercism.secrets.recaptcha_site_key),
             links: {
               settings: Exercism::Routes.donations_settings_url,
-              donate: Exercism::Routes.donate_url
+              confirm_params_return_url: Exercism::Routes.donate_url
```
ErikSchierboom: I'm not entirely sure about the name. What is this link used for? We usually don't use a `_url` suffix

app/javascript/components/modals/student/FinishMentorDiscussionModal.tsx
```diff
@@ -72,23 +72,36 @@ const Inner = ({
...
+          <Step.DonationStep
+            donation={donation}
+            links={links}
+            onSuccessfulDonation={(_, amount) => {
+              setDonatedAmount(amount)
+              send('SUCCESSFUL_DONATION')
+            }}
+          />
+        )
+      } else
```
ErikSchierboom: You could lose the `else` and just directly `return` here

app/javascript/components/types.ts
```diff
@@ -453,6 +453,26 @@ export type MentoredTrackExercise = {
   completedByMentor: boolean
 }
 
+export type MentoringSessionDonation = {
+  showDonationModal: boolean
+  request: {
+    endpoint: string
+    options: {
+      initialData: string
```
ErikSchierboom: Is the initial data a `string`?

app/models/user.rb
```diff
@@ -221,6 +221,10 @@ def formatted_reputation(*args)
     User::FormatReputation.(rep)
   end
 
+  def donated_in_last_35_days?
```
ErikSchierboom, line 225:
```suggestion
  memoize
  def donated_in_last_35_days?
```

.github/labels.yml
dem4ron: This was merged from main at some point:
https://github.com/exercism/website/pull/6202/commits/28e1729898de403603625521195a01b501c07d55

CODE_OF_CONDUCT.md
dem4ron: same as above: https://github.com/exercism/website/pull/6202/commits/28e1729898de403603625521195a01b501c07d55

app/helpers/react_components/donations/footer_form.rb
dem4ron: This goes into `stripe.confirmPayment.options.confirmParams.[returnUrl | return_url]`

app/javascript/components/types.ts
dem4ron: Updated, thanks!

app/helpers/react_components/student/mentoring_session.rb
```diff
@@ -95,6 +110,14 @@ def iterations
 
         SerializeIterations.(solution.iterations, comment_counts:)
       end
+
+      def show_donation_modal
+        return false if current_user.insider?
+        return false if current_user.donated_in_last_35_days?
+
+        num_testimonials = current_user.provided_testimonials.count
+        num_testimonials.zero? || (num_testimonials % 3).zero?
```
ErikSchierboom: 0 % 3 = 0

```suggestion
        (num_testimonials % 3).zero?
```

app/css/ui-kit/filters.css
dem4ron: This was actually unused. It might have been used at some point in this PR, but upon doing a grep, I couldn't find any references to it. I've removed it.

.github/labels.yml
iHiD:
```suggestion
  
```

CODE_OF_CONDUCT.md
iHiD:
```suggestion
This policy applies to the Exercism website, the Exercism GitHub organization, any other Exercism-related communication channels (e.g.  Slack, Twitter, email) and any other Exercism entity or event._
```

app/helpers/react_components/donations/footer_form.rb
ErikSchierboom: I would probably rename it slightly to something like `donate_return_url`.

iHiD: I think I'm ok with the `confirm_params_return` bit as it's in the scope of the form, but agree it probably shouldn't have the `_url` prefix. Pragmatically I don't care if its hard to change it, but if it's just a couple of renames it would be better to be consistent.

dem4ron: I agree, but the reason I chose this name is because this is exactly what the API accepts. so it is pretty specific.

ErikSchierboom: Well, the reason I was suggesting `donate_return_url` (actually, `redirect_to_after_donation` or something like that might be better) is because it isn't coupled to Stripe. I feel that's an implementation detail 🤷

PR #2977, opened 2022-09-23
Fix minor visual bugs on Representer Preview

dem4ron:
This PR is part of this issue: https://github.com/exercism/exercism/issues/6547

- Adds scrolling to the RHS
- Adds syntax highlighting.

### UI Changes:

Screenshot 2022-09-23 at 13 15 01


app/javascript/hooks/index.ts
```diff
@@ -1,2 +1,3 @@
 export { useLogger } from './use-logger'
 export { useList } from './use-list'
+export { useHighlighting } from './use-sytnax-highlighting'
```
iHiD: use-sytnax-highlighting

There's a typo here - it should be `use-syntax-highlighting` (`n` and `t` the other way around)

PR #2983, opened 2022-09-26
Fix refreshing bug on Representation-tables

dem4ron:
This PR takes care of these from this issue https://github.com/exercism/exercism/issues/6547:
- Fix bug that number of requests doesn't go down when representations have been provided.
- Fix this caching bug: https://www.loom.com/share/f50400a550484b2889693b911cdee80b?from_recorder=1&focus_title=1


iHiD:
Thanks @dem4ron! At some point soon, it'd be good to start adding system tests to verify these behaviours and stop regressions. That's something you can maybe work through with @ErikSchierboom in the week I'm away - this PR would be a good example to pair on.

PR #2984, opened 2022-09-26
Fix XSS in profile bio

### LINKED ISSUE #6553: Stored XSS in profile bio
## Description

Cross-site scripting vulnerabilities arise when user input is incorporated in application responses in an unsafe way. An attacker can use this vulnerability to inject malicious JavaScript code that will be executed within the context of the victim's browser.

> The reflection endpoint is `/profiles/<user>`.

## Impact

An attacker can use the vulnerability to inject malicious JavaScript code into the application, which will execute within the browser of any user who views the affected endpoint(s).

As the session cookies are protected by the `httponly` and `secure` flags, it is not possible to steal a user's session. Although more sophisticated attacks are possible such as a wormable browser exploit (see [BeEF](https://beefproject.com/)).

## Remediation

* Output encoding should be applied in order to safely display data provided by a user.
* User controlled input should **not** be considered as trusted and therefore input validation checks should be performed where it is necessary.
* Perform input sanitization (e.g: DOMPurify)

As a defense-in-depth strategy, the following measures could also be incorporated into the web application:

* Use and enforce a strict `Content-Security-Policy`
* Install a Web Application Firewall (WAF)

## Analysis

A user is able to change his bio at `/settings` endpoint. It can be observed in `app/javascript/components/settings/ProfileForm.tsx:L85-L93`:

```tsx
// Lines 6-10
type User = {
  name: string
  location: string
  bio: string
}

[...]

const [user, setUser] = useState<User>(defaultUser)

[...]

// Lines 85-93
	<div className="bio field">
	<label htmlFor="user_bio" className="label">
	  Bio
	</label>
	<textarea
	  id="user_bio"
	  value={user.bio || ''}
	  onChange={(e) => setUser({ ...user, bio: e.target.value })}
	/>
```

When we declare a state variable with `useState`, it returns an array with two items. The first item is the current value, and the second is a function to update that value.

The following function is responsible to send the data to the backend controller and it is found at `app/javascript/components/settings/useSettingsMutation.tsx:L23-L27`:

```tsx
sendRequest({
	endpoint: endpoint,
	method: method,
	body: JSON.stringify(body),
})
```

The template in `app/views/profiles/_header.html.haml:L27` :

```haml
- if user.bio.present?
	= ReactComponents::Common::Expander.new(user.bio, button_text_compressed: "Read more", button_text_expanded: "Read less", class_name: "bio")
```

In `app/helpers/react_components/common/expander.rb:L4-L11`:

```ruby
def initialize(content, button_text_compressed:, button_text_expanded:, class_name: nil)
	super()
	
	@content = content
	@button_text_compressed = button_text_compressed
	@button_text_expanded = button_text_expanded
	@class_name = class_name
end
```

image

iHiD:
Closes https://github.com/exercism/exercism/issues/6553

@dem4ron Please do the following:

- [x] Add a boolean parameter to `Expander` (in app/javascript/components/common/Expander.tsx) for `contentIsSafe`
- [x] If `contentIsSafe` is `true`, continue to use `dangerouslySetInnerHTML`. If it's `false`, set the content as a string.
- [x] Proxy that parameter through the bootstrapper.

### DIFF
--- a/app/commands/exercise/representation/search.rb
+++ b/app/commands/exercise/representation/search.rb
@@ -2,7 +2,7 @@ class Exercise::Representation::Search
   include Mandate
 
   # Use class method rather than constant for easier stubbing during testing
-  def self.requests_per_page = 20
+  def self.requests_per_page = 21
 
   initialize_with with_feedback: Mandate::NO_DEFAULT, mentor: Mandate::NO_DEFAULT, criteria: nil, track: nil,
     order: :most_submissions, page: 1, paginated: true, sorted: true, only_mentored_solutions: false do
--- a/app/commands/exercise/representation/search.rb
+++ b/app/commands/exercise/representation/search.rb
@@ -2,7 +2,7 @@ class Exercise::Representation::Search
   include Mandate
 
   # Use class method rather than constant for easier stubbing during testing
-  def self.requests_per_page = 21
+  def self.requests_per_page = 20
 
   initialize_with with_feedback: Mandate::NO_DEFAULT, mentor: Mandate::NO_DEFAULT, criteria: nil, track: nil,
     order: :most_submissions, page: 1, paginated: true, sorted: true, only_mentored_solutions: false do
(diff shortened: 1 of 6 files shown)


iHiD:
@dem4ron This is another one that could do with a regression test (cc @ErikSchierboom for when I'm away)

dem4ron (approved):

PR #2985, opened 2022-09-26
Only show representations with >1 submissions

iHiD:


app/commands/exercise/representation/search.rb
```diff
@@ -10,7 +10,9 @@ def self.requests_per_page = 20
   end
 
   def call
-    @representations = Exercise::Representation.includes(:exercise, :track)
+    @representations = Exercise::Representation.
+      includes(:exercise, :track).
+      where('num_submissions > 1')
```
ErikSchierboom, line 15: How does this work with indexing?

iHiD: On filtering its faster by about 80ms. On count, it's slower by about 2ms.

🤷

ErikSchierboom: Okay.

ErikSchierboom: Do we need to update our indexes?

ErikSchierboom (approved):
LGTM

app/commands/exercise/representation/search.rb
iHiD: No. It's just fast 🤷 

I mean, adding extra indexes might help but it's not slower without them from what I can tell.

iHiD:
@ErikSchierboom Any chance you could get this over the line. I've tried 4 times to get it green and am failing. It'll just be changing a total_count down 1, or something probably. Thanks!

ErikSchierboom:
Should be fixed with: https://github.com/exercism/website/pull/2985/commits/b14c079044d158dab6cc7b60389316f0c7d80ff6
I also added a test for the num_submissions filtering: https://github.com/exercism/website/pull/2985/commits/462d714c159804c89a5db3de6393c89634996570

iHiD:
Thank you!! 💙

PR #2986, opened 2022-09-26
Reset page in Representers on criteria and order change

dem4ron:


iHiD:
Works well, thanks! Can you make yourself an issue to go through anywhere else pagination is used and check this please. That's one you can do in the week I'm away.

PR #2987, opened 2022-09-26
Persist filters on Representation table

dem4ron:
part of this issue: https://github.com/exercism/exercism/issues/6547

Queries are stored in local storage. 
- This makes it available to persist state of filters while switching from one table to another. (If you don't like this, it can be disabled.) Clicking on the **Automation** button opens up an unfiltered state of the table, since it is opened from Ruby. However **Feedback submitted** button opens up your previously saved filter results.
- When clicking on *X*, the table will be loaded in with your previous  filters.


iHiD:
@ErikSchierboom One for you to review pls :)

ErikSchierboom:
@dem4ron This is superseded by https://github.com/exercism/website/pull/2996 right?

PR #2994, opened 2022-09-28
Fix rack attack sidekiq issue

ErikSchierboom:
- Add test to verify sidekiq works
- Fix error


config/initializers/rack_attack.rb
```diff
@@ -9,6 +9,8 @@ def routed_to
     "#{route[:controller]}##{route[:action]}"
   rescue ActionController::RoutingError
     nil
+  rescue NoMethodError
+    nil
```
ErikSchierboom: This is an ugly fix, but I've looked into the source code and couldn't find a way around this :(

iHiD (approved):

PR #3017, opened 2022-10-06
Track-dashboard UI

dem4ron:
### Updated UI:

Screenshot 2022-10-13 at 15 09 43

Screenshot 2022-10-13 at 15 11 41

Screenshot 2022-10-13 at 15 14 00

Screenshot 2022-10-13 at 15 11 57

Screenshot 2022-10-13 at 15 12 35

Screenshot 2022-10-13 at 15 13 41


app/css/pages/track-build.css
```diff
@@ -1,5 +1,329 @@
...
+        @apply filter-textColor6 ml-12;
+        width: 12px;
+        height: 12px;
+    }
+}
+
+.action-required {
+    @apply py-8 px-16 mb-24;
+    @apply label-large text-16;
+    background-color: rgb(250, 250, 255);
```
dem4ron:
```suggestion

```

app/css/ui-kit/filters.css
```diff
@@ -19,4 +19,32 @@
     .filter-orange {
         filter: var(--orange-filter);
     }
+
+    /* I don't think we need duplicates of all the filters - 
+    one here, on in colors.css, or do we? */
```
dem4ron:
```suggestion
    one here, one in colors.css, or do we? */
```

app/views/tracks/build.html.haml
```diff
@@ -1,3 +1,200 @@
...
+        = link_to "", class: "btn btn-primary btn-large border-1 border-textColor1 rounded-8 shadow-xsZ1v3" do
+          Learn more
+
+
+      .flex.mb-16
+        = graphical_icon "building", css_class: "h-[64px] w-[64px] mr-20", hex: true
+        .flex.flex-col
+          %h2.text-h2 Build
+          %p.text-18.text-textColor1.leading-150 Use your knowledge to build the #{@track.title} track itself
+        -# TODO add React component with authors/contributors here
```
dem4ron: and everywhere else

ErikSchierboom:
Data is cached in Redis
Data is updated once a day
Precalculate data for _all_ tracks

Rolling average for +num students (last 30 days or so)

Started: solution exists
Attempts: submissions (average number)
Completed: percentage of started

Consider optimizing num analyzer comments (maybe store last updated at)

Is analysis data[:comments] present and has content?

app/commands/track/unimplemented_practice_exercises.rb
```diff
@@ -0,0 +1,15 @@
+class Track::UnimplementedPracticeExercises
```
ErikSchierboom: At some point, it might be nice to show the unimplemented exercises.
Something for a follow-up PR I feel

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,359 @@
...
+  end
+
+  def health
+    num_exemplar = component_health_statuses[:exemplar].to_i
+    num_healthy = component_health_statuses[:healthy].to_i
+
+    return :exemplar if num_exemplar == 5
+    return :healthy if num_exemplar + num_healthy == 5
+
+    :needs_attention
```
ErikSchierboom: I'm being quite strict, where the overall status is `:needs_attention` when at least one component has the `:needs_attention` or `:missing` status.

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,359 @@
...
+    # TODO: use error status to determine health (unhealthy if everything fails)
+    return :missing unless track.has_test_runner?
+    return :needs_attention if test_runner_version < 2 && track.course?
+    return :healthy if test_runner_version < 3 && track.course?
+
+    :exemplar
+  end
+
+  memoize
+  def test_runner_version = [1, Submission::TestRun.pluck(:version).last.to_i].max
```
ErikSchierboom, line 116: I'm infering the test runner version from the last test run's `version`

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,359 @@
...
+
+  def test_runner_version_target
+    return 1 unless track.has_test_runner?
+    return 2 if test_runner_version < 2
+    return 3 if test_runner_version < 3 && track.course?
```
ErikSchierboom, line 121: There is no need to upgrade to version 3 if the track does not have learning mode enabled

app/models/track.rb
```diff
@@ -1,15 +1,19 @@
 class Track < ApplicationRecord
   extend FriendlyId
   extend Mandate::Memoize
+  include Track::BuildStatus
```
ErikSchierboom, line 4: I've struggled with how I wanted to access the build status. Via a command? Maybe, but we don't do that often.
I've opted for an approach similar to `User::Roles`, where a module is included.

app/views/tracks/build/analyzer_tooltip.html.haml
```diff
@@ -0,0 +1,23 @@
...
+      %p.text-p-base
+      - case health
+      - when :exemplar
+        Gives feedback on #{@status.analyzer.display_rate_percentage}% of the submitted iterations!
+      - when :healthy
+        Gives feedback on #{@status.analyzer.display_rate_percentage}% of the submitted iterations.
+      - when :needs_attention
+        No feedback has been given.
+      - when :missing
+        The analyzer has not yet been implemented.
```
ErikSchierboom: We can tweak this information later on, but this gives us something to work with (same for the other tooltips)

ErikSchierboom:
We need to run once dark-deployed:

```ruby
Track.find_each {|t| Track::UpdateBuildStatus.(t)}
```

app/commands/track/unimplemented_practice_exercises.rb
iHiD: Commands should be verbs.

```suggestion
class Track::RetrieveUnimplementedPracticeExercises
```

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,359 @@
+class Track::UpdateBuildStatus
+  include Mandate
+
+  initialize_with :track
+
+  def call
+    track.build_status = build_status
```
iHiD, line 7: Should this not be `track.update(...)`. If it's not updating, I'd rather this command was `DetermineTrackStatus` and then the `track.build_status = ` was part of a wider command.

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,359 @@
...
+      representer:,
+      analyzer:
+    }
+  end
+
+  def health
+    num_exemplar = component_health_statuses[:exemplar].to_i
+    num_healthy = component_health_statuses[:healthy].to_i
+
+    return :exemplar if num_exemplar == 5
```
iHiD: Can you set `5` to a constant here pls, with a comment as to what it entails?

iHiD: Or maybe to `component_health_statuses.length` or something

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,359 @@
...
+    syllabus_health,
+    practice_exercises_health
+  ].tally
+
+  def volunteers
+    volunteer_user_ids = User::ReputationPeriod.where(
+      period: :forever,
+      about: :track,
+      track_id: track.id
+    ).select(:user_id)
```
iHiD, line 51: I think you want to also set `category: :any` here. IIRC the index can only work if that is explicitely set.

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,359 @@
...
+    volunteers = User.with_attached_avatar.where(id: volunteer_user_ids).order(reputation: :desc).take(12)
```
iHiD, line 52: A lot of users have `avatar_url` field set, rather than having had an avatar uploaded (e.g. if we took the data from the GH oauth). So we might want some sort of OR here.

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,359 @@
...
+  end
+
+  def mentor_discussions
+    {
+      num_discussions: track.mentor_discussions.count
+    }
+  end
+
+  def test_runner
+    status_counts = track.submissions.where(tests_status: %i[passed failed errored exceptioned]).group(:tests_status).count
```
iHiD: This will probably be quicker without the `where` and then just exclude `pending` from the totalling bit.

app/commands/track/update_build_status.rb
iHiD: This isn't what you want :) 

`pluck` immediately executes. So you're getting all the versions from all test runs then getting the last in the array. You want to probably use `order(id: :desc).pick(:version).to_i` instead.

app/commands/track/update_build_status.rb
iHiD: Add this as a comment?

app/commands/track/update_build_status.rb
ErikSchierboom: No, as `build_status` is not a materialized column of `Track`, but instead automatically assigned to Redis: https://github.com/exercism/website/pull/3017/files#diff-ccd907c771daebc66834912a9e0dca52fe42ac4580297fd82f8e29debdfa6a83

iHiD: Ah - not got there yet :)

app/commands/track/update_build_status.rb
iHiD: OK, so from our chat on Slack I understand this just does `include(...)` effectively.

What I was trying to suggest was that if we only want to show faces here, then we'd want to filter to get records either with an attached_avatar, or with `avatar_url` being a proper value. I can't remember how to do that efficiently though.

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,368 @@
...
+
+  def test_runner_version_target
+    return 1 unless track.has_test_runner?
+    return 2 if test_runner_version < 2
+    return 3 if test_runner_version < 3 && track.course?
+  end
+
+  def representer
+    {
+      num_runs: Submission::Representation.joins(submission: :exercise).where('exercises.track_id': track.id).count,
```
iHiD: Use the same style as what you have on L152?

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,368 @@
...
+      num_comments: Submission::Analysis.joins(submission: :exercise).where(submission: { exercises: { track: } }).sum(:num_comments),
+      display_rate_percentage: analyzer_display_rate_percentage,
+      volunteers: serialize_tooling_volunteers(track.analyzer_repo_url),
+      health: analyzer_health
+    }
+  end
+
+  memoize
+  def analyzer_display_rate_percentage
+    percentage(Submission::Analysis.with_comments.where(submission: track.submissions).count, track.submissions.count)
```
iHiD: Materialise this into a memoized variable in here as it's used a few times?

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,368 @@
...
+  memoize
+  def taught_concepts = Concept.where(id: concept_taught_exercise.keys).to_a
+
+  memoize
+  def concept_taught_exercise
+    Exercise::TaughtConcept.
+      where(exercise: active_concept_exercises).
+      pluck(:track_concept_id, :exercise_id).
+      map { |concept_id, exercise_id| { concept_id => exercise_id } }.
+      reduce({}, :merge)
```
iHiD: I think you can just do:

```suggestion
      to_h
```

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,368 @@
...
+      num_exercises: active_practice_exercises.size,
+      num_exercises_target: practice_exercises_num_exercises_target,
+      created: active_practice_exercises.map { |exercise| serialize_exercise(exercise) },
+      volunteers: practice_exercises_volunteers,
+      health: practice_exercises_health
+    }
+  end
+
+  def practice_exercises_num_exercises_target
+    num_unimplemented = Track::UnimplementedPracticeExercises.(track.reload).size
```
iHiD:
```suggestion
    num_unimplemented = Track::RetrieveUnimplementedPracticeExercises.(track.reload).size
```

iHiD: Why are you reloading here?

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,368 @@
...
+  memoize
+  def active_concept_exercises = track.concept_exercises.where(status: %i[active beta]).order(:position).to_a
+
+  memoize
+  def active_practice_exercises = track.practice_exercises.where(status: %i[active beta]).order(:position).to_a
+
+  def average_number_per_day(query, model)
+    total_count = query.where("#{model.table_name}.created_at >= ?", Time.current - NUM_DAYS_FOR_AVERAGE.days).count
+    (total_count / NUM_DAYS_FOR_AVERAGE.to_f).ceil
+  end
```
iHiD, line 289: Iterating by date is slow. This may be quicker if you do something like:

```
first_id = query.where("#{model.table_name}.created_at >= ?", Time.current - NUM_DAYS_FOR_AVERAGE.days).order(:id).pick(:id)
total_count = query.where("#{model.table_name}.id >= ?", first_id).count
```

But it might also not be, as you still have to look at all the rows. But 🤷

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,368 @@
...
+
+  memoize
+  def active_concept_exercises = track.concept_exercises.where(status: %i[active beta]).order(:position).to_a
+
+  memoize
+  def active_practice_exercises = track.practice_exercises.where(status: %i[active beta]).order(:position).to_a
+
+  def average_number_per_day(query, model)
+    total_count = query.where("#{model.table_name}.created_at >= ?", Time.current - NUM_DAYS_FOR_AVERAGE.days).count
```
iHiD: Iterating by date is slow. This may be quicker if you do something like:

```
first_id = query.where("#{model.table_name}.created_at >= ?", Time.current - NUM_DAYS_FOR_AVERAGE.days).order(:id).pick(:id)
total_count = query.where("#{model.table_name}.id >= ?", first_id).count
```

But it might also not be, as SQL might not be able to optimise it enough. But I think it will be.

iHiD: Cool method though.

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,368 @@
...
+  def percentage(count, total_count)
+    return 0.0 if total_count.zero?
+
+    ((count / total_count.to_f) * 100.0).round(1)
+  end
+
+  def average(count, total_count)
+    return 0 if total_count.zero?
+
+    (count.to_f / total_count).round(1)
```
iHiD, line 300: Experience tells me that nearly everything will be `1` or `2`. So I suspect rounding this to 1DP will be useful.

iHiD: Also in the UI can you say Mean or Median, not Average pls.

app/commands/track/update_build_status.rb
```diff
@@ -0,0 +1,368 @@
...
+  end
+
+  def percentage(count, total_count)
+    return 0.0 if total_count.zero?
+
+    ((count / total_count.to_f) * 100.0).round(1)
+  end
+
+  def average(count, total_count)
+    return 0 if total_count.zero?
```
iHiD:
```suggestion
    return 0.0 if total_count.zero?
```

app/css/ui-kit/filters.css
```diff
@@ -20,13 +20,31 @@
         filter: var(--orange-filter);
     }
 
+    /* I don't think we need duplicates of all the filters - 
+    one here, one in colors.css, or do we? */
```
iHiD, line 24: No, they should move to this file. This file is just newer.

app/helpers/view_components/track/build/health.rb
```diff
@@ -0,0 +1,34 @@
...
+  def text
+    case health_status.to_sym
+    when :exemplar
+      "#{person} exemplar ✨"
+    when :healthy
+      "#{person} healthy ✅"
+    when :needs_attention
+      "needs attention ⚠️"
+    else
+      "#{person}  missing ❓"
```
iHiD:
```suggestion
      "#{person} missing ❓"
```

app/helpers/view_components/track/build/health.rb
```diff
@@ -0,0 +1,34 @@
...
+    when :healthy
+      "text-healthyGreen"
+    when :needs_attention
+      "text-warning"
+    else
+      "text-textColor6"
+    end
+  end
+
+  def person = plural ? 'are' : 'is'
```
iHiD:
```suggestion
  def verb = plural ? 'are' : 'is'
```

app/images/icons/graph-stats.svg
```diff
@@ -1,9 +1,3 @@
...
-<defs>
-<linearGradient id="paint0_linear_7509_373583" x1="12.4995" y1="4.25537" x2="12.4995" y2="19.9956" gradientUnits="userSpaceOnUse">
-<stop stop-color="#2200FF"/>
-<stop offset="1" stop-color="#9E00FF"/>
-</linearGradient>
-</defs>
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M0.75 11.75H4.25C4.85567 11.7502 5.40202 11.3861 5.635 10.827L8.235 4.712C8.35657 4.42597 8.64196 4.24456 8.95255 4.25587C9.26313 4.26719 9.53456 4.46889 9.635 4.763L14.369 19.488C14.4694 19.7821 14.7409 19.9838 15.0515 19.9951C15.362 20.0065 15.6474 19.825 15.769 19.539L18.364 12.673C18.597 12.1139 19.1433 11.7498 19.749 11.75H23.249" stroke="black" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
```
iHiD, line 2: Can you make sure anywhere that this is used is updated pls (you probably have already 🙂)

iHiD (approved):
I've reviewed the big class. The rest all looks pretty straight-forward code (although lots of work!) so I'll just judge that by my eyes once it's deployed 😁 

This is fantastic work by you both. Thank you 🙂

app/commands/track/update_build_status.rb
ErikSchierboom: For no reason 🤦

app/commands/track/update_build_status.rb
ErikSchierboom: I can if I change the above line too:

```ruby
map { |concept_id, exercise_id| [concept_id, exercise_id] }.
      to_h
```

app/images/icons/graph-stats.svg
dem4ron: Yes, I've checked

PR #3022, opened 2022-10-06
Fallback to smaller thumbnails

ErikSchierboom:


kytrinyx (approved):

app/commands/community_video/retrieve_from_youtube.rb
```diff
@@ -43,6 +43,12 @@ def youtube_id
       params["v"]
     end
 
+    def thumbnail_url
+      %w[standard high medium default].each do |size|
+        return snippet["thumbnails"][size]["url"] if snippet["thumbnails"][size]
+      end
```
iHiD, line 49: You could probably refactor this to use `find` and `dig`. e.g. something like:
```ruby
%w[standard high medium default].find { |size| snippet.dig("thumbnails", size, "url") }
```

or
```ruby
thumbnails = snippet["thumbnails"]
%w[standard high medium default].find { |size| thumbnails.dig(size, "url") }
```

ErikSchierboom: I don't think I can. I started out with that, but what `find` appears to be doing (which matches [the docs](https://ruby-doc.org/core-3.1.2/Enumerable.html#method-i-find)) is to return the first element that is being queried on depending on the first truthy, not the value returned in the `find` call.

I could do this:

```
size = %w[standard high medium default].find { |size| snippet.dig("thumbnails", size, "url") }
snippet.dig("thumbnails", size, "url")
```

Let me know what you prefer

PR #3023, opened 2022-10-07
Allow representation comments to be celebratory

ErikSchierboom:
- Add celebratory status to representation feedback type
- Verify that celebratory representer comments count as non_actionable for submission
- Allow adding celebratory comment for representation


kytrinyx (approved):
💙

PR #3026, opened 2022-10-07
Add notifications for new representer feedback

ErikSchierboom:
- Add setting for new analysis feedback
- Add notification
- Add notification for analysis feedback added
- Add partially working email notification for analysis feedback added
- Send feedback added notifications


app/commands/exercise/representation/send_new_feedback_notifications.rb
```diff
@@ -0,0 +1,36 @@
...
+      raise "Incorrect feedback type"
+    end
+  end
+
+  def submission_representations = representation.submission_representations
+  def latest_active_iterations = submission_representations.map(&:iteration).select(&:latest?)
+
+  def latest_recent_active_iterations
+    latest_active_iterations.select { |iteration| iteration.created_at >= Time.zone.now - 2.weeks }
+  end
```
ErikSchierboom: I still need to check the performance of this

app/mailers/notifications_mailer.rb
```diff
@@ -118,4 +118,18 @@ def mentor_finished_discussion
...
+    notification = params[:notification]
+    @user = notification.user
+    @iteration = notification.iteration
+    @exercise = @iteration.exercise
+    @track = @iteration.track
+    @emphasis = notification.emphasis
+
+    @unsubscribe_key = :email_on_analysis_feedback_added_notification
+    @title = "New feedback is added to your solution on #{@track.title}/#{@exercise.title}"
+    subject = "[Mentoring] New feedback is added to your solution on #{@track.title}/#{@exercise.title}" # rubocop:disable Layout/LineLength
```
ErikSchierboom: I'm open to any suggestions for the copy

config/locales/communication_preferences/en.yml
```diff
@@ -18,6 +18,8 @@ en:
     email_on_student_finished_discussion_notification: "Email me when a student finished a discussion I am mentoring"
     email_on_student_added_iteration_notification: "Email me when a student adds an iteration to a solution I am mentoring"
 
+    email_on_analysis_feedback_added_notification: "Email me when new analysis feedback is added to my solution"
```
ErikSchierboom: I've made the naming quite generic as to potentially allow us to re-use this if we ever want to re-run analyzers.

config/locales/notifications/en.yml
```diff
@@ -16,6 +16,10 @@ en:
     added_to_contributors_page:
       1: >
         You now appear on our Contributors page. Thank you for contributing to Exercism!
+    analysis_feedback_added:
+      1: >
+        New feedback is added to iteration (#%{iteration_idx}) of your solution to <strong>%{exercise_title}</strong> in <strong>%{track_title}</strong>.
```
ErikSchierboom: Once again, feedback on the copy is appreciated

test/fixtures/notifications_mailer/analysis_feedback_added
```diff
@@ -0,0 +1,3 @@
+Hi handle-6b48cf20,
+New feedback is added to iteration (#1) of your solution to Lasagna on the Ruby track. We strongly recommend taking a look.
+Go to the iteration
```
ErikSchierboom: The test that uses this is currently failing. I'll need to fix that

app/commands/exercise/representation/send_new_feedback_notifications.rb
```diff
@@ -0,0 +1,36 @@
...
+    if representation.has_essential_feedback?
+      latest_active_iterations
+    elsif representation.has_actionable_feedback?
+      latest_recent_active_iterations
+    else
+      raise "Incorrect feedback type"
+    end
+  end
+
+  def submission_representations = representation.submission_representations
```
iHiD: This should `delegate`. I would have expected rubocop to sort that for you automatically. Maybe it can't for one line methods? Or maybe you still have it not running on commit?

app/commands/exercise/representation/send_new_feedback_notifications.rb
```diff
@@ -0,0 +1,36 @@
...
+  def latest_active_iterations = submission_representations.map(&:iteration).select(&:latest?)
```
iHiD: This feels like it's going to be hellish from a performance perspective. Imagine for where there are thousands of latest iterations.

app/commands/exercise/representation/send_new_feedback_notifications.rb
```diff
@@ -0,0 +1,36 @@
...
+  def call = iterations.each { |iteration| send_notification(iteration) }
+
+  private
+  def iterations
+    if representation.has_essential_feedback?
+      latest_active_iterations
+    elsif representation.has_actionable_feedback?
+      latest_recent_active_iterations
+    else
+      raise "Incorrect feedback type"
```
iHiD: I'd have a small preference for this being a noop and removing the guard below. Then we're encapsulating the knowledge into one place, whereas currently we have two different places we have to remember to keep in sync.

app/commands/exercise/representation/submit_feedback.rb
```diff
@@ -25,6 +26,12 @@ def award_reputation_token!(token_type)
       def mentor_is_author?
         mentor == representation.feedback_author
       end
+
+      def send_notifications!
+        return unless representation.has_essential_feedback? || representation.has_actionable_feedback?
```
iHiD: (This is the guard I refered to above)

app/mailers/notifications_mailer.rb
iHiD:
```suggestion
    @title = "New feedback added to your solution on #{@track.title}/#{@exercise.title}"
    subject = "[Mentoring] New feedback has been added to your solution on #{@track.title}/#{@exercise.title}" # rubocop:disable Layout/LineLength
```

config/locales/communication_preferences/en.yml
iHiD:
```suggestion
    email_on_automated_feedback_added_notification: "Email me when new automated feedback is added to my solution"
```

config/locales/notifications/en.yml
iHiD:
```suggestion
        New feedback has been added to iteration (#%{iteration_idx}) of your solution to <strong>%{exercise_title}</strong> in <strong>%{track_title}</strong>.
```

app/commands/exercise/representation/send_new_feedback_notifications.rb
ErikSchierboom: No, it shouldn't have. This is in the `private` section, which is when Rubucop doesn't suggest `delegate` as I think that would make it public.

app/commands/exercise/representation/send_new_feedback_notifications.rb
ErikSchierboom: Yes, this was just a quick and dirty first pass

app/commands/exercise/representation/send_new_feedback_notifications.rb
ErikSchierboom: Excellent suggestion

app/commands/exercise/representation/send_new_feedback_notifications.rb
ErikSchierboom: I've now refactored this to only run one query and the result appears to be quite fast. It ran in production in 0.02 seconds when tested on my Elixir high-score representation (248 matches). The query cost is around 980.

The resulting query is:

```sql
SELECT 
    `iterations`.`id`
FROM
    `iterations`
        LEFT JOIN
    `iterations` AS `i` ON `i`.`solution_id` = `iterations`.`solution_id`
        AND `i`.`deleted_at` IS NULL
        AND `i`.`idx` > `iterations`.`idx`
WHERE
    `iterations`.`deleted_at` IS NULL
        AND (`i`.`id` IS NULL)
        AND `iterations`.`id` IN (SELECT 
            `iterations`.`id`
        FROM
            `submission_representations`
                INNER JOIN
            `submissions` ON `submissions`.`id` = `submission_representations`.`submission_id`
                INNER JOIN
            `iterations` ON `iterations`.`submission_id` = `submissions`.`id`
        WHERE
            `submission_representations`.`ast_digest` = '590c2e4d266ec9a7ec5967db46145391640dcbbf');
```

app/models/user/notifications/automated_feedback_added_notification.rb
```diff
@@ -0,0 +1,29 @@
...
+          iteration_idx: iteration.idx,
+          track_title: track.title,
+          exercise_title: exercise.title,
+          emphasis:
+        }
+      end
+
+      def emphasis
+        representation.has_essential_feedback? ? 'strongly ' : nil
+      end
```
iHiD, line 24:
```suggestion
    end
    def email_should_send?
      super && representation.created_at > "2022-10-12"
    end
```

ErikSchierboom:
Essential notifications:

```ruby
Exercise::Representation.where(feedback_type: %i[essential actionable]).find_each {|r| Exercise::Representat
ion::SendNewFeedbackNotifications.(r)}
```

app/models/iteration.rb
```diff
@@ -7,9 +7,15 @@ class Iteration < ApplicationRecord
 
   has_one :exercise, through: :solution
   has_one :track, through: :exercise
+  has_one :user, through: :solution
   has_one :test_run, through: :iteration
 
   scope :not_deleted, -> { where(deleted_at: nil) }
+  scope :latest, lambda {
```
ErikSchierboom, line 14: I made the latest iterations logic into a scope, which makes it easier to unit test and use.

iHiD: Really nice.

PR #3029, opened 2022-10-11
Fix track header caching when enabling/disabling learning mode

ErikSchierboom:


app/helpers/view_components/track/header.rb
```diff
@@ -84,6 +85,8 @@ def guard!
         raise "Incorrect track nav tab" unless TABS.include?(selected_tab)
       end
 
+      def practice_mode = !!user_track&.practice_mode?
```
iHiD, line 88: Reuse this on L27 too?

iHiD (approved):
One small comment

PR #3030, opened 2022-10-12
Allow staff to see automation page

ErikSchierboom:


app/controllers/application_controller.rb
```diff
@@ -66,6 +66,7 @@ def ensure_mentor!
   def ensure_supermentor!
```
ErikSchierboom, line 65: We have a similar method named `ensure_mentor!`, do we also want to add the `staff` check there?

app/models/user/roles.rb
```diff
@@ -1,9 +1,10 @@
 module User::Roles
   def founder? = roles.include?(:founder)
   def admin? = roles.include?(:admin)
-  def staff? = roles.include?(:staff)
-  def maintainer? = roles.include?(:maintainer)
-  def supermentor? = roles.include?(:supermentor)
+  def staff? = roles.include?(:staff) || admin?
+  def maintainer? = roles.include?(:maintainer) || admin?
```
iHiD: And staff here too?

ErikSchierboom: I was unsure of that at first, but it's probably fine. Updated

iHiD (approved):

PR #3031, opened 2022-10-12
Don't automatically unlock help

ErikSchierboom:
We'll also need to fix the existing values in the database

```ruby
Solution.where(unlocked_help: true).in_batches do |batch|
    batch.update_all(unlocked_help: false)
  end
end
```


iHiD:
I think we can do:
`Solution.update_all(unlocked_help: false)`
which should be quicker I think.

iHiD (approved):

PR #3044, opened 2022-10-13
Show page for individual approaches

dem4ron:
## UI changes:

Screenshot 2022-10-13 at 14 19 57

Screenshot 2022-10-13 at 14 20 42


app/controllers/tracks/approaches_controller.rb
```diff
@@ -23,9 +23,11 @@ def show
...
         avatar_url: "https://avatars.githubusercontent.com/u/135246?v=4",
-        handle: "Erik",
+        handle: "ErikSchierboom",
+        name: "Erik Schierboom",
         formatted_reputation: 999
       ),
+      content_html: "<h1>Hello</h1>",
       title: "Utilising In-built Date Functions for Consistency",
       published_at: DateTime.now
     )
```
iHiD: @ErikSchierboom I presume this shouldn't be hardcoded? :)

app/helpers/react_components/track/credits.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      def to_s
+        super("track-credits", {
+          avatar_urls:,
+          top_count:,
+          top_label:,
+          bottom_count:,
+          bottom_label:,
+          max:
+        }
```
iHiD: You probably don't need the `{` and `}`, but if you do, then this one and the line below should be on the same line (to match how they are on L7)

app/views/tracks/approaches/show.html.haml
```diff
@@ -1,17 +1,23 @@
...
-      .py-8.px-24.bg-gray.border-2.border-gradient.rounded-100.flex.flex-row.items-center{ class: "h-[68px]" }
-        = graphical_icon "balloons", category: :graphics, css_class: "h-[40px] mr-12"
-        = graphical_icon "building", category: :graphics, css_class: "h-48 mr-12"
-        -# = track_icon css_class: "h-[40px] w-[40px] mr-12"
-          . = exercise_icon "amusement-park", css_class: "h-48 mr-12"
-        .flex.flex-col
-          .text-h5.whitespace-nowrap Amusement park
-          .textColor-6.font-normal.leading-150.text-16 Csharp
+      = link_to Exercism::Routes.track_exercise_path(@track, @exercise) do
+        .py-8.px-24.bg-gray.border-2.border-gradient-bgGray.rounded-100.flex.flex-row.items-center{ class: "h-[68px]" }
```
iHiD: Merge these two lines. If you ever have a single div within a link, then normally the div's classes can be inlined into the link.

iHiD:
This is looking really good, @dem4ron!

app/controllers/tracks/approaches_controller.rb
dem4ron: this was me :) I just used it as dummy data to show something.

app/views/tracks/approaches/show.html.haml
dem4ron: true!

app/assemblers/assemble_exercise_approach_share_panel.rb
```diff
@@ -0,0 +1,13 @@
+class AssembleExerciseApproachSharePanel
```
iHiD: This should be a serializer, not an assembler - there's no assembling happening.

This should be `SerializeApproach`.

iHiD: This should be a serializer, not an assembler - there's no assembling happening.

This should be `SerializeApproach`.

app/commands/git/sync_exercise_approach_introduction.rb
```diff
@@ -0,0 +1,23 @@
+class Git::SyncExerciseApproachIntroduction
+  include Mandate
+
+  initialize_with :exercise, :config
+
+  def call
+    exercise.update!(
```
iHiD, line 7: Is this update actually needed? It feels like the `Create` below is actually doing the work of creating. What extra does this do?

app/commands/user/combine_authors_and_contributors.rb
```diff
@@ -0,0 +1,11 @@
+class User::CombineAuthorsAndContributors
```
iHiD: The doesn't feel like a command to me as it doesn't actually do anything mutating, and it doesn't feel like it's part of the `User` scope. I'll have a think about where it should go.

app/commands/git/sync_exercise_approaches.rb
```diff
@@ -8,47 +8,23 @@ def initialize(exercise)
     end
 
     def call
+      exercise.update(approaches:)
```
iHiD, line 12:
```suggestion
      # This removes any approaches that aren't read from the config below
      exercise.update(approaches:)
```

app/commands/user/combine_authors_and_contributors.rb
ErikSchierboom: Yeah I know. Maybe an assembler?

iHiD (approved):

PR #3051, opened 2022-10-14
Move approaches git logic into separate model

ErikSchierboom:


app/controllers/tracks/approaches_controller.rb
```diff
@@ -52,28 +52,21 @@ def introduction
...
-  memoize
-  def introduction_num_contributors = @exercise.approach_introduction_contributors.count
-
   def introduction_avatar_urls
     avatar_urls = proc { |users, limit| users.order("RAND()").limit(limit).select(:avatar_url).to_a.map(&:avatar_url) }
 
     target = 3
     urls = avatar_urls.(@exercise.approach_introduction_authors, target)
-    if urls.size < 3 && introduction_num_contributors.positive?
+    if urls.size < 3 && @exercise.approach_introduction_contributors.count.positive?
```
iHiD:
```suggestion
    if urls.size < 3 && @exercise.approach_introduction_contributors.length.positive?
```

If you're materialising them, it's better to use `length` rather than `count` as `length` will use the materialized array whereas `count` will do a query.

app/models/git/exercise/approaches.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  memoize
+  def commit = repo.lookup_commit(git_sha)
+
+  memoize
+  def track = Git::Track.new(repo:)
+
+  # Yes, the filename should be duplicated. I think this is a GitHub bug.
+  NEW_GITHUB_URL = "https://github.com/exercism/%<track_slug>s/new/main/%<filepath>s?filename=%<filename>s".freeze
+  EDIT_GITHUB_URL = "https://github.com/exercism/%<track_slug>s/edit/main/%<filepath>s".freeze
```
iHiD: I much prefer `#{...}` rather than `%<...>s`

iHiD: (We should add a rubocop rule for this if there's one we can use)

ErikSchierboom: I know, but this is parameterized using a `format` call. Doesn't `#{...}` require concrete values to be specified?

ErikSchierboom:
Closing in favor of #3044

PR #3052, opened 2022-10-14
Hide concept exercises when learning mode disabled

### LINKED ISSUE #6559: Concept exercises are visible even when the syllabus is hidden
Currently on the Haskell track the work-in-progress concept exercises and concept pages are available to students, even though the syllabus itself is still hidden. In [exercism\haskell#1109](https://github.com/exercism/haskell/pull/1109#issuecomment-1260547963) @ErikSchierboom suggests this might be a bug.

ErikSchierboom:
- Calculate track's num_exercises based on course?
- Add published scope to exercise
- Add available exercise methods to track
- Use exercise filtering for external user track
- Simplify tests for external user track
- Add started scope on exercise
- Use started scope in user_track
- Add available scope to exercise
- Use available scope in user_track

Closes https://github.com/exercism/exercism/issues/6559

### DIFF
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -47,8 +47,8 @@ class Track < ApplicationRecord
   end
 
   def recache_num_exercises!
-    filtered_exercises = course ? exercises : practice_exercises
-    filtered_exercises = filtered_exercises.where(status: %i[active beta])
+    filtered_exercises = exercises.where(status: %i[active beta])
+    filtered_exercises = filtered_exercises.where(type: PracticeExercise.model_name.name) unless course?
 
     update_column(:num_exercises, filtered_exercises.count)
   end
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -48,7 +48,7 @@ class Track < ApplicationRecord
 
   def recache_num_exercises!
     filtered_exercises = exercises.where(status: %i[active beta])
-    filtered_exercises = filtered_exercises.where(type: PracticeExercise.model_name.name) unless course?
+    filtered_exercises = filtered_exercises.where(type: PracticeExercise.to_s) unless course?
 
     update_column(:num_exercises, filtered_exercises.count)
   end
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -47,7 +47,10 @@ class Track < ApplicationRecord
   end
 
   def recache_num_exercises!
-    update_column(:num_exercises, exercises.where(status: %i[active beta]).count)
+    filtered_exercises = course ? exercises : practice_exercises
+    filtered_exercises = filtered_exercises.where(status: %i[active beta])
+
+    update_column(:num_exercises, filtered_exercises.count)
   end
 
   def to_param = slug
--- a/app/models/user_track.rb
+++ b/app/models/user_track.rb
@@ -81,6 +81,7 @@ class UserTrack < ApplicationRecord
     status = %i[active beta]
     status << :wip if maintainer?
 
+    exercises = exercises.where(type: PracticeExercise.model_name.name) unless track.course?
     exercises.where(status:).or(exercises.where(id: solutions.select(:exercise_id)))
   end
 
--- a/app/models/user_track.rb
+++ b/app/models/user_track.rb
@@ -81,7 +81,7 @@ class UserTrack < ApplicationRecord
     status = %i[active beta]
     status << :wip if maintainer?
 
-    exercises = exercises.where(type: PracticeExercise.model_name.name) unless track.course? || maintainer?
+    exercises = exercises.where(type: PracticeExercise.to_s) unless track.course? || maintainer?
     exercises.where(status:).or(exercises.where(id: solutions.select(:exercise_id)))
   end
 
--- a/app/models/user_track.rb
+++ b/app/models/user_track.rb
@@ -159,6 +159,8 @@ class UserTrack < ApplicationRecord
 
   memoize
   def maintainer?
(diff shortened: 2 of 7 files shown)


app/models/exercise.rb
```diff
@@ -58,6 +58,9 @@ class Exercise < ApplicationRecord
     source: :contributor
 
   scope :sorted, -> { order(:position) }
+  scope :published, -> { where(status: %i[active beta]) }
+  scope :started, ->(user_track) { where(id: user_track.solutions.select(:exercise_id)) }
+  scope :available, ->(user_track) { published.or(started(user_track)) }
```
ErikSchierboom: I've had a _lot_ of issues coming up with sensible names.

The idea is that `published` returns exercises which the track intends to be available to students now.
The `started` scope includes all exercises for which the user has submitted a solution.
The `available`  scope combines these two scopes.

Let me know if you can come up with better names.

app/models/track.rb
```diff
@@ -29,6 +29,10 @@ class Track < ApplicationRecord
   delegate :debugging_instructions, :representer_normalizations, to: :git
   delegate :content, :edit_url, to: :mentoring_notes, prefix: :mentoring_notes
 
+  after_commit do
+    recache_num_exercises! if saved_changes.key?("course")
```
ErikSchierboom: We now need to recalculate the number of exercises in a track when the `course`  field changes.

ErikSchierboom: Changing a track's `course` status should now update the number of exercises available.

app/models/track.rb
```diff
@@ -42,7 +46,7 @@ def self.for_repo(repo)
   end
 
   def recache_num_exercises!
-    update_column(:num_exercises, exercises.where(status: %i[active beta]).count)
+    update_column(:num_exercises, exercises.published(self).count)
```
ErikSchierboom: I've moved the `active`/`beta` logic into a scope on `Exercise`. It was repeated in a couple of places.

app/models/user_track/external.rb
```diff
@@ -10,46 +10,29 @@ def initialize(track)
...
 
     def user = nil
     def tutorial_exercise_completed? = false
     def anonymous_during_mentoring? = true
 
-    memoize
-    def exercises
-      enabled_exercises(track.exercises)
-    end
+    def solutions = Solution.none
```
ErikSchierboom: This is a trick to have the `scope :started, ->(user_track) { where(id: user_track.solutions.select(:exercise_id)) }` scope work for external user tracks too. It generates SQL that the query optimizer will detect as never giving any results and thus it will be skipped (`SELECT `solutions`.`exercise_id` FROM `solutions` WHERE (1=0))`).

app/models/exercise.rb
```diff
@@ -58,6 +58,10 @@ class Exercise < ApplicationRecord
     source: :contributor
 
   scope :sorted, -> { order(:position) }
+  scope :course, ->(track) { where(type: PracticeExercise.model_name.name) unless track.course? }
```
ErikSchierboom: Note: this is a conditional scope

app/models/exercise.rb
```diff
@@ -58,6 +58,10 @@ class Exercise < ApplicationRecord
...
+  scope :published, ->(track) { where(status: %i[active beta]).course(track) }
+  scope :started, ->(user_track) { where(id: user_track.solutions.select(:exercise_id)) }
+  scope :available, ->(user_track) { published(user_track.track).or(started(user_track)) }
```
ErikSchierboom: Naming is hard...

app/models/exercise.rb
iHiD: Can you explain what this does pls?

app/models/track.rb
iHiD: Nice.

app/models/exercise.rb
ErikSchierboom: Sure. If the track does _not_ have learning mode enabled, it will only return practice exercises. However, if the track _does_ have learning mode enabled, it'll not do anything so both concept and practice exercises are returned.

ErikSchierboom:
Rewrite this an use logic from https://github.com/exercism/website/pull/3260

app/models/user_track.rb
```diff
@@ -78,6 +78,7 @@ def unlocked_exercises_for_exercise(exercise)
   end
 
   def enabled_exercises(exercises)
+    exercises = exercises.where(type: PracticeExercise.model_name.name) unless track.course?
```
ErikSchierboom: We probably want track maintainers to see these exercises. I can rebase/update once https://github.com/exercism/website/pull/3260 is merged

app/models/track.rb
```diff
@@ -47,7 +47,10 @@ def self.for_repo(repo)
   end
 
   def recache_num_exercises!
-    update_column(:num_exercises, exercises.where(status: %i[active beta]).count)
+    filtered_exercises = exercises.where(status: %i[active beta])
+    filtered_exercises = filtered_exercises.where(type: PracticeExercise.model_name.name) unless course?
```
iHiD: Maybe this would be simpler?

```suggestion
    filtered_exercises = filtered_exercises.where(type: PracticeExercise.to_s) unless course?
```

iHiD (approved):
Seems sane!

app/models/track.rb
ErikSchierboom: It would! I'll update

PR #3056, opened 2022-10-17
Responsive tweaks on dashboard and landing page

dem4ron:
Will resolve this issue: https://github.com/exercism/exercism/issues/6581

https://user-images.githubusercontent.com/66035744/196198129-391f70bc-2cdd-4ad2-bee0-1d7ad17ae1d5.mov

### Fix image size here:

Screenshot 2022-10-17 at 16 06 49

### DIFF
--- a/app/css/pages/dashboard.css
+++ b/app/css/pages/dashboard.css
@@ -66,7 +66,7 @@
         }
     }
     & header.dashboard-header {
-        @apply flex lg:flex-row flex-col gap-48 items-center pt-32 mb-48;
+        @apply flex lg:flex-row flex-col gap-y-48 items-center pt-32 mb-48;
 
         & .intro {
             @apply flex;
--- a/app/css/pages/dashboard.css
+++ b/app/css/pages/dashboard.css
@@ -66,7 +66,7 @@
         }
     }
     & header.dashboard-header {
-        @apply flex items-center pt-32 mb-48;
+        @apply flex lg:flex-row flex-col gap-48 items-center pt-32 mb-48;
 
         & .intro {
             @apply flex;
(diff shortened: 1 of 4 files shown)


app/views/pages/index.html.haml
```diff
@@ -75,7 +75,7 @@
                         = track_icon(track)
                     .font-semibold.text-textColor7.ml-16.whitespace-nowrap +#{sample[:num_tracks]} more
 
-        = image_tag "screenshots/landing-page-exercise.jpg", alt: "", role: :presentation, category: :graphics, class: "main-graphic lg:max-w-1-2 lg:ml-80 mt-32 lg:mt-96 shadow-lgZ1", width: 668, height: 611
+        = image_tag "screenshots/landing-page-exercise.jpg", alt: "", role: :presentation, category: :graphics, class: "main-graphic lg:max-w-1-2 lg:ml-80 mt-32 lg:mt-96 shadow-lgZ1", width: '100%', height: 611
```
iHiD: I believe you should leave the `width` in still. The browser will use the width/height to calculate the aspect ratio, then respect the CSS to render at the correct size for the user's viewport.

I've verified this works and will push :)

app/css/pages/dashboard.css
```diff
@@ -66,7 +66,7 @@
         }
     }
     & header.dashboard-header {
-        @apply flex items-center pt-32 mb-48;
+        @apply flex lg:flex-row flex-col gap-48 items-center pt-32 mb-48;
```
iHiD: This causes the page to render differently.

Compare these:
Screenshot 2022-10-19 at 11 57 16
 
Screenshot 2022-10-19 at 11 57 12

I think what you want is to either remove the gap at the higher viewport, or just do `gap-y`

```suggestion
        @apply flex lg:flex-row flex-col gap-y-48 items-center pt-32 mb-48;
```

PR #3064, opened 2022-10-19
Build community-page

dem4ron:
## Current UI:

Screenshot 2022-10-19 at 14 27 13
Screenshot 2022-10-19 at 14 27 47
Screenshot 2022-10-19 at 14 27 58
Screenshot 2022-10-19 at 14 32 04
Screenshot 2022-10-19 at 14 29 04
Screenshot 2022-10-19 at 14 30 09

## Individual podcast page - `community-show`:

Screenshot 2022-10-20 at 13 36 44

- some tweaks are needed in the podcast-grid header so it matches the design
- refactor react component, DRY


app/views/community/index.html.haml
```diff
@@ -0,0 +1,111 @@
...
+            = graphical_icon "avatar-placeholder", css_class: 'w-[32px] h-[32px] mr-8'
+            .flex.flex-col.mr-auto
+              %h5.text-h5 What&apos;s in Ruby 3.6?
+              .text-textColor6.leading-150
+                Last replied ErikSchierboom
+            .message-count
+              = graphical_icon "message-bubble-square"
+              56
+      .grid
+        = image_tag "forum.png"
```
iHiD: When using `image_tag` you either need to set `alt: "Something"` or `alt: "", role: :presentation` depending on whether the image has any presentational value or not. In this situation, I'd say it doesn't have value, so I'd go for the second option.

PR #3068, opened 2022-10-20
Fix other approaches css

iHiD:
My approach
- Stop putting wrapper divs inside links 😛 
- Use flex to vertically align and space.
- Fixed height on the top.
- `mt-auto` to anchor something to the bottom

Screenshot 2022-10-20 at 15 52 24


dem4ron:
> Stop putting wrapper divs inside links

@iHiD Yeah, sorry, you told me this already 2 times, and this piece of code was written between those two 😄

PR #3070, opened 2022-10-20
Set trust_level for over 20 rep users after 30secs

iHiD:


app/commands/user/set_discourse_trust_level.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    discourse_user_id = client.by_external_id(user.id)['id']
+    client.update_trust_level(discourse_user_id, level: 2)
+  end
+
+  def client
+    DiscourseApi::Client.new("https://forum.exercism.org").tap do |client|
+      client.api_key = Exercism.secrets.discourse_api_key
+      client.api_username = "system"
+    end
```
ErikSchierboom, line 20: Did you consider making this part of `config`, like the GitHub client?

app/commands/user/set_discourse_trust_level.rb
```diff
@@ -0,0 +1,19 @@
+class User::SetDiscourseTrustLevel
+  include Mandate
+
+  initialize_with :user
+
+  def call
+    return unless user.reputation >= 20
+
+    discourse_user_id = client.by_external_id(user.id)['id']
```
ErikSchierboom, line 9: Do we need to guard for the exceptional case where the client could not be found?

ErikSchierboom (approved):

app/commands/user/set_discourse_trust_level.rb
iHiD: As this is the only place we're using it, not so far. But if we start doing more with it then yeah, that'd make sense. So if you're retrieving stuff elsewhere, feel free to do that 🙂

app/commands/user/set_discourse_trust_level.rb
iHiD: Yeah, good shout. I'll do that.

PR #3071, opened 2022-10-21
Sync community stories from blog

ErikSchierboom:
- Add community story model
- Sync community stories
- Gracefully handle missing stories or posts keys


db/migrate/20221021084927_create_community_stories.rb
```diff
@@ -0,0 +1,23 @@
...
+      t.belongs_to :interviewee, null: false, foreign_key: {to_table: :users}
+
+      t.string :uuid, null: false
+      t.string :slug, null: false
+      t.string :title, null: false      
+      t.string :blurb, null: false, limit: 280
+      t.string :thumbnail_url
+      t.string :image_url
+      t.string :youtube_id
+      t.integer :length_in_minutes, limit: 2, null: false
```
ErikSchierboom: Not sure if any of these could be `null`

app/commands/git/sync_blog.rb
```diff
@@ -36,6 +40,31 @@ def create_or_update_post(data)
...
+        title: data[:title],
+        blurb: data[:blurb],
+        published_at: data[:published_at],
+        thumbnail_url: data[:thumbnail_url],
+        image_url: data[:image_url],
+        youtube_id: data[:youtube_id],
+        length_in_minutes: data[:length_in_minutes]
+      }
+
+      story = CommunityStory.create_or_find_by!(uuid: data[:uuid]) do |d|
```
iHiD:
```suggestion
      story = CommunityStory.find_create_or_find_by!(uuid: data[:uuid]) do |d|
```

db/migrate/20221021084927_create_community_stories.rb
iHiD: I think they should all be `null: false`

db/migrate/20221021084927_create_community_stories.rb
```diff
@@ -0,0 +1,23 @@
+class CreateCommunityStories < ActiveRecord::Migration[7.0]
+  def change
+    create_table :community_stories do |t|
+      t.belongs_to :interviewer, null: false, foreign_key: {to_table: :users}
+      t.belongs_to :interviewee, null: false, foreign_key: {to_table: :users}
+
+      t.string :uuid, null: false
```
iHiD:
```suggestion
      t.string :uuid, null: false, index: {unique: true}
```

(Pretty sure this works, rather than having it as a seperate line below)

app/models/git/blog.rb
```diff
@@ -5,31 +5,26 @@ class Blog
...
+
+    def post_content_for(slug) = content_for("posts", slug)
+    def story_content_for(slug) = content_for("stories", slug)
 
-    def content_for(slug)
-      filepath = "posts/#{slug}.md"
+    def content_for(dir, slug)
+      filepath = "#{dir}/#{slug}.md"
       repo.read_text_blob(head_commit, filepath)
     end
```
iHiD: You could probably one line this.

iHiD (approved):
I'd set `null: false` for them all, but other than that, let's get this merged!

PR #3078, opened 2022-10-24
Community page patch

dem4ron:
this PR is mostly about VideoGrid component on the Community page

- adds missing types
- refactors components, imports
- adds loading state to videoGrid
- avoids refetching on mount
- creates new query params for VideoGrid filters in Ruby with `video_` prefix
- adds a new `useQueryParams` hook, that uses the browser API to manage queries, does the filtering on the URL not on a copy of request.query obj. This way you can use query params for multiple React components
- resets page the same way Representation-table does
- Adds fallback component if there are no search results.
- Back to community link is now white

### UI Changes:

Screenshot 2022-10-27 at 10 41 50

Screenshot 2022-10-24 at 17 11 28


app/javascript/components/community/video-grid/useVideoGrid.tsx
```diff
@@ -72,9 +156,22 @@ export function useVideoGrid(
...
+function pushQueryParams(key: string, value: string | number): void {
+  const url = new URL(window.location.toString())
+
+  if (typeof value === 'string') {
+    if (value && value.length > 0) {
+      url.searchParams.set(key, value)
+    } else {
+      url.searchParams.delete(key)
+    }
+  }
```
dem4ron:
```suggestion
  
    if (value && value.length > 0) {
      url.searchParams.set(key, value)
    } else {
      url.searchParams.delete(key)
    }
  
```

dem4ron:
@iHiD Please take a look at this. If you'd like you can merge this but there are parts that must be covered in a next PR. e.g. ContributorList params behaviour.

app/helpers/react_components/community/video_grid.rb
```diff
@@ -39,7 +38,7 @@ def tracks
   def query
     {
       page: current_page,
-      track_slug: params[:track_slug],
+      track_slug: params[:video_track_slug],
```
ErikSchierboom: Why isn't this `video_track_slug`?

app/javascript/components/community/video-grid/TrackFilterList.tsx
```diff
@@ -61,14 +58,13 @@ export const TrackFilterList = ({
...
   isFetching: boolean
-  value: AutomationTrack
-  setValue: (value: AutomationTrack) => void
-  cacheKey: string
+  value: VideoTrack
+  setValue: (value: VideoTrack) => void
   sizeVariant?: 'large' | 'multi' | 'inline' | 'single' | 'automation'
   countText: string
-  // TODO remove this
+  // TODO this is present in the design but not yet implemented
```
ErikSchierboom, line 67: I like the todo!

app/javascript/components/community/video-grid/index.tsx
```diff
@@ -1,13 +1,25 @@
...
-import { useVideoGrid } from './useVideoGrid'
+import { HandleTrackChangeType, useVideoGrid } from './useVideoGrid'
+import type { CommunityVideoType } from '@/components/types'
 
 type VideoGridProps = {
-  data: any
+  data: {
+    tracks: VideoTrack[]
+    request: Request
+    selectedTrackSlug: string | null
```
ErikSchierboom: I can't find any place where this field is used

app/javascript/components/community/video-grid/useVideoGrid.tsx
```diff
@@ -14,15 +21,36 @@ export type APIResponse = {
...
-): any {
+  tracks: VideoTrack[]
+): UseVideoGridReturnType {
   const initialTrack =
-    tracks.find((track) => track.slug == selectedTrackSlug) || tracks[0]
+    tracks.find(
+      (t) =>
+        t.slug ===
+        new URLSearchParams(window.location.search).get('video_track_slug')
+    ) || tracks[0]
```
ErikSchierboom: Could't this use the `videoRequest` object?

app/javascript/components/community/video-grid/useVideoGrid.tsx
```diff
@@ -32,37 +60,61 @@ export function useVideoGrid(
...
+        videoTrackSlug: track.slug,
+        videoPage: 1,
+      })
     },
     [setPage, setQuery, request.query]
   )
 
+  useQueryParams(request.query)
+
+  const handlePageTurn = useCallback(
```
ErikSchierboom: Any reason not to use:

```suggestion
  const handlePageChange = useCallback(
```

which seems more in line with what we're doing elsewhere

app/javascript/components/types.ts
```diff
@@ -795,3 +795,42 @@ export type Modifier =
   | 'placeholder-shown'
   | 'autofill'
   | 'read-only'
+
+export type CommunityVideoAuthorLinks = {
+  profile?: string
+  channel_url?: string
```
ErikSchierboom: An underscore in the property name?

app/javascript/components/community/video-grid/index.tsx
dem4ron: ok, I will remove this!

app/javascript/components/community/video-grid/useVideoGrid.tsx
dem4ron: I think it could, I'll change this.

app/javascript/components/types.ts
dem4ron: removed as it wasn't used. probably I copied it from the return object to transform it into a type, and left it like that. Thanks for spotting!

app/helpers/react_components/community/video_grid.rb
dem4ron: fixed it!

ErikSchierboom (approved):

PR #3079, opened 2022-10-24
Fix approach

iHiD:
@dem4ron The approach page is a bit messy. Feels very not-polished compared to the other stuff you've made - I'm guessing it was just rushed at the end? 

I've tidied it enough that I'm happy to deploy it, but it needs work still this week. You can take a look and see what is right/wrong but two points that come out of this PR specifically:
1. I've removed the shapes as they were different to the designs in both style and placement.
2. I've added a fixed height to the "Credits" component to avoid CLS, but does this actually need to be React at all. Why? It's also different visually from Tai's design (in terms of content and styling/spacing)

Screenshot 2022-10-24 at 12 46 22
vs
Screenshot 2022-10-24 at 12 46 26


dem4ron:
The reason why the `Credits` is a react component is that we thought there'd be a modal that is opened when the Credits button is pressed, but the data is not there yet for that.

PR #3091, opened 2022-10-25
Don't show component if data is missing on video modal

dem4ron:
## Before
Screenshot 2022-10-25 at 10 18 25

## After
Screenshot 2022-10-25 at 10 18 34

### DIFF
--- a/app/javascript/components/track/approaches-elements/community-videos/CommunityVideoModal.tsx
+++ b/app/javascript/components/track/approaches-elements/community-videos/CommunityVideoModal.tsx
@@ -39,25 +39,21 @@ export function CommunityVideoModal({
         <VideoCredits links={video.links} author={video.author} />
       )}
 
-      {video.submittedBy.handle.length > 0 ||
-        (video.createdAt.length > 0 && (
-          <div className="text-center text-textColor6 leading-160 text-16">
-            {video.submittedBy.links.profile &&
-              video.submittedBy.handle.length > 0 && (
-                <>
-                  Posted by{' '}
-                  <a
-                    href={video.submittedBy.links.profile}
-                    className="underline"
-                  >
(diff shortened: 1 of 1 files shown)


app/javascript/components/track/approaches-elements/community-videos/CommunityVideoModal.tsx
```diff
@@ -38,17 +38,26 @@ export function CommunityVideoModal({
...
+                    className="underline"
+                  >
+                    @{video.submittedBy.handle}
+                  </a>
+                </>
+              )}
+            {video.createdAt.length > 0 &&
+              `&middot; ${dayjs(video.createdAt).format('D MMM YYYY')}`}
+          </div>
+        ))}
```
dem4ron: @ErikSchierboom 
This is a bit too noisy. Is there a truth that could make this simpler?
By this I mean "if `submittedBy` is missing, `createdAt` won't be there too."
In that case we can just return null if either is missing and don't need to do many so many conditional branching inside.

app/javascript/components/track/approaches-elements/community-videos/CommunityVideoModal.tsx
```diff
@@ -39,25 +39,21 @@ export function CommunityVideoModal({
...
-                  >
-                    @{video.submittedBy.handle}
-                  </a>
-                </>
-              )}
-            {video.createdAt.length > 0 &&
-              `&middot; ${dayjs(video.createdAt).format('D MMM YYYY')}`}
-          </div>
-        ))}
+      {(video.submittedBy.handle.length > 0 || video.createdAt.length > 0) && (
```
ErikSchierboom, line 42: I don't entirely understand this. The `submittedBy` field should never be empty as it is non-nullable in the database.

PR #3092, opened 2022-10-25
Open external links in new tab

dem4ron:
We probably want this. :)

### DIFF
--- a/app/views/community/show.html.haml
+++ b/app/views/community/show.html.haml
@@ -85,7 +85,7 @@
       %h2.text-h2.text-white.text-unnamed15.mb-8 Explore our swag
       .leading-150.text-20.text-borderLight.mb-24
         Enhance your desk. Enhance your look.
-      = link_to "https://swag.exercism.org", target: "_blank", rel: "noreferrer", class: 'btn-primary btn-l' do
+      = link_to "https://swag.exercism.org", class: 'btn-primary btn-l' do
         %span Get some swag
         = graphical_icon "arrow-right"
 
--- a/app/views/community/show.html.haml
+++ b/app/views/community/show.html.haml
@@ -10,7 +10,7 @@
             Accelerate your learning with Exercism&apos;s community.
           Ask questions, explore ideas, and get to know millions of other developers upskilling on Exercism.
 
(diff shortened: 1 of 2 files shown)


app/views/community/show.html.haml
```diff
@@ -10,7 +10,7 @@
             Accelerate your learning with Exercism&apos;s community.
           Ask questions, explore ideas, and get to know millions of other developers upskilling on Exercism.
 
-        = link_to 'https://forum.exercism.org/', class: "btn-primary btn-l mb-20 w-[fit-content]" do
+        = link_to 'https://forum.exercism.org/', target: "_blank", rel: "noreferrer", class: "btn-primary btn-l mb-20 w-[fit-content]" do
```
dem4ron: I don't know if we are worried about omitting `Referer`_(sic!)_ header, but as a best practice I did it like this.

app/views/community/stories/show.html.haml
```diff
@@ -14,7 +14,7 @@
             %h1.text-h1.mb-12.text-white= @story.title
             %p.text-p-large.mb-28.theme-dark= @story.blurb
             .flex.items-center
-              = link_to @story.youtube_external_url, class: 'youtube-link mr-32' do
+              = link_to @story.youtube_external_url, target: "_blank", rel: "noreferrer", class: 'youtube-link mr-32' do
```
iHiD, line 17: Keep this one.

iHiD:
I actually want the forum links to stay inside the website and not be external links, so I think all but one (which I've commented on) should stay inside. I'm 50/50 on the Swag one.

app/views/community/show.html.haml
```diff
@@ -85,7 +85,7 @@
       %h2.text-h2.text-white.text-unnamed15.mb-8 Explore our swag
       .leading-150.text-20.text-borderLight.mb-24
         Enhance your desk. Enhance your look.
-      = link_to "https://swag.exercism.org", class: 'btn-primary btn-l' do
+      = link_to "https://swag.exercism.org", target: "_blank", rel: "noreferrer", class: 'btn-primary btn-l' do
```
iHiD:
```suggestion
      = link_to "https://swag.exercism.org", class: 'btn-primary btn-l' do
```

PR #3093, opened 2022-10-25
Use rubygems not github for mysql2 gem

ErikSchierboom:
Just noticed that there finally is a new version released for the mysql2 gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem 'kaminari'
 gem 'oj'
 
 # Setup dependencies
-gem 'exercism-config', '>= 0.86.0' # gem 'exercism-config', path: '../exercism_config'
+gem 'exercism-config', '>= 0.86.0' # path: '../exercism_config'
 
 # Model-level dependencies
 gem 'image_processing', '~> 1.2'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -37,8 +34,7 @@ gem 'kaminari'
 gem 'oj'
 
 # Setup dependencies
-gem 'exercism-config', '>= 0.86.0'
-# gem 'exercism-config', path: '../exercism_config'
+gem 'exercism-config', '>= 0.86.0' # gem 'exercism-config', path: '../exercism_config'
```
ErikSchierboom: This is rubocop. I'm happy to revert if you'd like me to

Gemfile
```diff
@@ -9,10 +9,7 @@ gem 'jsbundling-rails'
 gem 'turbo-rails'
 gem 'propshaft', '0.4.0' # Changing this will break stuff. Work needed.
 
-gem 'mysql2',
-  github: 'brianmario/mysql2',
-  ref: '56c0ced2afe2cabe0b3f1a05e8e5e86713ebbc87'
-
+gem 'mysql2'
```
iHiD: Could you specify the specific version here pls (You can preceed with `~>` so that we get patch updates)

Gemfile
iHiD:
```suggestion
gem 'exercism-config', '>= 0.86.0' # path: '../exercism_config'
```

PR #3098, opened 2022-10-25
Improve Credits component

dem4ron:
## Before
Screenshot 2022-10-25 at 18 53 12

## After
Screenshot 2022-10-25 at 18 52 07

Screenshot 2022-10-25 at 18 53 57

Screenshot 2022-10-25 at 18 54 20

Component handles a lot of stuff

if `handle` is missing, falls back to 1 `[top_label]`

Screenshot 2022-10-25 at 18 56 26


iHiD (approved):
This looks good to me on the Rails side. I'll let Erik check the JS. Nice work! :)

app/javascript/packs/application.tsx
```diff
@@ -462,6 +461,7 @@ import { highlightAll } from '../utils/highlight'
 import { AutomationLockedTooltipProps } from '../components/tooltips/AutomationLockedTooltip.js'
 import { ApproachesProps } from '@/components/track/Approaches'
 import { ChartData } from '@/components/impact/Chart'
+import { CreditsUsersProp } from '@/components/common/Credits.js'
```
dem4ron: just noticing this `.js` at the end..

app/helpers/react_components/track/credits.rb
```diff
@@ -1,16 +1,15 @@
 module ReactComponents
   module Track
     class Credits < ReactComponent
```
ErikSchierboom: At some point we should probably rename this to something more generic, as it doesn't necessarily have to do anything with a track. Maybe it should be `Common::Credits`, which is also how it is named in React.

app/helpers/react_components/track/credits.rb
```diff
@@ -1,16 +1,15 @@
...
-      initialize_with :avatar_urls, :top_count, :top_label, :bottom_count, :bottom_label, max: 2, css_class: nil
+      initialize_with :users, :top_count, :top_label, :bottom_count, :bottom_label, css_class: nil
 
       def to_s
         super("track-credits", {
```
ErikSchierboom: If you apply the above suggestion, please also change this value.

app/javascript/components/common/AvatarGroup.tsx
```diff
@@ -1,30 +1,30 @@
...
-          <Avatar
-            className="face"
-            src={avatarUrl}
-            key={`${avatarUrl}${Math.random()}`}
-          />
-        ))}
-      {max > 0 && avatarUrls.length - max > 0 ? (
-        <div className="counter">+{Math.min(99, avatarUrls.length - max)}</div>
+    <div className={`c-faces-with-overflow-counter ${className}`}>
+      {users.slice(0, 2).map((user) => (
```
ErikSchierboom: You're guaranteed that users only ever has a maximum of three elements, so the slice is not needed.

app/javascript/components/common/AvatarGroup.tsx
```diff
@@ -1,30 +1,30 @@
 import React from 'react'
 import { Avatar } from '../common'
+import { CreditsUsersProp } from './Credits'
```
ErikSchierboom: This is giving me a TS error that no such type is exported

app/javascript/components/common/Credits.tsx
```diff
@@ -1,38 +1,48 @@
...
-          {pluralizeWithNumber(topCount, topLabel)}
-        </div>
+    <div className={`flex items-center ${className}`}>
+      <AvatarGroup
+        className={`mr-${overflow > 0 ? 12 : 8}`}
+        overflow={overflow}
+        users={users}
+      />
+      <div>
+        {topCount + Number(bottomCount) > 1 || users[0]['handle'].length === 0
```
ErikSchierboom, line 38: `users[0]['handle'].length === 0` I don't think this is ever true. Every user has a handle. Or are you considering the fact that the data might not be there?

app/javascript/packs/application.tsx
ErikSchierboom: I see a `.js` three lines above this line too

app/javascript/components/common/Credits.tsx
dem4ron: yes, just a fallback in case something would happen.

dem4ron:
Once this will be merged: https://github.com/exercism/website/pull/3116, the flakey test won't fail here.

PR #3099, opened 2022-10-25
Make community page responsive

iHiD:


iHiD:
@dem4ron I don't think anything here will be unexpected to you, but give it a check anyway. I like your pattern of using grid laying out two things as it works really well for responsive. I hated it at first, but I like it now 😁

PR #3106, opened 2022-10-27
Add support for exercise articles

ErikSchierboom:
- Update repo
- Rename approach
- Add schema and migrations for exercise articles
- Add models for exercise articles
- Add factories for exercise articles
- Add git models for exercise articles
- Add tests for filepaths for exercise with approaches
- Reduce nesting of class
- Sync articles


app/controllers/tracks/approaches_controller.rb
```diff
@@ -2,15 +2,12 @@ class Tracks::ApproachesController < ApplicationController
   include UseTrackExerciseSolutionConcern
   before_action :use_solution
   before_action :use_approach, only: :show
-  before_action :guard_accessible!, except: :tooltip_locked
+  before_action :guard_accessible!
```
iHiD, line 5: Is this deliberate?

Could you also move the `skip_...` above these `before_action`s pls.

iHiD: Ah - I see why this has changed :+1:

config/routes.rb
```diff
@@ -166,7 +166,11 @@
           get :tooltip_locked
         end
       end
-      resources :approaches, only: %i[index show], controller: "tracks/approaches" do
+
+      resources :articles, only: %i[index show], controller: "tracks/articles"
+      resources :approaches, only: %i[index show], controller: "tracks/approaches"
+
+      resources :dig_deeper, only: %i[index], controller: "tracks/dig_deeper" do
```
iHiD:
```suggestion
      resource :dig_deeper, only: %i[index], controller: "tracks/dig_deeper" do
```

This should be a single resource (as there can't be multiple dig-deepers as the resource name is singular) and the action should be `show` not index. You can then remove the `collection do` line below this too.

iHiD (approved):
I've only skimmed this but other than my routes comment it looks great and good to merge.

config/routes.rb
ErikSchierboom: Neat!

iHiD (approved):

PR #3116, opened 2022-10-28
Try to fix metrics test failures

ErikSchierboom:


test/commands/metric/create_test.rb
```diff
@@ -55,9 +55,9 @@ class Metric::CreateTest < ActiveSupport::TestCase
     action = :start_solution
     solution = create :concept_solution
     occurred_at = Time.current - 2.seconds
-    Exercism.request_context = { remote_ip: nil }
+    request_context = { remote_ip: nil }
```
ErikSchierboom, line 58: There is no need to set the global `Exercism.request_context` value in this test, and it might even have caused tests to fail due to the context changing whilst they are running. 🤞

test/commands/metric/create_test.rb
```diff
@@ -25,8 +25,8 @@ class Metric::CreateTest < ActiveSupport::TestCase
     solution = create :concept_solution
     track = solution.track
     user = solution.user
-    remote_ip = '127.0.0.1'
-    country_code = 'US'
+    remote_ip = SecureRandom.compact_uuid
```
ErikSchierboom, line 28: I'm now using random IP to stub as to now overwrite the default stub that is setup in the test helper

iHiD:
Let's try it 🤷

PR #3132, opened 2022-11-01
Fix overflow and height of snippets on Approaches

dem4ron:
# UI changes:

### shows 8 lines of code in a snippet.
Screenshot 2022-11-01 at 12 28 55

### gets rid of overflowing
Screenshot 2022-11-01 at 12 31 34

### DIFF
--- a/app/javascript/components/track/approaches-elements/ArticleSnippet.tsx
+++ b/app/javascript/components/track/approaches-elements/ArticleSnippet.tsx
@@ -16,7 +16,7 @@ export function ArticleSnippet({ article }: { article: Article }): JSX.Element {
         ref={codeBlockRef}
       >
         <div
-          className="c-cli-walkthrough overflow-hidden block"
+          className="overflow-hidden block"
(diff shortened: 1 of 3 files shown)


app/javascript/components/track/approaches-elements/ArticleSnippet.tsx
```diff
@@ -16,8 +16,12 @@ export function ArticleSnippet({ article }: { article: Article }): JSX.Element {
         ref={codeBlockRef}
       >
         <div
-          className="c-cli-walkthrough"
+          className="c-cli-walkthrough overflow-hidden block"
```
iHiD: Why is this using cli-walkthrough as a class?

dem4ron: I don't know, it doesn't have any effect here at the moment. Can I remove it, @ErikSchierboom?

ErikSchierboom: I don't think I created that file, so 🤷

PR #3137, opened 2022-11-02
Fixing editor layout shifting

dem4ron:
It is just a random proposal, as it was bugging me a bit. If you don't like it, I'm fine with that.

## Before:

https://user-images.githubusercontent.com/66035744/199544133-c0d811df-0651-4b98-a97a-2b5225b14ba5.mov

## After:

https://user-images.githubusercontent.com/66035744/199544165-8b7df5d8-721c-4cb7-a6bd-097be6fcaf27.mov

Screenshot 2022-11-02 at 16 08 49

Screenshot 2022-11-02 at 15 45 00


app/javascript/components/Editor.tsx
```diff
@@ -182,6 +185,7 @@ export default ({
 
       setSubmission(submission.uuid, { ...submission, testRun: testRun })
     },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
iHiD, line 198: What's the reasoning for disabling these here? Should they be disabled everywhere?

iHiD:
This is good, thanks. I've commented on the JS.

app/javascript/components/Editor.tsx
```diff
@@ -5,39 +5,41 @@ import React, {
...
+  EditorStatusSummary,
+  RunTestsButton,
+  SubmitButton,
+  InstructionsTab,
+  TestsTab,
+  ResultsTab,
+  InstructionsPanel,
+  TestsPanel,
+  ResultsPanel,
+} from './editor/index'
```
ErikSchierboom, line 42: This is nice

app/javascript/components/Editor.tsx
ErikSchierboom: I'm curious about that too. What is the reason we can't be exhaustive here?

dem4ron: This is exhaustive here.
The reason why the linter warns here is because it would like to see `[submission, setSubmission]` as deps, as the body of `useCallback` contains those state getter and setter hook-functions. 
But `submission` is wrapped in a `JSON.stringify`, so it no longer recognises/accepts it. 

If you don't stringify the object, as `SameObject!==SameObject` in JS, it will infinitely redeclare this function, because the hook will think that the `SameObject` has been updated, so it reinvokes itself.

This workaround wouldn't be necessary if React provided an opportunity to add a custom comparison function, but they don't -  despite the fact it is being requested for a long time.

Some of the dependency arrays _can_ be more exhaustive, but doesn't really have any effect, because those are static information that would never change on the fly, but are relevant information on the mount. 
e.g. `track.slug` - which in this case doesn't change, because when this page is loaded, track.slug - or the URL that fetches it is already determined by Ruby. This could change if everything was handled by React and a big state machine/store.

iHiD: Maybe add this to each of them:
```
# TODO: Remove stringify. See https://github.com/exercism/website/pull/3137/files#r1015500657 for details
```

ErikSchierboom (approved):

iHiD (approved):

PR #3141, opened 2022-11-03
Store num_comments for analysis

ErikSchierboom:


db/migrate/20221103092412_add_num_comments_to_submission_analyses.rb
```diff
@@ -0,0 +1,11 @@
+class AddNumCommentsToSubmissionAnalyses < ActiveRecord::Migration[7.0]
+  def change
+    add_column :submission_analyses, :num_comments, :integer, limit: 1, default: 0, null: false
```
ErikSchierboom: I've assumed that we'll never have an analyzer with more than 255 comments :)

db/schema.rb
```diff
@@ -1142,6 +1143,8 @@
   add_foreign_key "cohorts", "tracks"
   add_foreign_key "community_stories", "users", column: "interviewee_id"
   add_foreign_key "community_stories", "users", column: "interviewer_id"
+  add_foreign_key "community_videos", "exercises"
+  add_foreign_key "community_videos", "tracks"
```
ErikSchierboom, line 1151: These lines were automatically added and it seems like they should indeed have been there.

db/migrate/20221103092412_add_num_comments_to_submission_analyses.rb
iHiD: Can we specify algorithm and lock pls to avoid locking the db while this happens.

I think this is right, but check it:
```suggestion
    execute "ALTER TABLE submission_analyses ADD COLUMN num_comments TINYINT DEFAULT 0 NOT NULL, ALGORITHM=INPLACE, LOCK=NONE"
```

iHiD (approved):

PR #3144, opened 2022-11-04
Add CSP policy

iHiD:
This adds a reporting-only CSP policy (via a header) which shows any errors in Chrome Dev Tools. If we're happy there are no errors, we can then uncomment the meta tag.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -81,20 +81,25 @@ class ApplicationController < ActionController::Base
 
   # rubocop:disable Lint/PercentStringArray
   def csp_policy
-    ws_host = Rails.env.production? ? "exercism.org" : "local.exercism.io:3334"
-    stripe_base = "https://js.stripe.com"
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -78,11 +79,40 @@ def set_request_context
...
+      font: %w[https://fonts.gstatic.com],
+      style: %w['self' 'unsafe-inline' https://exercism.org https://d24y9kuxp2d7l2.cloudfront.net https://fonts.googleapis.com],
+      child: %w['none']
+
+    }.map do |type, domains|
+      "#{type}-src #{domains.join(' ')}"
+    end.join("; ")
+  end
+  helper_method :csp_policy
+  # rubocop:enable Lint/PercentStringArray
```
iHiD, line 110: (Rubocop bawks at the quotes around `'self'`)

ErikSchierboom (approved):

PR #3157, opened 2022-11-08
Add `Build` tab element

dem4ron:
Screenshot 2022-11-08 at 11 12 05

### TODO:
- Make the header visible when visiting `/build` page

### DIFF
--- a/app/views/tracks/build/show.html.haml
+++ b/app/views/tracks/build/show.html.haml
@@ -1,5 +1,3 @@
-- track_meta_tags(@user_track)
-
 #page-tracks-build.bg-white
   = render ViewComponents::Track::Header.new(@track, :build)
 
--- a/app/views/tracks/build/show.html.haml
+++ b/app/views/tracks/build/show.html.haml
@@ -1,7 +1,9 @@
 - track_meta_tags(@user_track)
 
 #page-tracks-build.bg-white
-  %header.absolute.top-0
+  = render ViewComponents::Track::Header.new(@track, :build)
+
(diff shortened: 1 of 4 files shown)


iHiD:
Let's not merge this for now. That will enable us to dark-deploy without this bit.

(this might be why you've done it as a seperate PR 🙂)

app/views/tracks/build/show.html.haml
```diff
@@ -1,6 +1,9 @@
+- track_meta_tags(@user_track)
```
ErikSchierboom: I've snuck this in as well

ErikSchierboom (commented):
@iHiD This PR is now ready for review.

One thing we need to remember is to clear the Rails cache for the header, as otherwise the build tab won't show up. What would be the most efficient way of doing that?

app/views/tracks/build/show.html.haml
iHiD: Nice

iHiD: You'll actually need to change this, else we'll have duplicate meta-titles which hurt us from an SEO perspective.

So you'll need to manually do:

```
content_for :meta_title, "#{user_track.track_title} build page on Exercism"
```

etc (same for description, image, etc)

ErikSchierboom: I'll then remove it and fix it in another PR (other track pages have this too)

iHiD (approved):
LGTM

PR #3159, opened 2022-11-08
Add credits to toolings

dem4ron:
Adds Credits component under tooling info sections.

### DIFF
--- a/app/javascript/components/common/Credits.tsx
+++ b/app/javascript/components/common/Credits.tsx
@@ -23,12 +23,12 @@ export function Credits({
 }: CreditsProps): JSX.Element | null {
   const [overflow] = useState<number>(topCount + Number(bottomCount) - 2)
 
-  if (users.length === 0 || !users) {
+  if (!users || users.length === 0) {
     return null
   }
 
   return (
-    <div className={`flex items-center ${className}`}>
+    <div className={`flex items-center ${className ? className : ''}`}>
       <AvatarGroup
         className={`mr-${overflow > 0 ? 12 : 8}`}
         overflow={overflow}
--- a/app/views/tracks/build/show.html.haml
+++ b/app/views/tracks/build/show.html.haml
@@ -200,6 +200,7 @@
                 = exercise_icon exercise
                 = exercise.title
               .record-value #{exercise.num_completed} completions (30% of attempts)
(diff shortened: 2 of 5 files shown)


app/views/tracks/build/show.html.haml
```diff
@@ -200,7 +201,6 @@
                 = exercise.title
               .record-value #{exercise.num_completed} completions (30% of attempts)
 
-        = render ReactComponents::Track::Credits.new(Array.new(5, 'https://avatars.githubusercontent.com/u/66035744?v=4'), 1, "volunteer", 3, "hero")
```
ErikSchierboom, line 214: Shouldn't this be replaced with a call to the renamed component?

dem4ron: I've already replaced those. check above.

app/javascript/components/common/Credits.tsx
```diff
@@ -23,12 +23,12 @@ export function Credits({
...
   const [overflow] = useState<number>(topCount + Number(bottomCount) - 2)
 
-  if (users.length === 0 || !users) {
+  if (!users || users.length === 0) {
     return null
   }
 
   return (
-    <div className={`flex items-center ${className}`}>
+    <div className={`flex items-center ${className ? className : ''}`}>
```
ErikSchierboom, line 31: Why is this change needed?

app/views/tracks/build/show.html.haml
ErikSchierboom: Ah

app/javascript/components/common/Credits.tsx
dem4ron: in this case it fails 'quickerly', and doesn't error out with a `cannot read propertly 'length' of undefined`.

ErikSchierboom: I think that was fixed in https://github.com/exercism/website/pull/3159/files#diff-46fb86f1af44dd4f4a5165f8fba3c09d544a6651a560bffb1a9b8c0f5b63e3ecR26. Could you try reverting and see if it works?

ErikSchierboom:
```suggestion
    <div className={`flex items-center ${className}`}>
```

PR #3175, opened 2022-11-11
Fix representer display rate percentage

ErikSchierboom:


app/commands/track/update_build_status.rb
```diff
@@ -123,22 +123,32 @@ def test_runner_version_target
...
       num_comments: representer_num_submissions_with_feedback,
       display_rate_percentage: representer_display_rate_percentage,
       volunteers: serialize_tooling_volunteers(track.representer_repo_url),
       health: representer_health
     }
   end
 
+  memoize
+  def representer_num_runs
+    Submission::Representation.joins(submission: :exercise).where(submission: { exercises: { track: } }).count
```
iHiD: I guess this will change now track is materialized?

ErikSchierboom: Yep!

PR #3187, opened 2022-11-15
Tweak build status a bit

ErikSchierboom:
- Make track healthy/exemplar when not having learning mode
- Use track-specific reputation for build status
- Show number of mentoring requests

### DIFF
--- a/app/commands/track/update_build_status.rb
+++ b/app/commands/track/update_build_status.rb
@@ -131,7 +131,7 @@ class Track::UpdateBuildStatus
 
   def representer
     {
-      num_runs: Submission::Representation.where(track:).count,
+      num_runs: representer_num_runs,
       num_comments: representer_num_submissions_with_feedback,
       display_rate_percentage: representer_display_rate_percentage,
       volunteers: serialize_tooling_volunteers(track.representer_repo_url),
@@ -140,13 +140,23 @@ class Track::UpdateBuildStatus
   end
 
   memoize
+  def representer_num_runs
+    Submission::Representation.where(track:).count
+  end
+
+  memoize
   def representer_num_submissions_with_feedback
     Exercise::Representation.with_feedback.joins(:submission_representations).count
   end
 
   memoize
+  def representer_num_submissions
+    Exercise::Representation.joins(:submission_representations).count
+  end
+
+  memoize
   def representer_display_rate_percentage
-    percentage(representer_num_submissions_with_feedback, num_submissions)
+    percentage(representer_num_submissions_with_feedback, representer_num_submissions)
   end
 
   memoize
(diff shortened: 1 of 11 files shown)


app/commands/track/update_build_status.rb
```diff
@@ -38,22 +38,16 @@ def component_health_statuses = [
...
-      track_id: track.id
-    ).select(:user_id)
-    volunteers = User.with_attached_avatar.where(id: volunteer_user_ids).order(reputation: :desc).take(12)
+    volunteers = AssembleContributors.({ track_slug: track.slug, page: 1 })
 
     {
-      num_volunteers: volunteer_user_ids.distinct.count,
-      users: SerializeAuthorOrContributors.(volunteers)
+      num_volunteers: volunteers.dig(:meta, :total_count),
+      users: volunteers[:results].take(NUM_VOLUNTEERS)
```
iHiD: While I like this idea, I don't think you want to do it. That AssembleContibutors thing is really slow due to the contextual data.

Instead, I'd go with your existing code, but replace the `where...select` with:
```
    User::ReputationPeriod::Search.(track_id: track.id)
```

which I think does what you want.

ErikSchierboom: But then I don't have the right, per-track reputation for volunteers.

iHiD: I think this is the going to be the most efficient way:

```
users = User::ReputationPeriod::Search.(track_id: track.id)
track_rep_mapping = User::ReputationToken.where(track_id: track.id).where(user: users).group(:user_id).sum(:value)
```

app/commands/track/update_build_status.rb
```diff
@@ -38,22 +38,25 @@ def component_health_statuses = [
...
+    track_volunteers = User::ReputationPeriod::Search.(track_id: track.id)
+    num_volunteers = track_volunteers.total_count
+    top_track_volunteers = track_volunteers.take(NUM_VOLUNTEERS)
+
+    contextual_data = User::ReputationToken.
+      where(track:).
+      where(user: top_track_volunteers).
+      group(:user_id).
+      sum(:value).
+      transform_values { |reputation| { activity: '', reputation: } }
```
ErikSchierboom: Let me know what you think of this new implementation

app/commands/track/update_build_status.rb
ErikSchierboom: I've changed the implementation. Let me know what you think

app/commands/track/update_build_status.rb
```diff
@@ -220,12 +225,20 @@ def num_concepts_target
   end
 
   memoize
-  def taught_concepts = Concept.where(id: concept_taught_exercise.keys).to_a
+  def taught_concepts
+    Concept.where(id: concept_taught_exercise.keys).
```
iHiD:
```suggestion
    Concept.where(id: taught_concepts_ids).
```

iHiD (approved):

PR #3188, opened 2022-11-15
Materialize track on submission tooling jobs

ErikSchierboom:
- Add track to submission_{analyses/representations/test_run} including index
- Store track with submission {analyses,representations,test_runs} model
- Use track field for filtering submission analyses/test_runs/representations

### DIFF
--- a/db/migrate/20221115123455_add_track_to_submission_test_runs.rb
+++ b/db/migrate/20221115123455_add_track_to_submission_test_runs.rb
@@ -4,9 +4,9 @@ class AddTrackToSubmissionTestRuns < ActiveRecord::Migration[7.0]
     add_index :submission_test_runs, %i[track_id id], order: {track_id: :asc, id: :desc}, unique: false, if_not_exists: true
 
     unless Rails.env.production?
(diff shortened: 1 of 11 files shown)


db/migrate/20221115123455_add_track_to_submission_test_runs.rb
```diff
@@ -0,0 +1,14 @@
+class AddTrackToSubmissionTestRuns < ActiveRecord::Migration[7.0]
+  def change
+    add_reference :submission_test_runs, :track, null: true, foreign_key: true, if_not_exists: true
+    add_index :submission_test_runs, %i[track_id id], order: {track_id: :asc, id: :desc}, unique: false, if_not_exists: true
+
+    unless Rails.env.production?
+      Submission::TestRun.includes(submission: :track).find_in_batches do |batch|
+        batch.each do |test_run|
```
iHiD: So the transaction can go between L7 and L8 now to make one transaction per batch, rather than 1,000 update transactions per batch.

iHiD: (Two different uses of transactions)

iHiD (approved):

PR #3191, opened 2022-11-16
Require min number of solutions mentored for automation feedback

ErikSchierboom:
- Tiny refactoring
- Tiny refactoring
- Only include tracks in automation when mentored at least 100 solutions

### DIFF
--- a/app/assemblers/assemble_representation_tracks_for_select.rb
+++ b/app/assemblers/assemble_representation_tracks_for_select.rb
@@ -4,15 +4,34 @@ class AssembleRepresentationTracksForSelect
   initialize_with :mentor, with_feedback: Mandate::NO_DEFAULT
 
   def call
-    tracks.map { |track| SerializeTrackForSelect.(track).merge(num_submissions: track_counts[track.id]) }
+    tracks.map do |track|
+      SerializeTrackForSelect.(track).merge(num_submissions: track_num_representations[track.id])
+    end
   end
 
   private
   memoize
-  def track_counts = representations.group(:track_id).count
+  def tracks
+    Track.where(id: track_ids_with_representation).
+      where(id: track_ids_with_min_num_mentored_solutions).
+      order(title: :asc)
+  end
+
+  def track_ids_with_representation = track_num_representations.keys
+  def track_ids_with_min_num_mentored_solutions = track_num_mentored_solutions.keys
 
   memoize
-  def tracks = Track.where(id: track_counts.keys).order(title: :asc)
+  def track_num_representations = representations.group(:track_id).count
+
+  memoize
+  def track_num_mentored_solutions
+    mentor.mentor_discussions.
+      joins(:request).
+      finished_for_student.
+      having('COUNT(*) >= ?', Mentor::Supermentor::MIN_NUM_SOLUTIONS_MENTORED_PER_TRACK).
+      group(:track_id).
+      count
+  end
 
   memoize
   def representations
--- a/app/helpers/react_components/mentoring/representations/without_feedback.rb
+++ b/app/helpers/react_components/mentoring/representations/without_feedback.rb
@@ -46,7 +46,8 @@ module ReactComponents
         end
 
         def representations_with_feedback_count
-          Exercise::Representation::Search.(mentor:, with_feedback: true, sorted: false, paginated: false).count
+          0 # TODO: support counting only tracks of which user is supermentor
+          # Exercise::Representation::Search.(mentor:, with_feedback: true, sorted: false, paginated: false).count
         end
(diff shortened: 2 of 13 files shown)


app/assemblers/assemble_representation_tracks_for_select.rb
```diff
@@ -4,18 +4,38 @@ class AssembleRepresentationTracksForSelect
...
+  def tracks
+    Track.where(id: track_ids_with_representation).
+      where(id: track_ids_with_min_num_mentored_solutions).
+      order(title: :asc)
+  end
+
+  def track_ids_with_representation = track_num_representations.keys
+
+  def track_ids_with_min_num_mentored_solutions
+    track_num_mentored_solutions.filter_map { |track_id, count| track_id if count >= MIN_NUM_SOLUTIONS_MENTORED_PER_TRACK }
```
ErikSchierboom: I could also do this filtering in MySQL using a `HAVING` clause. Your preference?

app/assemblers/assemble_representation_tracks_for_select.rb
```diff
@@ -4,18 +4,38 @@ class AssembleRepresentationTracksForSelect
...
+  end
 
   memoize
-  def tracks = Track.where(id: track_counts.keys).order(title: :asc)
+  def track_num_representations = representations.group(:track_id).count
+
+  memoize
+  def track_num_mentored_solutions
+    Mentor::Discussion.joins(solution: :exercise).group(:track_id).count
+  end
```
ErikSchierboom: For now, I'm calculating the number of mentored solutions each time one of the automation pages is shown. We could consider caching the value somewhere.

ErikSchierboom (commented):
So far, this PR only hides tracks with < 100 solutions mentored for the solution.
For a more fool-proof solution, we should also guard against this in the controller and API, to not allow users manually specifying a track slug.
I've postponed implementing that until we know whether or not we want to cache number of mentored solutions.

app/assemblers/assemble_representation_tracks_for_select.rb
```diff
@@ -4,18 +4,38 @@ class AssembleRepresentationTracksForSelect
...
+    track_num_mentored_solutions.filter_map { |track_id, count| track_id if count >= MIN_NUM_SOLUTIONS_MENTORED_PER_TRACK }
+  end
 
   memoize
-  def tracks = Track.where(id: track_counts.keys).order(title: :asc)
+  def track_num_representations = representations.group(:track_id).count
+
+  memoize
+  def track_num_mentored_solutions
+    Mentor::Discussion.joins(solution: :exercise).finished_for_student.group(:track_id).count
```
ErikSchierboom: I'm currently only counting discussion where the student has finished the discussion

app/assemblers/assemble_representation_tracks_for_select.rb
iHiD: I'd presume HAVING is faster? Unless it somehow means some index can't be used. So happy to go with that. I don't have a big preference.

iHiD (approved):
This looks great to me.

app/assemblers/assemble_representation_tracks_for_select.rb
ErikSchierboom: Updated

iHiD:
@ErikSchierboom What's blocking this?

ErikSchierboom:
@iHiD I need to add the "has mentored at least 100 solutions in the track" guard to a couple of places.

app/helpers/react_components/mentoring/representations/without_feedback.rb
```diff
@@ -46,7 +46,8 @@ def representations
         end
 
         def representations_with_feedback_count
-          Exercise::Representation::Search.(mentor:, with_feedback: true, sorted: false, paginated: false).count
+          0 # TODO: support counting only tracks of which user is supermentor
+          # Exercise::Representation::Search.(mentor:, with_feedback: true, sorted: false, paginated: false).count
```
ErikSchierboom, line 50: Right now, this would return an invalid count, as we're only showing tracks for which the user is a supermentor (>= 100 solutions mentored), but `Exercise::Representation::Search` doesn't know about those tracks. I could theoretically pass them here, but that would require some more work (not to run the logic more than once) and I want to get this merged to "fix" the issue.

ErikSchierboom (commented):
There is one open issue with this implementation: if a supermentor has no tracks that match the criteria (>= mentored solutions for the track), then it will show all representations. I'm still working on fixing this.

iHiD (approved):

PR #3192, opened 2022-11-16
Track-dashboard chevrons

dem4ron:
Had a hard time centering/working with `::after` content chevrons, so replaced them with icons. It wasn't _perfectly_ centered until now.

Screenshot 2022-11-16 at 10 41 35


iHiD:
Merged. Thanks.

Would be interested to understand what didn't work with `:after`.

PR #3212, opened 2022-11-18
Remove new-tab icon from Approach page

dem4ron:
This PR removes some dead code, and removes a misleading new tab icon, as the button opens the link in the same tab.


iHiD:
@dem4ron What's blocking this?

PR #3213, opened 2022-11-18
Add track to submission

ErikSchierboom:
- Add migration to add track_id to submissions table
- Add track reference to submission
- Add track to submissions table

### DIFF
--- a/db/migrate/20221118125948_add_track_to_submissions.rb
+++ b/db/migrate/20221118125948_add_track_to_submissions.rb
@@ -1,0 +1,16 @@
+class AddTrackToSubmissions < ActiveRecord::Migration[7.0]
+  def change
+    unless Rails.env.production?
+      add_column :submissions, :track_id, :smallint, null: true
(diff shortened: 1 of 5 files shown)


db/migrate/20221118125948_add_track_to_submissions.rb
```diff
@@ -0,0 +1,16 @@
+class AddTrackToSubmissions < ActiveRecord::Migration[7.0]
+  def change
+    unless Rails.env.production?
+      add_column :submissions, :track_id, :smallint, null: true
```
ErikSchierboom, line 4: No foreign key, right?

iHiD:
Did we run this manually?

db/migrate/20221118125948_add_track_to_submissions.rb
iHiD: Correct. We can't because we've used `smallint` here anyway

iHiD (approved):

PR #3224, opened 2022-11-24
Optimize build status queries

ErikSchierboom:
- Calculate num_submissions per day from metrics
- Optimize query for num_discussions
- Add index on [track_id, exercise_id] for mentor_requests
- Add index on [track_id, exercise_id] for mentor_requests
- Add index on [track_id, category, external_url] for user_reputation_tokens
- Add index on [track_id, num_comments] for submission_analyses
- Add index on [track_id, tests_status] for submissions
- Include profile in author/contributor combining
- Short-circuit combining authors and contributors

### DIFF
--- a/app/assemblers/combine_authors_and_contributors.rb
+++ b/app/assemblers/combine_authors_and_contributors.rb
@@ -10,12 +10,12 @@ class CombineAuthorsAndContributors
   end
 
   private
-  def randomize(users, limit)
+  def randomize(users, num_users)
     if users.is_a?(ActiveRecord::Relation)
-      ids = users.pluck(:id).shuffle.take(limit)
-      users.where(id: ids).to_a
+      ids = users.pluck(:id).shuffle.take(num_users)
+      users.includes(:profile).where(id: ids).to_a
     else
-      users.shuffle.take(limit)
+      users.shuffle.take(num_users)
     end
   end
 end
--- a/app/assemblers/combine_authors_and_contributors.rb
+++ b/app/assemblers/combine_authors_and_contributors.rb
@@ -15,7 +15,7 @@ class CombineAuthorsAndContributors
   private
   def randomize(users, num_users)
     if users.is_a?(ActiveRecord::Relation)
-      ids = users.pluck(user_id_column).shuffle.take(num_users)
+      ids = users.distinct.pluck(user_id_column).shuffle.take(num_users)
       User.with_attached_avatar.includes(:profile).where(id: ids).to_a
     else
       users.shuffle.take(num_users)
--- a/app/assemblers/combine_authors_and_contributors.rb
+++ b/app/assemblers/combine_authors_and_contributors.rb
@@ -1,21 +1,22 @@
 class CombineAuthorsAndContributors
   include Mandate
 
-  initialize_with :authors, :contributors, limit: 3
+  initialize_with :authors, :contributors, limit: 3, user_id_column: :id
 
   def call
     users = randomize(authors, limit)
-    users += randomize(contributors, limit - users.count)
+
+    updated_limit = limit - users.count
+    users += randomize(contributors, updated_limit) if updated_limit.positive?
+
     users
   end
 
   private
   def randomize(users, num_users)
-    return [] unless num_users.positive?
-
     if users.is_a?(ActiveRecord::Relation)
-      ids = users.pluck(:id).shuffle.take(num_users)
-      User.includes(:profile).where(id: ids).to_a
+      ids = users.pluck(user_id_column).shuffle.take(num_users)
+      User.with_attached_avatar.includes(:profile).where(id: ids).to_a
     else
       users.shuffle.take(num_users)
     end
--- a/app/assemblers/combine_authors_and_contributors.rb
+++ b/app/assemblers/combine_authors_and_contributors.rb
@@ -11,9 +11,11 @@ class CombineAuthorsAndContributors
 
   private
   def randomize(users, num_users)
+    return [] unless num_users.positive?
+
     if users.is_a?(ActiveRecord::Relation)
       ids = users.pluck(:id).shuffle.take(num_users)
-      users.includes(:profile).where(id: ids).to_a
+      User.includes(:profile).where(id: ids).to_a
     else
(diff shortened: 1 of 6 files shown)


app/assemblers/combine_authors_and_contributors.rb
```diff
@@ -10,12 +10,14 @@ def call
...
 
   private
-  def randomize(users, limit)
+  def randomize(users, num_users)
+    return [] unless num_users.positive?
+
     if users.is_a?(ActiveRecord::Relation)
-      ids = users.pluck(:id).shuffle.take(limit)
-      users.where(id: ids).to_a
+      ids = users.pluck(:id).shuffle.take(num_users)
```
iHiD: Is there ever a risk with this code that we're passing in like 10k rows that we end up materialising, which takes forever?

app/assemblers/combine_authors_and_contributors.rb
```diff
@@ -10,12 +10,14 @@ def call
   end
 
   private
-  def randomize(users, limit)
+  def randomize(users, num_users)
+    return [] unless num_users.positive?
```
iHiD: I like the variable rename.

db/migrate/20221124094828_add_build_status_indexes.rb
```diff
@@ -0,0 +1,10 @@
+class AddBuildStatusIndexes < ActiveRecord::Migration[7.0]
+  def change
+    unless Rails.env.production?
+      add_index :mentor_requests, %i[track_id exercise_id], if_not_exists: true
+      add_index :user_reputation_tokens, %i[track_id category external_url], name: 'index_user_reputation_tokens_on_track_id_category_external_url', if_not_exists: true
+      add_index :submission_analyses, %i[track_id num_comments], if_not_exists: true
+      add_index :submissions, %i[track_id tests_status], if_not_exists: true
```
iHiD, line 7: You checked the order on these?

app/assemblers/combine_authors_and_contributors.rb
```diff
@@ -1,21 +1,24 @@
...
 
-  initialize_with :authors, :contributors, limit: 3
+  initialize_with :authors, :contributors, limit: 3, user_id_column: :id
 
   def call
     users = randomize(authors, limit)
-    users += randomize(contributors, limit - users.count)
+
+    updated_limit = limit - users.count
+    users += randomize(contributors, updated_limit) if updated_limit.positive?
```
ErikSchierboom, line 10: This is an especially useful micro optimization for the build pages, as most categories will have at least 3 authors so in all those cases the additional query won't be executed.

app/commands/track/update_build_status.rb
```diff
@@ -213,11 +218,11 @@ def syllabus_health
   end
 
   def syllabus_volunteers
-    authors = User.with_attached_avatar.where(id: Concept::Authorship.where(concept: taught_concepts).select(:user_id)).
-      or(User.with_attached_avatar.where(id: Exercise::Authorship.where(exercise: active_concept_exercises).select(:user_id)))
+    authors = User.where(id: Concept::Authorship.where(concept: taught_concepts).select(:user_id)).
```
ErikSchierboom, line 221: I haven't yet split this up into two queries, as that has proven slightly tricky as the `serialize_volunteers` method is designed to work with two queries. I could duplicate some code, but it might be that with the change in the `CombineAuthorsAndContributors` class, the perf is better.

app/commands/track/update_build_status.rb
```diff
@@ -287,12 +292,10 @@ def num_unimplemented_practice_exercises
...
   def practice_exercises_volunteers
-    authors = User.with_attached_avatar.where(id:
-      Exercise::Authorship.where(exercise: active_practice_exercises).select(:user_id))
-    contributors = User.with_attached_avatar.where(id:
-      Exercise::Contributorship.where(exercise: active_practice_exercises).select(:user_id))
+    authors = Exercise::Authorship.where(exercise: active_practice_exercises)
+    contributors = Exercise::Contributorship.where(exercise: active_practice_exercises)
 
-    serialize_volunteers(authors, contributors)
+    serialize_volunteers(authors, contributors, user_id_column: :user_id)
```
ErikSchierboom, line 298: We no longer need the additional join on users

app/commands/track/update_build_status.rb
```diff
@@ -360,17 +363,15 @@ def serialize_exercise(exercise)
...
 
   def serialize_tooling_volunteers(repo_url)
-    authors = User.with_attached_avatar.where(id: track.reputation_tokens.where(category: :building).where('external_url LIKE ?',
-      "#{repo_url}/%").select(:user_id))
-    volunteers = User.with_attached_avatar.where(id: track.reputation_tokens.where(category: :maintaining).where('external_url LIKE ?',
-      "#{repo_url}/%").select(:user_id))
-    serialize_volunteers(authors, volunteers)
+    authors = track.reputation_tokens.where(category: :building).where('external_url LIKE ?', "#{repo_url}/%")
+    contributors = track.reputation_tokens.where(category: :maintaining).where('external_url LIKE ?', "#{repo_url}/%")
+    serialize_volunteers(authors, contributors, user_id_column: :user_id)
```
ErikSchierboom, line 368: Extra users join is no longer needed

db/migrate/20221124094828_add_build_status_indexes.rb
ErikSchierboom: Yes

app/commands/track/update_build_status.rb
iHiD: I suspect the issue is it can't use an index with the `OR`, so I think it'll still be slow.

ErikSchierboom: Yeah probably. I'll have a think

app/assemblers/combine_authors_and_contributors.rb
ErikSchierboom: Probably not that many, as they have to be people that have authored/contributed things to a track. But there could definitely be hundreds.

iHiD (approved):
Let the chaos begin!

PR #3226, opened 2022-11-24
Materialize exercise_id on submissions

ErikSchierboom:
- Add exercise_id column to submissions table
- Use submission's exercise in build status update


db/migrate/20221124152329_add_exercise_to_submissions.rb
```diff
@@ -0,0 +1,16 @@
+class AddExerciseToSubmissions < ActiveRecord::Migration[7.0]
+  def change
+    unless Rails.env.production?
+      add_column :submissions, :exercise_id, :mediumint, null: true
```
ErikSchierboom, line 4: We currently have 4000+ exercises. :smallint has room for 32_000. Do we want to use that or be safer with :mediumint?

iHiD: mediumint

iHiD (approved):

PR #3229, opened 2022-11-25
Add metric for joining track

ErikSchierboom:
We can use this metric to calculate the build status page's data and on the impact page.

This should back-port the existing data (1.2M records):

```ruby
UserTrack.includes(:track, :user).find_in_batches do |batch|
  batch.each do |user_track|
    Metric::Create.(:join_track, user_track.created_at, user_track:, track: user_track.track, user: user_track.user)
  end
end
```


iHiD (approved):
Looks good.

PR #3235, opened 2022-11-28
Track dashboard frontend patch

dem4ron:
#### - Apply white background only for the column names.

Before:
Screenshot 2022-11-28 at 11 22 17

After:
Screenshot 2022-11-28 at 11 17 56

---

#### - Add fix height-width to avatars on volunteers panel

Before:
Screenshot 2022-11-28 at 11 26 43

After:
Screenshot 2022-11-28 at 11 25 22

### DIFF
--- a/app/views/tracks/build/show.html.haml
+++ b/app/views/tracks/build/show.html.haml
@@ -4,7 +4,7 @@
   .lg-container.grid.grid-cols-3.gap-48.relative.mt-40
     .col-span-2
       %h1.text-h1.mb-12.text-48.font-bold.z-1
-        .highlight-text= @status.students.num_students
+        .highlight-text= number_with_delimiter(@status.students.num_students)
         students have developed their #{@track.title} skills.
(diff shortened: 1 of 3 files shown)


app/views/tracks/build/show.html.haml
```diff
@@ -295,16 +295,18 @@
         - @status.volunteers.users[0, 3].to_a.each do |volunteer|
           .contributor-credit-block
             .contributor
-              = avatar(volunteer)
+              = avatar(volunteer, css_class: "h-32 w-32")
```
iHiD, line 298:
```suggestion
              = avatar(volunteer, css_class: "h-[32px] w-[32px]")
```

Also below. Or edit the tailwind file to have these?

dem4ron: I added this to the Tailwind config.
I still had the Integral-site tailwind setup in my cell-memory where height wasn't overridden. 😄

iHiD (approved):

PR #3241, opened 2022-11-29
Allow blocking users

ErikSchierboom:
- Add flow to test confirming using token
- Add User::BlockDomain model
- Users with blocked email domain are not confirmed
- Allow disabling users

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -274,4 +274,6 @@ class User < ApplicationRecord
 
   def may_create_profile? = reputation >= User::Profile::MIN_REPUTATION
   def profile? = profile.present?
+
+  def confirmed? = super && !User::BlockDomain.blocked?(self)
 end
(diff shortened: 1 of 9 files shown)


app/models/user.rb
```diff
@@ -274,4 +274,8 @@ def send_devise_notification(notification, *args)
 
   def may_create_profile? = reputation >= User::Profile::MIN_REPUTATION
   def profile? = profile.present?
+
+  def confirmed? = super && !disabled? && !blocked?
```
ErikSchierboom, line 278: I've ordered the checks to have the non-expensive checks first

app/models/user/block_domain.rb
```diff
@@ -0,0 +1,6 @@
+class User::BlockDomain < ApplicationRecord
+  def self.blocked?(user)
+    domain = Mail::Address.new(user.email).domain
```
ErikSchierboom, line 3: Is it possible for a user to not have an email?

app/models/user.rb
iHiD: Nice.

app/models/user/block_domain.rb
iHiD: No.

PR #3247, opened 2022-11-30
Cleanup stripe failure handling code

ErikSchierboom:
- Add tests for invoice failure handling
- Use mandate constructor initialization block
- Reduce nesting

@iHiD This might be best reviewed via the commits: https://github.com/exercism/website/pull/3247/commits

### DIFF
--- a/app/commands/donations/payment_intent/handle_invoice_failure.rb
+++ b/app/commands/donations/payment_intent/handle_invoice_failure.rb
@@ -13,19 +13,24 @@ module Donations
       # We guard against spammers here
       def call
         return unless user
-        return unless user.stripe_customer_id.present?
         return if user.uid # Return if the user has auth'd via GitHub
 
-        has_three_failed_invoiced_today = Stripe::Charge.search(
-          query: %(customer:"#{user.stripe_customer_id}" AND status:"failed" AND created>#{(Time.current - 24.hours).to_i}),
-          limit: 3
-        ).count == 3
-
-        user.update!(disabled_at: Time.current) if has_three_failed_invoiced_today
+        user.update!(disabled_at: Time.current) if too_many_failed_invoices_in_last_24_hours?
       end
 
       private
       attr_reader :id, :invoice
+
+      def too_many_failed_invoices_in_last_24_hours?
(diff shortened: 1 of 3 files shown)


app/commands/donations/payment_intent/handle_invoice_failure.rb
```diff
@@ -13,20 +13,25 @@ def initialize(id: nil, invoice: nil)
...
       attr_reader :id, :invoice
 
+      def too_many_failed_invoices_in_last_24_hours?
+        number_of_failed_invoices_in_last_24_hours == 3
+      end
+
+      def number_of_failed_invoices_in_last_24_hours
+        Stripe::Charge.search(
+          query: %(customer:"#{user.stripe_customer_id}" AND status:"failed" AND created>#{(Time.current - 24.hours).to_i}),
+          limit: 3
```
iHiD: It's important that the `3` here is constant across both this and L25, so maybe put this into a constant?

app/commands/donations/payment_intent/handle_invoice_failure.rb
```diff
@@ -1,38 +1,39 @@
...
+  attr_reader :id, :invoice
+
+  def too_many_failed_invoices_in_last_24_hours?
+    number_of_failed_invoices_in_last_24_hours >= MINIMUM_FAILED_INVOICES_IN_LAST_24_HOURS
   end
+
+  def number_of_failed_invoices_in_last_24_hours
+    Stripe::Charge.search(
+      query: %(customer:"#{user.stripe_customer_id}" AND status:"failed" AND created>#{(Time.current - 24.hours).to_i}),
+      limit: 3
```
iHiD:
```suggestion
      limit: MINIMUM_FAILED_INVOICES_IN_LAST_24_HOURS
```

iHiD: If we increased `MINIMUM_FAILED_INVOICES_IN_LAST_24_HOURS` to 4, then the logic would break as this only retrieves 3

ErikSchierboom: Lol

PR #3250, opened 2022-11-30
Give users with 1k rep pm-enabled permission in the forum

iHiD:
I don't know how to test this in a less terrible way 🤷


app/commands/user/set_discourse_groups.rb
```diff
@@ -0,0 +1,39 @@
...
+    set_trust_level!
+    set_pm_enabled!
+  rescue DiscourseApi::NotFoundError
+    # If the external user can't be found, then the
+    # oauth didn't complete so there's nothing to do.
+  end
+
+  private
+  def set_trust_level!
+    return unless user.reputation >= MIN_REP_FOR_TRUST_LEVEL
```
ErikSchierboom: `if` is arguably slightly easier to read than `unless`:

```suggestion
    return if user.reputation < MIN_REP_FOR_TRUST_LEVEL
```

app/commands/user/set_discourse_groups.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    client.update_trust_level(discourse_user_id, level: 2)
+  end
+
+  def set_pm_enabled!
+    return unless user.reputation >= MIN_REP_FOR_PM_ENABLED
```
ErikSchierboom:
```suggestion
    return if user.reputation < MIN_REP_FOR_PM_ENABLED
```

ErikSchierboom (approved):

PR #3254, opened 2022-12-01
Gracefully handle adding user to discourse group when already a member

ErikSchierboom:


app/commands/user/set_discourse_groups.rb
```diff
@@ -23,6 +23,9 @@ def set_pm_enabled!
 
     group_id = client.group("pm-enabled").dig(*%w[group id])
     client.group_add(group_id, user_id: [discourse_user_id])
+  rescue DiscourseApi::UnprocessableEntity
+    # If the user was already a member of the group,
+    # ignore the error
```
iHiD: I suggest doing this around the `group_add` LOC only. If the group_id lookup fails, we probably want that to hit bugsnag.

ErikSchierboom: That seems unlikely in this case, but yeah sure!

PR #3258, opened 2022-12-01
Add foregone and unimplemented exercises to track build page

ErikSchierboom:
This will show up like:

image

---

#### Current state of UI:

Screenshot 2022-12-01 at 16 14 37

#### Removed `new-tab` icons from links that won't be opened in a new tab

Screenshot 2022-12-01 at 16 21 58


app/commands/track/retrieve_unimplemented_practice_exercises.rb
```diff
@@ -4,12 +4,19 @@ class Track::RetrieveUnimplementedPracticeExercises
...
   private
   memoize
-  def prob_specs_repo = Git::ProblemSpecifications.new
+  def prob_specs = ProblemSpecifications.new
+
+  memoize
+  def implemented_exercises = track.practice_exercises.pluck(:slug).to_set
+
+  memoize
+  def foregone_exercises = track.git.foregone_exercises.to_set
```
iHiD, line 21: This is fine if it's called once in a request but will be slow if we ever show a page with all 60 tracks iterating over this. Just to be aware of that.

app/commands/track/update_build_status.rb
```diff
@@ -274,21 +274,33 @@ def practice_exercises
       num_exercises: active_practice_exercises.size,
       num_exercises_target: practice_exercises_num_exercises_target,
       created: active_practice_exercises.map { |exercise| serialize_exercise(exercise) },
+      num_unimplemented: num_unimplemented_practice_exercises,
+      unimplemented: unimplemented_practice_exercises.map { |exercise| serialize_prob_specs_exercise(exercise) },
```
iHiD, line 278: I feel like `num_unimplemented` could just be calculated client-side using `unimplemented.count`.

ErikSchierboom: It could. It is a pattern that's being used in other places too. We can change it, but I'd like to change it elsewhere too then

app/commands/track/retrieve_unimplemented_practice_exercises.rb
ErikSchierboom: Yeah, true. We could always materialize when we want to

app/commands/track/update_build_status.rb
ErikSchierboom: Let me know if you'd prefer that

iHiD: I'd rather that, yes. But do it as a follow up?

iHiD (approved):

PR #3260, opened 2022-12-01
Allow maintainers to do WIP exercises in the editor

iHiD:
The issue was that for non-maintainers the exercises where never even getting to the user_track summary in the first place.

There's two commits here. The second improvement is probably more correct. There's a comment that needs addressing before we can merge, and I need to check if this breaks tests elsewhere.

### DIFF
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -122,7 +122,7 @@ class Track < ApplicationRecord
     git.foregone_exercises.map { |slug| ProblemSpecifications::Exercise.new(slug) }
   end
 
-  def team_name = slug
+  def github_team_name = slug
 
   CATGEORIES = {
     paradigm: "Paradigm",
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -122,6 +122,8 @@ class Track < ApplicationRecord
     git.foregone_exercises.map { |slug| ProblemSpecifications::Exercise.new(slug) }
   end
 
+  def team_name = slug
+
   CATGEORIES = {
     paradigm: "Paradigm",
     typing: "Typing",
--- a/app/models/user_track.rb
+++ b/app/models/user_track.rb
@@ -158,7 +158,7 @@ class UserTrack < ApplicationRecord
 
   memoize
   def maintainer?
-    user.maintainer? && user.github_team_memberships.where(team_name: track.team_name).exists?
+    user.maintainer? && user.github_team_memberships.where(team_name: track.github_team_name).exists?
   end
 
   private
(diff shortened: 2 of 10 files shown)


app/models/user_track.rb
```diff
@@ -78,6 +78,12 @@ def unlocked_exercises_for_exercise(exercise)
   end
 
   def enabled_exercises(exercises)
+    # This should really be "If the user a maintainer FOR THIS TRACK".
+    # I don't know how well we keep that up to date. If we do, then
+    # let's take that approach and refine it a bit. Otherwise let's
+    # consider if we want this brute-force approach for all maintainers.
+    return exercises if user.maintainer?
```
ErikSchierboom: There is some code to keep track of that, but we never got around to get it working.
It would not be hard to do so though, so maybe we do that as a follow-up?

ErikSchierboom (approved):

app/models/user_track.rb
iHiD: Yep.

app/models/user_track.rb
```diff
@@ -78,7 +78,15 @@ def unlocked_exercises_for_exercise(exercise)
...
 
   def enabled_exercises(exercises)
-    exercises.where(status: %i[active beta]).or(exercises.where(id: solutions.select(:exercise_id)))
+    status = %i[active beta]
+
+    # This should really be "If the user a maintainer FOR THIS TRACK".
+    # I don't know how well we keep that up to date. If we do, then
+    # let's take that approach and refine it a bit. Otherwise let's
+    # consider if we want this brute-force approach for all maintainers.
+    status << :wip if user.maintainer?
```
ErikSchierboom: Do check with team member

app/models/track.rb
```diff
@@ -122,6 +122,8 @@ def foregone_exercises
     git.foregone_exercises.map { |slug| ProblemSpecifications::Exercise.new(slug) }
   end
 
+  def team_name = slug
```
ErikSchierboom: Even though it is just deferring to the slug, I still feel like it was worth introducing a separate method for readability purposes

app/models/user_track.rb
```diff
@@ -78,7 +78,10 @@ def unlocked_exercises_for_exercise(exercise)
   end
 
   def enabled_exercises(exercises)
-    exercises.where(status: %i[active beta]).or(exercises.where(id: solutions.select(:exercise_id)))
+    status = %i[active beta]
+    status << :wip if maintainer?
```
ErikSchierboom, line 82: I've made `maintainer?` a method on `UserTrack` as we could potentially use it elsewhere too

app/models/track.rb
iHiD: I don't love this name. "team" is too inspecific for my liking. Maybe `github_team_name`?

PR #3263, opened 2022-12-02
Use pre-calculated supermentor requirements

ErikSchierboom:
- Add tests for eligible
- Use num_finished_discussions to determine supermentorship


iHiD (approved):
Nice. This is much neater.

PR #3264, opened 2022-12-02
Award supermentor badge

ErikSchierboom:
- Add supermentor badge model
- Award supermentor badge when role is added


app/models/badges/supermentor_badge.rb
```diff
@@ -0,0 +1,9 @@
+class Badges::SupermentorBadge < Badge
+  seed 'Supermentor',
+    :legendary,
+    'supermentor',
+    'Mentored 100 students on a track'
```
ErikSchierboom, line 5: Do you want this to maybe also say something about the satisfaction rating?

Something like:

```suggestion
    'Satisfactorily mentored 100 students on a track'
```

ErikSchierboom (commented):
This will also require a badge to be designed :)

app/models/badges/supermentor_badge.rb
iHiD: I think I'm ok with just "Mentored"

iHiD (approved):
Nice :)

PR #3271, opened 2022-12-05
Store github team members

ErikSchierboom:
This PR adds support for adding Github team membership to our database. 
This allows us to (as an example) only show wip exercises for track's of which someone is a maintainer (i.e. a member of the correct team).
In the process, I've removed any references to the old "contributor team" code because:

a) that overlaps with what we're doing here
b) it wasn't used
c) it likely needs to change anyway when we look at our volunteering setup with fresh eyes


app/commands/github/team_member/destroy.rb
```diff
@@ -0,0 +1,7 @@
+class Github::TeamMember::Destroy
+  include Mandate
+
+  initialize_with :username, :team
+
+  def call = ::Github::TeamMember.where(username:, team:).delete_all
```
iHiD:
```suggestion
  def call = ::Github::TeamMember.where(username:, team:).destroy_all
```

(Safer to use `destroy` for callbacks. But it should definitely match whatever this Command is called)

db/migrate/20221205094552_create_github_team_members.rb
```diff
@@ -0,0 +1,12 @@
+class CreateGithubTeamMembers < ActiveRecord::Migration[7.0]
+  def change
+    create_table :github_team_members do |t|
+      t.string :username, null: false
+      t.string :team, null: false
```
ErikSchierboom: I opted to just store the team name in this table and not create another `Github::Team` table

app/commands/github/team_member/sync_members.rb
```diff
@@ -0,0 +1,22 @@
...
+  private
+  def add_team_members!
+    org_team_members.flat_map do |team, usernames|
+      usernames.map { |username| ::Github::TeamMember::Create.(username, team) }
+    end
+  end
+
+  def delete_team_members!(current_team_members)
+    Github::TeamMember.where.not(id: current_team_members.map(&:id)).delete_all
+  end
```
iHiD, line 18: Again, I'd probably use destroy here, in case we have callbacks.

app/models/github/organization.rb
```diff
@@ -89,12 +89,31 @@ def team_member_usernames
     end
   end
 
-  def fetch_team_member_usernames(cursor)
+  def team_members
+    cursor = nil
+    members = {}
+
+    loop do
```
iHiD, line 96: Any risk this becomes infinite?

db/migrate/20221205094552_create_github_team_members.rb
```diff
@@ -0,0 +1,12 @@
...
+    create_table :github_team_members do |t|
+      t.string :username, null: false
+      t.string :team, null: false      
+
+      t.timestamps
+
+      t.index %i[username team], unique: true
+    end
+  end
+end
```
iHiD, line 12: General question - should we be using username or an id? Presuming usernames are mutable and could cause us chaos.

iHiD (approved):
Read the code, but not the tests. Approving pending the comments about destroy. Nice work :)

app/models/github/organization.rb
ErikSchierboom: Well, we've used this pattern in several other places and things have worked well there, so 🤷

db/migrate/20221205094552_create_github_team_members.rb
ErikSchierboom: Good question. I thought usernames are immutable on github, but it looks like they're not. I'll update to github user ID.

PR #3282, opened 2022-12-06
Don't create Stripe payments when email is on block domain list

ErikSchierboom:
- Allow testing blocked domain with email string
- Don't create Stripe payment intents when email is on blocked domain

### DIFF
--- a/app/commands/donations/payment_intent/create.rb
+++ b/app/commands/donations/payment_intent/create.rb
@@ -31,7 +31,11 @@ module Donations
       end
 
       private
-      def invalid_user_or_email? = User::BlockDomain.blocked?(user_or_email)
+      def invalid_user_or_email?
+        user_or_email.is_a?(User) ?
+          User::BlockDomain.blocked?(user:) :
+          User::BlockDomain.blocked?(email: user_or_email)
+      end
 
       memoize
       def user
--- a/app/commands/donations/payment_intent/create.rb
+++ b/app/commands/donations/payment_intent/create.rb
@@ -9,7 +9,14 @@ module Donations
       initialize_with :user_or_email, :type, :amount_in_cents
 
       def call
-        return if invalid_user_or_email?
+        if invalid_user_or_email?
+          begin
+            raise "Invalid user or email trying to make donation: #{user_or_email}"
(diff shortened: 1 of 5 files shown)


app/models/user/block_domain.rb
```diff
@@ -1,6 +1,7 @@
 class User::BlockDomain < ApplicationRecord
-  def self.blocked?(user)
-    domain = Mail::Address.new(user.email).domain
+  def self.blocked?(user_or_email)
+    email = user_or_email.is_a?(User) ? user_or_email.email : user_or_email
+    domain = Mail::Address.new(email).domain
```
iHiD: Maybe this method signature could be `user: nil, email: nil`, to make the pass in more explicit?

See `app/commands/donations/payment_intent/handle_invoice_failure.rb` for an example of this pattern.

app/commands/donations/payment_intent/create.rb
```diff
@@ -9,6 +9,8 @@ class Create
       initialize_with :user_or_email, :type, :amount_in_cents
 
       def call
+        return if invalid_user_or_email?
```
iHiD: Let's log a bugsnag here instead so we can see it happening?

```ruby
if invalid_user_or_email?
  begin
    raise "Invalid user or email trying to make donation: #{user_or_email}"
  rescue => e 
    Bugsnag.notify(e)
    return
  end
end
```

app/commands/donations/payment_intent/create.rb
```diff
@@ -9,6 +9,11 @@ class Create
       initialize_with :user_or_email, :type, :amount_in_cents
 
       def call
+        if invalid_user_or_email?
+          Bugsnag.notify(PaymentIntentError.new("Invalid user or email trying to make donation: #{user_or_email}"))
```
ErikSchierboom, line 13: I noted that this class defined a custom `PaymentIntentError` class, so I thought: why not use that? (we also don't need the `begin/rescue` bit)

iHiD: Nice.

iHiD (approved):

PR #3283, opened 2022-12-06
Further optimize build status generation

ErikSchierboom:
- Optimize counting of mentor discussions
- Remove duplicate query from build status updating
- Greatly optimize user queries

### DIFF
--- a/app/commands/track/update_build_status.rb
+++ b/app/commands/track/update_build_status.rb
@@ -141,17 +141,12 @@ class Track::UpdateBuildStatus
 
   def representer
     {
-      num_runs: representer_num_runs,
+      num_runs: representer_num_submissions,
       num_comments: representer_num_submissions_with_feedback,
       display_rate_percentage: representer_display_rate_percentage,
       volunteers: serialize_tooling_volunteers(track.representer_repo_url),
       health: representer_health
     }
-  end
-
-  memoize
-  def representer_num_runs
-    Submission::Representation.where(track:).count
   end
 
   memoize
--- a/app/commands/track/update_build_status.rb
+++ b/app/commands/track/update_build_status.rb
@@ -45,7 +45,7 @@ class Track::UpdateBuildStatus
   def volunteers
     track_volunteers = User::ReputationPeriod::Search.(track_id: track.id)
     num_volunteers = track_volunteers.total_count
-    top_track_volunteers = track_volunteers.take(NUM_VOLUNTEERS)
+    top_track_volunteers = track_volunteers.take(NUM_TRACK_VOLUNTEERS)
 
     contextual_data = User::ReputationPeriod.
       where(category: :any).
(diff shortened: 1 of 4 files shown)


app/commands/track/update_build_status.rb
```diff
@@ -91,7 +91,7 @@ def num_submissions = track.submissions.count
 
   def mentor_discussions
     {
-      num_discussions: Mentor::Discussion.joins(:request).where(request: { track: }).count
+      num_discussions: track.mentor_requests.fulfilled.count
```
ErikSchierboom, line 94: Looking at the code, `fulfilled` seems to imply that a discussion was created for the request, If true, this query does not need a join and is quicker.

app/commands/track/update_build_status.rb
```diff
@@ -141,19 +141,14 @@ def test_runner_version_target
 
   def representer
     {
-      num_runs: representer_num_runs,
+      num_runs: representer_num_submissions,
```
ErikSchierboom, line 144: The submissions representations were counted twice. This fixes that

app/commands/track/update_build_status.rb
```diff
@@ -223,13 +218,15 @@ def syllabus_health
   end
 
   def syllabus_volunteers
-    authors = User.where(id: Concept::Authorship.where(concept: taught_concepts).select(:user_id)).
-      or(User.where(id: Exercise::Authorship.where(exercise: active_concept_exercises).select(:user_id)))
+    concept_author_ids = Concept::Authorship.where(concept: taught_concepts).distinct.pluck(:user_id)
+    exercise_author_ids = Exercise::Authorship.where(exercise: active_concept_exercises).distinct.pluck(:user_id)
+    author_ids = (concept_author_ids | exercise_author_ids).uniq
```
ErikSchierboom, line 223: The new pattern to serialize volunteers is to fetch the unique ids first via individual queries, and then to randomize and fetch the actual users using those ids.

app/commands/track/update_build_status.rb
```diff
@@ -391,15 +388,33 @@ def serialize_prob_specs_exercise(exercise)
...
+    serialize_volunteers(author_ids, contributor_ids)
   end
 
-  def serialize_volunteers(authors, contributors, user_id_column: :id)
+  def serialize_volunteers(author_ids, contributor_ids)
+    random_author_ids = author_ids.shuffle.take(NUM_VOLUNTEERS)
+    random_authors = User.where(id: random_author_ids).to_a
+
+    random_contributor_ids = (contributor_ids - author_ids).shuffle.take(NUM_VOLUNTEERS - author_ids.size)
+    random_contributors = User.where(id: random_contributor_ids).to_a
```
ErikSchierboom, line 411: If there are three authors, `random_contributor_ids` will be an empty array causing `User.where(id: random_contributor_ids)` to be a noop.

app/commands/track/update_build_status.rb
```diff
@@ -391,15 +388,33 @@ def serialize_prob_specs_exercise(exercise)
...
+
+    num_users = (author_ids | contributor_ids).uniq.size
```
ErikSchierboom, line 413: No additional count query needed

app/commands/track/update_build_status.rb
iHiD: Nice.

app/commands/track/update_build_status.rb
iHiD: Good.

app/commands/track/update_build_status.rb
iHiD: Love a bit of `|`

iHiD: Another way of doing this that I do elsewhere is to create a Set

iHiD (approved):
Scanned it. Looks solid. LGTM!

app/commands/track/update_build_status.rb
ErikSchierboom: Yes that would be an option too. I didn't know what was idiomatic 🤷

PR #3311, opened 2022-12-13
Tweak build status page

ErikSchierboom:
- Sort foregone/unimplemented exercises by title in build status
- Inline num_foregone
- Inline num_unimplemented
- Use better name for active entities
- Inline active entities size
- Normalize naming
- Show deprecated exercises
- Add tests for num active target
- Don't show syllabus info on build page if track does not have course


iHiD (approved):
Approving blind. Merge if/when you're happy.

PR #3313, opened 2022-12-14
Listen for workflow_run webhook events

ErikSchierboom:
This PR allows us to listen to `workflow_run` events, which in turn lets us do things when e.g. tooling has deployed successfully.

### DIFF
--- a/app/commands/webhooks/process_workflow_run_update.rb
+++ b/app/commands/webhooks/process_workflow_run_update.rb
@@ -1,0 +1,10 @@
+class Webhooks::ProcessWorkflowRunUpdate
+  include Mandate
+
+  initialize_with params: Mandate::KWARGS
+
+  def call
+    return unless params[:action] == 'completed'
+    return unless params[:conclusion] == 'success'
+  end
+end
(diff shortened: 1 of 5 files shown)


app/commands/webhooks/process_workflow_run_update.rb
```diff
@@ -0,0 +1,10 @@
+class Webhooks::ProcessWorkflowRunUpdate
+  include Mandate
+
+  initialize_with params: Mandate::KWARGS
+
+  def call
+    return unless params[:action] == 'completed'
+    return unless params[:conclusion] == 'success'
```
ErikSchierboom, line 8: This is in preparation for the representers re-running logic

test/commands/webhooks/process_workflow_run_update_test.rb
```diff
@@ -0,0 +1,4 @@
+require "test_helper"
+
+class Webhooks::ProcessWorkflowRunUpdateTest < ActiveSupport::TestCase
```
ErikSchierboom, line 3: I'll add tests once the class actually does something :)

iHiD (approved):
Looks good.

PR #3314, opened 2022-12-14
Only consider successful test runs for determining of test runner version

ErikSchierboom:
CC @bethanyg

### DIFF
--- a/app/commands/track/update_build_status.rb
+++ b/app/commands/track/update_build_status.rb
@@ -130,7 +130,7 @@ class Track::UpdateBuildStatus
   memoize
   def test_runner_version = [
     1,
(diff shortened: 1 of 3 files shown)


app/commands/track/update_build_status.rb
```diff
@@ -130,7 +130,7 @@ def test_runner_health
   memoize
   def test_runner_version = [
     1,
-    Submission::TestRun.where(track:).order(id: :desc).pick(:version).to_i
+    Submission::TestRun.ops_successful.where(track:).order(id: :desc).pick(:version).to_i
```
ErikSchierboom, line 133: I've verified that this is still fast (20ms)

iHiD (approved):

PR #3041, opened 2022-10-12
Handle representers being updated

iHiD:
@kytrinyx @ErikSchierboom (and @dem4ron) I've written some code and am happy that this method is going to work.

This PR adds the basic functionality (which is very small changes - go us thinking this through 18months ago!)

I've added all the tests into one file so we can just sanity check this in one place. We can then explode them out into the relevant actual test files if we're happy that the key cases are covered.

@ErikSchierboom There are three TODOs for you (including the whole "Sync when representers sync" thing). I'll leave you to get this over the line now, but shout if you want to pair on any of it 🙂 
- [ ] Add version to representer output 
  - [x] Spec
  - [ ] Docs
  - [x] Website reading code
  - [x] Sample examplar (C# or F#?)
- [ ] Add version to exercise
  - [x] Spec
  - [ ] Docs
  - [x] Website reading code
  - [x] Sample examplar (C# or F#?)
- [ ] Trigger rerun of all existing representations with comments on representer push.

PR #3316, opened 2022-12-14
Re-run representer after deploy

ErikSchierboom:
- Add `representer_version` column to `tracks` table
- Initialize representer version to 1 for tracks with representer
- Queue job to handle representer deploy when successfully deployed representer


app/commands/webhooks/process_workflow_run_update.rb
```diff
@@ -4,7 +4,20 @@ class Webhooks::ProcessWorkflowRunUpdate
...
-    return unless params[:conclusion] == 'success'
+    Tooling::HandleRepresenterDeploy.defer(track) if handle_representer_deploy?
   end
+
+  private
+  def handle_representer_deploy?
+    params[:action] == 'completed' &&
+      params[:conclusion] == 'success' &&
+      params[:head_branch] == 'main' &&
+      params[:path] == '.github/workflows/deploy.yml' &&
```
ErikSchierboom, line 15: This is safe to do as the `deploy.yml` file is automatically synced to all repos via the `org-wide-files` repo and is thus guaranteed to have the same name across repos

app/commands/tooling/handle_representer_deploy.rb
```diff
@@ -0,0 +1,12 @@
+class Tooling::HandleRepresenterDeploy
+  include Mandate
+
+  queue_as :default
+
+  initialize_with :track
+
+  def call
+    # TODO: update track's representer version
+    # TODO: trigger re-running of representer if representer version changes
```
iHiD: OK, I think you want this:

```ruby
Exercise::Representation.where(track:).with_feedback.includes(source_submission: :exercise).find_each do |representation|
  submission = representation.source_submission
  Submission::Representation::Init.(submission, type: :exercise, git_sha: submission.exercise.git_sha, run_in_background: true)
end
```

I suggest rebasing this on top of #3041 and then that **should** work. We're going to need to pair tomorrow to test all this though - it's really hard to work out whether anything is working as expected as there's so many bits that need setting (e.g. versions!)

app/commands/tooling/handle_representer_deploy.rb
```diff
@@ -0,0 +1,12 @@
+class Tooling::HandleRepresenterDeploy
+  include Mandate
+
+  queue_as :default
```
iHiD: I'd say we definitely want this on one of the slower background queues.

PR #3357, opened 2022-12-23
Add system flow tests for automation pages

ErikSchierboom:
- Add system test for need feedback tracks
- Add system test for switching tracks on automation page
- Add system test for with feedback tracks


iHiD (approved):
Brilliant.

PR #3364, opened 2023-01-03
Award new years resolution badge when iteration is created, not solution

ErikSchierboom:


app/commands/iteration/create.rb
```diff
@@ -25,6 +25,7 @@ def call
       AwardBadgeJob.perform_later(user, :die_unendliche_geschichte, context: iteration)
       AwardBadgeJob.perform_later(user, :growth_mindset)
       record_activity!(iteration)
+      award_badge!(iteration)
```
iHiD: Look two lines above - move those into here and pluralise the method name?

app/models/badges/new_years_resolution_badge.rb
```diff
@@ -5,13 +5,11 @@ class NewYearsResolutionBadge < Badge
       'new-years-resolution',
       'Submitted a solution on January 1st'
```
iHiD: `Submitted an iteration on January 1st`

iHiD (approved):
LGTM

app/models/badges/new_years_resolution_badge.rb
```diff
@@ -3,15 +3,15 @@ class NewYearsResolutionBadge < Badge
     seed "New Year's resolution",
       :rare,
       'new-years-resolution',
-      'Submitted a solution on January 1st'
+      'Submitted an iteration on January 1st'
 
-    def self.worth_queuing?(solution:)
-      solution.created_at.yday == 1
+    def self.worth_queuing?(iteration:)
+      [1, 2, 365, 366].include?(iteration.created_at.yday)
```
ErikSchierboom, line 9: 366 is for leap years

iHiD (approved):

PR #3366, opened 2023-01-03
12in23

iHiD:
Firstly, I accidentally pushed this to `main`... I don't want to undo that now as I don't want to risk breaking stuff, so I've created an `old-main` branch that this is based against.

This adds the #12in23 challenge (I've renamed it everywhere to that now). It has an external page, an internal page, changes to the dashboard and also an introducer. I'll launch this and then follow up with an email to everyone.

### External page

Screenshot 2023-01-03 at 19 49 00
Screenshot 2023-01-03 at 19 49 05

### Dashboard

Screenshot 2023-01-03 at 19 48 48

### Actual Challenge page

Screenshot 2023-01-03 at 19 47 23

## Introducer

Screenshot 2023-01-03 at 22 33 20

---

## TODO Before Deploy
- [x] Add thumbnail to youtube
- [x] Add end-screen to youtube
- [ ] Make the video public on youtube

## TODO After Deploy
- [ ] Post announcement on forum (using tag)
- [ ] Post announcement on Twitter (using hashtag)
- [ ] Post announcement on LinkedIn (using hashtag)
- [ ] Post announcement on Slack
- [ ] Add blog post

## TODO Eventually

We'll need a completion state, like adding a badge etc if you complete it. But I'm hoping no-one manages that in the next 24hours (probably a stupid assumption admitedly!)


app/controllers/challenges_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  def load_data_for_12in23 # rubocop:disable Naming/VariableNumber
+    @track_counts = Solution.joins(:exercise).
+      where(
+        id: current_user.iterations.where('iterations.created_at >= ?', Date.new(2022, 12, 31)).select(:solution_id)
+      ).
+      where.not('exercises.slug': "hello-world").
+      group(:track_id).count
+    @track_counts = @track_counts.sort_by(&:second).reverse.to_h
+    @tracks = Track.where(id: @track_counts.keys).index_by(&:id)
+  end
```
iHiD: @ErikSchierboom This is the only bit that's probably worth you giving much brain energy to :)

db/schema.rb
```diff
@@ -331,14 +331,10 @@
     t.integer "draft_feedback_type", limit: 1
     t.text "draft_feedback_markdown"
     t.index ["exercise_id", "ast_digest", "representer_version", "exercise_version"], name: "exercise_representations_guard", unique: true
-    t.index ["feedback_author_id", "exercise_id", "last_submitted_at"], name: "index_exercise_representation_author_exercise_last_submitted_at", order: { last_submitted_at: :desc }
-    t.index ["feedback_author_id", "exercise_id", "num_submissions"], name: "index_exercise_representation_author_exercise_num_submissions", order: { num_submissions: :desc }
```
ErikSchierboom, line 335: Is this deletion intentional? It might be worth checking production to see if there is a mismatch between your DB and what is in production.

db/schema.rb
```diff
@@ -331,14 +331,10 @@
...
     t.index ["feedback_author_id", "track_id", "last_submitted_at"], name: "index_exercise_representation_author_track_last_submitted_at", order: { last_submitted_at: :desc }
     t.index ["feedback_author_id", "track_id", "num_submissions"], name: "index_exercise_representation_author_track_num_submissions", order: { num_submissions: :desc }
     t.index ["feedback_author_id"], name: "index_exercise_representations_on_feedback_author_id"
     t.index ["feedback_editor_id"], name: "index_exercise_representations_on_feedback_editor_id"
-    t.index ["feedback_type", "exercise_id", "last_submitted_at"], name: "index_exercise_representation_type_exercise_last_submitted_at", order: { last_submitted_at: :desc }
-    t.index ["feedback_type", "exercise_id", "num_submissions"], name: "index_exercise_representation_type_exercise_num_submissions", order: { num_submissions: :desc }
```
ErikSchierboom, line 341: Same comment as above

db/schema.rb
```diff
@@ -1163,8 +1168,6 @@
   add_foreign_key "cohorts", "tracks"
   add_foreign_key "community_stories", "users", column: "interviewee_id"
   add_foreign_key "community_stories", "users", column: "interviewer_id"
-  add_foreign_key "community_videos", "exercises"
-  add_foreign_key "community_videos", "tracks"
```
ErikSchierboom, line 1167: Same comment as above

app/controllers/challenges_controller.rb
ErikSchierboom: It's quite a slow query in production at the moment. This is likely due to the `created_at >=` clause. I added a new key on the `created_at` and `solution_id` fields on the `iterations` table, and that greatly reduced the query cost locally. Might be worth trying that in production too?

Alternatively, the query could be rewritten to:

```ruby
current_user.iterations.
  joins(:exercise).
  where('iterations.created_at >= ?', Date.new(2021, 12, 31)).
  where.not('exercises.slug': "hello-world").
  group(:track_id).select(:solution_id).distinct.count
```

This loses the solution join and has the same result. It has a slightly lower query cost (around 15% lower).

app/views/challenges/12in23.html.haml
```diff
@@ -0,0 +1,35 @@
...
+        Below you'll see a list of all the languages you've used during 2023.
+        Aim to complete 12 languages by the end of the year.
+
+        Join in the conversation by using the
+        = link_to "#12in23 tag", "https://forum.exercism.org/tag/12in23"
+        on our forums or the
+        %strong.font-semibold #12in23
+        hashtag around social media!
+      %p.cta
+        Complete 5 exercises (not including "Hello World") in a track to tick off that language!
```
ErikSchierboom: The track counts query does not check for completed exercises, but only for solutions with a submitted iteration. Is the query wrong?

app/views/challenges/external.html.haml
```diff
@@ -0,0 +1,56 @@
...
+        - Track.active.map(&:icon_url).shuffle.each do |icon_url|
+          = image_tag icon_url, alt: '', role: :presentation, class: "c-icon c-track-icon h-[24px] w-[24px] lg:h-[48px] lg:w-[48px]"
+
+      %h1.text-h0.mb-16 The #12in23 Challenge
+      %p.mb-12
+        Expand your horizons by trying out
+        %strong.font-semibold 12 different programming languages
+        in 2023.
+      %p
+        Go old-school with COBOL, cutting edge with Unison and esoteric with Prolog. Explore low-level code with Assembly, expressions with a Lisp and functional with Haskell!
```
ErikSchierboom: Maybe change "and" to "or" to make it clear to students that these are choices, not requirements.

```suggestion
        Go old-school with COBOL, cutting edge with Unison or esoteric with Prolog. Explore low-level code with Assembly, expressions with a Lisp or functional with Haskell!
```

app/css/components/youtube-container.css
```diff
@@ -0,0 +1,16 @@
+.c-youtube-container {
+    @apply relative rounded-8 overflow-hidden;
+    padding-bottom: 56.25%;
+    padding-top: 30px;
+    height: 0;
```
dem4ron, line 5: This is nice, it is used in many places, so I'm glad this class will replace the handcrafted containers.

app/javascript/components/modals/Challenge12in23Modal.tsx
```diff
@@ -0,0 +1,119 @@
...
+        </FormButton>
+        <ErrorBoundary resetKeys={[status]}>
+          <ErrorMessage error={error} defaultError={DEFAULT_ERROR} />
+        </ErrorBoundary>
+      </div>
+      <div className="rhs">
+        <h2 className="text-h4 mb-12">Watch Jeremy's introduction...</h2>
+        <div
+          className="video relative rounded-8 overflow-hidden !mb-24"
+          style={{ padding: '56.25% 0 0 0', position: 'relative' }}
```
dem4ron, line 99: here this could be a `c-youtube-container`

app/javascript/components/modals/Challenge12in23Modal.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <h2>What do I have to do?</h2>
+        <p className="mb-12">
+          It's simple - complete 5 exercises (not including "Hello World!") in
+          12 languages throughout the year. We'll give you a nice progress page
+          that shows you how you're getting on, and if you complete the
+          challenge we'll give you access to exclusive badges and SWAG.
+        </p>
+
+        <p className="mb-12">
+          Each month we'll focus on a different paradign (e.g. functional, OOP,
```
dem4ron:
```suggestion
          Each month we'll focus on a different paradigm (e.g. functional, OOP,
```

app/views/blog_posts/show.html.haml
```diff
@@ -24,7 +24,7 @@
 
       .p-16.md:py-32.md:px-40.flex.flex-col
         - if @post.video?
-          .video-container
+          .c-youtube-container.mb-32
             %iframe{ width: "560", height: "315", src: "https://www.youtube-nocookie.com/embed/#{@post.youtube_id}", title: "YouTube video player", frameborder: "0", allow: "accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture", allowfullscreen: true }
```
dem4ron, line 28: I think here you don't need the width and height, because it is defined by the ratio and width of the container

app/controllers/challenges_controller.rb
iHiD: I've rewritten this to be done mainly in Ruby which is much faster. Basically any grouping seems to be pretty slow in SQL.

app/css/components/youtube-container.css
iHiD: Thanks!

app/views/challenges/12in23.html.haml
iHiD: Updated the text.

app/controllers/challenges_controller.rb
ErikSchierboom: > I've rewritten this to be done mainly in Ruby which is much faster.

Not pushed yet?

db/schema.rb
iHiD: Do you have an easy way of doing this? As I've pushed this to main accidently I'm not sure of the best route forward here.

ErikSchierboom (approved):

iHiD:
@dem4ron Could you follow up with another PR with the suggestions you've made pls? I don't have time to test them all properly today. Thanks! :)

PR #3394, opened 2023-01-20
Fix issue with exercise updates not triggering concept recount

iHiD:
Currently, changing an exercise from wip to beta/active doesn't recache the number of concepts on a track. This fixes that, and also moves things into their own commands.


ErikSchierboom (approved):
Nice!

PR #3412, opened 2023-01-31
Award 12 in 23 badge

ErikSchierboom:


app/models/badges/participant_in_12_in_23_badge.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def award_to?(user)
+      challenge = user.challenges.find_by(challenge_id: User::Challenge::CHALLENGE_12_IN_23)
+      return false if challenge.nil?
+
+      user.solutions.completed.joins(exercise: :track).
+        where('completed_at >= ?', challenge.created_at).
+        group(:track_id).
+        having("count(*) >= 5").
+        exists?
```
ErikSchierboom: This is quite slow (1s), due to there not being an appropriate index.

iHiD: Hmm, it's only 5ms when I check...

Screenshot 2023-01-31 at 22 55 23

ErikSchierboom: Probably depends on the number of solutions I guess

iHiD: I guess. That's your user above though, who has a lot!

ErikSchierboom: True. Ah well

iHiD (approved):

PR #3473, opened 2023-02-15
Configure CSS nesting correctly

dem4ron:


postcss.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   plugins: [
     require('postcss-import'),
-    require('tailwindcss/nesting'),
+    require('tailwindcss/nesting')(require('postcss-nesting')),
```
iHiD, line 4: What does this line do?

dem4ron: following the [docs](https://tailwindcss.com/docs/using-with-preprocessors#nesting) you'd need to write it in an object format like this: 
```javascript
// postcss.config.js
module.exports = {
  plugins: {
    'postcss-import': {},
    'tailwindcss/nesting': 'postcss-nesting',
    tailwindcss: {},
    autoprefixer: {},
  }
}
```

but as it was already configured in this commonjs format, this is just the analogue of the above. It invokes the required `tailwindcss/nesting` plugin with the required `postcss-nesting` plugin as an argument.

ErikSchierboom (approved):
Does this remove the warning in the logs?

dem4ron:
> Does this remove the warning in the logs?

yes

PR #3477, opened 2023-02-16
Add prefer-color-scheme default

iHiD:
This sort of works. It'll work properly if you can get the postcss-reuse thing working, but I failed and don't have time to dig further. But you'll work it out.

### DIFF
--- a/app/controllers/concerns/body_class_concern.rb
+++ b/app/controllers/concerns/body_class_concern.rb
@@ -9,9 +9,9 @@ module BodyClassConcern
     classes << "namespace-#{namespace_name}"
     classes << "controller-#{controller_name || 'none'}"
     classes << "action-#{action_name}"
-    classes << "theme-light"
+    classes << theme
     classes << (user_signed_in? ? "user-signed_in" : "user-signed_out")
-    classes.join(" ")
+    classes.compact.join(" ")
   end
 
   memoize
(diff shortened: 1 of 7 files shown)


app/controllers/concerns/body_class_concern.rb
```diff
@@ -24,6 +24,10 @@ def landing_page?
     namespace_name.nil? && controller_name == "pages" && action_name == "index"
   end
 
+  def theme
+    "theme-unset"
```
iHiD, line 28: So this can be `theme-light` or `theme-dark`. If it's `theme-unset` then we should go with the browser default.

For now you can just manually change the string here to experiment.

app/css/ui-kit/colors.css
```diff
@@ -338,3 +338,10 @@
 
     @apply text-textColor5 bg-backgroundColorB;
 }
+
+@media (prefers-color-scheme: dark) {
+    body.theme-unset {
+        background: red;
+        @reuse .theme-dark;
+    }
+}
```
iHiD, line 352: If, in the bit above this where you set `.theme-dark`, you could instead do:
```
.theme-dark, @media (prefers-color-scheme: dark) .theme-unset {
...
}
```

then that would be better. But I don't think you can use the `,` and media queries.

If `@reuse` works then this is a fine alternative though.

PR #3490, opened 2023-02-22
Always include selected testimonial on page

iHiD:
This is a spike of an idea for @dem4ron to test and see if it works (I've not run the code, let alone tested it in a browser). Then it needs tests adding and maybe the code refining (which @ErikSchierboom can do!)

@dem4ron Can you lead on getting this over the line pls.

### DIFF
--- a/app/assemblers/assemble_profile_testimonials_list.rb
+++ b/app/assemblers/assemble_profile_testimonials_list.rb
@@ -19,11 +19,11 @@ class AssembleProfileTestimonialsList
 
   memoize
   def paginated_testimonials
-    page = testimonials.order(id: :desc).page(params[:page]).per(64)
+    page = testimonials.order(id: :desc).page(params[:page]).per(1)
     return page unless params[:uuid]
-    return page if page.map(&:id).include?(params[:uuid])
+    return page if page.map(&:uuid).include?(params[:uuid])
 
-    selected = user.mentor_testimonials.published.find_by(id: params[:uuid])
+    selected = user.mentor_testimonials.published.find_by(uuid: params[:uuid])
     return page unless selected
 
     (page + [selected])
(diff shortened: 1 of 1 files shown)


app/assemblers/assemble_profile_testimonials_list.rb
```diff
@@ -19,11 +19,11 @@ def call
 
   memoize
   def paginated_testimonials
-    page = testimonials.page(params[:page]).per(64)
+    page = testimonials.page(params[:page]).per(1)
```
iHiD: You probably didn't mean to commit this.

ErikSchierboom:
The code doesn't work because the serializer calls `.includes` on what gets passed in. For the old code, that was an active record "thingy", but for the new code, it could also be an array, which doesn't support the `.includes` method. The serializer is used in a lot of places though, so I was unsure how to best fix this.

Side note: wouldn't it be weird to somehow have a different number of items on the page?

iHiD:
> The code doesn't work because the serializer calls .includes on what gets passed in. 

Ah! Makes sense. I'll run up a `rails s` and try it myself and see if I can fix it if I get chance over the weekend.

> wouldn't it be weird to somehow have a different number of items on the page?

Probably, but I'm not sure it would matter enough as a quick hotfix. If you'd like to both find a better way of doing it that can be done quickly, that'd be great. As this is a bug though, I'd like to get something shipped asap, even if we then refine it further later.

app/assemblers/assemble_profile_testimonials_list.rb
```diff
@@ -18,7 +20,22 @@ def call
...
+    return page if page_uuids.include?(params[:uuid])
+
+    selected = user.mentor_testimonials.published.find_by(uuid: params[:uuid])
+    return page unless selected
+
+    page_uuids.unshift(selected.uuid) # Show selected testimonial first
+
+    testimonials.where(uuid: page_uuids).
+      order(Arel.sql("FIND_IN_SET(uuid, '#{page_uuids.join(',')}')")).
+      page(page.current_page).per(page.size + 1)
```
ErikSchierboom, line 37: I've kept the same ordering to make things consistent-like

iHiD: Lovely.

PR #3492, opened 2023-02-23
Keep track of last visited on date

ErikSchierboom:
- Add last_visited_on column to users table
- Update last_visited_on when requesting HTML page

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,6 +11,7 @@ class ApplicationController < ActionController::Base
   before_action :set_request_context
   after_action :set_body_class_header
   after_action :set_csp_header
+  after_action :updated_last_visited_on!
 
   def process_action(*args)
     super
@@ -130,6 +131,14 @@ class ApplicationController < ActionController::Base
 
   def store_user_location!
     store_location_for(:user, request.fullpath)
+  end
+
+  def updated_last_visited_on!
+    return unless user_signed_in?
+    return unless request.format == 'html'
+    return if current_user.last_visited_on == Time.zone.today
+
+    current_user.update(last_visited_on: Time.zone.today)
   end
 
   def render_template_as_json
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -1164,6 +1164,7 @@
     t.datetime "updated_at", null: false
     t.boolean "show_on_supporters_page", default: true, null: false
     t.datetime "disabled_at"
+    t.date "last_visited_on"
```
ErikSchierboom, line 1167: Should this be indexed?

app/controllers/application_controller.rb
```diff
@@ -11,6 +11,7 @@ class ApplicationController < ActionController::Base
   before_action :set_request_context
   after_action :set_body_class_header
   after_action :set_csp_header
+  after_action :updated_last_visited_on!
```
ErikSchierboom, line 14: I don't know if you want this to be a `before_action` 🤷

app/controllers/application_controller.rb
```diff
@@ -132,6 +133,14 @@ def store_user_location!
     store_location_for(:user, request.fullpath)
   end
 
+  def updated_last_visited_on!
+    return unless user_signed_in?
+    return unless request.format == 'html'
```
ErikSchierboom: Not sure if this is the "proper" way to do it, but the aim was to not updated the visited on date when doing an API request

db/schema.rb
iHiD: Yeah, probably.

app/controllers/application_controller.rb
iHiD: I think request.format is a symbol, not a string (based on `config/initializers/devise.rb`)

app/controllers/application_controller.rb
iHiD: This is fine.

iHiD:
Needs rebasing and being green :)

iHiD (approved):

PR #3499, opened 2023-02-28
Add mechanical march badge

ErikSchierboom:


ee7:
I've checked that the badge doesn't have a typo :)

iHiD:
@ee7 I was quite honestly relying on you for this 😁

PR #3500, opened 2023-02-28
Fix apostrophe in meta title

dem4ron:
Screenshot 2023-02-28 at 14 34 51


app/views/profiles/show.html.haml
```diff
@@ -1,4 +1,4 @@
-- content_for :meta_title, "#{@user.handle}'s profile on Exercism"
+- content_for :meta_title, "#{@user.handle}'s profile on Exercism".html_safe
```
iHiD: I don't think we can do this quite like this, as someone's handle might contain code injection.

I think instead we need:
```
"#{sanitize(@user.handle)}'s profile on Exercism".html_safe
```

But I can't remember what the actual `sanitize` method is, so you'll need to check and get the right one :)

dem4ron: true, checking it!

iHiD (approved):
Nice. Thanks. Feel free to merge when green :)

PR #3511, opened 2023-03-01
Add arbitary site updates

iHiD:
We want to allow maintainers to create arbitrary site updates (not to do with an exercise or concept).

I've started the branch off with a bit of basic code, but need a few next steps implementing.

As a useful line of code, to create an arbitrary site update, you can do:

```ruby
SiteUpdates::ArbitaryUpdate.create!(track: Track.find('csharp'), title: "Some title", description: "Some description", author: User.first, published_at: Time.now)
```

### Rendering (@dem4ron):
-  [x] Adding this ^^ breaks the React rendering. Probably because some expected data (e.g. `makers` isn't part of these updates). I imagine you just need to make a few extra bits optional.
-  [x] This update should render as per this loom: https://www.loom.com/share/cbcaa81ac576456f8d7390689ae9d39a?from_recorder=1&focus_title=1

#### Updated UI:
Screenshot 2023-03-02 at 13 01 29

#### Notes:
- the "[handle] published a new update" comes from the backend (`update.text`) in an html form. 
 #### TODO:
- [x] Add markdown support - this can come from the backend in parsed form

### Markdown (@ErikSchierboom)
-  [x]  I want to change `description` to be `description_markdown` and `description_html` and the normal code (maybe a Concern?) to be attached to Site Update (This is for **all** site updates).

### Maintainers UI (@dem4ron and @ErikSchierboom - but Erik, teach Aron how to do this so he can do it next time).
- [x] Update the maintainers/site_updates UI to have the description as a markdown field not text (I think this is probably just changing the font - we could add a preview, but that's probably a step too far at this stage!)
- [x] Add a "new" page so that maintainers can create arbitary updates.
  - Add a `new` template
  - Extract `edit` to a shared form etc.
  - Reuse them.
  - Check admin/streaming_events for a recent version I added of this. (The PR for that is [here](https://github.com/exercism/website/pull/3457/files) as a reference)
- [x] Arbitrary updates should have `published_at` set to `Time.current` (not the future time as currently works)
- [x] Add pagination.

### Current state of `.c-admin-table`:
Tiny changes, maybe a bigger upgrade in another PR? if necessary.
Screenshot 2023-03-09 at 14 10 25


app/models/site_updates/new_exercise_update.rb
```diff
@@ -5,7 +5,7 @@ def i18n_params
     {
       exercise_title: exercise.title,
       exercise_url: Exercism::Routes.track_exercise_url(track, exercise),
-      maker_handles:
+      author_handle: author.handle
```
ErikSchierboom: Did you intend to make this change? The site update itself isn't updated to use this field, so this will actually break. With this change we only show the site update author right, not the authors and contributors of the exercise. Double checking this.

iHiD: No, I didn't intend to change anything to do with new exercises. Thanks for spotting.

ErikSchierboom: Great

app/controllers/maintaining/site_updates_controller.rb
```diff
@@ -26,4 +39,14 @@ def update
 
     redirect_to action: :index
   end
+
+  private
+  def maintained_tracks
+    Track.where(slug: Github::TeamMember.where(user_id: current_user.id).select(:team_name)).order(:title)
+  end
```
iHiD:
```suggestion
  def setup_new_form
    @tracks = Track.where(slug: Github::TeamMember.where(user_id: current_user.id).select(:team_name)).order(:title)
  end
```

app/controllers/maintaining/site_updates_controller.rb
```diff
@@ -7,6 +7,19 @@ def index
 
   def new
     @update = SiteUpdates::ArbitraryUpdate.new
+    @tracks = maintained_tracks
```
iHiD:
```suggestion
    setup_new_form
```

app/controllers/maintaining/site_updates_controller.rb
```diff
@@ -7,6 +7,19 @@ def index
...
+
+  def create
+    create_params = site_update_params.merge(author: current_user, published_at: Time.current)
+    @update = SiteUpdates::ArbitraryUpdate.new(create_params)
+
+    if @update.save
+      flash[:site_updates_notice] = "Site update was successfully created."
+      redirect_to %i[maintaining site_updates]
+    else
+      render :new, status: :unprocessable_entity
```
iHiD, line 22:
```suggestion
      setup_new_form
      render :new, status: :unprocessable_entity
```

PR #3526, opened 2023-03-03
Add feedback tab

dem4ron:
Screenshot 2023-03-08 at 13 42 12

#### if `iteration` === `null`:
Screenshot 2023-03-09 at 14 46 00

#### When requested mentoring
Screenshot 2023-03-13 at 16 11 23


app/helpers/react_components/editor.rb
```diff
@@ -32,9 +32,10 @@ def to_s
             }
           },
           iteration: {
-            analyzer_feedback: @iteration.analyzer_feedback,
-            representer_feedback: @iteration.representer_feedback
+            analyzer_feedback: iteration&.analyzer_feedback,
+            representer_feedback: iteration&.representer_feedback
           },
```
iHiD: Initial instinct is that the `iteration` block should be nil if there's not an iteration.

app/css/pages/editor.css
```diff
@@ -304,10 +304,11 @@
 
             section.feedback-pane {
                 --header-height: 32px;
+                @apply pb-16 px-24;
```
iHiD, line 306: I moved this from lower down. Always best to keep these declarations at the start of a section as it's hard to understand what they're nested in when they're amongst children.

app/css/pages/editor.css
```diff
@@ -304,10 +304,11 @@
...
 
                 & .timeline-entry {
                     @apply flex;
-                    @apply py-8 px-24;
+                    @apply py-8;
```
iHiD: Once we DRY this up, we might need this as an override.

iHiD: This and the 40->16 one below just move everything left a bit.

app/javascript/components/editor/FeedbackPanel.tsx
```diff
@@ -46,19 +46,29 @@ export const FeedbackPanel = ({
...
-                  />
-                )
-              })
+              <div>
+                <p className="text-p-base">
+                  This is your latest mentoring session for this exercise. To
+                  continue the discussion, switch to{' '}
+                  <a href="font-semibold text-blue">mentoring mode</a>.
+                </p>
+                <div>
```
iHiD: We need this extra div here so that first-child selectors work.

app/javascript/components/editor/FeedbackPanel.tsx
```diff
@@ -46,19 +46,29 @@ export const FeedbackPanel = ({
...
-                    prevIterationIdx={
-                      index === 0
-                        ? 0
-                        : data.items[index >= 1 ? index - 1 : 0].iterationIdx
-                    }
-                    post={post}
-                  />
-                )
-              })
+              <div>
```
iHiD: This div can probably actually get removed.

app/helpers/react_components/editor.rb
ErikSchierboom: Done. @dem4ron the `iteration` field can now be `null`

iHiD:
@dem4ron This is red now.

ErikSchierboom (approved):

PR #3540, opened 2023-03-07
Add db-driven mailshots

iHiD:
Doing PRs and deploying the website every time we want to send a mailshot is becoming ridiculous, so I'm moving it into the database. 

## TODO
- [x] Create models
- [x] Create admin section
- [x] Update `user_mailshot` sending code
- [x] Add `mailshot` sending code
- [x] Add magic image link support (via @ErikSchierboom's magic)


iHiD:
(cc @dem4ron too :))

app/commands/mailshot/send_to_audience_segment.rb
```diff
@@ -0,0 +1,42 @@
+class Mailshot
+  class SendToAudienceSegment
+    include Mandate
+
+    initialize_with :mailshot, :audience_type, :audience_slug, :limit, :offset
+
+    # TOOD: This is where we can add throttling to check how many
+    # mails have be sent today and reschedule ourselves for the next
```
ErikSchierboom:
```suggestion
    # mails have been sent today and reschedule ourselves for the next
```

app/commands/mailshot/send_to_audience_segment.rb
```diff
@@ -0,0 +1,42 @@
...
+      records.each do |record|
+        user = user_extractor.(record)
+        next unless user
+
+        User::Mailshot::Send.(user, mailshot)
+      end
+    end
+
+    def schedule_next_segment!
+      Mailshot::SendToAudienceSegment.defer(
```
ErikSchierboom, line 34: Do we want to explicitly specify the queue this command runs on?

app/models/mailshot.rb
```diff
@@ -4,4 +4,35 @@ class Mailshot < ApplicationRecord
...
+  def audience_for(type, slug)
+    send("audience_for_#{type}", slug)
+  end
+
+  # rubocop:disable Lint/NonLocalExitFromIterator
+
+  # This is pretty terribly slow and should only be used rarely.
+  def audience_for_admins(_)
+    [
+      User.where("JSON_CONTAINS(roles, ?, '$')", %("admin")),
```
ErikSchierboom, line 23: I spent some time trying to figure out a way to make this fast, but couldn't find it. I was slightly amazed by that

ErikSchierboom (approved):

app/commands/mailshot/send_to_audience_segment.rb
iHiD: Fuck yes. Good man!

app/models/mailshot.rb
iHiD: I didn't acutally check, but I expect it to be terrible. The newer version of MySQL is apparently better, but the one in production will be especially terrible.

iHiD: All the JSON stuff is bad in this version of MySQL basically.

PR #3541, opened 2023-03-08
Support non-Stripe donors

ErikSchierboom:
- Add donated column to users
- Set donate to true when creating payment
- Use correct name for Supporter badge
- Remove unused delegation from site footer
- Fix test class name for site footer
- Show support prompt based on donated field
- Award supporter badge based on donated? field
- Use donated? instead of donated
- Add guard for ensuring user is staff
- Add command to register as donor
- Update column and index
- Move registering of being donor to command


app/helpers/view_components/site_footer.rb
```diff
@@ -2,9 +2,6 @@ module ViewComponents
   class SiteFooter < ViewComponent
     extend Mandate::Memoize
 
-    delegate :namespace_name, :controller_name,
-      to: :view_context
-
```
ErikSchierboom, line 7: This wasn't used

app/views/about/individual_supporters.html.haml
```diff
@@ -43,8 +43,8 @@
           %h2.text-15.leading-140.text-textColor5.font-semibold.mb-4 Why am I not on this page?
           %p.text-14.leading-140.text-textColor5
             This page aims to show all individuals who have financially supported Exercism.
-            Technically, it lists everyone who have been awarded the Supporters Badge and who have enabled the "Appear on supporters page?" toggle in their #{link_to 'Donations Settings', donations_settings_path, class: 'text-lightBlue font-medium'}.
-            Donating through the Exercism website should automatically award you the Supporters badge.
+            Technically, it lists everyone who have been awarded the Supporter badge and who have enabled the "Appear on supporters page?" toggle in their #{link_to 'Donations Settings', donations_settings_path, class: 'text-lightBlue font-medium'}.
```
ErikSchierboom, line 45: These were using the wrong, pluralized version of the badge's name

db/migrate/20230308095911_add_donated_at_to_users.rb
```diff
@@ -0,0 +1,13 @@
+class AddDonatedAtToUsers < ActiveRecord::Migration[7.0]
+  def change
+    unless Rails.env.production?
+      add_column :users, :first_donated_at, :datetime, null: true      
+      remove_index :users, name: "users-supporters-page", if_not_exists: true
+      add_index :users, [:first_donated_at, :show_on_supporters_page], name: "users-supporters-page", if_not_exists: true
+
+      User::AcquiredBadge.joins(:user).where(badge: Badges::SupporterBadge.first).each do |badge|
+        badge.user.update(first_donated_at: badge.created_at)
+      end
```
ErikSchierboom, line 10: I checked in the database, and all users that have made a donation via Stripe have the badge

app/commands/donations/payment/create.rb
```diff
@@ -24,7 +24,7 @@ def call
       amount_in_cents: stripe_data.amount
     ).tap do |payment|
       user.update(total_donated_in_cents: user.donation_payments.sum(:amount_in_cents))
-      AwardBadgeJob.perform_later(user, :supporter)
+      Donations::RegisterUserAsDonor.(user, Time.current)
```
ErikSchierboom: I was unsure whether to put this command here, with the donations, or to put it with the user commands

iHiD: I'd probably go for `User::RegisterAsDonor` but could go either way.

app/controllers/application_controller.rb
```diff
@@ -34,6 +34,12 @@ def ensure_admin!
     redirect_to maintaining_root_path
   end
 
+  def ensure_staff!
+    return if current_user&.staff?
+
+    redirect_to maintaining_root_path
+  end
```
iHiD, line 41: This isn't currently used as far as I can see.

app/models/user.rb
```diff
@@ -120,6 +120,9 @@ class User < ApplicationRecord
 
   scope :random, -> { order('RAND()') }
 
+  scope :donor, -> { where.not(first_donated_at: nil) }
+  scope :supporter, -> { donor.where(show_on_supporters_page: true) }
```
iHiD: Maybe `public_supporter`?

iHiD: (It wasn't clear to me the difference between donor + supporter without reading the code)

iHiD (approved):

app/controllers/application_controller.rb
ErikSchierboom: It is: https://github.com/exercism/website/pull/3541/files#diff-d59db167813afb7fac0474ae3886da0f3d3f3775be9aafb4d13f3994d5200978R2

PR #3562, opened 2023-03-13
Add mentoring rep-gate

dem4ron:
Screenshot 2023-03-13 at 17 27 46


app/views/mentoring/external/show.html.haml
```diff
@@ -11,7 +12,10 @@
         Some of our best mentors worried too.
 
       - if user_signed_in?
-        = ReactComponents::Mentoring::TryMentoringButton.new.to_s
+        - if current_user.reputation >= 20
+          = ReactComponents::Mentoring::TryMentoringButton.new.to_s
```
iHiD:
```suggestion
          = ReactComponents::Mentoring::TryMentoringButton.new
```

You shouldn't need `to_s`. It should get automatically called (I know this was in the bit before - maybe you can check for other places to remove it to?)

dem4ron: ok!

app/views/mentoring/external/show.html.haml
```diff
@@ -11,7 +11,7 @@
         Some of our best mentors worried too.
 
       - if user_signed_in?
-        - if current_user.reputation >= 20
+        - if current_user.reputation >= 20 || current_user.mentor?
```
iHiD: If they current user is a mentor they should never see this page in the first place.

app/commands/user/become_mentor.rb
```diff
@@ -4,7 +4,7 @@ class User::BecomeMentor
   initialize_with :user, :track_slugs
 
   def call
-    return if user.mentor?
+    return if user.mentor? || user.reputation < 20
```
iHiD: Let's put this 20 into a constant (maybe in `User` called `MIN_REP_TO_MENTOR`) and then update everywhere that's 20 in this PR to `User::MIN_REP_TO_MENTOR`.

PR #3588, opened 2023-03-17
Add automation/admin tab

dem4ron:
This PR adds an Admin tab where all the representations where `feedback_type` is not `nil` can be seen.

[See this forum post for reference](https://forum.exercism.org/t/can-i-see-somewhere-all-the-representer-comments-created-by-other-mentors/4213)

Screenshot 2023-03-21 at 10 52 58


app/javascript/components/mentoring/automation/useAutomation.tsx
```diff
@@ -196,7 +210,7 @@ export function useAutomation(
     setChecked,
     order: request.query.order,
     setOrder,
-    page: request.query.page,
+    page: request.query.page || 0,
```
ErikSchierboom: Shouldn't page default to `1`?

app/commands/exercise/representation/search.rb
```diff
@@ -24,8 +24,10 @@ def call
 
   private
   def filter_with_feedback!
-    if with_feedback
+    if with_feedback && mentor
       @representations = @representations.with_feedback_by(mentor)
+    elsif with_feedback && !mentor
```
ErikSchierboom: You can omit the second part, as otherwise the previous clause would have matched

```suggestion
    elsif with_feedback
```

app/commands/exercise/representation/search.rb
```diff
@@ -24,8 +24,10 @@ def call
 
   private
   def filter_with_feedback!
-    if with_feedback
+    if with_feedback && mentor
```
ErikSchierboom: Maybe be explicit and not rely on truthiness:

```suggestion
    if with_feedback && mentor.present?
```

app/helpers/react_components/mentoring/representations/admin.rb
```diff
@@ -0,0 +1,91 @@
...
+              representations_with_feedback_count:,
+              is_introducer_hidden:
+            }
+          )
+        end
+
+        private
+        def representations_request
+          {
+            endpoint: Exercism::Routes.with_feedback_api_mentoring_representations_url,
```
ErikSchierboom:
```suggestion
            endpoint: Exercism::Routes.admin_api_mentoring_representations_url,
```

app/commands/exercise/representation/search.rb
```diff
@@ -24,8 +24,10 @@ def call
 
   private
   def filter_with_feedback!
-    if with_feedback
+    if with_feedback && mentor.present?
       @representations = @representations.with_feedback_by(mentor)
+    elsif with_feedback
+      @representations = @representations.with_feedback
```
ErikSchierboom, line 30: We should add tests to verify this behavior

app/assemblers/assemble_exercise_representations_admin.rb
```diff
@@ -0,0 +1,37 @@
...
+        unscoped_total: Exercise::Representation.with_feedback.count
+      }
+    )
+  end
+
+  private
+  memoize
+  def representations
+    Exercise::Representation::Search.(
+      mentor: nil,
```
iHiD: Let's change this so that we don't need to express it. (ie change the initializer to have `nil` as the default)

app/assemblers/assemble_exercise_representations_admin.rb
```diff
@@ -0,0 +1,37 @@
...
+      page: params.fetch(:page, 1),
+      order: params.fetch(:order, :most_submissions)
+    )
+  end
+
+  def track
+    return if params[:track_slug].blank?
+
+    Track.find(params[:track_slug])
+  end
```
iHiD: I think this is wrong. I think it should require a track and blow up if one is missing.

iHiD:
@ErikSchierboom @dem4ron Added a couple of small comments. This looks looks great though :)

app/assemblers/assemble_exercise_representations_admin.rb
dem4ron: I think it is set up this way to show all the tracks that have a representation. The initial state is to show everything the person mentors unfiltered.

iHiD: > The initial state is to show everything the person mentors unfiltered.

I don't think it is actually. The params passed in are: ```track_slug: params.fetch(:track_slug, track_slugs.first)``` (e.g. in `app/helpers/react_components/mentoring/representations/with_feedback.rb`). I might be wrong, but I don't think there's a scenario where we show everything anymore.

iHiD (approved):

app/commands/exercise/representation/search.rb
```diff
@@ -24,15 +24,21 @@ def call
...
     else
       @representations = @representations.without_feedback
     end
   end
 
   def filter_track!
-    return if track.blank?
+    if track.blank?
+      # Force no records being returned when no track was specified
+      @representations = @representations.where('false')
```
iHiD: I believe you can do `.none`

PR #3591, opened 2023-03-17
Add live streaming events API

ErikSchierboom:
- Add scopes to streaming event
- Add streaming events

### DIFF
--- a/app/assemblers/assemble_streaming_events.rb
+++ b/app/assemblers/assemble_streaming_events.rb
@@ -4,18 +4,18 @@ class AssembleStreamingEvents
   initialize_with :params
 
   def self.events_per_page = 20
-  def self.keys = %i[page per_page scheduled]
+  def self.keys = %i[page per_page live]
 
   def call = SerializePaginatedCollection.(events, serializer: SerializeStreamingEvents)
 
   private
   memoize
   def events
-    rows = scheduled? ? StreamingEvent.scheduled : StreamingEvent.live
+    rows = live? ? StreamingEvent.live : StreamingEvent.scheduled
     rows.order(:starts_at).page(page).per(per_page)
   end
 
-  def scheduled? = params[:scheduled].present?
+  def live? = params[:live].present?
   def page = params.fetch(:page, 1).to_i
   def per_page = params.fetch(:per_page, self.class.events_per_page).to_i
 end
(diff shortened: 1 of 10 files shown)


app/models/streaming_event.rb
```diff
@@ -1,22 +1,19 @@
...
 
-  def self.next_featured
-    where(featured: true).
-      where('starts_at > NOW()').
-      order(:starts_at).
-      first
-  end
+  def self.current_live = live.first
+  def self.next_featured = scheduled.featured.order(:starts_at).first
+  def self.next_5 = scheduled.order(:starts_at).first(5)
```
ErikSchierboom: I've _slightly_ changed this to now use scopes. Let me know what you think

app/serializers/serialize_streaming_events.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    initialize_with :event
+
+    def call
+      {
+        title: event.title,
+        description: event.description,
+        starts_at: event.starts_at,
+        ends_at: event.ends_at,
+        featured: event.featured,
```
iHiD:
```suggestion
        featured: event.featured?,
```

app/models/streaming_event.rb
```diff
@@ -1,22 +1,19 @@
...
+  scope :scheduled, -> { where('starts_at > NOW()') }
+  scope :featured, -> { where(featured: true) }
 
-  def self.next_featured
-    where(featured: true).
-      where('starts_at > NOW()').
-      order(:starts_at).
-      first
-  end
+  def self.current_live = live.first
```
iHiD: This is the only bit I'd probably change. I'd probably remove this and just use `live.first` in the controller. Because remembering the different between "live" and "current_live" isn't intuitive to me without looking at the code, and as it's currently only used in one place.

app/models/streaming_event.rb
iHiD: Yep, good. You could change next_5 to be `next(x)` if you wanted?

app/assemblers/assemble_streaming_events.rb
```diff
@@ -0,0 +1,21 @@
...
+  def call = SerializePaginatedCollection.(events, serializer: SerializeStreamingEvents)
+
+  private
+  memoize
+  def events
+    rows = scheduled? ? StreamingEvent.scheduled : StreamingEvent.live
+    rows.order(:starts_at).page(page).per(per_page)
+  end
+
+  def scheduled? = params[:scheduled].present?
```
iHiD: I'd be sort of tempted to invert this. I feel like `scheduled` should be the default. So I wonder if the param should be `live` instead?

PR #3597, opened 2023-03-21
Add command to update roles

ErikSchierboom:
- Add command to update a user's roles
- Use User::UpdateRoles command to update roles
- Set discord roles when roles change

Reminder: run the following code when this is merged:

```ruby
User.where.not(discord_uid: nil).each{|u|User::SetDiscordRoles.defer(u)}
````

### DIFF
--- a/app/commands/user/add_roles.rb
+++ b/app/commands/user/add_roles.rb
@@ -1,0 +1,10 @@
+class User::AddRoles
+  include Mandate
+
+  initialize_with :user, :roles
+
+  def call
+    user.update(roles: user.roles + roles)
+    User::SetDiscordRoles.(user)
+  end
+end
--- a/app/commands/user/reset_account.rb
+++ b/app/commands/user/reset_account.rb
@@ -19,7 +19,7 @@ class User::ResetAccount
       became_mentor_at: nil
     )
 
-    User::UpdateRoles.(user, [])
+    User::RemoveRoles.(user, user.roles)
   end
 
   def reset_tracks!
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: app/commands/user/update_roles.rb [hunk 1]
```diff
@@ -0,0 +1,10 @@
+class User::UpdateRoles
+  include Mandate
+
+  initialize_with :user, :roles
```
iHiD: How about we do it like this instead, and then encapsulate the actual manipulation within this command?

```suggestion
  initialize_with :user, add: [], remove: []
```

iHiD: Maybe with `remove_all: false` or something to handle the reset_account scenario?

ErikSchierboom: Do three commands

app/commands/user/reset_account.rb
```diff
@@ -12,13 +12,14 @@ def call
...
       reputation: 0,
-      roles: [],
       bio: nil,
       avatar_url: nil,
       location: nil,
       pronouns: nil,
       became_mentor_at: nil
     )
+
+    User::RemoveRoles.(user, user.roles)
```
ErikSchierboom, line 22: I figured I could clear the roles like this too 🤷

app/commands/user/add_roles.rb
```diff
@@ -0,0 +1,10 @@
+class User::AddRoles
+  include Mandate
+
+  initialize_with :user, :roles
```
ErikSchierboom, line 4: I considered making roles a splat argument, but this version is a bit more "obvious." Let me know what you think.

iHiD (approved):

PR #3600, opened 2023-03-21
Add chatterbox badge

ErikSchierboom:
- Fix invalid test file names
- Add chatterbox badge


app/images/icons/chatterbox.svg
ErikSchierboom, line 1: TODO

app/models/badges/chatterbox_badge.rb
```diff
@@ -0,0 +1,12 @@
+module Badges
+  class ChatterboxBadge < Badge
+    seed "Chatterbox",
+      :common,
+      :chatterbox,
+      "Linked your Exercism account to Discord"
+
+    def award_to?(user) = user.discord_uid.present?
+
+    def send_email_on_acquisition? = false
```
ErikSchierboom: Not sure if we wanted to send an email

app/models/badges/chatterbox_badge.rb
```diff
@@ -0,0 +1,12 @@
+module Badges
+  class ChatterboxBadge < Badge
+    seed "Chatterbox",
+      :common,
```
ErikSchierboom, line 4: I was unsure about the rarity

iHiD: common is good.

app/models/badges/chatterbox_badge.rb
iHiD: I'd say `false` in general, but I'd like to email it when we batch award it. So maybe deploy it as true, do the batch, then change it to false. Same for the forum one.

app/models/badges/chatterbox_badge.rb
```diff
@@ -0,0 +1,12 @@
...
+      :chatterbox,
+      "Linked your Exercism account to Discord"
```
iHiD:
```suggestion
      "Joined Exercism's Discord server"
```

iHiD (approved):

PR #3607, opened 2023-03-23
Allow skipping requeueing head test runs

ErikSchierboom:
This PR adds support for skipping the requeueing of tests when the important files hash changes. This should only be used when the maintainer "knows" that the changes won't actually affect the test status. As an example, this would be the case when an invalidator file is added or when comments are adding to a test file.

See https://github.com/exercism/exercism/issues/6678 and https://github.com/exercism/python/pull/3373#issuecomment-1480853252 for more context


app/models/git/exercise.rb
```diff
@@ -258,6 +258,9 @@ def approaches = Git::Exercise::Approaches.new(exercise_slug, exercise_type, git
     memoize
     def articles = Git::Exercise::Articles.new(exercise_slug, exercise_type, git_sha, repo:)
 
+    memoize
+    def skip_test_runs? = commit.message.downcase.include?(SKIP_TEST_RUNS_MAGIC_MARKER)
```
ErikSchierboom: We could consider supporting multiple markers 🤷

app/models/git/exercise.rb
```diff
@@ -285,5 +288,8 @@ def commit = repo.lookup_commit(git_sha)
 
     memoize
     def track = Track.new(repo:)
+
+    SKIP_TEST_RUNS_MAGIC_MARKER = "[skip tests]".freeze
```
ErikSchierboom: I've tried to make it use a similar syntax to GitHub's magic markers (e.g. `[skip ci]`). If we think this is a good plan, we could consider having this be a list of magic markers (e.g. `[skip test runs]` or `[tests skip]`).

iHiD: I like this. I'd like to use `exercism` in the message though.

`[exercism: skip test reruns]` maybe?

app/models/exercise.rb
```diff
@@ -96,7 +96,7 @@ def self.for(track_slug, exercise_slug)
   after_update_commit do
     if saved_changes.include?(:git_important_files_hash)
       Exercise::MarkSolutionsAsOutOfDateInIndex.defer(self)
-      Exercise::QueueSolutionHeadTestRuns.defer(self)
+      Exercise::QueueSolutionHeadTestRuns.defer(self) unless skip_test_runs?
```
iHiD: I think I'd like to move this into `Exercise::QueueSolutionHeadTestRuns` itself. It feels better encapsulated to have the logic there, in case we end up running a batch script later etc.

kytrinyx (approved):

kytrinyx (approved):

kytrinyx (approved):

kytrinyx (approved):

app/commands/exercise/process_git_important_files_changed.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      Exercise::QueueSolutionHeadTestRuns.defer(self)
+      Exercise::MarkSolutionsAsOutOfDateInIndex.defer(self)
+    end
+
+    private
+    def testable_files_have_changed?
+      # If the maintainer has used the manual flag, then
+      # don't run anything here
+      return false if git_no_important_files_changed?
```
ErikSchierboom:
```suggestion
      return false if exercise.git_no_important_files_changed?
```

app/commands/exercise/process_git_important_files_changed.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    initialize_with :exercise
+
+    def call
+      RecalculateImportantFilesHashWithSolutions.(exercise)
+
+      return unless testable_files_have_changed?
+
+      Exercise::QueueSolutionHeadTestRuns.defer(self)
+      Exercise::MarkSolutionsAsOutOfDateInIndex.defer(self)
```
ErikSchierboom:
```suggestion
      Exercise::QueueSolutionHeadTestRuns.defer(exercise)
      Exercise::MarkSolutionsAsOutOfDateInIndex.defer(exercise)
```

PR #3612, opened 2023-03-28
Add copy button to Your iterations

dem4ron:
https://forum.exercism.org/t/line-numbers-in-copied-code/4786

https://user-images.githubusercontent.com/66035744/230125578-bcb77e42-ad90-4b51-8e56-e9aaeb5f83eb.mov

TODO:
- [x] Review it at all the places where this copy button is used
- [x] Remove copy button from `tabs` on Mentoring Discussion page
- [x] Break this css out into a separate classname, and let `CopyToClipboardButton.tsx` use the already existing (on `main`) `c-copy-text-to-clipboard`. 
For this - to work well - the closest `position:relative` parent must be the iteration header, for `CopyToClipboardButtons.tsx` the `c-copy-text-to-clipboard` must be `position:relative` in order to show the "Copied!" animation correctly.


iHiD:
Screenshot 2023-04-05 at 12 35 26

Can you move the button to the RHS of these pls (just to the left of the line). I think the tests and analysis status should go together.

iHiD:
Screenshot pls!

ErikSchierboom:
> Can you move the button to the RHS of these pls (just to the left of the line). I think the tests and analysis status should go together.

But what about tracks that have multiple solution files? Which file's contents should be copied? In Aron's original loom, that was obvious.

iHiD:
@dem4ron Let's get this over the line next week. Can you rebase this pls. I'm then going to play with moving the button a little bit before we merge. I feel like where it is at the moment is too **in line** with the tabs, not in a particular tab. I've got some ideas but easier for me just to play.

PR #3615, opened 2023-03-29
Make tabs scrollable on the x-axis in editor

dem4ron:
This PR adds an invisible scrollbar to tabs in Editor.

https://user-images.githubusercontent.com/66035744/228518587-9af6c4bd-d66a-4f84-bc15-b88946d0bc7f.mov

Update:

https://user-images.githubusercontent.com/66035744/228536988-785a625f-dfd3-457d-9ee8-227879632414.mov

### DIFF
--- a/app/css/ui-kit/scroll.css
+++ b/app/css/ui-kit/scroll.css
@@ -1,0 +1,11 @@
+@layer base {
+    /* Makes element scrollable on the x-axis but hides the scrollbar */
+    .scroll-x-hidden {
+        @apply overflow-x-auto;
+        -ms-overflow-style: none; /* IE and Edge */
+        scrollbar-width: none; /* Firefox */
+        &::-webkit-scrollbar {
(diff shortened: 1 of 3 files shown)


app/css/ui-kit/scroll.css
```diff
@@ -0,0 +1,11 @@
...
+    /* Makes element scrollable on the x-axis but hides the scrollbar */
+    .scroll-x-hidden {
+        @apply overflow-x-auto;
+        -ms-overflow-style: none; /* IE and Edge */
+        scrollbar-width: none; /* Firefox */
+        &::-webkit-scrollbar {
+            display: none; /* Chrome, Opera, Safari */
+        }
+    }
+}
```
iHiD, line 19: So I like this conceptually. My only concern is whether people will know there's extra tabs they **can** scroll to if there's no scroll bar? Do we need to make that more obvious somehow?

iHiD (approved):
Nice work :)

PR #3619, opened 2023-03-30
Update to 3.2.1 and latest gems

iHiD:
This stops the warning I get on my new mac.

**Note: This needs the Dockerfile updating to 3.2.1. before we deploy this!**


app/commands/donations/subscription/cancel.rb
```diff
@@ -7,7 +7,7 @@ class Donations::Subscription::Cancel
 
   def call
     begin
-      Stripe::Subscription.delete(subscription.stripe_id)
+      Stripe::Subscription.cancel(subscription.stripe_id)
```
ErikSchierboom, line 10: See https://github.com/stripe/stripe-ruby/blob/master/CHANGELOG.md#%EF%B8%8F-removed

app/commands/track/determine_concept_map_layout.rb
```diff
@@ -92,9 +92,7 @@ def initialize(user_track)
     end
 
     memoize
-    def empty?
-      nodes.empty?
-    end
+    delegate :empty?, to: :nodes
```
ErikSchierboom: Would this (memoize) still work when using `delegate`?

config/initializers/devise.rb
```diff
@@ -1,17 +1,3 @@
-class TurboFailureApp < Devise::FailureApp
```
ErikSchierboom, line 1: The changes in this file are taken from https://github.com/heartcombo/devise/issues/5560

app/commands/track/determine_concept_map_layout.rb
iHiD: No.

PR #3624, opened 2023-03-31
Add editor to representer feedback

ErikSchierboom:
Screenshot 2023-03-31 at 14 51 57

### DIFF
--- a/app/javascript/components/student/iterations-list/RepresenterFeedback.tsx
+++ b/app/javascript/components/student/iterations-list/RepresenterFeedback.tsx
@@ -31,13 +31,11 @@ function EditedBy({
   author,
   editor,
 }: Pick<Props, 'author' | 'editor'>): JSX.Element | null {
-  if (!editor || editor.name === author.name) {
-    return null
-  } else {
-    return (
-      <em>
-        &nbsp;(edited by <strong>{editor.name}</strong>)
-      </em>
-    )
-  }
+  if (!editor || editor.name === author.name) return null
+
+  return (
+    <em>
+      &nbsp;(edited by <strong>{editor.name}</strong>)
+    </em>
+  )
 }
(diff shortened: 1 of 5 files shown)


app/models/submission.rb
```diff
@@ -159,7 +160,15 @@ def representer_feedback
         reputation: author.reputation,
         avatar_url: author.avatar_url,
         profile_url: author.profile ? Exercism::Routes.profile_url(author) : nil
-      }
+      },
+      editor: if editor.present?
```
ErikSchierboom, line 164: This is rubocop's doing (the if/end)

iHiD (approved):
Nice.

app/javascript/components/student/iterations-list/RepresenterFeedback.tsx
```diff
@@ -21,3 +26,16 @@ export const RepresenterFeedback = ({ html, author }: Props): JSX.Element => {
     </div>
   )
 }
+
+function EditedBy({
+  author,
+  editor,
+}: Pick<Props, 'author' | 'editor'>): JSX.Element | null {
+  if (editor && editor.name !== author.name) {
```
iHiD: Maybe early return is nicer here?
`if (!editor || editor.name == author.name) { return null }`

app/javascript/components/student/iterations-list/RepresenterFeedback.tsx
```diff
@@ -31,11 +31,13 @@ function EditedBy({
   author,
   editor,
 }: Pick<Props, 'author' | 'editor'>): JSX.Element | null {
-  if (editor && editor.name !== author.name) {
+  if (!editor || editor.name === author.name) {
+    return null
+  } else {
```
iHiD: You don't need this else now (that's the beauty of the early return is that everything else stops needing to be indented).

PR #3633, opened 2023-04-04
Prefer US English spelling of ize

ee7:
Not all of these are user-facing. Please take over and revert anything that we don't want. I erred towards changing more, rather than changing less, on the basis that it could make future instances of *ise easier to search for.

The most visible one is "familiarise", which everyone sees while solving the hello-world exercise.

From the [Exercism Markdown Style Guide][1]:

>This document acts as a Style Guide for the language and wording used in exercises.
>
>[...]
>
>All content should be written in US English, which differs from UK English in a variety of ways.
>In the future other translations may occur, but the "official" Exercism language is US English.

[1]: https://github.com/exercism/docs/blob/73835d5d918f/building/markdown/style-guide.md#language

I'm assuming that we generally do want US English everywhere.


iHiD:
Thanks! :)

PR #3652, opened 2023-04-10
Fix issue with people missing introductions

iHiD:
From http://forum.exercism.org/t/becoming-maintainer-of-the-rust-track/5051/26 I've learnt that people are missing the introductions on Concept Exercises. I've updated the template to fix that.


ErikSchierboom (approved):
How weird!

PR #3666, opened 2023-04-12
Support feature flag for insiders

ErikSchierboom:


config/initializers/feature_flags.rb
```diff
@@ -0,0 +1,3 @@
+module FeatureFlag
```
ErikSchierboom, line 1: I went back and forth on this being singular or plural.
In the end I think `if FeatureFlag::INSIDERS` reads a tiny bit better than `if FeatureFlags::INSIDERS`.

PR #3667, opened 2023-04-12
Launch basic Insiders page

ErikSchierboom:
- Build algorithm
- Add insiders_status field to users table


app/commands/insiders/activate.rb
```diff
@@ -0,0 +1,13 @@
+class Insiders::Activate
+  include Mandate
+
+  initialize_with :user
+
+  def call
+    return unless user.insiders_status == :eligible
```
iHiD: (This needs to pivot on the `lifetime_` stuff too)

app/commands/insiders/grant_access.rb
```diff
@@ -0,0 +1,12 @@
...
+  include Mandate
+
+  initialize_with :user
+
+  def call
+    return if user.insiders_status == :lifetime_active
+    return if user.insiders_status == :active
+
+    user.update(insiders_status: :eligible)
+  end
```
iHiD: Add a lock

app/commands/insiders/activate.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    user.update(insiders_status: :active)
```
iHiD: Add a lock to this and the lines above.

app/commands/user/add_roles.rb
```diff
@@ -6,5 +6,6 @@ class User::AddRoles
   def call
     user.update(roles: user.roles + roles)
     User::SetDiscordRoles.(user)
+    User::InsidersStatus::TriggerUpdate.(user)
```
ErikSchierboom, line 9: Now I need to remember to use the command when making someone a maintainer :)

app/commands/user/insiders_status/activate.rb
```diff
@@ -0,0 +1,17 @@
...
+  def call
+    return unless %i[eligible eligible_lifetime].include?(user.insiders_status)
+
+    if user.insiders_status == :eligible
+      user.update(insiders_status: :active)
+    else
+      user.update(insiders_status: :lifetime_active)
+    end
+
+    User::Notification::Create.(user, :joined_insiders)
```
ErikSchierboom: We probably want a different notification for lifetime insiders I guess?

app/commands/user/insiders_status/update.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  private
+  memoize
+  def eligibility_status = User::InsidersStatus::DetermineEligibilityStatus.(user)
+
+  def update_eligible_lifetime
+    return if user.insiders_status == :lifetime_active
+
+    user.update(insiders_status: :eligible_lifetime)
```
ErikSchierboom: I assume that if someone is currently active, we do want the status to become eligible lifetime? We do need to consider that to be "active" in the UI probably

app/commands/user/insiders_status/update.rb
```diff
@@ -0,0 +1,48 @@
...
+  def update_eligible
+    return if user.insiders_status == :lifetime_active
+    return if user.insiders_status == :active
+
+    user.update(insiders_status: :eligible)
+  end
+
+  def update_ineligible
+    return if user.insiders_status == :lifetime_active
+    return if user.insiders_status == :expired
```
ErikSchierboom: Expired should remain expired I think?

app/jobs/recheck_active_insiders_job.rb
```diff
@@ -0,0 +1,11 @@
+class RecheckActiveInsidersJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    User.where(insiders_status: :active).find_each do |user|
```
ErikSchierboom: Do we only want to check active insiders?

app/mailers/notifications_mailer.rb
```diff
@@ -132,4 +132,13 @@ def automated_feedback_added
     subject = "There's new feedback on your solution to #{@track.title}/#{@exercise.title}"
     mail_to_user(@user, subject)
   end
+
+  def joined_insiders
+    notification = params[:notification]
+    @user = notification.user
+
+    @hide_header = true
+    subject = "Welcome to Exercism Insiders"
```
ErikSchierboom, line 141: This is just a stub

db/migrate/20230412075814_add_insiders_status_to_users.rb
```diff
@@ -0,0 +1,8 @@
+class AddInsidersStatusToUsers < ActiveRecord::Migration[7.0]
+  def change
+    return if Rails.env.production?
+
+    add_column :users, :insiders_status, :tinyint, null: false, default: 0
```
ErikSchierboom, line 5: Default is `unset` or `ineligible`?

app/models/user/notifications/joined_insiders_notification.rb
```diff
@@ -0,0 +1,9 @@
+class User::Notifications::JoinedInsidersNotification < User::Notification
+  # TODO: determine what values to use
+  def url = Exercism::Routes.dashboard_url
+  def image_type; end
+  def image_url; end
+
+  # Users should only have this notification once
+  def guard_params = ""
```
ErikSchierboom, line 8: Is this correct? As in: this prevents someone from re-receiving the notification?

app/commands/user/insiders_status/trigger_update.rb
```diff
@@ -0,0 +1,10 @@
+class User::InsidersStatus::TriggerUpdate
+  include Mandate
+
+  initialize_with :user
+
+  def call
+    user.update(insiders_status: :unset)
```
ErikSchierboom: Don't unset active (or lifetime_active)

app/commands/user/insiders_status/update.rb
iHiD: I've updated this. Check if you're happy.

app/commands/user/insiders_status/update.rb
iHiD: I wonder if we actually need `expired`, or if we just need to change back to `illegible` and trigger an email when that happens.

app/jobs/recheck_active_insiders_job.rb
iHiD: Yes.

iHiD: We could probably add a scope here for `User.insiders`

app/models/user/notifications/joined_insiders_notification.rb
iHiD: I think it should be `nil`. I also think we can move this into `app/models/concerns/is_paramaterised_sti.rb` under the `def i18n_params` line. If the tests still pass, I think we're fine to do this.

app/jobs/recheck_active_insiders_job.rb
ErikSchierboom: > User.insiders

That should include `lifetime_active` users too, and those we don't want to check here. Even `User.active_insiders` feels like it is not the right name.

app/commands/user/insiders_status/update.rb
ErikSchierboom: I like it more when it is removed

app/commands/user/insiders_status/update.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  queue_as :default
+
+  initialize_with :user, :status_before_unset
+
+  def call
+    return unless user.insiders_status == :unset
+
+    user.with_lock do
+      case User::InsidersStatus::DetermineEligibilityStatus.(user)
```
iHiD: This need to be outside of the lock, as it's expensive.

app/commands/user/insiders_status/update.rb
```diff
@@ -0,0 +1,62 @@
...
+      case eligibility_status
+      when :eligible_lifetime
+        update_eligible_lifetime
+      when :eligible
+        update_eligible
+      when :ineligible
+        update_ineligible
+      end
+    end
+  end
```
iHiD, line 21: This isn't essential now, but I'd quite to not include the notification sending bit within the lock.

ErikSchierboom: Follow-up PR to fix this: https://github.com/exercism/website/pull/3700

PR #3683, opened 2023-04-14
Sync Stripe subscriptions

ErikSchierboom:
- Add Deactivate command for subscriptions
- Add syncing of subscriptions

### DIFF
--- a/app/commands/donations/stripe/sync_subscriptions.rb
+++ b/app/commands/donations/stripe/sync_subscriptions.rb
@@ -12,7 +12,7 @@ class Donations::Stripe::SyncSubscriptions
     missing_active_subscription_ids.each do |subscription_id|
       subscription = active_stripe_subscriptions[subscription_id]
 
-      user = User.find_by(stripe_customer_id: subscription.customer)
+      user = subscription_user(subscription)
       next unless user
 
       Donations::Subscription::Create.(user, subscription)
@@ -31,7 +31,8 @@ class Donations::Stripe::SyncSubscriptions
   def active_stripe_subscriptions
     subscriptions = Stripe::Subscription.search({
       query: "status:'active'",
-      limit: 100
+      limit: 100,
+      expand: ["data.customer"]
     })
 
     subscriptions.auto_paging_each.index_by(&:id)
@@ -45,4 +46,13 @@ class Donations::Stripe::SyncSubscriptions
 
   memoize
   def local_subscription_ids = local_subscriptions.keys
+
+  def subscription_user(subscription)
+    stripe_user = User.find_by(stripe_customer_id: subscription.customer.id)
+    return stripe_user if stripe_user
+
+    User.find_by(email: subscription.customer.email)&.tap do |user|
+      user.update(stripe_customer_id: subscription.customer.id)
+    end
+  end
 end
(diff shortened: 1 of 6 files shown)


app/commands/donations/subscription/cancel.rb
```diff
@@ -15,10 +15,6 @@ def call
       raise if data.status == 'active'
     end
 
-    subscription.update!(active: false)
-
-    # Update based on whether there is another different active subscription
-    user = subscription.user
-    User::SetActiveDonationSubscription.(user, user.donation_subscriptions.active.exists?)
+    Donations::Subscription::Deactivate.(subscription)
```
ErikSchierboom, line 18: I've extracted this logic to a separate command as I want to use its functionality in the syncing command too

app/commands/donations/stripe/sync_subscriptions.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      Donations::Subscription::Create.(user, subscription)
+    end
+  end
+
+  def deactivate_subscriptions!
+    inactive_subscription_ids = local_subscription_ids - active_stripe_subscription_ids
+    inactive_subscription_ids.each do |subscription_id|
+      subscription = local_subscriptions[subscription_id]
+      Donations::Subscription::Deactivate.(subscription) if subscription.active
```
ErikSchierboom, line 26: We need to figure out which statuses we'll deactivate for. The available values are:
`incomplete`, `incomplete_expired`, `trialing`, `active`, `past_due`, `canceled`, or `unpaid`.

app/commands/donations/stripe/sync_subscriptions.rb
```diff
@@ -0,0 +1,48 @@
...
+    missing_active_subscription_ids = active_stripe_subscription_ids - local_subscription_ids
+    missing_active_subscription_ids.each do |subscription_id|
+      subscription = active_stripe_subscriptions[subscription_id]
+
+      user = User.find_by(stripe_customer_id: subscription.customer)
+      next unless user
+
+      Donations::Subscription::Create.(user, subscription)
+    end
+  end
```
ErikSchierboom: This will create a subscription for any active subscription in Stripe that is not yet in our database.
One caveat: this only works if the user already has a stripe customer ID.

app/commands/donations/stripe/sync_subscriptions.rb
```diff
@@ -0,0 +1,48 @@
...
+    deactivate_subscriptions!
+  end
+
+  def create_subscriptions!
+    missing_active_subscription_ids = active_stripe_subscription_ids - local_subscription_ids
+    missing_active_subscription_ids.each do |subscription_id|
+      subscription = active_stripe_subscriptions[subscription_id]
+
+      user = User.find_by(stripe_customer_id: subscription.customer)
+      next unless user
```
iHiD: If this is missing, I wonder if we should set the `stripe_customer_id` by syncing on the email field coming back from Stripe, rather that doing a noop?

app/commands/donations/subscription/deactivate.rb
```diff
@@ -0,0 +1,14 @@
...
+  initialize_with :subscription
+
+  def call
+    subscription.update!(active: false)
+
+    # Update based on whether there is another different active subscription
+    user = subscription.user
+    User::SetActiveDonationSubscription.(user, user.donation_subscriptions.active.exists?)
+  end
+end
```
iHiD, line 14: Can you add a TODO to the top of this file to write tests for this.

app/commands/donations/stripe/sync_subscriptions.rb
ErikSchierboom: Yeah we can do that

PR #3685, opened 2023-04-14
Tweak subscription structure

ErikSchierboom:
- Add Deactivate command for subscriptions
- Add syncing of subscriptions
- Sync subscriptions
- Add tests for deactivate
- Automatically create user
- Add status column to subscriptions
- Update status correctly


db/migrate/20230414124641_add_status_to_subscriptions.rb
```diff
@@ -0,0 +1,10 @@
+class AddStatusToSubscriptions < ActiveRecord::Migration[7.0]
+  def change
+    return if Rails.env.production?
+
+    add_column :donations_subscriptions, :status, :tinyint, null: false, default: 0
```
ErikSchierboom: I've not yet removed the column as we don't want to end up in a situation where the DB is updated but the old code still runs which would fail when the column isn't there

app/models/donations/subscription.rb
```diff
@@ -1,7 +1,9 @@
 class Donations::Subscription < ApplicationRecord
   belongs_to :user
 
-  scope :active, -> { where(active: true) }
+  scope :active, -> { where(status: :active) }
+
+  enum status: { canceled: 0, overdue: 1, active: 2 }, _suffix: true
```
iHiD: I believe if you remove the `_suffix` here, you can remove the scope above. 

Then you get:
`Donations::Subscription.active` and `subscription.active?` from the `enum` for free.

iHiD: I only use the suffix where it reads more clearly that way (e.g. if we wanted to have `subscription.active_status?` for some reason because gramatically it works better)

app/commands/donations/stripe/sync_subscriptions.rb
```diff
@@ -47,6 +59,12 @@ def local_subscriptions = Donations::Subscription.find_each.index_by(&:stripe_id
   memoize
   def local_subscription_ids = local_subscriptions.keys
 
+  def update_subscription_status(subscription, status)
+    subscription.update(status:)
+
+    Donations::Subscription::Deactivate.(subscription) if subscription.status == :canceled
```
iHiD:
```suggestion
    Donations::Subscription::Deactivate.(subscription) if subscription.canceled?
```

app/commands/donations/stripe/sync_subscriptions.rb
```diff
@@ -19,11 +20,22 @@ def create_subscriptions!
...
+      subscription = local_subscriptions[subscription_id]
+      stripe_subscription = active_stripe_subscriptions[subscription_id]
+
+      update_subscription_status(subscription, SUBSCRIPTION_STATUSES[stripe_subscription.status])
+    end
+  end
+
   def deactivate_subscriptions!
     inactive_subscription_ids = local_subscription_ids - active_stripe_subscription_ids
     inactive_subscription_ids.each do |subscription_id|
```
iHiD:
```suggestion
    inactive_subscription_ids.each do |subscription_id|
      next unless local_subscriptions[subscription_id].active?
```

If it's already inactive, we don't need to redo this block for this subcription.

app/commands/donations/stripe/sync_subscriptions.rb
```diff
@@ -19,11 +20,22 @@ def create_subscriptions!
...
       subscription = local_subscriptions[subscription_id]
-      Donations::Subscription::Deactivate.(subscription) if subscription.active
+
+      update_subscription_status(subscription, SUBSCRIPTION_STATUSES[subscription.status])
```
iHiD: This should use the `stripe_subscription` for the updating bit (same as the update block above does)

```suggestion
      stripe_subscription = Stripe::Subscription.retrieve(subscription_id) # **This line should change to this**

      update_subscription_status(subscription, SUBSCRIPTION_STATUSES[stripe_subscription.status])
```

app/commands/donations/stripe/sync_subscriptions.rb
ErikSchierboom: I think we do, as it can still be in different inactive states (canceled and overdue)

app/commands/donations/stripe/create_payment.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  private
+  def external_id = stripe_data.id
+  def external_receipt_url = stripe_data.charges.first.receipt_url
+  def amount_in_cents = stripe_data.amount
+
+  memoize
+  def subscription
+    return @subscription if @subscription
+
```
iHiD, line 28:
```suggestion
```

app/commands/donations/stripe/create_payment.rb
```diff
@@ -0,0 +1,40 @@
...
+    return unless stripe_data.invoice
```
iHiD:
```suggestion
    return nil unless stripe_data.invoice
```

I just prefer being explicit here as the return type actually matters.

app/commands/donations/stripe/create_payment.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    invoice = Stripe::Invoice.retrieve(stripe_data.invoice)
+    return unless invoice.subscription
```
iHiD:
```suggestion
    return nil unless invoice.subscription
```

app/commands/donations/stripe/sync_subscriptions.rb
```diff
@@ -15,15 +16,27 @@ def create_subscriptions!
...
-      Donations::Subscription::Create.(user, subscription)
+      Donations::Stripe::CreateSubscription.(user, subscription)
+    end
+  end
+
+  def update_subscriptions!
+    subscription_ids = local_subscription_ids & active_stripe_subscription_ids
+    subscription_ids.each do |subscription_id|
+      subscription = local_subscriptions[subscription_id]
+      stripe_subscription = active_stripe_subscriptions[subscription_id]
```
iHiD: Can you update L14 to also have the variable called `stripe_subscription` just to make the whole file more consistently easy to read pls.

PR #3698, opened 2023-04-20
GitHub sponsor subscriptions

ErikSchierboom:
Add GitHub sponsor subscriptions


app/commands/donations/stripe/subscription/update_amount.rb
```diff
@@ -25,9 +25,5 @@ def call
     )
 
     subscription.update!(amount_in_cents:)
-
-    # Update based on whether there is another different active subscription
-    user = subscription.user
-    User::SetActiveDonationSubscription.(user, user.donation_subscriptions.active.exists?)
```
ErikSchierboom, line 31: As we're only updating the amount of the subscription, `user.donation_subscriptions.active` doesn't change

app/controllers/webhooks/github_sponsors_controller.rb
```diff
@@ -6,7 +6,11 @@ def create
       ProcessGithubSponsorUpdateJob.perform_later(
         # params[:action] does not work as it is populated by Rails with the action method name
         request.request_parameters[:action],
-        params[:sponsorship][:sponsor][:login]
+        params[:sponsorship][:sponsor][:login],
+        params[:sponsorship][:node_id],
+        params[:sponsorship][:privacy_level],
```
ErikSchierboom: I was wondering how we want to handle the privacy level. I can of course update the user's `show_on_supporters_page` property, but I might be overriding something that was already set (the field can't be  `null` so I can't know if the user changed this value, can I?)

db/migrate/20230418133624_change_external_receipt_url_to_null_in_donations_payments.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeExternalReceiptUrlToNullInDonationsPayments < ActiveRecord::Migration[7.0]
+  def change
+    change_column_null :donations_payments, :external_receipt_url, true
```
ErikSchierboom, line 3: AFAICT there is no receipt URL for GitHub sponsorships

app/commands/user/set_active_donation_subscription.rb
ErikSchierboom, line 1: I've refactored this to capture the logic of how to determine the active donation subscription (via `user.donation_subscriptions.active.exists?`) within the command, which means that the caller doesn't have to do that (there was some repeated logic). Then I renamed the command to `UpdateActiveDonationSubscription`.

app/jobs/process_github_sponsor_update_job.rb
```diff
@@ -1,12 +1,19 @@
 class ProcessGithubSponsorUpdateJob < ApplicationJob
   queue_as :default
 
-  def perform(action, gh_username)
-    return unless action == 'created'
-
+  def perform(action, gh_username, node_id, privacy_level, is_one_time, monthly_price_in_cents)
     user = User.find_by(github_username: gh_username)
```
ErikSchierboom, line 5: Do we want to attempt to automatically update a user's github_username if we can't find it? We try to etch the user's email via another API call (it is not in the payload I think)

app/jobs/process_github_sponsor_update_job.rb
```diff
@@ -1,12 +1,19 @@
...
+    case action
+    when 'cancelled'
+      Donations::Github::Sponsorship::HandleCancelled.(user, node_id, privacy_level, is_one_time, monthly_price_in_cents)
+    when 'created'
+      Donations::Github::Sponsorship::HandleCreated.(user, node_id, privacy_level, is_one_time, monthly_price_in_cents)
+    when 'edited'
+      Donations::Github::Sponsorship::HandleEdited.(user, node_id, privacy_level, is_one_time, monthly_price_in_cents)
+    when 'tier_changed'
+      Donations::Github::Sponsorship::HandleTierChanged.(user, node_id, privacy_level, is_one_time, monthly_price_in_cents)
+    end
```
ErikSchierboom: I'll cleanup unused parameters once we've decided what we want to do with the privacy level (see other comment)

app/commands/donations/github/sponsorship/handle_cancelled.rb
```diff
@@ -0,0 +1,16 @@
+# This responds to a Github sponsorship webhook event with
+# the action being 'cancelled'
+class Donations::Github::Sponsorship::HandleCancelled
+  include Mandate
+
+  initialize_with :user, :node_id, :privacy_level, :is_one_time, :monthly_price_in_cents
```
iHiD: We don't need/use `monthly_price_in_cents` or `privacy_level ` here.

app/commands/donations/github/sponsorship/handle_created.rb
```diff
@@ -0,0 +1,14 @@
+# This responds to a Github sponsorship webhook event with
+# the action being 'created'
+class Donations::Github::Sponsorship::HandleCreated
+  include Mandate
+
+  queue_as :default
+
+  initialize_with :user, :node_id, :privacy_level, :is_one_time, :monthly_price_in_cents
```
iHiD: We don't need `privacy_level` here.

app/commands/donations/github/sponsorship/handle_edited.rb
```diff
@@ -0,0 +1,12 @@
+# This responds to a Github sponsorship webhook event with
+# the action being 'edited'
+class Donations::Github::Sponsorship::HandleEdited
+  include Mandate
+
+  initialize_with :user, :node_id, :privacy_level, :is_one_time, :monthly_price_in_cents
```
iHiD: Do we need to update the subscriptions' `monthly_price_in_cents` here?

iHiD: We can just get rid of this?

app/commands/donations/github/sponsorship/handle_tier_changed.rb
```diff
@@ -0,0 +1,16 @@
+# This responds to a Github sponsorship webhook event with
+# the action being 'tier_changed'
+class Donations::Github::Sponsorship::HandleTierChanged
+  include Mandate
+
+  initialize_with :user, :node_id, :privacy_level, :is_one_time, :monthly_price_in_cents
```
iHiD: Remove `privacy_level`

app/commands/donations/github/sponsorship/handle_cancelled.rb
```diff
@@ -0,0 +1,16 @@
+# This responds to a Github sponsorship webhook event with
+# the action being 'cancelled'
+class Donations::Github::Sponsorship::HandleCancelled
+  include Mandate
+
+  initialize_with :user, :node_id, :is_one_time, :monthly_price_in_cents
```
iHiD: Get rid of `monthly_price_in_cents`

iHiD (approved):

PR #3702, opened 2023-04-21
Listen for PayPal webhook events

ErikSchierboom:
Add Paypal support


app/commands/donations/subscription/activate.rb
```diff
@@ -0,0 +1,7 @@
+class Donations::Subscription::Activate
+  include Mandate
+
+  initialize_with :subscription
+
+  def call = subscription.update!(status: :active)
```
ErikSchierboom: This command is currently a bit silly. I felt like this might be warranted still as it is called from multiple places, but 🤷

app/commands/donations/paypal/customer/find_or_update.rb
```diff
@@ -0,0 +1,14 @@
...
+  initialize_with :paypal_payer_id, :email
+
+  def call
+    paypal_user = User.find_by(paypal_payer_id:)
+    return paypal_user if paypal_user
+
+    User.find_by(email:)&.tap do |user|
+      user.update(paypal_payer_id:)
+    end
+  end
```
iHiD, line 13:
```suggestion
    User.where(email:).update(paypal_payer_id:)
  end
```

app/commands/donations/subscription/activate.rb
iHiD: Do we need to recall the insiders check here too?

app/commands/donations/subscription/cancel.rb
```diff
@@ -0,0 +1,7 @@
+class Donations::Subscription::Cancel
+  include Mandate
+
+  initialize_with :subscription
+
+  def call = subscription.update!(status: :canceled)
```
iHiD: And insiders check here too.

app/commands/donations/subscription/overdue.rb
```diff
@@ -0,0 +1,7 @@
+class Donations::Subscription::Overdue
+  include Mandate
+
+  initialize_with :subscription
+
+  def call = subscription.update!(status: :overdue)
```
iHiD: Insiders check here (in case we decide to invalidate in the future)

app/commands/donations/subscription/update_amount.rb
```diff
@@ -0,0 +1,7 @@
+class Donations::Subscription::UpdateAmount
+  include Mandate
+
+  initialize_with :subscription, :amount_in_cents
+
+  def call = subscription.update!(amount_in_cents:)
```
iHiD: Do an insiders check here too, in case we end up having a threshold for insiders (we will do with Pro for example)

app/commands/webhooks/paypal/request_access_token.rb
```diff
@@ -0,0 +1,41 @@
...
+  rescue StandardError
+    raise RequestAccessTokenError
+  end
+
+  private
+  memoize
+  def http
+    http = Net::HTTP.new(OAUTH2_TOKEN_URI.host, OAUTH2_TOKEN_URI.port)
+    http.use_ssl = true
+    http
```
iHiD:
```suggestion
    Net::HTTP.new(OAUTH2_TOKEN_URI.host, OAUTH2_TOKEN_URI.port).tap do |http|
      http.use_ssl = true
    end
```

app/commands/webhooks/paypal/request_access_token.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def request
+    request = Net::HTTP::Post.new(OAUTH2_TOKEN_URI.request_uri)
+    request.basic_auth(paypal_client_id, paypal_client_secret)
+    request.set_content_type('application/x-www-form-urlencoded')
+    request.set_form_data({ 'grant_type' => 'client_credentials' })
+    request
```
iHiD: You could use tap here too.

app/commands/webhooks/paypal/request_access_token.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def paypal_client_id = ENV.fetch('PAYPAL_CLIENT_ID', Exercism.secrets.paypal_client_id)
+  def paypal_client_secret = ENV.fetch('PAYPAL_CLIENT_SECRET', Exercism.secrets.paypal_client_secret)
```
iHiD: Can we get rid of ENV here and use the new `~/.config/exercism/secrets.yml` thing instead pls?

app/controllers/webhooks/paypal_controller.rb
```diff
@@ -0,0 +1,18 @@
+# This controller listens for webhooks from PayPal.
+module Webhooks
+  class PaypalController < BaseController
+    def create
+      Webhooks::Paypal::VerifySignature.(request.headers, payload_body)
+
+      Webhooks::ProcessPaypalUpdate.(
+        params[:event_type],
+        request.request_parameters[:resource] # params[:resource] does not work as it is populated by Rails
```
iHiD, line 9: Bet this was fun to debug.

db/migrate/20230421064441_add_paypal_payer_id_to_users.rb
```diff
@@ -0,0 +1,7 @@
+class AddPaypalPayerIdToUsers < ActiveRecord::Migration[7.0]
+  def change
+    return if Rails.env.production?
+
+    add_column :users, :paypal_payer_id, :string, null: true, index: {unique: true}
```
iHiD: Let's remember to migrate this to `data` in that PR.

app/controllers/webhooks/paypal_controller.rb
ErikSchierboom: Yep. But I remembered having a similar issue with the `action` so it didn't take very long

app/commands/webhooks/paypal/request_access_token.rb
ErikSchierboom: Ah nice: https://github.com/exercism/config/commit/4a8e0100c9dbeb294900adb91420c9910593e08a

PR #3712, opened 2023-04-24
Add user data record

iHiD:
This adds a user_data table. It proxies things from user to user_data.

It's much more complex than it needs to be as we need to handle the migration phase without breaking the website. Once we've migrated we can merge #3713 which will make the code much simpler.

PR #3713, opened 2023-04-24
Simplify user data

iHiD:
**⚠️ This PR will lead to data loss unless 3712 is fully migrated first ⚠️**

---

Builds on #3712 to get us to the final simpler position


app/commands/user/reputation_token/award_for_pull_request_reviewers.rb
```diff
@@ -17,7 +17,7 @@ def call
     reviewer_usernames.delete(params[:author_username])
 
     # Only award reviewer reputation to organization members
-    reviewer_usernames &= ::Github::OrganizationMember.pluck(:username)
+    reviewer_usernames &= ::Github::OrganizationMember.map(&:username)
```
iHiD:
```suggestion
    reviewer_usernames &= ::Github::OrganizationMember.pluck(:username)
```

PR #3717, opened 2023-04-25
Add admin section to find user

ErikSchierboom:
- Add admin index page
- Add admin section for finding users


app/controllers/admin/admin_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::AdminController < ApplicationController
```
iHiD:
```suggestion
class Admin::BaseController < ApplicationController
```

etc

iHiD (approved):
I'd like the base controller renaming as per my comment (and updating in the other places too). This is the standard we have elsewhere. Other than that, great :)

app/controllers/admin/admin_controller.rb
ErikSchierboom: And rename to DashboardController

PR #3732, opened 2023-04-27
Remove old tables

iHiD:


db/schema.rb
```diff
@@ -33,7 +33,7 @@
     t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
   end
 
-  create_table "active_storage_variant_records", charset: "utf8mb4", force: :cascade do |t|
+  create_table "active_storage_variant_records", charset: "utf8mb4", collation: "utf8mb4_0900_ai_ci", force: :cascade do |t|
```
iHiD, line 36: @ErikSchierboom Not sure how we fix/standardise this

ErikSchierboom: Wow, that's odd

ErikSchierboom (approved):

iHiD:
> Wow, that's odd
@ErikSchierboom (and @dem4ron): FYI, that collation is the new one in Mysql 8.1, but as our DB isn't that new, it's still the old one. And also CI apparently doesn't support it either.

PR #5323, opened 2023-05-01
Add a cache to remove three SQL calls from every request

iHiD:
@ErikSchierboom This will probably be a simple but satisfying one, and removes there SQL queries from every request. This mainly speeds things up for power users.

We need things adding for the creation and revealing of badges, testimonials and badges, and tests adding.

This is low priority but if you just want 30mins of coding at any point, it's a valuable one.


iHiD:
(I also apologise for the schema diff. My db is total trash at this point...!)

app/commands/mentor/discussion/finish_by_student.rb
```diff
@@ -64,6 +64,7 @@ def block!
     ).update!(blocked_by_student: true)
   end
 
+  # TODO: Move to command and call User::ResetCache
```
ErikSchierboom: Where can I find this User::ResetCache file?

iHiD: In my latest commit. *sigh*

app/commands/mentor/testimonial/create.rb
```diff
@@ -0,0 +1,15 @@
...
+  initialize_with :discussion, :testimonial
+
+  def call
+    Mentor::Testimonial.create!(
+      mentor: discussion.mentor,
+      student: discussion.student,
+      discussion:,
+      content: testimonial
+    )
+    User::ResetCache.(discussion.mentor)
```
ErikSchierboom: I've not made this into a job yet. Do we want that? I think the queries are quite fast.

app/commands/user/reputation_token/mark_as_seen.rb
```diff
@@ -0,0 +1,12 @@
+class User::ReputationToken::MarkAsSeen
+  include Mandate
+
+  initialize_with :token
+
+  def call
+    return if token.seen?
```
ErikSchierboom, line 7: This is a micro optimization to prevent unnecessary cache resetting. Is this okay or should I use the return value of `seen!` (or some other state)

iHiD: This is good.

app/commands/donations/stripe/customer/create_for_user.rb
```diff
@@ -22,7 +22,7 @@ def call
           email: user.email,
           metadata: { user_id: user.id }
         )
-        user.update_column(:stripe_customer_id, customer.id)
+        user.update(stripe_customer_id: customer.id)
```
iHiD, line 25:
```suggestion
        user.update!(stripe_customer_id: customer.id)
```

app/commands/mentor/testimonial/create.rb
iHiD: Yeah, job. But you can just call `defer` instead of `.()` here.

PR #5324, opened 2023-05-02
Update to modern Yarn

ErikSchierboom:
https://yarnpkg.com/getting-started/qa


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: pnp
```
ErikSchierboom: This should make things a lot faster. You can also do zero installs, but that would mean committing a lot of packages (about 100MB) into version control, which I don't think we want

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
ErikSchierboom, line 1: Ideally, we use `pnp` for this, but that seemed to be breaking things and I don't want to spend time of fixing it

PR #5327, opened 2023-05-02
:lipstick: Fix code snippet overflow in tracks about section

homersimpsons:
On Ballerina track (https://exercism.org/tracks/ballerina) the code snippet is not readable because it overflows to the right.

I am not sure about the selector but it looks like this component css is designed like this.

Before:
image

After:
image
image

### DIFF
--- a/app/css/pages/track-about.css
+++ b/app/css/pages/track-about.css
@@ -144,6 +144,11 @@
                 @screen md {
                     max-width: 450px;
                 }
+
+                & > code {
+                    display: block;
+                    overflow-x: auto;
+                }
             }
         }
     }


iHiD:
I would rather shorten the ballerina snippet than make this box scrollable personally. I think having horizontally scrolling boxes isn't a great UI pattern and quite ugly with scrollbars on windows etc. This is aimed to be more of an "image" than it is useful text really here.

homersimpsons:
> I would rather shorten the ballerina snippet than make this box scrollable personally.

Okay fine, I think this example is like this because it clearly highlights the fact that the language is designed around web services.

> I think having horizontally scrolling boxes isn't a great UI pattern and quite ugly with scrollbars on windows etc.

Maybe on some mobile phones the width won't be enough for other tracks snippets.

I did attempt to allow word wrap too but it does not really work.

> This is aimed to be more of an "image" than it is useful text really here.

I understand that it should just show a sample for a hello world, but I guess having some specificities of the language is interesting too.

iHiD:
> I understand that it should just show a sample for a hello world, but I guess having some specificities of the language is interesting too.

Yeah - I don't disagree with that. Maybe below we could show more code where we have more space, but I think with the existing design, this is intended as a "graphic" rathe than something designed to be comprehensive..

PR #5329, opened 2023-05-03
Disable selected theme selector

dem4ron:
This PR disables functionally the currently selected theme-selector button, but keeps a locked state visually for `non-insiders` disabled level.

### DIFF
--- a/app/javascript/components/settings/theme-preference-form/InfoMessage.tsx
+++ b/app/javascript/components/settings/theme-preference-form/InfoMessage.tsx
@@ -1,0 +1,39 @@
+import React from 'react'
+
+export function InfoMessage({
+  insidersStatus,
+  insidersPath,
+}: {
+  insidersStatus: string
+  insidersPath: string
+}): JSX.Element {
+  switch (insidersStatus) {
+    case 'active':
+    case 'active_lifetime':
+      return (
+        <p className="text-p-base mb-16">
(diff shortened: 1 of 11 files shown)


app/javascript/components/settings/theme-preference-form/InfoMessage.tsx
```diff
@@ -0,0 +1,39 @@
...
+        </p>
+      )
+    case 'eligible':
+    case 'eligible_lifetime':
+      return (
+        <p className="text-p-base mb-16">
+          You&apos;re eligible to join Insiders.{' '}
+          <a href={insidersPath}>Get started here.</a>
+        </p>
+      )
```
ErikSchierboom, line 25: Maybe we have two separate messages here? As a lifetime insider is something slightly different and one could already be an insider.

app/javascript/components/settings/theme-preference-form/InfoMessage.tsx
```diff
@@ -0,0 +1,39 @@
...
+    case 'ineligible':
+      return (
+        <p className="text-p-base mb-16">
+          Dark theme is only available to Exercism Insiders.
+        </p>
+      )
+    default:
+      return (
+        <p className="text-p-base mb-16">
+          [Learn more about Exercism Insiders](...).
```
ErikSchierboom, line 35: Maybe add a TODO here?

PR #5334, opened 2023-05-03
Refine notifications and emails

ErikSchierboom:
- Rename original-insiders to lifetime-insiders
- Award lifetime insider
- Auto-award lifetime-insider badge


app/commands/user/insiders_status/activate.rb
```diff
@@ -10,17 +10,22 @@ def call
...
+    return unless FeatureFlag::INSIDERS
+
     user.update(flair: :insider) unless %i[founder staff original_insider].include?(user.flair)
-    User::Notification::Create.(user, @notification_key) if FeatureFlag::INSIDERS
-    AwardBadgeJob.perform_later(user, :insider) if FeatureFlag::INSIDERS
-    User::SetDiscordRoles.(user)
-    User::SetDiscourseGroups.(user)
+    User::Notification::Create.(user, @notification_key)
+    AwardBadgeJob.perform_later(user, :insider)
+    AwardBadgeJob.perform_later(user, :lifetime_insider) if user.insiders_status_active_lifetime? && FeatureFlag::INSIDERS
```
ErikSchierboom: I think the feature flag test is always `true` here due to the preceding `return unless FeatureFlag::INSIDERS`

```suggestion
    AwardBadgeJob.perform_later(user, :lifetime_insider) if user.insiders_status_active_lifetime?
```

app/commands/user/insiders_status/activate.rb
```diff
@@ -10,17 +10,22 @@ def call
...
+    User::SetDiscordRoles.defer(user)
+    User::SetDiscourseGroups.defer(user)
```
ErikSchierboom, line 29: Deferring makes total sense. Should have come up with that myself

iHiD (approved):

PR #5345, opened 2023-05-04
Fix tabs in editor

iHiD:
The DOM is currently broken in the editor for panels - there's an extra incorrect wrapper div:
Screenshot 2023-05-04 at 13 51 50

This fixes it:
Screenshot 2023-05-04 at 13 50 18

And adds back a border that got lost:
Screenshot 2023-05-04 at 13 50 00
Screenshot 2023-05-04 at 13 50 04


iHiD:
I'm merging this. I recommend this is refactored to be better though.

PR #5367, opened 2023-05-05
Assign lifetime insiders flair

ErikSchierboom:
- Update lifetime_insider flair
- Add lifetime insiders flair


app/commands/user/insiders_status/update_flair.rb
```diff
@@ -0,0 +1,20 @@
+class User::InsidersStatus::UpdateFlair
```
iHiD:
```suggestion
class User::UpdateFlair
```

PR #5371, opened 2023-05-05
:lipstick: Fix community solution summary background in dark mode

homersimpsons:
I am not sure that this is the correct `@apply` as I do not have the project setup.

Before:
image

After:
image


homersimpsons:
(CI failure does not look related to this changes)

PR #5415, opened 2023-05-09
Add debugging to paypal

ErikSchierboom:


app/commands/webhooks/paypal/debug.rb
```diff
@@ -0,0 +1,7 @@
+class Webhooks::Paypal::Debug
+  include Mandate
+
+  initialize_with :message
+
+  def call = IO.write("/mnt/efs/debug/paypal.txt", message, mode: 'a')
```
iHiD: It needs to be `/mnt/efs/repos/debug` (with the `repos` in). The `repos` is the bit that gets mounted.

iHiD (approved):
Approved pending my comment.

PR #5441, opened 2023-05-11
Cache individual columns, not all the data

iHiD:
@ErikSchierboom This builds on your work to just cache individual columns, not everything. I think this would be best refactored into different commands rather than using the `key` but I want to do this so I can add an extra bit into the cache and it's much quicker for me to do this.

I've also refactored the tests to check the thing is actually cached (which I think you'll approve of).

This also adds Optimistic merging to user data, which I think is wise to have. It does make it a little risky to add at first though, as we might need to add some Pessimistic locks in place in places where we update things. However, the combination is really solid. https://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html

I'll build on this with another branch adding the extra key.

### DIFF
--- a/app/commands/mentor/testimonial/create.rb
+++ b/app/commands/mentor/testimonial/create.rb
@@ -10,6 +10,6 @@ class Mentor::Testimonial::Create
       discussion:,
       content: testimonial
     )
-    User::ResetCache.defer(discussion.mentor)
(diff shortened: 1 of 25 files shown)


ErikSchierboom (approved):
LGTM

PR #5442, opened 2023-05-11
Add extra cache keys

iHiD:
Builds on #5441. One of these queries takes 18s for bobahop sometimes and that's inline when he loads his profile using indexes... So this will help performance lots.

@ErikSchierboom Please sanity check and add tests for the reset cache bits.


app/models/user/profile.rb
```diff
@@ -20,4 +20,10 @@ def contributions_tab?
 
   memoize
   def badges_tab? = user.revealed_badges.exists?
+
+  memoize
+  def num_solutions_published = user.solutions.published.count
+
+  memoize
+  def num_testimonials = user.mentor_testimonials.published.count
```
iHiD: We could cache these on user_data too, but I'm working on the basis of one thing at a time!

app/commands/user/reset_cache.rb
```diff
@@ -24,4 +27,11 @@ def cache
   def value_for_has_unrevealed_testimonials? = user.mentor_testimonials.unrevealed.exists?
   def value_for_has_unrevealed_badges? = user.acquired_badges.unrevealed.exists?
   def value_for_has_unseen_reputation_tokens? = user.reputation_tokens.unseen.exists?
+  def value_for_num_solutions_mentored = user.mentor_discussions.count
+  def value_for_num_testimonials = user.mentor_testimonials.published.count
+
+  # This one is sloooooow!
```
ErikSchierboom, line 45: Maybe we could join on request instead of solution? Might be faster

iHiD: I considered that too. But discovered that `request` would have to join to `solution` to get the `user_id`.

ErikSchierboom: But couldn't we use the `student_id` field in `Mentor::Request`?

ErikSchierboom: I looked into this a bit, and found something I don't understood. How could there be a mentoring request for a solution where the request's student is different from the solution's user?

ErikSchierboom: Joining on request would be way faster (mentor ID redacted):

```
irb(main):028:0> user.mentor_discussions.joins(:request).distinct.count(:student_id)
D, [2023-05-18T12:06:37.624318 #71023] DEBUG -- :   Mentor::Discussion Count (4281.5ms)  SELECT COUNT(DISTINCT student_id) FROM `mentor_discussions` INNER JOIN `mentor_requests` ON `mentor_requests`.`id` = `mentor_discussions`.`request_id` WHERE `mentor_discussions`.`mentor_id` = X
=> 7642
irb(main):029:0> user.mentor_discussions.joins(:solution).distinct.count(:user_id)
D, [2023-05-18T12:07:13.302365 #71023] DEBUG -- :   Mentor::Discussion Count (19616.3ms)  SELECT COUNT(DISTINCT user_id) FROM `mentor_discussions` INNER JOIN `solutions` ON `solutions`.`id` = `mentor_discussions`.`solution_id` WHERE `mentor_discussions`.`mentor_id` = X
=> 7526
```

app/commands/user/reset_cache.rb
```diff
@@ -24,4 +27,11 @@ def cache
   def value_for_has_unrevealed_testimonials? = user.mentor_testimonials.unrevealed.exists?
   def value_for_has_unrevealed_badges? = user.acquired_badges.unrevealed.exists?
   def value_for_has_unseen_reputation_tokens? = user.reputation_tokens.unseen.exists?
+  def value_for_num_solutions_mentored = user.mentor_discussions.count
```
ErikSchierboom: There already is a `num_solutions_mentored` field on `User::Data`. How does this relate to that field?

iHiD: Interesting. I didn't know that. That should be removed then I think and replaced with this.

PR #5453, opened 2023-05-15
Unique key for corrupted data

dem4ron:
Ensures React keys are even more unique, providing additional protection in case of corrupted data.


ErikSchierboom (approved):
👍

PR #5456, opened 2023-05-17
Add premium

ErikSchierboom:
- Add product column to donation payment and subscription
- Rename Donations module to Payments
- Pass in product when creating subscription
- Pass in product when creating payment
- Add premium scope
- Award lifeime premium when activating insiders
- Add audience for premium users
- Add audience for insiders
- Use proper urls for insiders notifications
- Send notifications when joining premium
- Check expiration
- Work on updating premium
- Update factory name
- Add premium flair


app/commands/payments/payment/create.rb
```diff
@@ -0,0 +1,33 @@
...
+    Payments::Payment.create!(
+      user:,
+      provider:,
+      product:,
+      external_id:,
+      external_receipt_url:,
+      subscription:,
+      amount_in_cents:
+    ).tap do |payment|
+      User::UpdateTotalDonatedInCents.(user) if product == :donation
```
ErikSchierboom, line 21: This assumes that payments for premium don't count towards one's total donation amount

app/commands/payments/stripe/payment/create.rb
```diff
@@ -14,27 +14,38 @@ def initialize
...
   end
 
   private
   def external_id = stripe_data.id
   def external_receipt_url = stripe_data.charges.first.receipt_url
   def amount_in_cents = stripe_data.amount
 
+  def product
+    # TODO: is a payment intent guaranteed to have an invoice?
+    return :donation unless invoice
```
ErikSchierboom: It looks like a one-off payment does not have an invoice. After some googling, it looks like linking a product to a paymentintent is not supported, unless we pass in some metadata ourselves

ErikSchierboom:
The one failing test requires the `config` gem to be updated (PR is up, but still needs tweaking)

PR #5457, opened 2023-05-17
Add new site nav

dem4ron:
This PR adds a shell for a new site nav. Content can be added as a submenu-object with a submenu-view

Screenshot 2023-05-17 at 12 32 22


app/javascript/components/settings/theme-preference-form/utils.ts
```diff
@@ -16,7 +16,7 @@ export function setThemeClassName(theme: string): void {
   }
 }
 
-function grabCurrentTheme():
+export function grabCurrentTheme():
```
dem4ron, line 19: TODO: remove this fn as it is deprecated and replaced by `getCurrentTheme` from `hooks/use-theme-observer.tsx`

app/css/components/site-header.css
```diff
@@ -18,6 +18,11 @@ body:not(.namespace-.controller-insiders) {
     --exercism-link-filter: var(--textColor1-filter);
 }
 
+body.keyboard-navigation .nav-element:hover .nav-element-dropdown,
+body.keyboard-navigation .nav-element:hover .arrow {
+    display: none !important;
```
iHiD, line 23: I believe you can do `@apply !hide`; or something if we want to use Tailwind here,

app/helpers/react_components/common/theme_toggle_button.rb
```diff
@@ -1,8 +1,17 @@
 module ReactComponents
   module Common
     class ThemeToggleButton < ReactComponent
+      initialize_with :disabled
```
iHiD, line 4: Normally it's best to have positive names for variables. So this might be better as `:enabled`. You might also want to do `enabled: true` (or `enabled: false`) as a default.

app/helpers/view_components/site_header.rb
```diff
@@ -53,22 +54,27 @@ def signed_in_section
     end
 
     def signed_in_nav
```
iHiD, line 56: Let's use Rails Cache to cache the HTML for this so we don't have to render it on every view (as it's going to be expensive).

You can wrap the content of this method in something like:
```
Rails.cache.fetch("top-nav-signed-in") do 
...
end
```

Check if we use any sort of naming schema for Rails cache by looking at other places in the UI

iHiD: You also need to make sure that any variables are part of the cache key (e.g. `current_user.insiders_status`).

app/helpers/view_components/theme_toggle_button.rb
```diff
@@ -0,0 +1,12 @@
+module ViewComponents
+  module ThemeToggleButton
+    def toggle_button
+      tag.div class: 'toggle-button' do
```
iHiD, line 4: Should this be `c-toggle-button`?

iHiD:
I've merged this into #5468 so we can get it all working together :)

PR #5467, opened 2023-05-24
All the Premium Stuff Except the Renames.

iHiD:


app/commands/payments/github/sponsorship/handle_created.rb
```diff
@@ -0,0 +1,14 @@
...
+class Payments::Github::Sponsorship::HandleCreated
+  include Mandate
+
+  queue_as :default
+
+  initialize_with :user, :node_id, :is_one_time, :monthly_price_in_cents
+
+  def call
+    subscription = Payments::Github::Subscription::Create.(user, node_id, monthly_price_in_cents) unless is_one_time
+    Payments::Github::Payment::Create.(user, node_id, monthly_price_in_cents, subscription:)
```
iHiD: How do you find the fact that subscription is nil despite never being defined or in scope on this line. I semi-like it and semi find it very weird in Ruby! (The code is idiomatic though!)

app/commands/payments/stripe/payment/create.rb
```diff
@@ -14,14 +14,21 @@ def initialize
...
 
   private
   def external_id = stripe_data.id
   def external_receipt_url = stripe_data.charges.first.receipt_url
   def amount_in_cents = stripe_data.amount
 
+  def product
+    return subscription.product if subscription
+
+    # Premium payments are always linked to a subscription, even one-off payments
```
iHiD:
```suggestion
    # Premium payments are always linked to a subscription
```

As we no-longer have one-off product payments?

app/commands/payments/subscription/create.rb
```diff
@@ -2,22 +2,25 @@
...
       provider:,
+      product:,
+      interval:,
       external_id:,
       amount_in_cents:,
       status: :active
     ).tap do
-      User::UpdateActiveDonationSubscription.(user)
+      User::UpdateActiveDonationSubscription.(user) if product == :donation
+      User::Premium::Update.(user) if product == :premium
```
iHiD: Maybe a case on `product` or something here, just as they're mutually exclusive? I'm not fussed though.

app/commands/user/insiders_status/activate.rb
```diff
@@ -19,6 +19,7 @@ def call
 
     User::UpdateFlair.(user)
     User::Notification::CreateEmailOnly.defer(user, @notification_key)
+    User::Premium::Update.defer(user)
```
iHiD:
```suggestion
    User::Premium::Update.(user)
```

Move this above and undefer it as we want this happening in-band?

app/commands/user/insiders_status/determine_eligibility_status.rb
```diff
@@ -37,6 +39,13 @@ def period_reputation(period)
...
-    user.donation_payments.where('created_at < ?', Date.new(2023, 5, 1)).sum(:amount_in_cents)
+    user.payment_payments.donation.where('created_at < ?', LAUNCH_DATE).sum(:amount_in_cents)
   end
+
+  def lifetime_premium_donation?
+    user.payment_payments.where('amount_in_cents >= ?', LIFETIME_PREMIUM_AMOUNT).exists?
+  end
+
+  # TODO: update launch date with actual launch date
+  LAUNCH_DATE = Date.new(2023, 5, 1).freeze
```
iHiD:
```suggestion
  LAUNCH_DATE = Date.new(2023, 5, 30).freeze
```

app/commands/user/insiders_status/trigger_update.rb
```diff
@@ -11,6 +11,6 @@ def call
       user.update!(insiders_status: :unset) if user.insiders_status_ineligible?
     end
 
-    User::InsidersStatus::Update.defer(user)
+    User::InsidersStatus::Update.(user)
```
iHiD: I think this one should generally be `defer` actually. But we might just an exception for the donation route. If you revert this for now, I'll work it out when I test everything e2e properly.

app/commands/user/insiders_status/update.rb
```diff
@@ -27,6 +27,7 @@ def call
     User::SetDiscordRoles.defer(user)
     User::SetDiscourseGroups.defer(user)
     User::Notification::CreateEmailOnly.defer(user, @notification_key) if @notification_key
+    User::Premium::Update.defer(user)
```
iHiD: This is the case where someone **loses** Insiders + Premium, right?

app/commands/user/premium/expire.rb
```diff
@@ -0,0 +1,7 @@
+class User::Premium::Expire
+  include Mandate
+
+  initialize_with :user
+
+  def call = User::Notification::CreateEmailOnly.defer(user, :expired_premium)
```
iHiD: Do we call this anywhere yet? I'm guessing not?

app/commands/user/premium/join.rb
```diff
@@ -0,0 +1,10 @@
+class User::Premium::Join
+  include Mandate
+
+  initialize_with :user, :premium_until
+
+  def call
+    user.update!(premium_until:)
+    User::Notification::CreateEmailOnly.defer(user, :joined_premium) if FeatureFlag::PREMIUM
+  end
+end
```
iHiD, line 19: This is to be called after someone completes the payment? I can't see where it's called atm (but struggling with the massive diff)

app/commands/user/premium/update.rb
```diff
@@ -0,0 +1,48 @@
...
+    return last_payment_premium_until if last_payment_premium_until.present?
+
+    nil
+  end
+
+  def lifetime_premium? = user.insider?
+
+  memoize
+  def last_payment_premium_until
+    return nil if last_payment.nil?
```
iHiD, line 35:
```suggestion
```

The line below covers this case.

app/commands/user/premium/update.rb
```diff
@@ -0,0 +1,48 @@
...
+  memoize
+  def last_payment_grace_period
+    return nil if last_payment.nil?
+    return nil if last_payment.subscription.nil?
+
+    return 45.days if last_payment.subscription.active?
+    return 45.days if last_payment.subscription.overdue?
+
+    30.days
+  end
```
iHiD: We need to pivot here on monthly/annual. (Either 30 days or 365 days (+15))

app/commands/payments/github/sponsorship/handle_created.rb
ErikSchierboom: I'm okay with it. It looked idiomatic. I could also create a `subscription` method and be more explicit about it?

app/commands/payments/payment/create.rb
```diff
@@ -0,0 +1,33 @@
...
+      external_id:,
+      external_receipt_url:,
+      subscription:,
+      amount_in_cents:
+    ).tap do |payment|
+      User::UpdateTotalDonatedInCents.(user) if product == :donation
+
+      # TODO: the donation guard means that a premium user doesn't get the supporter badge
+      # We probably do want that, right?
+      User::RegisterAsDonor.(user, Time.current) if product == :donation
```
ErikSchierboom: @iHiD Should premium users get the supporter badge?

app/commands/payments/github/sponsorship/handle_created.rb
ErikSchierboom: I did end up refactoring it

app/commands/user/insiders_status/update.rb
ErikSchierboom: Correct

app/commands/user/premium/join.rb
ErikSchierboom: I had still to add that. My bad. Just did that.

app/commands/user/premium/expire.rb
ErikSchierboom: We don't yet. We do now :)

app/commands/payments/payment/create.rb
iHiD: Not for their premium subscription payments.

app/models/user_track/mentoring_slots.rb
```diff
@@ -7,6 +7,8 @@ def has_available_mentoring_slot? = num_available_mentoring_slots.positive?
 
   memoize
   def num_locked_mentoring_slots
+    return 0 if user.premium?
```
iHiD, line 10: Nice.

db/migrate/20230516122816_add_premium_until_to_user_data.rb
```diff
@@ -0,0 +1,8 @@
+class AddPremiumUntilToUserData < ActiveRecord::Migration[7.0]
+  def change
+    return if Rails.env.production?
+
+    add_column :user_data, :premium_until, :datetime, null: true
+    add_index :user_data, [:premium_until], if_not_exists: true
```
iHiD:
```suggestion
    add_index :user_data, :premium_until, if_not_exists: true
```

app/commands/payments/payment/update_amount.rb
```diff
@@ -8,6 +8,7 @@ def call
     return unless payment
 
     payment.update!(amount_in_cents:)
-    User::UpdateTotalDonatedInCents.(payment.user)
+    User::UpdateTotalDonatedInCents.(payment.user) if payment.donation?
+    User::Premium::Update.(user) if payment.premium?
```
iHiD:
```suggestion
    if payment.donation?
      User::UpdateTotalDonatedInCents.(payment.user) 
    elsif payment.premium?
      User::Premium::Update.(user) 
    end
```

I just find this a bit clearer.

app/commands/payments/paypal/interval.rb
```diff
@@ -0,0 +1,14 @@
...
+    case payment_cycle
+    when 'Monthly'
+      :month
+    when 'Yearly'
+      :year
+    else
+      raise UnsupportedPaypalIntervalError
+    end
+  end
+end
```
iHiD: I'd rather this either: 
1. Followed our command naming. So module `Payments::Paypal::Interval::CycleToString.(...)`
2. Was a model in app/models/payments/paypal/interval. Probably one that was initialize_with :payment_cycle then `from_payment_cycle` was `to_s`. Or something.

This doesn't block merging to be clear, but I think it would be good to tidy/refactor to use an existing pattern.

app/commands/payments/paypal/payment/handle_web_accept.rb
```diff
@@ -42,4 +44,5 @@ def external_id = payload["txn_id"]
   def payment_status = payload["payment_status"]
   def payer_id = payload["payer_id"]
   def payer_email = payload["payer_email"]
+  def product = :donation
```
iHiD, line 47: Do we know this is a donation because `web_accept` is always a donation?

app/commands/payments/paypal/product.rb
```diff
@@ -0,0 +1,14 @@
...
+    case name
+    when Exercism.secrets.paypal_donation_product_name
+      :donation
+    when Exercism.secrets.paypal_premium_product_name
+      :premium
+    else
+      raise UnknownPaypalProductError
+    end
+  end
+end
```
iHiD: Same comment as on Interval.

iHiD: The other option, which maybe nicer/simpler, would be to have these as `self.product_from_name` on `Payments::Paypal` instead.

app/commands/payments/stripe/price.rb
```diff
@@ -0,0 +1,34 @@
+module Payments::Stripe::Price
+  class UnknownStripeIntervalError < RuntimeError; end
+  class UnknownStripePriceError < RuntimeError; end
+
+  MONTH_AMOUNT_IN_CENTS = 999
+  YEAR_AMOUNT_IN_CENTS = 9999
+  LIFETIME_AMOUNT_IN_CENTS = 49_900
```
iHiD:
```suggestion
  MONTH_AMOUNT_IN_CENTS = 9_99
  YEAR_AMOUNT_IN_CENTS = 99_99
  LIFETIME_AMOUNT_IN_CENTS = 499_00
```

If rubocop tried to "fix" this, tell it to fuck off

app/commands/payments/stripe/subscription/create.rb
```diff
@@ -3,13 +3,28 @@
...
 
-  def call = Payments::Subscription::Create.(user, :stripe, external_id, amount_in_cents)
+  def call = Payments::Subscription::Create.(user, :stripe, product, interval, external_id, amount_in_cents)
 
   private
   def external_id = stripe_data.id
-  def amount_in_cents = stripe_data.items.data[0].price.unit_amount
+  def price = stripe_data.items.data[0].price
+  def amount_in_cents = price.unit_amount
+  def product = Payments::Stripe::Product.from_product_id(price.product)
```
iHiD:
```suggestion
  def amount_in_cents = price.unit_amount
  def product = Payments::Stripe::Product.from_product_id(price.product)
  def price = stripe_data.items.data[0].price
```

app/commands/payments/subscription/update_status.rb
```diff
@@ -0,0 +1,11 @@
+class Payments::Subscription::UpdateStatus
+  include Mandate
+
+  initialize_with :subscription, :status
+
+  def call
+    subscription.update!(status:)
+    User::UpdateActiveDonationSubscription.(subscription.user) if subscription.donation?
+    User::Premium::Update.(subscription.user) if subscription.premium?
```
iHiD: Switch to if/elseif here?

app/commands/user/insiders_status/determine_eligibility_status.rb
```diff
@@ -2,7 +2,8 @@ class User::InsidersStatus::DetermineEligibilityStatus
   include Mandate
 
   LIFETIME_REPUTATION_THRESHOLD = 1_000
-  LIFETIME_DONATIONS_THRESHOLD = 500_00
+  LIFETIME_DONATIONS_THRESHOLD = 499_00
+  LIFETIME_PREMIUM_AMOUNT = 499_00
```
iHiD: This should be unified with the constant in `Payments::Stripe::Price` to be just one constant.

app/commands/user/insiders_status/determine_eligibility_status.rb
```diff
@@ -37,6 +39,12 @@ def period_reputation(period)
...
   def prelaunch_donation_total
-    user.donation_payments.where('created_at < ?', Date.new(2023, 5, 1)).sum(:amount_in_cents)
+    user.payments.donation.where('created_at < ?', LAUNCH_DATE).sum(:amount_in_cents)
   end
+
+  def lifetime_premium_donation?
+    user.payments.where('amount_in_cents >= ?', LIFETIME_PREMIUM_AMOUNT).exists?
+  end
+
+  LAUNCH_DATE = Date.new(2023, 5, 30).freeze
```
iHiD:
```suggestion
  LAUNCH_DATE = Date.new(2023, 5, 31).freeze
```

app/commands/user/insiders_status/update.rb
```diff
@@ -24,14 +24,20 @@ def call
       end
     end
 
+    # These things are being called for when someone
+    # LOSES Insiders access, or when they change from normal,
```
iHiD, line 28:
```suggestion
    # LOSES Insiders access, or when they CHANGE from normal,
```

app/commands/user/insiders_status/update.rb
```diff
@@ -24,14 +24,20 @@ def call
...
+    # to lifetime. This code is not called when they GAIN Insiders.
+    # Look at Activate.() for that instead.
```
iHiD:
```suggestion
    # These things are being called for when someone:
    # - LOSES Insiders access; or
    # - UPGRADES from normal to lifetime
    # 
    # This code is not called when they GAIN Insiders.
    # Look at Activate.() for that instead.
```

app/commands/user/premium/join.rb
```diff
@@ -0,0 +1,18 @@
...
+    update_theme!
+    User::Notification::CreateEmailOnly.defer(user, :joined_premium) if FeatureFlag::PREMIUM
+  end
+
+  def update_theme!
+    existing_theme = user.preferences.theme
+    return unless existing_theme.nil? || existing_theme == 'light'
+
+    user.preferences.update!(theme: :dark)
+  end
```
iHiD, line 18: This needs a test adding.

app/commands/user/premium/update.rb
```diff
@@ -0,0 +1,43 @@
...
+  include Mandate
+
+  initialize_with :user
+
+  def call
+    return if user.premium_until == premium_until
+
+    if expire?
+      User::Premium::Expire.(user)
+    else
```
iHiD: Should this be:
```suggestion
    elsif user.premium_until.nil?
```

app/commands/user/premium/update.rb
```diff
@@ -0,0 +1,43 @@
...
+      User::Premium::Join.(user, premium_until)
+    end
+  end
+
+  private
+  def premium_until
```
iHiD, line 22:
```suggestion
  memoize
  def premium_until
```

app/helpers/react_components/premium/paypal_status.rb
```diff
@@ -0,0 +1,4 @@
+class ReactComponents::Premium::PaypalStatus < ReactComponents::ReactComponent
+  def to_s = super("premium-paypal-status", { endpoint: Exercism::Routes.api_user_path,
+                                              premium_redirect_path: Exercism::Routes.premium_path })
+end
```
iHiD, line 8: I'd rather this was a normal method, not an inlined one.

app/helpers/view_components/handle_with_flair.rb
```diff
@@ -59,13 +59,15 @@ def ml_in_px
...
+      premium: 'premium',
       insider: 'insiders',
       lifetime_insider: 'lifetime-insiders',
       founder: 'exercism-face-gradient',
       staff: 'exercism-face-gradient'
     }.freeze
 
     TITLES = {
+      premium: 'Premium User',
       insider: 'An Insider',
```
iHiD:
```suggestion
      insider: 'An Exercism Insider',
```

app/models/payments/payment.rb
```diff
@@ -15,7 +15,7 @@ def email_communication_preferences_key; end
 
   def amount_in_dollars = amount_in_cents / BigDecimal(100)
   def self.total_supporters = select(:user_id).distinct.count
-  def self.total_donated_in_dollars = sum(:amount_in_cents) / BigDecimal(100)
+  def self.total_donated_in_dollars = donation.sum(:amount_in_cents) / BigDecimal(100)
```
iHiD: This looks wrong to me (singular.sum)

app/models/user.rb
```diff
@@ -131,6 +132,8 @@ class User < ApplicationRecord
 
   scope :with_data, -> { joins(:data) }
   scope :donor, -> { with_data.where.not(user_data: { first_donated_at: nil }) }
```
iHiD: This should be plural.

app/models/user.rb
```diff
@@ -131,6 +132,8 @@ class User < ApplicationRecord
...
+  scope :premium, -> { with_data.where('user_data.premium_until > ?', Time.current) }
+  scope :insiders, -> { with_data.where(user_data: { insiders_status: %i[active active_lifetime] }) }
```
iHiD, line 136: We need indexes on both of these columns.

app/commands/user/premium/update.rb
ErikSchierboom: No, because of the "sliding window" aspect of it. If a new payment was made, the `premium_until` time needs to be updated. And the preceding `return if user.premium_until == premium_until` prevents unnecessary updating.

app/commands/user/premium/join.rb
ErikSchierboom: I guess we also need to do the inverse when expiring?

app/models/user.rb
ErikSchierboom: They were already there:

https://github.com/exercism/website/blob/main/db/migrate/20230412075814_add_insiders_status_to_users.rb#L6
https://github.com/exercism/website/pull/5467/files#diff-b12c7e45b80830692065f74867506a622dbe92f05ac9099de2ff63015d6808b6R6

app/commands/payments/paypal/payment/handle_web_accept.rb
ErikSchierboom: Yep. Otherwise it is a `handle_recurring_payment`

app/commands/payments/paypal/product.rb
ErikSchierboom: I've refactored things a bit

app/commands/user/insiders_status/determine_eligibility_status.rb
ErikSchierboom: We should unify, but this is not stripe specific though (it also holds for paypal donations). Any ideas for a better place?

app/commands/payments/subscription/create.rb
```diff
@@ -22,6 +22,7 @@ def call
         User::UpdateActiveDonationSubscription.(user)
       when :premium
         User::Premium::Update.(user)
+        User::InsidersStatus::Update.(user)
```
iHiD: I don't understand this line. Creating a Premium subscription shouldn't affect someone's insider status should it?

app/commands/user/premium/update.rb
iHiD: But I don't understand why we call `Join` a second time in that case. Shouldn't we be calling something to update `premium_until` instead?

app/commands/user/premium/join.rb
iHiD: Yes, true.

app/commands/payments/subscription/create.rb
ErikSchierboom: No that line can be removed. It was a test for the flair bit

app/commands/user/premium/update.rb
ErikSchierboom: That's basically what `Join` does

PR #5470, opened 2023-05-24
Check email status

ErikSchierboom:
- Add email_status column to user_data
- Don't send emails when email_status is invalid
- Remove unneeded .keep file
- Update config gem
- Add command to verify email
- Verify email

### DIFF
--- a/app/commands/user/verify_email.rb
+++ b/app/commands/user/verify_email.rb
@@ -1,0 +1,27 @@
+class User::VerifyEmail
+  include Mandate
+
+  initialize_with :user
+
+  def call = user.update(email_status:)
+
+  private
+  def email_status
+    case check_email_status!
+    when 'valid'
+      :verified
+    when 'neutral'
+      :verified
+    else
+      :invalid
+    end
+  end
+
+  memoize
+  def check_email_status!
+    response = JSON.parse(RestClient.get(api_url).body, symbolize_names: true)
+    response.dig(:results, :result)
+  end
+
+  def api_url = "https://api.sparkpost.com/api/v1/recipient-validation/single/#{user.email}"
+end
--- a/app/commands/user/verify_email.rb
+++ b/app/commands/user/verify_email.rb
@@ -3,7 +3,7 @@ class User::VerifyEmail
 
   initialize_with :user
 
-  def call = user.update(email_status:)
+  def call = user.update!(email_status:)
 
   private
   def email_status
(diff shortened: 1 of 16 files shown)


app/commands/user/verify_email.rb
```diff
@@ -0,0 +1,27 @@
+class User::VerifyEmail
+  include Mandate
+
+  initialize_with :user
+
+  def call = user.update(email_status:)
```
iHiD:
```suggestion
  def call = user.update!(email_status:)
```

app/mailers/application_mailer.rb
```diff
@@ -15,6 +15,8 @@ def user_email_with_name(user)
   end
 
   def mail_to_user(user, subject, **options)
+    return if user.email_status_invalid?
```
iHiD: This would be better in `app/commands/user/send_email.rb` alongside `users.noreply.github.com` as that will correctly set the status of the email to `skipped` not `sent`.

app/models/user.rb
```diff
@@ -156,6 +156,10 @@ class User < ApplicationRecord
     after_confirmation if confirmed?
   end
 
+  after_save_commit do
+    User::VerifyEmail.defer(self) if previous_changes.key?('email')
```
iHiD: Nice.

app/mailers/application_mailer.rb
iHiD: I've done this :)

ErikSchierboom: Ah, that's much better. I'm still fairly new to the email sending and how that works.

app/models/user.rb
```diff
@@ -154,6 +154,11 @@ class User < ApplicationRecord
     create_communication_preferences
 
     after_confirmation if confirmed?
+    verify_email! if email_status_unverified?
```
ErikSchierboom: This is a little trick to ensure that none of the existing unit tests break, otherwise we'd have to stub the verification call in any unit tests that creates a user and runs background jobs

test/factories/users.rb
```diff
@@ -16,6 +16,10 @@
       Rack::Test::UploadedFile.new(tempfile.path, 'image/png')
     end
 
+    after(:build) do |user|
+      user.data.email_status = :verified
+    end
```
ErikSchierboom, line 21: Setting `email_status { :verified }` in the `User::Data` factory didn't work, so 🤷 
I had a nicer solution, but that broke some other tests.

app/models/user.rb
iHiD: We could also just set verified to true in the main user factory?

iHiD (approved):

iHiD (approved):

PR #5476, opened 2023-05-25
Dark-only premium/insiders site-header

dem4ron:
##### Everything inside a `view` is rendered with its theme-dark variant
Screenshot 2023-05-25 at 22 29 02

### DIFF
--- a/app/javascript/packs/application.tsx
+++ b/app/javascript/packs/application.tsx
@@ -483,6 +483,11 @@ import { PriceOption, PriceOptionProps } from '@/components/premium/PriceOption'
 
 document.addEventListener('turbo:load', () => {
   highlightAll()
+
+  if (window.location.pathname === '/dashboard') {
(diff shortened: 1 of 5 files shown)


app/javascript/packs/application.tsx
```diff
@@ -483,6 +483,11 @@ import { PriceOption, PriceOptionProps } from '@/components/premium/PriceOption'
 
 document.addEventListener('turbo:load', () => {
   highlightAll()
+
+  if (window.location.pathname === '/dashboard') {
+    const siteHeader = document.querySelector('#site-header')
+    siteHeader?.classList.remove('theme-dark')
+  }
```
iHiD, line 490: I don't get what this does. What's special about the dashboard?

dem4ron: I think some caching goes on, because when I logged in request.path in the ruby console, nothing appeared when I visited dashboard, but worked fine for the other paths.

PR #5489, opened 2023-05-29
Add published at to videos

iHiD:
We'll use this to order

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_05_02_172935) do
+ActiveRecord::Schema[7.0].define(version: 2023_05_29_130106) do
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -164,6 +166,7 @@
     t.datetime "updated_at", null: false
     t.integer "email_status", limit: 1, default: 0, null: false
     t.integer "provider", limit: 1, default: 0, null: false
+    t.integer "product", limit: 1, default: 0, null: false
```
iHiD:
```suggestion
```

db/schema.rb
```diff
@@ -178,6 +181,8 @@
     t.integer "email_status", limit: 1, default: 0, null: false
     t.integer "status", limit: 1, default: 0, null: false
     t.integer "provider", limit: 1, default: 0, null: false
+    t.integer "product", limit: 1, default: 0, null: false
+    t.integer "interval", limit: 1, default: 0, null: false
```
iHiD:
```suggestion
```

ErikSchierboom (approved):

PR #5519, opened 2023-06-01
Create premium PayPal subscriptions via API

ErikSchierboom:
- [ ] Change IPN URL to add `webhooks/paypal/ipn`
- [ ] Setup webhooks URL to `webhooks/paypal`
- [x] Update copy of `app/views/premium/paypal_cancelled.haml`
- [x] Ensure the following secrets are defined:
  - [x] `stripe_premium_product_id`
  - [x] `stripe_premium_monthly_price_id`
  - [x] `stripe_premium_yearly_price_id`
  - [x] `stripe_premium_lifetime_price_id`
  - [x] `paypal_donation_product_name`
  - [x] `paypal_donation_hosted_button_id`
  - [x] `paypal_premium_product_name`
  - [x] `paypal_premium_monthly_plan_id`
  - [x] `paypal_premium_yearly_plan_id`
  - [x] `paypal_lifetime_insiders_hosted_button_id`
- [x] Ensure the following config values are defined:
  - [x] `paypal_api_url`
  - [x] `paypal_url`


app/commands/payments/paypal.rb
ErikSchierboom, line 1: This was moved to models (commands is the wrong place I think)

app/models/payments/subscription.rb
```diff
@@ -3,7 +3,7 @@ class Payments::Subscription < ApplicationRecord
 
   has_many :payments, dependent: :destroy
 
-  enum status: { canceled: 0, overdue: 1, active: 2 }
+  enum status: { canceled: 0, overdue: 1, active: 2, pending: 3 }
```
ErikSchierboom, line 6: We probably need to think if we need to account for `pending` anywhere. I've at least accounted for it in the premium determination, where pending subscriptions won't count.

app/models/premium.rb
```diff
@@ -0,0 +1,39 @@
...
+  MONTH_AMOUNT_IN_DOLLARS = MONTH_AMOUNT_IN_CENTS / 100.0
+  YEAR_AMOUNT_IN_DOLLARS = YEAR_AMOUNT_IN_CENTS / 100.0
+  LIFETIME_AMOUNT_IN_DOLLARS = LIFETIME_AMOUNT_IN_CENTS / 100.0
+
+  def self.amount_in_cents_from_interval(interval)
+    case interval.to_sym
+    when :month
+      MONTH_AMOUNT_IN_CENTS
+    when :monthly
+      MONTH_AMOUNT_IN_CENTS
```
ErikSchierboom, line 15: The `month/monthly` bit is due to PayPal using both month and monthly for intervals.

app/views/donations/_alternatives.html.haml
```diff
@@ -2,7 +2,7 @@
   %h2.text-h5.mb-4 Looking for another payment provider?
   %p.text-p-base.mb-8 We currently support Paypal and GitHub Sponsors as alternative payment providers. Please contact #{link_to 'Loretta', 'mailto:loretta@exercism.org'} if there is another provider you feel we should add.
   .flex.flex-col.items-start.gap-12
-    = link_to "https://www.paypal.com/donate/?hosted_button_id=KDRW2GL6FVVFS", class: "btn-m btn-secondary" do
+    = link_to "#{Exercism.config.paypal_url}/donate/?hosted_button_id=#{Exercism.secrets.paypal_donation_hosted_button_id}&custom=#{current_user&.email}", class: "btn-m btn-secondary" do
```
ErikSchierboom: This adds the user in the custom field for the regular donation button. I've chosen the user's email instead of their ID/uuid to not expose that information. Let me know if you'd like this to be changed.

app/views/premium/paypal_cancelled.haml
ErikSchierboom, line 1: @iHiD TODO for you: come up with some sensible content. This page is shown when someone cancels a PayPal payment , which means that the payment/subscription isn't actually created.

config/routes.rb
```diff
@@ -32,7 +32,9 @@
   # ######## #
   namespace :webhooks do
     resource :stripe, only: [:create], controller: "stripe"
-    resource :paypal, only: [:create], controller: "paypal"
+    resource :paypal, only: [:create], controller: "paypal" do
+      post :ipn
```
ErikSchierboom, line 36: I've changed the routing pattern. The IPN events will be at `webhooks/paypal/ipn`.
Currently they're at `webhooks/paypal`, but I felt it might make more sense to have `webhooks/paypal` be used for paypal webhook events.

app/controllers/api/payments/subscriptions_controller.rb
```diff
@@ -41,4 +41,16 @@ def update_plan
       }
     }
   end
+
+  def create_paypal_premium
+    subscription = ::Payments::Paypal::Subscription::CreateForPremium.(current_user, params[:interval])
```
ErikSchierboom, line 46: What should be do if no data is returned? That would mean there is a paypal error somehow.

app/controllers/webhooks/base_controller.rb
```diff
@@ -11,6 +11,7 @@ class BaseController < ApplicationController
     skip_before_action :authenticate_user!
     skip_after_action :set_body_class_header
     skip_around_action :mark_notifications_as_read!
+    skip_after_action :updated_last_visited_on!
```
ErikSchierboom, line 14: I also don't think we want this for webhooks

app/commands/payments/paypal/subscription/create.rb
```diff
@@ -3,9 +3,9 @@
 class Payments::Paypal::Subscription::Create
   include Mandate
 
-  initialize_with :user, :id, :amount, :product, :interval
+  initialize_with :user, :id, :amount, :product, :interval, status: :active
```
ErikSchierboom, line 6: The added `status` field is there to be able to initialize a subscription with the `pending` status, which we need when manually initiating a paypal subscription.

app/controllers/api/base_controller.rb
```diff
@@ -9,6 +9,7 @@ class BaseController < ApplicationController
     skip_before_action :verify_authenticity_token
     skip_after_action :set_body_class_header
     skip_around_action :mark_notifications_as_read!
+    skip_after_action :updated_last_visited_on!
```
ErikSchierboom, line 12: I assume we don't want this for API requests? Or do we?

app/commands/payments/paypal/request_access_token.rb
```diff
@@ -0,0 +1,37 @@
...
+  CACHE_KEY = :paypal_access_token
+
+  def call
+    access_token = Rails.cache.read(CACHE_KEY)
+    return access_token if access_token.present?
+
+    token = request_token!
+    token[:access_token].tap do |new_access_token|
+      expires_in = token[:expires_in].to_i - EXPIRY_TIME_OFFSET
+      Rails.cache.write(CACHE_KEY, new_access_token, expires_in:)
```
iHiD: I'm not sure our production cache supports `expires_in`. Is it needed?

app/commands/payments/paypal/request_access_token.rb
```diff
@@ -0,0 +1,37 @@
+class Payments::Paypal::RequestAccessToken
+  include Mandate
+
+  CACHE_KEY = :paypal_access_token
```
iHiD:
```suggestion
  CACHE_KEY = "payments/paypal_access_token".freeze
```

iHiD: Probably also move this to be a private constant?

test/commands/payments/paypal/request_access_token_test.rb
```diff
@@ -0,0 +1,61 @@
+require_relative '../test_base'
+
+class Payments::Paypal::RequestAccessTokenTest < Payments::TestBase
+  def setup = Rails.cache.clear(Payments::Paypal::RequestAccessToken::CACHE_KEY)
```
iHiD: You could just clear the whole cache here rather than exposing the cache key. It should really be clearerd for each test anyway!

app/commands/payments/paypal/request_access_token.rb
ErikSchierboom: I hadn't consider that our Redis cache provider might not support this. 
And it doesn't: https://edgeapi.rubyonrails.org/classes/ActiveSupport/Cache/RedisCacheStore.html

It is not technically needed, we'll just need to request an access token dynamically. It doesn't really influence the flow much at the moment, but we'll be making two request access token calls in the process of creating a paypal subscription instead of one.

app/controllers/api/base_controller.rb
iHiD: I think we want to skip.

app/controllers/api/payments/subscriptions_controller.rb
iHiD: That sort of depends on what is hitting this API endpoint. What's it used for?

app/commands/payments/paypal/request_access_token.rb
iHiD: I really meant, do we need the expires in. Can we just leave it in the cache and leave it to auto-expire.

ErikSchierboom: When would it auto-expire? We detect the error response?

kytrinyx (approved):

kytrinyx (approved):

kytrinyx (approved):

kytrinyx (approved):

PR #5523, opened 2023-06-01
Fix number of featured exercises for #12in23

vaeng:
The total number of featured exercises was not updated with the June update.


iHiD:
Thanks! (@ErikSchierboom we should generate this number :))

PR #5526, opened 2023-06-02
Add course to track header caching

ErikSchierboom:


app/helpers/view_components/track/header.rb
```diff
@@ -96,11 +97,11 @@ def guard!
       end
 
       def practice_mode = !!user_track&.practice_mode?
+      def external = !!user_track&.external?
+      def course = !!user_track&.course?
```
iHiD, line 101: Small nit - can you make these end in `?` pls.

ErikSchierboom: I didn't want to because then I couldn't use the `external:` shorthand. But I'll update

iHiD: Ah, ok. Fine. Let's leave it as it is then.

PR #5542, opened 2023-06-06
Inline spinning animations

dem4ron:
- Add couple inline animate classes
- Update spinners


iHiD:
@dem4ron Any reason not to merge this? (It's still a draft atm)

PR #5548, opened 2023-06-06
Improve Automation tables

dem4ron:
Screenshot 2023-06-13 at 11 29 55


app/css/components/mentor/representer-row.css
```diff
@@ -67,14 +67,15 @@
...
+
+        .--occurencies {
+            @apply mr-32;
         }
 
         & .--feedback-glimpse {
             @apply text-right mr-60 leading-150 text-textColor6;
             @apply truncate;
-            width: 400px;
+            width: 300px;
```
iHiD, line 78: Careful not to edit the CSS here that also edits it in the main table.

dem4ron: I wanted to edit the width on all tables. I've checked it on main table too, and it looked fine. I think it was just a bit too wide overall.

app/serializers/serialize_exercise_representations.rb
```diff
@@ -32,6 +32,9 @@ def call
         },
         num_submissions: representation.num_submissions,
         appears_frequently: representation.appears_frequently?,
+        author: User.where(id: representation.feedback_author_id).pick(:handle),
+        editor: User.where(id: representation.feedback_editor_id).pick(:handle),
```
ErikSchierboom: I would change this to:

```suggestion
        author: representation.feedback_author.handle,
        editor: representation.feedback_editor.handle,
```

Your version will do two queries for each representation, whereas the suggested code could use `.includes` (you should get a warning) or `joins`

app/serializers/serialize_exercise_representations.rb
```diff
@@ -20,6 +20,8 @@ class SerializeRepresentation
     delegate :exercise, :track, to: :representation
 
     def call
+      user_ids = [representation.feedback_author_id, representation.feedback_editor_id].compact
+      user_handles = user_ids.present? ? User.where(id: user_ids).pluck(:id, :handle).to_h : {}
```
ErikSchierboom: This has the same issue. You're still doing a query for each representation. Any reason why you didn't use my suggestion?

app/helpers/react_components/mentoring/representations/admin.rb
```diff
@@ -80,8 +80,8 @@ def links
         def is_introducer_hidden = mentor.introducer_dismissed?(INTRODUCER_SLUG)
 
         SORT_OPTIONS = [
-          { value: :most_submissions, label: 'Sort by highest occurance' },
-          { value: :most_recent, label: 'Sort by recent first' }
+          { value: :feedback_added_at_desc, label: 'Most recent feedbacks' },
+          { value: :feedback_added_at_asc, label: 'Least recent feedbacks' }
```
ErikSchierboom: Maybe do:

```suggestion
          { value: :most_recent_feedback, label: 'Most recent feedbacks' },
          { value: :least_recent_feedback, label: 'Least recent feedbacks' }
```

I feel like we sorting option should be more "high-level"-ish, less about the actual columns used

app/assemblers/assemble_representation_tracks_for_select.rb
```diff
@@ -20,9 +20,16 @@ def supermentored_tracks
     Track.where(id: mentor.track_mentorships.supermentor_frequency.select(:track_id)).order(title: :asc)
   end
 
+  def current_mentor
+    return nil if is_admin
+
+    mentor
+  end
```
ErikSchierboom: I'm not sure about the naming. How is `current_mentor` different from mentor? Maybe it might make sense to inline this inside the `track_num_representations` method:

`num_representations_mentor = is_admin ? nil : mentor`

Or something like that

PR #5554, opened 2023-06-08
Use lock for cache updates

iHiD:


app/commands/user/reset_cache.rb
```diff
@@ -22,9 +22,11 @@ def call
...
     # )
 
-    user.data.reload
-    user.data.cache[key] = new_value
-    user.data.save!
+    User::Data.transaction do
+      data = User::Data.lock(true).find(user.data.id)
+      data.cache[key] = new_value
+      data.save!
+    end
```
ErikSchierboom, line 29: Hey, I tried to do something like this for the other PR, but this is a bit more fancy!

ErikSchierboom (approved):

PR #5659, opened 2023-07-05
Don't count publishing reputation to insiders eligibility

ErikSchierboom:


app/commands/user/insiders_status/determine_eligibility_status.rb
```diff
@@ -32,11 +32,11 @@ def annual_reputation = period_reputation(:year)
...
 
   def period_reputation(period)
-    User::ReputationPeriod::Search.new(
+    user.reputation_periods.where(
       period:,
-      category: :any, # Auto-excludes publishing
-      user_handle: user.handle
-    ).period_records.to_a.sum(&:reputation)
+      category: %i[building maintaining authoring mentoring],
+      about: :everything
```
iHiD:
```suggestion
      about: :everything,
      track_id: 0
```

This also needs specifying else it'll sum all the records for one user for every track.

iHiD: (This is why I use `User::ReputationPeriod::Search` btw as it guarantees these constraints - but I think it's fine not to use it here).

PR #5660, opened 2023-07-05
Don't allow wildcards in searching for contributors

ErikSchierboom:


iHiD:
We could actually remove this search by user_handle entirely as we don't use it?

ErikSchierboom:
I was afraid of breaking anything :shrug

iHiD (approved):

PR #5682, opened 2023-07-07
Show announcement bar for logged in users only

dem4ron:


app/helpers/view_components/site_header.rb
```diff
@@ -17,7 +17,7 @@ def to_s
     end
 
     def announcement_bar
-      return tag.span("") if current_user&.donated?
+      return tag.span("") if current_user&.donated? || !user_signed_in?
```
iHiD:
```suggestion
      return tag.span("") if user_signed_in? && !current_user.donated?
```

iHiD: This reads more like a sentence.

PR #5700, opened 2023-07-11
Move more internal packages to be internal

iHiD:


ErikSchierboom (approved):

dem4ron (approved):

iHiD:
@dem4ron Could you see why this isn't green please, and add commits to fix it. Specifically, why has this PR caused these tests to fail?

app/javascript/packs/internal.tsx
```diff
@@ -1,6 +1,7 @@
 import 'easymde/dist/easymde.min.css'
 
 import React, { lazy, Suspense } from 'react'
+import currency from 'currency.js'
```
iHiD, line 5: @dem4ron can we remove this from application.tsx now?

dem4ron: Done

iHiD:
@dem4ron I think this is causing https://forum.exercism.org/t/unable-to-subscribe-to-premium/6560

Could you work out why the tests didn't pick this up pls?

PR #5708, opened 2023-07-13
Add PHP codemirror support

ErikSchierboom:
@dem4ron It doesn't yet work. This is the console error:

```
Error: Unrecognized extension value in extension set ([object Object]). This sometimes happens because multiple instances of @codemirror/state are loaded, breaking instanceof checks.
```

Could you maybe look into it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@codemirror/lang-cpp": "^0.19.0",
     "@codemirror/lang-java": "^0.19.0",
     "@codemirror/lang-javascript": "^0.19.0",
-    "@codemirror/lang-php": "6.0.0",
+    "@codemirror/lang-php": "^0.19.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
     "@codemirror/lang-cpp": "^0.19.0",
     "@codemirror/lang-java": "^0.19.0",
     "@codemirror/lang-javascript": "^0.19.0",
+    "@codemirror/lang-php": "6.0.0",
```
dem4ron:
```suggestion
    "@codemirror/lang-php": "^0.19.0",
```

The version number for PHP should match the other modules. When different versions are used, it results in two instances of `@codemirror/state`, which is the underlying cause of the problem

iHiD (approved):

PR #5712, opened 2023-07-14
Delete ghost user documents

ErikSchierboom:
- Remove comment
- Remove solution from index when solution user is ghost

### DIFF
--- a/app/commands/solution/sync_to_search_index.rb
+++ b/app/commands/solution/sync_to_search_index.rb
@@ -6,9 +6,6 @@ class Solution::SyncToSearchIndex
   initialize_with :solution
 
   def call
-    # p solution.id
-    # p Solution::CreateSearchIndexDocument.(solution)
-
     Exercism.opensearch_client.index(
       index: Solution::OPENSEARCH_INDEX,
       type: 'solution',
--- a/app/commands/solution/sync_to_search_index.rb
+++ b/app/commands/solution/sync_to_search_index.rb
@@ -6,11 +6,28 @@ class Solution::SyncToSearchIndex
   initialize_with :solution
 
   def call
+    if solution.user.ghost?
+      delete_document!
+    else
+      create_document!
+    end
+  end
+
+  private
+  def create_document!
     Exercism.opensearch_client.index(
       index: Solution::OPENSEARCH_INDEX,
       type: 'solution',
       id: solution.id,
       body: Solution::CreateSearchIndexDocument.(solution)
+    )
+  end
+
+  def delete_document!
+    Exercism.opensearch_client.delete(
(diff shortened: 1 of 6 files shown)


app/commands/solution/sync_to_search_index.rb
```diff
@@ -6,14 +6,28 @@ class Solution::SyncToSearchIndex
...
     )
   end
+
+  def delete_document!
+    Exercism.opensearch_client.delete(
+      index: Solution::OPENSEARCH_INDEX,
+      type: 'solution',
+      id: solution.id
+    )
+  end
```
iHiD, line 32: Does this raise if the document isn't in the index or just ignore it?

app/commands/solution/sync_to_search_index.rb
```diff
@@ -6,14 +6,28 @@ class Solution::SyncToSearchIndex
   initialize_with :solution
 
   def call
-    # p solution.id
-    # p Solution::CreateSearchIndexDocument.(solution)
+    if solution.user.ghost?
```
iHiD, line 9: This won't get called at the point that solutions become owned by the ghost as we use `update_all` to do that (See `app/commands/user_track/reset.rb#L8`)

So I suggest wrapping that line in `UserTrack::Reset`:
1. Get all solution_ids
2. Update all the records (as currently)
3. Defer this command for all the solutions using the ids from (1)

ErikSchierboom: We're sort of already doing that:

`app/commands/user_track/reset.rb:16`

```ruby
Solution::RemoveUserSolutionsForTrackFromSearchIndex.defer(user_track)
```

And that command runs:

```ruby
Exercism.opensearch_client.delete_by_query(index: Solution::OPENSEARCH_INDEX, body: {
      query: {
        bool: {
          must: [
            { term: { 'user.id': user_track.user_id } },
            { term: { 'track.id': user_track.track_id } }
          ]
        }
      }
    })
```

Is there a path where the user is reset elsewhere? `User::ResetAccount` calls `UserTrack::Destroy`, which then calls `UserTrack::Reset`.

iHiD: OK, so `Solution::RemoveUserSolutionsForTrackFromSearchIndex.defer(user_track)` should work theoretically but it's really racey as the user_track might not exist by the time it gets called.

So I'd suggest changing:
```
Solution::RemoveUserSolutionsForTrackFromSearchIndex.defer(user_track)
```

to 
```
Solution::RemoveUserSolutionsForTrackFromSearchIndex.defer(user, track)
```

at which stage I think this'll be ok.

PR #5745, opened 2023-07-19
Materialize highlightjs_language on tracks table

ErikSchierboom:


app/models/track.rb
```diff
@@ -95,7 +95,7 @@ def num_mentors
   def icon_url = "#{Exercism.config.website_icons_host}/tracks/#{slug}.svg"
 
   def highlightjs_language
-    git.highlightjs_language || slug
+    super || slug
```
ErikSchierboom, line 98: Currently, I've chosen to make the field nullable, which means that we can easily get a list of the tracks that don't have a highlightjs language.

PR #5746, opened 2023-07-19
Fix issue with tests being run multiple times

iHiD:


app/commands/iteration/create.rb
```diff
@@ -31,7 +31,9 @@ def call
   end
 
   def init_services
-    Submission::TestRun::Init.(submission) if submission.tests_not_queued? && solution.exercise.has_test_runner?
+    return unless solution.exercise.has_test_runner?
```
ErikSchierboom, line 34: We don't want the analyzer or representer being run when there is no test runner (unlikely scenario, I admit)?

app/commands/solution/queue_head_test_run.rb
```diff
@@ -37,16 +30,28 @@ def handle_latest!
     end
   end
 
+  def should_run_latest?
+    # Always run if we force
+    return true if force
+
+    # Don't run if we're already running this
```
ErikSchierboom:
```suggestion
    # Don't run if we're already running this
```

Maybe expand a little on what "this" means? In particular, what the important files hash condition does?

ErikSchierboom (approved):

app/commands/iteration/create.rb
iHiD: No, we don't. But thanks for checking!

PR #5758, opened 2023-07-20
Speed up profile

iHiD:
This knocks 23seconds of @bobahop's profile load time.

### DIFF
--- a/app/commands/solution/search_user_solutions.rb
+++ b/app/commands/solution/search_user_solutions.rb
@@ -34,9 +34,8 @@ class Solution::SearchUserSolutions
 
     solution_ids = results["hits"]["hits"].map { |hit| hit["_source"]["id"] }
     solutions = solution_ids.present? ?
-      Solution.where(id: solution_ids).
-        order(Arel.sql("FIND_IN_SET(id, '#{solution_ids.join(',')}')")).
-        to_a : []
+      Solution.where(id: solution_ids).sort_by { |s| solution_ids.index(s.id) }
+        : []
 
     total_count = results["hits"]["total"]["value"].to_i
     Kaminari.paginate_array(solutions, total_count:).
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -8,10 +8,7 @@ class ProfilesController < ApplicationController
   end
 
   def show
-    @solutions = @user.solutions.published.
-      order(num_stars: :desc, updated_at: :desc).
-      includes(:exercise, :track).
-      first(3)
+    @solutions = Solution::SearchUserSolutions.(@user, status: :published, per: 3)
 
     # TODO: Order by most prominent first (what is the most prominent testimonial?)
     @testimonials = @user.mentor_testimonials.published.first(3)
--- a/db/migrate/20230720125540_add_num_tokens_to_reputation_periods.rb
+++ b/db/migrate/20230720125540_add_num_tokens_to_reputation_periods.rb
@@ -1,5 +1,7 @@
 class AddNumTokensToReputationPeriods < ActiveRecord::Migration[7.0]
   def change
+    return if Rails.env.production?
+
     add_column :user_reputation_periods, :num_tokens, :integer, null: false, default: 0
   end
 end
--- a/db/migrate/20230720125540_add_num_tokens_to_reputation_periods.rb
+++ b/db/migrate/20230720125540_add_num_tokens_to_reputation_periods.rb
@@ -1,0 +1,5 @@
+class AddNumTokensToReputationPeriods < ActiveRecord::Migration[7.0]
+  def change
+    add_column :user_reputation_periods, :num_tokens, :integer, null: false, default: 0
+  end
+end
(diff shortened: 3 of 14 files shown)


test/assemblers/assemble_contributions_summary_test.rb
```diff
@@ -45,10 +61,8 @@ class AssembleContributionsSummaryTest < ActiveSupport::TestCase
           title: "All Tracks",
           icon_url: "ICON",
           categories: [
-            # TODO: Reputation
-            { id: :publishing, reputation: 0, metric_full: "6 solutions published", metric_short: "6 solutions" },
-            # TODO: Reputation
-            { id: :mentoring, reputation: 0, metric_full: "9 students mentored", metric_short: "9 students" },
+            { id: :publishing, reputation: 10, metric_full: "5 solutions published", metric_short: "5 solutions" },
+            { id: :mentoring, reputation: 45, metric_full: "9 students mentored", metric_short: "9 students" },
```
iHiD, line 62: Apparently I just never found a way to even implement this before in a time that was sane. (Apparently 22s was sane so goodness knows how long this took...)

test/commands/user/reputation_token/calculate_contextual_data_test.rb
```diff
@@ -17,8 +17,11 @@ class User::ReputationToken::CalculateContextualDataTest < ActiveSupport::TestCa
     5.times { create :user_exercise_contribution_reputation_token, user: }
     6.times { create :user_mentored_reputation_token, user: }
     7.times { create :user_published_solution_reputation_token, user: }
+
+    generate_reputation_periods!
+
     data = User::ReputationToken::CalculateContextualData.(user.id)
-    expected_activity = "1 PR created • 3 PRs reviewed • 2 PRs merged • 4 exercises authored • 5 exercise contributions • 6 solutions mentored • 7 solutions published" # rubocop:disable Layout/LineLength
+    expected_activity = "1 PR created • 5 PRs reviewed and/or merged • 9 exercise contributions • 6 solutions mentored • 7 solutions published" # rubocop:disable Layout/LineLength
```
iHiD, line 24: This is the only "lossy" bit of this PR. But it's like 4 OOM faster so I don't care.

app/controllers/profiles_controller.rb
```diff
@@ -8,10 +8,7 @@ def index
   end
 
   def show
-    @solutions = @user.solutions.published.
-      order(num_stars: :desc, updated_at: :desc).
-      includes(:exercise, :track).
-      first(3)
+    @solutions = Solution::SearchUserSolutions.(@user, status: :published, per: 3)
```
iHiD, line 11: This is much faster (like 1s -> 0.02sec) so we should check if there are other places that could use this too.

app/commands/solution/search_user_solutions.rb
```diff
@@ -34,9 +34,8 @@ def call
 
     solution_ids = results["hits"]["hits"].map { |hit| hit["_source"]["id"] }
     solutions = solution_ids.present? ?
-      Solution.where(id: solution_ids).
-        order(Arel.sql("FIND_IN_SET(id, '#{solution_ids.join(',')}')")).
-        to_a : []
+      Solution.where(id: solution_ids).sort_by { |s| solution_ids.index(s.id) }
+        : []
```
iHiD, line 38: This moves the sorting to Ruby and knocks off 100ms.

bobahop:
Thank you very much! I sincerely appreciate it.

app/commands/solution/search_user_solutions.rb
ErikSchierboom: We have more places where we do the same. Maybe also apply the same strategy there?

test/commands/user/reputation_token/calculate_contextual_data_test.rb
ErikSchierboom: Yeah that's fine

test/test_helper.rb
```diff
@@ -327,6 +327,13 @@ def stub_latest_track_forum_threads(track)
     stub_request(:get, "https://forum.exercism.org/c/programming/#{track.slug}/l/latest.json")
   end
 
+  def generate_reputation_periods!
```
ErikSchierboom, line 330: Convenient!

db/migrate/20230720125540_add_num_tokens_to_reputation_periods.rb
```diff
@@ -0,0 +1,5 @@
+class AddNumTokensToReputationPeriods < ActiveRecord::Migration[7.0]
+  def change
+    add_column :user_reputation_periods, :num_tokens, :integer, null: false, default: 0
```
ErikSchierboom, line 5: Should this have:

```suggestion
    return if Rails.env.production?

    add_column :user_reputation_periods, :num_tokens, :integer, null: false, default: 0
```

app/controllers/profiles_controller.rb
ErikSchierboom: Is that due to the fact that the sorting is now changing to just sorting by `num_stars` and no longer using `updated_at`?

ErikSchierboom (approved):

app/controllers/profiles_controller.rb
iHiD: It's that it uses cloudsearch not Sql

app/commands/solution/search_user_solutions.rb
iHiD: Yeah, I'll go through and check.

PR #5762, opened 2023-07-20
Don't send icon requests to Rails

iHiD:


iHiD:
@IsaacG This look ok to you?

IsaacG:
The syntax looks correct. But nginx has a test flag to validate the syntax :D 

Is the goal primarily to reduce logging? If so, this looks fine. If not, I'm not sure why this is needed.

Is there a reason there's no favicon or robots.txt? Should there be one?

iHiD:
Thanks for reviewing. I don't want these requests being passed to the webservers from nginx basically- it just adds more effort to those, so I want to 404 them at the nginx level.

Favicon etc are specified at different locations in the HEAD of the HTML, so real users won't be looking for that file here.

IsaacG:
Sounds reasonable to me!

PR #5763, opened 2023-07-20
Improve logging

iHiD:
This knocks 400ms off the dashboard and gives us verbose query logs.

### DIFF
--- a/app/serializers/serialize_tracks.rb
+++ b/app/serializers/serialize_tracks.rb
@@ -1,7 +1,9 @@
 class SerializeTracks
   include Mandate
 
-  initialize_with :tracks, :user
+  initialize_with :tracks, :user do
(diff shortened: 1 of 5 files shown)


app/serializers/serialize_tracks.rb
```diff
@@ -1,7 +1,9 @@
 class SerializeTracks
   include Mandate
 
-  initialize_with :tracks, :user
+  initialize_with :tracks, :user do
+    @tracks = tracks.to_a if tracks.is_a?(ActiveRecord::Relation)
```
iHiD, line 6:
```suggestion
    # Materialise this so we're not ending up with nested queries
    @tracks = tracks.to_a if tracks.is_a?(ActiveRecord::Relation)
```

PR #5767, opened 2023-07-21
Trigger backup event when pushed to main

ErikSchierboom:
- Dispatch backup repo event
- Rename org wide files repo event
- Simplify DispatchOrgWideFilesRepoUpdateEvent arguments
- Dispatch org-wide-files repo update event in background job
- Add command to dispatch event

### DIFF
--- a/app/commands/github/dispatch_event.rb
+++ b/app/commands/github/dispatch_event.rb
@@ -1,0 +1,11 @@
+class Github::DispatchEvent
+  include Mandate
+
+  initialize_with :repo, :event_type, :client_payload
+
+  def call = Exercism.octokit_client.post(api_url, body)
+
+  private
+  def api_url = "https://api.github.com/repos/exercism/#{repo}/dispatches"
+  def body = { event_type:, client_payload: }.to_json
+end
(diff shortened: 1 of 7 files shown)


app/commands/github/dispatch_event.rb
```diff
@@ -0,0 +1,11 @@
+class Github::DispatchEvent
```
ErikSchierboom, line 1: I added this command to simplify the command a bit and to re-use it in the other event dispatching command

app/commands/webhooks/process_push_update.rb
```diff
@@ -16,8 +16,10 @@ def call
     else
       track = Track.find_by(slug: repo_name)
       Git::SyncTrack.defer(track) if track
-      Github::DispatchEventToOrgWideFilesRepo.(:repo_update, [repo], pusher_username) if trigger_repo_update?
+      Github::DispatchOrgWideFilesRepoUpdateEvent.defer(repo, pusher_username) if trigger_repo_update?
```
ErikSchierboom, line 19: Deferring this makes sense, as we don't need it to happen instantaneously

iHiD (approved):
I don't fully get what all this does, but the code looks good! :)

PR #5773, opened 2023-07-21
Remove solutions from index when resetting/destroying account

ErikSchierboom:
- Remove solutions from index when resetting account
- Remove solutions from index when resetting account
- Gracefully handle deleting opensearch document that isn't there

### DIFF
--- a/app/commands/user/reset_account.rb
+++ b/app/commands/user/reset_account.rb
@@ -37,6 +37,8 @@ class User::ResetAccount
     user.user_tracks.each do |user_track|
       UserTrack::Destroy.(user_track)
     end
+
+    Solution::RemoveUserSolutionsFromSearchIndex.defer(user)
   end
 
   def reset_mentoring!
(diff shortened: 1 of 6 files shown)


app/commands/user/reset_account.rb
```diff
@@ -37,6 +37,8 @@ def reset_tracks!
     user.user_tracks.each do |user_track|
       UserTrack::Destroy.(user_track)
     end
+
+    Solution::RemoveUserSolutionsFromSearchIndex.defer(user)
```
iHiD, line 41: What's unclear to me is why the `UserTrack::Destroy` isn't doing this though. Do you understand that?

ErikSchierboom: I don't. It is quite weird

iHiD:
I think this can be closed in favour of https://github.com/exercism/website/pull/5809

PR #5787, opened 2023-07-24
Add comments preferences

dem4ron:
Screenshot 2023-07-27 at 12 57 12
Screenshot 2023-07-27 at 12 57 07


db/migrate/20230724113620_add_allow_comments_by_default_to_user_preferences.rb
```diff
@@ -0,0 +1,5 @@
+class AddAllowCommentsByDefaultToUserPreferences < ActiveRecord::Migration[7.0]
+  def change
+    add_column :user_preferences, :allow_comments_by_default, :boolean
```
iHiD:
```suggestion
    return if Rails.env.production?
    
    add_column :user_preferences, :allow_comments_on_published_solutions, :boolean, null: false, default: false
```

PR #5809, opened 2023-07-27
Fix race conditions around deleting users

iHiD:
Supersedes https://github.com/exercism/website/pull/5773. That shouldn't be needed as the actual issue was race conditions. It's a complex PR - I've added comments to the code, but I'll also add to the PR.

### DIFF
--- a/app/commands/solution/remove_user_solutions_for_track_from_search_index.rb
+++ b/app/commands/solution/remove_user_solutions_for_track_from_search_index.rb
@@ -3,18 +3,27 @@ class Solution::RemoveUserSolutionsForTrackFromSearchIndex
 
   queue_as :background
 
-  initialize_with :user, :track
+  # Use ids here as we're calling this because either the
+  # user or the track has been deleted, so we can't retrieve
+  # the record from the database
+  initialize_with :user_id, :track_id
 
   def call
     Exercism.opensearch_client.delete_by_query(index: Solution::OPENSEARCH_INDEX, body: {
       query: {
         bool: {
           must: [
-            { term: { 'user.id': user.id } },
-            { term: { 'track.id': track.id } }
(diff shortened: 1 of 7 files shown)


app/commands/solution/remove_user_solutions_for_track_from_search_index.rb
```diff
@@ -3,18 +3,27 @@ class Solution::RemoveUserSolutionsForTrackFromSearchIndex
...
           ]
         }
       }
     })
+  rescue Elasticsearch::Transport::Transport::Errors::Conflict
+    raise if @retries.to_i >= 3
+
+    sleep(0.5)
+    @retries = @retries.to_i + 1
+    retry
```
iHiD, line 27: This might just happen in the tests but if this record is still be persisted internally at cloudsearch when the delete is requested, we get this conflict. I don't see any harm in having this code here.

app/jobs/application_job.rb
```diff
@@ -6,8 +6,28 @@ class ApplicationJob < ActiveJob::Base
...
+    sleep(0.5)
+    begin
+      exception.cause.model.constantize.find(exception.cause.id)
+      retry_job
+    rescue NoMethodError
+      # The model has gone. Kill the job.
+    rescue ActiveRecord::RecordNotFound
+      # The record is gone. Kill the job.
+    end
+  end
```
iHiD, line 30: This bit is a bit complex. We're looking at three issues:
1. The record is deleted (the final AR::RecordNotFound). In which case we just silently end the job.
2. The model is deleted (ie we've removed the file, which is the NoMethodError). In which case we just silently end the job.
3. The job has run before the db transaction has finished, in which case we sleep for half a second and then retry it.

ErikSchierboom (approved):
Nice work

PR #5828, opened 2023-08-01
Fix youtube video links not working when using underscore

ErikSchierboom:


app/commands/markdown/preprocess.rb
```diff
@@ -51,7 +51,7 @@ def convert_inline_links!
         node.url = "https://player.vimeo.com/video/#{m[1]}"
       end
 
-      link_text.match(%r{^video:youtube-mail/(\w+)$}) do |m|
+      link_text.gsub('\\_', '_').match(%r{^video:youtube-mail/([\w_]+)$}) do |m|
```
ErikSchierboom, line 54: The `gsub` call is needed as somewhere in the pipeline, an underscore is escaped by prepending it with a backslash.

PR #5855, opened 2023-08-04
Attempt to award reputation token even when authorship/contributorship exists

ErikSchierboom:
I'm not yet sure why this happens, but it looks like sometimes the author is added to the concept/exercise but the reputation token isn't. This PR wil "solve" this by just trying to add the reputation token regardless if the authorship/contributorship already exists.

We can run this code to fix this:

```ruby
Concept::Contributorship.find_each do |contributorship|
  User::ReputationToken::Create.defer(
    contributorship.contributor,
    :concept_contribution,
    contributorship:
  )
end

Concept::Authorship.find_each do |authorship|
  User::ReputationToken::Create.defer(
    authorship.author,
    :concept_author,
    authorship:
  )
end

Exercise::Contributorship.find_each do |contributorship|
  User::ReputationToken::Create.defer(
    contributorship.contributor,
    :exercise_contribution,
    contributorship:
  )
end

Exercise::Authorship.find_each do |authorship|
  User::ReputationToken::Create.defer(
    authorship.author,
    :exercise_author,
    authorship:
  )
end
```

### DIFF
--- a/app/commands/concept/authorship/create.rb
+++ b/app/commands/concept/authorship/create.rb
@@ -4,16 +4,12 @@ class Concept::Authorship::Create
   initialize_with :concept, :author
 
   def call
-    begin
-      authorship = concept.authorships.create!(author:)
-    rescue ActiveRecord::RecordNotUnique
-      return nil
+    concept.authorships.find_or_create_by(author:).tap do |authorship|
+      User::ReputationToken::Create.defer(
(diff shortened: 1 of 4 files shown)


iHiD (approved):
I'm fine with this. Seems odd. But easier than debugging deeply :)

PR #5861, opened 2023-08-07
Fix the world

iHiD:


app/javascript/components/donations/StripeForm.tsx
```diff
@@ -60,14 +60,17 @@ export function StripeForm({
...
+  }
 
   return (
-    <form
-      data-turbo="false"
-      onSubmit={
-        paymentIntentType === 'payment' ? handlePaymentSubmit : handleSubmit
-      }
-    >
+    <form data-turbo="false" onSubmit={handlePaymentSubmit}>
```
iHiD, line 64: You can get rid of `handleSubmit` entirely now, and probably a whole ton of unused code

PR #5869, opened 2023-08-09
Fix mentoring info on community solutions

dem4ron:
Screenshot 2023-08-09 at 16 15 43

### DIFF
--- a/app/css/pages/community-solution-show.css
+++ b/app/css/pages/community-solution-show.css
@@ -139,7 +139,7 @@
                             @apply mb-4 flex;
                             @apply text-16 leading-150 text-textColor2;
                             & strong {
-                                @apply font-semibold;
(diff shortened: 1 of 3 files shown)


app/css/pages/community-solution-show.css
```diff
@@ -139,7 +139,7 @@
                             @apply mb-4 flex;
                             @apply text-16 leading-150 text-textColor2;
                             & strong {
-                                @apply font-semibold;
+                                @apply font-semibold flex gap-4;
```
iHiD: You could also maybe just do this as inline or inline flex and let the browser handle the spacing. But I'm ok with this too :)

dem4ron: yeah, that is nicer!

iHiD:
This is red.

PR #5874, opened 2023-08-09
Fix race condition in creating reputation tokens

iHiD:
@ErikSchierboom This issue with this is that exercise and concept authors are called in a single transaction (`git grep transaction app/commands/git`). That's creating a race condition where Sidekiq gets the records before they're materialised to the database.

The reason this is hurting us now is due to the ignoring of ActiveRecord::NotFound exceptions in the ApplicationJob, to avoid us having tons of dead (but forever-retrying) jobs from deleted users (and their records).

Two fixes:
1. Wait 30s before doing these specific jobs
2. Give us a bit more breathing room in the application job code.

### DIFF
--- a/app/commands/concept/authorship/create.rb
+++ b/app/commands/concept/authorship/create.rb
@@ -8,7 +8,10 @@ class Concept::Authorship::Create
       User::ReputationToken::Create.defer(
         author,
         :concept_author,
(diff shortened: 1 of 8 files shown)


ErikSchierboom (approved):
Makes sense

PR #5890, opened 2023-08-11
Skip modal on Hello World

dem4ron:
https://github.com/exercism/website/assets/66035744/56968af0-1453-40f9-92d2-7c6249f40850


iHiD:
This is good. If we expose the `tutorial` key to React, I'd rather use that than the slug, but if we don't I'm 100% fine with the slug for now.

PR #5916, opened 2023-08-16
Delete site updates with concepts

iHiD:
Fixes this:

Screenshot 2023-08-16 at 14 05 26


iHiD:
@ErikSchierboom Could you follow up with a seperate PR that just does the same for anything else that we have site updates about pls (Exercises and maybe others)

PR #5919, opened 2023-08-16
Delete exercise site updates on destroy

ErikSchierboom:
Other than new exercise and concept updates, we have arbitrary site updates but those aren't linked AFAICT.


iHiD (approved):
Great. Thanks.

PR #5930, opened 2023-08-18
Fix issues with edge case in representations

iHiD:


app/commands/submission/representation/create.rb
```diff
@@ -4,14 +4,17 @@ class Submission::Representation::Create
   initialize_with :submission, :tooling_job, :ast_digest, :exercise_representer_version
 
   def call
-    Submission::Representation.create!(
+    representation = Submission::Representation.find_create_or_find_by!(
       submission:,
-      tooling_job_id: tooling_job.id,
       ops_status: tooling_job.execution_status.to_i,
```
ErikSchierboom, line 9: Do we also want to include the `ops_status` here?

app/commands/exercise/representation/create_or_update.rb
```diff
@@ -16,10 +16,13 @@ def call
...
     end
 
+    # Do this straight away before anything else
+    representation_is_new = representation.id_previously_changed?
+
     update_cache_columns!
 
-    # Now copy the old feedback and trigger runs if appropriate
-    if old_representation && representation != old_representation
+    # Now copy the old feedback and trigger runs if we've created a new represenation that's different from the old one
```
ErikSchierboom:
```suggestion
    # Now copy the old feedback and trigger runs if we've created a new representation that's different from the old one
```

ErikSchierboom (approved):

app/commands/submission/representation/create.rb
iHiD: Yeah, I think so. If the previous was a 400 and now this is a 200, we want a new record. Good question though - I pondered it too.

PR #5945, opened 2023-08-21
Change text based on completed_percentage

dem4ron:
https://forum.exercism.org/t/track-overview-page-does-not-update-for-100-completion/6985
Screenshot 2023-08-21 at 11 14 45


iHiD:
Updated. I prefer not polluting the model and using a helper instead.

PR #5946, opened 2023-08-21
Fix wrong nesting

dem4ron:
https://forum.exercism.org/t/in-line-code-is-rendered-too-tall/6904

Screenshot 2023-08-21 at 12 52 01

#### Current state:
Screenshot 2023-08-21 at 17 02 58

### DIFF
--- a/app/css/components/code-pane.css
+++ b/app/css/components/code-pane.css
@@ -5,12 +5,6 @@
         @apply flex-grow;
     }
 
-    #panel-instructions {
-        code {
-            @apply py-0;
-        }
-    }
-
     code {
         @apply flex;
         @apply py-8;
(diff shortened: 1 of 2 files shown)


app/css/components/code-pane.css
```diff
@@ -5,6 +5,12 @@
         @apply flex-grow;
     }
 
+    #panel-instructions {
+        code {
+            @apply py-0;
+        }
+    }
```
iHiD: What makes this different in this one specific panel?

iHiD: Or maybe to put it another way, why does a `c-code-pane` contain `#panel-instructions`. Is that semantically correct in the first place?

PR #5951, opened 2023-08-21
Use only one event listener

dem4ron:


app/javascript/packs/application.tsx
```diff
@@ -518,12 +518,11 @@ import { TrophiesProps, Trophy } from '@/components/track/Trophies.js'
 
 document.addEventListener('turbo:load', () => {
   highlightAll()
+  showSiteFooterOnTurboLoad()
```
iHiD: This should be renamed to `showSiteFooter()` if it no longer includes the `onTurboLoad` logic.

PR #5953, opened 2023-08-21
Add step to wait for opensearch

iHiD:
Try and fix this!

Screenshot 2023-08-21 at 15 02 46

### DIFF
--- a/.github/workflows/rerun-failed.yml
+++ b/.github/workflows/rerun-failed.yml
@@ -1,0 +1,29 @@
+name: Rerun failed
+
+on:
+  workflow_run:
+    workflows:
+      - Tests
+    types:
+      - completed
+
+jobs:
+  on-failure:
+    runs-on: ubuntu-22.04
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
+      - name: Check if OpenSearch failed to start
+        id: check-opensearch
+        run: |
+          jobs=$(gh api ${{ github.event.workflow_run.jobs_url }})
+          failed=$(echo "${jobs}" | jq '.jobs | any(.steps | any(.name == "Wait for open search to become ready" and .conclusion == "failure"))')
+          echo "failed=${failed}" >> "$GITHUB_OUTPUT"
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+      - name: Rerun job
+        if: ${{ steps.check-opensearch.outputs.failed == 'true' }}
+        run: |
+          gh api --method POST ${{ github.event.workflow_run.rerun_url }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.WORKFLOWS_PAT }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -234,7 +234,7 @@ jobs:
         id: wait-for-opensearch
         env:
           OPENSEARCH_HOST: https://127.0.0.1:${{ job.services.opensearch.ports['9200'] }}
-        run: ./bin/wait_for_opensearch.sh
+        run: ./bin/wait_for_opensearch
 
       ###
       # Run the tests
@@ -398,7 +398,7 @@ jobs:
         id: wait-for-opensearch
         env:
           OPENSEARCH_HOST: https://127.0.0.1:${{ job.services.opensearch.ports['9200'] }}
-        run: ./bin/wait_for_opensearch.sh
+        run: ./bin/wait_for_opensearch
 
       ###
       # Run the tests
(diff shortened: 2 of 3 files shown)


iHiD:
Conclusion is that opensearch sometimes just never becomes ready. We could ideally conditionally find a way to restart the job in this situation. @ErikSchierboom Any ideas?

ErikSchierboom:
My initial though is to detect that opensearch isn't running and then to re-trigger the workflow via a `workflow_dispatch` message. https://www.eliostruyf.com/restart-github-actions-workflow-failed/

bin/wait_for_opensearch
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+timeout="10s"
+status="yellow"
+url="'${OPENSEARCH_HOST}/_cluster/health?wait_for_status=${status}&timeout=${timeout}'"
```
ErikSchierboom: We can use opensearch's built-in functionality for this: https://opensearch.org/docs/latest/api-reference/cluster-api/cluster-health/

.github/workflows/tests.yml
```diff
@@ -229,6 +228,21 @@ jobs:
...
+      ###
+      # Wait for everything to be ready
+      - name: Wait for open search to become ready
+        id: wait-for-opensearch
+        env:
+          OPENSEARCH_HOST: https://127.0.0.1:${{ job.services.opensearch.ports['9200'] }}
+        run: ./bin/wait_for_opensearch
+
+      - name: Re-run job if open-search failed to become ready
+        if: ${{ failure() }}
```
ErikSchierboom: This means it only runs when the previous step failed.

.github/workflows/tests.yml
```diff
@@ -229,6 +228,21 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ secrets.RERUN_WORKFLOWS_PAT }}
```
ErikSchierboom: We need a separate token to be able to re-run workflow runs

.github/workflows/tests.yml
```diff
@@ -229,6 +228,21 @@ jobs:
...
+        run: |
+          gh run rerun --failed ${{ github.run_id }}
```
ErikSchierboom: This should only re-run the current run, not the entire workflow

.github/workflows/rerun-failed.yml
```diff
@@ -0,0 +1,25 @@
+name: Rerun failed
+
+on:
+  workflow_run:
+    workflows: [Run]
+    types: [completed]
```
ErikSchierboom: This ensures that we can safely restart the workflow, as it fires when the entire run completes (could be in a failed state)

.github/workflows/rerun-failed.yml
```diff
@@ -0,0 +1,25 @@
...
+  on-failure:
+    runs-on: ubuntu-22.04
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
+      - name: Check if OpenSearch failed to start
+        id: check-opensearch
+        run: |
+          jobs=$(gh api ${{ github.event.workflow_run.jobs_url }})
+          failed=$(echo "${jobs}" | jq '.jobs | any(.steps | any(.name == "Wait for open search to become ready" and .conclusion == "failure"))')
+          echo "failed=${failed}" >> "$GITHUB_OUTPUT"
```
ErikSchierboom, line 20: We don't just want to restart any run, but only those where opensearch failed to start

.github/workflows/rerun-failed.yml
```diff
@@ -0,0 +1,29 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+      - name: Rerun job
+        if: ${{ steps.check-opensearch.outputs.failed == 'true' }}
+        run: |
+          gh api --method POST ${{ github.event.workflow_run.rerun_url }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.WORKFLOWS_PAT }}
```
ErikSchierboom, line 29: A fine-grained token worked! Wouldn't you know

.github/workflows/tests.yml
```diff
@@ -229,6 +228,13 @@ jobs:
           bundle exec rails r bin/monitor-manifest
           bundle exec rails test:prepare
 
+      ###
+      # Wait for everything to be ready
+      - name: Wait for open search to become ready
+        env:
+          OPENSEARCH_HOST: https://127.0.0.1:${{ job.services.opensearch.ports['9200'] }}
+        run: ./bin/wait_for_opensearch
```
ErikSchierboom, line 236: I wonder at what step we should be having this check. We're not doing a bunch of other things first that will be meaningless if opensearch won't start

.github/workflows/rerun-failed.yml
iHiD: Wow!

PR #5954, opened 2023-08-21
Add source code pro

iHiD:
I think this got lost when self-hosting fonts.

@ErikSchierboom @dem4ron Do either of you know if we use other font-weights anywhere (e.g. in highlightJS)?


dem4ron:
We use couple bold font-weights for mono. 
eg.: in `text.css`
```postcss
.text-label-code-caps {
    @apply font-mono text-14 leading-regular uppercase text-darkGray font-bold;
}
```
or
```postcss
.text-label {
    @apply block;
    @apply text-13 leading-150 text-textColor2 font-mono font-bold uppercase;
}
```
and more

ErikSchierboom:
I'll leave this to @dem4ron, he'll know this much better than I do

PR #5960, opened 2023-08-22
Order of unseen reputation tokens descendingly

ErikSchierboom:


iHiD:
Can you check this is still fast for you/bob/etc in production pls (`AssembleReputationToken.(User.for!('erikschierboom'), for_header: true)` or whatever). The first run is the one that matters. MySQL will cache the data in itself for 10mins or so afterward, so the first benchmark is the one to check. 

(My sense based on the comment that's in the code, is that this is one of the things that was really slowly bob down)

iHiD:
Merging as a result of Erik's testing. 

---

@bobahop Tiny worry this might slow things down to you, but Erik doesn't think it will. If you notice any pain, please let Erik know and we can always revert :)

bobahop:
Thanks for the heads up! Sorry, didn't get onto my computer today until about 2 PM. Just did a bunch of Exercism stuff. Did not seem noticeably slower.

PR #5961, opened 2023-08-22
Convert the React bootloader to TypeScript

dem4ron:
This PR turns `react-bootloader.jsx` into `react-bootloader.tsx` in `utils`.


app/javascript/utils/react-bootloader.tsx
```diff
@@ -0,0 +1,181 @@
...
+      }
+
+      event.setUser(tag.content)
+    },
+  })
+  const reactPlugin = Bugsnag.getPlugin('react')
+  if (reactPlugin) {
+    ErrorBoundary = reactPlugin.createErrorBoundary(React)
+  } else {
+    throw new Error('Bugsnag React plugin is null')
```
dem4ron:
```suggestion
    throw new Error('Failed to load Bugsnag\'s react plugin')
```

app/javascript/utils/react-bootloader.tsx
```diff
@@ -0,0 +1,181 @@
...
+import BugsnagPluginReact from '@bugsnag/plugin-react'
+import { ExercismTippy } from '../components/misc/ExercismTippy'
+import { ReactQueryCacheProvider } from 'react-query'
+
+type ErrorBoundaryType = React.ComponentType<any>
+
+type GeneratorFunc = (data: any, elem: HTMLElement) => JSX.Element
+type Mappings = Record<string, GeneratorFunc>
+
+type TurboFrameRenderDetail = {
```
dem4ron, line 13: I couldn't find @types for turbo, so inferred this from its usage.

app/javascript/utils/react-bootloader.tsx
```diff
@@ -0,0 +1,181 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import Bugsnag from '@bugsnag/js'
+import BugsnagPluginReact from '@bugsnag/plugin-react'
+import { ExercismTippy } from '../components/misc/ExercismTippy'
+import { ReactQueryCacheProvider } from 'react-query'
+
+type ErrorBoundaryType = React.ComponentType<any>
```
dem4ron, line 8: Once updated to React 18, hopefully I can get rid of the `any` here

app/javascript/utils/react-bootloader.tsx
```diff
@@ -0,0 +1,181 @@
...
+    </React.StrictMode>,
+    elem
+  )
+  document.addEventListener('turbo:before-frame-render', () => {
+    ReactDOM.unmountComponentAtNode(elem)
+  })
+}
+
+export function renderComponents(
+  parentElement: HTMLElement = document.body,
```
dem4ron: changed all
```js
 if (!parentElement) {
    parentElement = document.body
  }
  ```
  to default parameters

File: app/javascript/utils/react-bootloader.tsx [hunk 5]
```diff
@@ -0,0 +1,181 @@
...
+      }
+
+      event.setUser(tag.content)
+    },
+  })
+  const reactPlugin = Bugsnag.getPlugin('react')
+  if (reactPlugin) {
+    ErrorBoundary = reactPlugin.createErrorBoundary(React)
+  } else {
+    throw new Error('Failed to load Bugsnag\'s react plugin')
```
ErikSchierboom: This is a change from the original code. Does this mean that when the Bugsnag plugin doesn't load, nothing works? The `throw` does not seem to be caught, which means it would stop executing any other code in the file. But I guess that was also what would have happened previously?

app/javascript/utils/react-bootloader.tsx
ErikSchierboom: Isn't that changing the behavior though? Previously, these calls would end up with `parentElement` being set to `document.body`, but they won't anymore:

```js
renderComponents(null, ...)
renderComponents(undefined, ...)
```

app/javascript/utils/react-bootloader.tsx
```diff
@@ -0,0 +1,181 @@
...
+  }
```
ErikSchierboom, line 49: Previously there was an `else`:

```
else {
  ErrorBoundary = <></>
}
```

Did you remove that on purpose?

app/javascript/utils/react-bootloader.tsx
dem4ron: > But I guess that was also what would have happened previously?

Yes, it is more descriptive, and shows the error when it occurs.

app/javascript/utils/react-bootloader.tsx
dem4ron: Yes, I added that as a default value

app/javascript/utils/react-bootloader.tsx
ErikSchierboom: @dem4ron Did you see this comment?

app/javascript/utils/react-bootloader.tsx
ErikSchierboom: Where can I find that?

app/javascript/utils/react-bootloader.tsx
```diff
@@ -0,0 +1,181 @@
...
+  fetchResponse: {
+    response: {
+      headers: {
+        get: (name: string) => string | null
+      }
+    }
+  }
+}
+
+let ErrorBoundary: ErrorBoundaryType = () => <></>
```
dem4ron, line 23: @ErikSchierboom here

app/javascript/utils/react-bootloader.tsx
dem4ron: L23

app/javascript/utils/react-bootloader.tsx
dem4ron: In our case this doesn't change the behaviour, because we don't pass in falsy values.
I made sure things wouldn't fail, by passing in a default parameter a level above too, which used to be simply `undefined`
```diff
- const renderThings = (parentElement) => {
+ const renderThings = (parentElement = document.body) => {
    renderComponents(parentElement, mappings)
    renderTooltips(parentElement, mappings)
  }

  // This adds rendering for all future turbo clicks
  document.addEventListener('turbo:load', () => {
    renderThings()
  })
  ...
```


Another place where `renderComponents` is used is in `LazyTippy.tsx`:

```ts
export const LazyTippy = (props: LazyTippyProps): JSX.Element => {
  const { renderReactComponents, ...tippyProps } = props
  const [mounted, setMounted] = React.useState(false)

  const lazyPlugin = {
    fn: () => ({
      onMount: () => setMounted(true),
      onHidden: () => setMounted(false),
      onAfterUpdate: (instance: Instance) => {
        if (renderReactComponents) {
          renderComponents(instance.popper, mappings)
        }
      },
    }),
  }
  ```
  
In the worst-case scenario here, the value would be `undefined` too
  
On the other hand - given how mission critical this code is - I'm leaning towards reverting these changes to the input validation approach, because that is significantly more robust.

Comment (dem4ron) [reply app/javascript/utils/react-bootloader.tsx]: But here:
```ts
  const renderThings = (parentElement = document.body) => {

    renderComponents(parentElement, mappings)
    renderTooltips(parentElement, mappings)
  }
  ```
  
  I'd just pass in `document.body`:

  ```ts
   const renderThings = () => {

    renderComponents(document.body, mappings)
    renderTooltips(document.body, mappings)
  }
  ```
  
  ```ts
  export function initReact(mappings: Mappings): void {
  const renderThings = () => {

    renderComponents(document.body, mappings)
    renderTooltips(document.body, mappings)
  }

  // This adds rendering for all future turbo clicks
    renderThings()
  })

  // This renders if turbo has already finished at the
  // point at which this calls. See packs/core.tsx
  if (window.turboLoaded) {
    renderThings()
  }
}
```

ErikSchierboom (approved):

PR #5966, opened 2023-08-22
Remove EasyMDE from loading every time

iHiD:
Loading things into these index.ts files means things get grouped together in chunks. By removing them from the index.ts, they don't seem to then get chunked with the other things, and get 

---

Before:
Screenshot 2023-08-22 at 17 31 39

After:
Screenshot 2023-08-22 at 17 32 13


ErikSchierboom (approved):
@dem4ron do we need to do this in more places?

iHiD:
I've sent Aron an hour long loom about this going into to waaaaay too much detail about it (I've put it on Slack for you too). I'll reproduce the key message here though:
> I think the one lesson I've taken from today, and this would be the absolutely key thing I'd like you to consider, is that if something is only called from one other place, then it seems that esbuild will put it in its own file. As soon as it's referenced from two places, it gets put in a chunk with lots of other things. So if we can ensure that things are only called in one place are not unnecessarily exported or put in index files, I think that will actually make a lot of positive difference.

But the negative tradeoff is that it can be a lot of effort for diminishing returns.

dem4ron:
> @dem4ron do we need to do this in more places?

Probably :)

PR #5980, opened 2023-08-23
Proxy images

iHiD:
This PR changes the way we use avatar images.

Currently we use url_for(avatar), which creates a rails-redirect URL, which means our webserver gets hit for each request. This is quite painful as it doesn't have caching, it means we have to n+1 load avatars and blobs. 

Instead, we're now going via cloudfront (assets.exercism.org/avatars/{user_id}) and that gets proxied to Rails, which proxies from S3. But it's cached in cloudfront forever so we only have to do this once. This removes all the db lookups and saves us a ton of bandwidth, load time, and pain.

As part of it, I've added a `version` column to users which means we don't have to worry about expiring caches (which is impossible to do on people's local computers) when someone changes their image. We can use the same field for other image generation - e.g. on solution OG images, which is something I've been pondering previously. We still do invalidate cloudfront to remove old images from our caches though (for data privacy concerns).


app/commands/user/update_avatar.rb
```diff
@@ -0,0 +1,12 @@
+class User::UpdateAvatar
+  include Mandate
+
+  initialize_with :user, :avatar
+
+  def call
+    user.update!(
+      avatar:,
+      version: user.version + 1
```
iHiD: I figure if this happens simultaneously with something else enough that the db clashes, then cloudfront won't have got round to caching the new one yet anyway, so I'm not worried about the race.

app/models/user.rb
```diff
@@ -284,11 +284,9 @@ def may_view_solution?(solution)
     solution.viewable_by?(self)
   end
 
-  memoize
+  # Don't memoize this as it can change when someone uploads their photo
   def avatar_url
-    return Rails.application.routes.url_helpers.url_for(avatar.variant(:thumb)) if avatar.attached?
-
-    super.presence || "#{Exercism.config.website_icons_host}/placeholders/user-avatar.svg"
+    "#{Exercism.config.website_avatars_host}/avatars/#{id}/#{version}"
```
iHiD: This is the key change that means rather than looking up the url or file ourselves, we always just push someone to `assets.exercism.org`, which either renders the cached version or hits the new Avatars controller above (but obviously as a seperate request)

app/commands/user/update_avatar.rb
```diff
@@ -0,0 +1,14 @@
...
```
ErikSchierboom: I somehow had to giggle when I saw we were incrementing the version of a user. Sounds like a firmware update 😄

app/controllers/avatars_controller.rb
```diff
@@ -0,0 +1,31 @@
+require 'open-uri'
+
+# This should inhert from a name ActionController::Base, not ApplicationController
```
ErikSchierboom:
```suggestion
# This should inherit from ActionController::Base, not ApplicationController
```

app/controllers/profiles_controller.rb
```diff
@@ -1,7 +1,6 @@
 class ProfilesController < ApplicationController
-  skip_before_action :authenticate_user!, except: %i[intro new create]
```
ErikSchierboom: Does removing this mean that someone can't access some of the actions (like the `index` and `solutions` actions) without being logged in?

app/controllers/profiles_controller.rb
```diff
@@ -1,7 +1,6 @@
...
-  before_action :use_user, except: %i[index intro new create]
-  before_action :use_profile, except: %i[index intro new create tooltip]
+  before_action :use_user, except: %i[index intro new create avatar]
+  before_action :use_profile, except: %i[index intro new create tooltip avatar]
```
ErikSchierboom: Where is this `avatar` action defined? I don't see it in this controller.

app/commands/user/invalidate_avatar_in_cloudfront.rb
```diff
@@ -0,0 +1,36 @@
+class User::InvalidateAvatarInCloudfront
```
ErikSchierboom, line 1: This isn't yet used, correct?

iHiD: Used here: https://github.com/exercism/website/pull/5980/files#diff-aee3db72f1aac6b9ee5196b908407de2b4cd53e11ae655c10ee2aa401b2a8255R12

PR #5988, opened 2023-08-24
Refactor MentoringRequestForm

dem4ron:
Refactored the two very similar MentoringRequestForms.


app/javascript/components/student/mentoring-session/mentoring-request/MentoringRequestFormComponents/TrackObjectivesTextArea.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import type { MentorSessionTrack } from '@/components/types'
+
+export const TrackObjectivesTextArea = React.forwardRef<
```
ErikSchierboom, line 4: Any reason why you didn't go for:

```suggestion
export const TrackObjectives = React.forwardRef<
```

That way the component name is not tied to the implementation.

This same comment applies to the other component.

ErikSchierboom (approved):

iHiD (approved):

app/javascript/components/student/mentoring-session/mentoring-request/MentoringRequestFormComponents/TrackObjectivesTextArea.tsx
dem4ron: I agree with this, but it is harder to see in the markup that this is a textarea. that's the only reason I chose this name.

ErikSchierboom: Do you mean in the generated HTML or at the calling site? Is it important to know that it is a textarea?

PR #6008, opened 2023-08-26
Preload a single instance of the font to stop dramatic CLS

iHiD:
This preloads a single version of Poppins and sets it on body, which results in us making the initial render with Poppins, not the fallback font. The browser makes "guesstimate" versions of the other weights etc. This means that when Poppins loads, we don't get as dramatic layout shifts.

There's details here, although my version is a bit different: https://css-tricks.com/the-best-font-loading-strategies-and-how-to-execute-them/

---

https://www.loom.com/share/f92f54f56a31437ebe605a111271591a

### DIFF
--- a/app/css/components/site-header.css
+++ b/app/css/components/site-header.css
@@ -263,6 +263,9 @@ body.keyboard-navigation .nav-element:hover .arrow {
             @apply ml-8;
         }
     }
+    .auth-buttons + .c-react-wrapper-dropdowns-dropdown {
+        @apply min-w-[135px] lg:min-w-[0];
+    }
     & .auth-buttons a,
     & .explore-menu {
         height: 40px;
--- a/app/css/fonts.css
+++ b/app/css/fonts.css
@@ -64,14 +64,6 @@
     font-weight: 500;
     src: url("source-code-pro-v22-latin_latin-ext-500.woff2") format("woff2"); /* Chrome 36+, Opera 23+, Firefox 39+, Safari 12+, iOS 10+ */
 }
-/* source-code-pro-600 - latin_latin-ext */
-@font-face {
-    font-display: swap; /* Check https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display for other options. */
-    font-family: "Source Code Pro";
-    font-style: normal;
-    font-weight: 600;
-    src: url("source-code-pro-v22-latin_latin-ext-600.woff2") format("woff2"); /* Chrome 36+, Opera 23+, Firefox 39+, Safari 12+, iOS 10+ */
-}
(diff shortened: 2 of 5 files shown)


app/css/components/site-header.css
```diff
@@ -263,6 +263,9 @@ body.keyboard-navigation .nav-element:hover .arrow {
             @apply ml-8;
         }
     }
+    .auth-buttons + .c-react-wrapper-dropdowns-dropdown {
+        @apply min-w-[135px] lg:min-w-[0];
+    }
```
iHiD, line 268: This stops another dramatic layout shift.

app/css/fonts.css
```diff
@@ -64,14 +64,6 @@
...
     src: url("source-code-pro-v22-latin_latin-ext-500.woff2") format("woff2"); /* Chrome 36+, Opera 23+, Firefox 39+, Safari 12+, iOS 10+ */
 }
-/* source-code-pro-600 - latin_latin-ext */
-@font-face {
-    font-display: swap; /* Check https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display for other options. */
-    font-family: "Source Code Pro";
-    font-style: normal;
-    font-weight: 600;
-    src: url("source-code-pro-v22-latin_latin-ext-600.woff2") format("woff2"); /* Chrome 36+, Opera 23+, Firefox 39+, Safari 12+, iOS 10+ */
-}
```
iHiD, line 74: This wasn't actually being used anywhere.

app/css/components/site-header.css
iHiD: This should probably have been a seperate PR. Sorry @ErikSchierboom

app/views/layouts/_font_loading.html.haml
```diff
@@ -0,0 +1,16 @@
...
+    src: url(#{asset_path('poppins-v20-latin-regular.woff2')}) format('woff2');
+  }
+  body {
+    --body-font: Poppins, PoppinsInitial, serif;
+    font-family: var(--body-font);
+    -webkit-font-smoothing: antialiased;
+  }
+  body.fonts-loaded {
+    --body-font: Poppins, serif;
+  }
```
dem4ron, line 16: What does this do?

ErikSchierboom: @iHiD Something for you to answer when you're back

ErikSchierboom (approved):

app/views/layouts/_font_loading.html.haml
iHiD: It's not used any more. This is the route taken in the article I linked to, but my way is better and doesn't require JavaScript. I'll simplify this now.

PR #6050, opened 2023-08-29
Fix mail tests

ErikSchierboom:
@ihid CI was broken due to this change: https://github.com/exercism/website/commit/8108640e79c5d063d9d11557076818760b2e7ce5#diff-8f79448a02e89f9d3c4c7cf5cadb6c0bb492b7549586ba5b9ae85fd6c29b5d60R23 For now, I've updated the tests. If this was an unintended side-effect of the above commit, let me know and I'll try fixing it.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -134,14 +134,17 @@ else
 end
 
 class ActionMailer::TestCase
-  def assert_email(email, to, subject, fixture)
+  def assert_email(email, to, subject, fixture, bulk: false)
     # Test email can send ok
     assert_emails 1 do
       email.deliver_now
     end
 
+    tld = bulk ? "io" : "org"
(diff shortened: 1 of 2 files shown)


test/test_helper.rb
```diff
@@ -134,14 +134,17 @@ def ==(other)
...
 class ActionMailer::TestCase
-  def assert_email(email, to, subject, fixture)
+  def assert_email(email, to, subject, fixture, bulk: false)
     # Test email can send ok
     assert_emails 1 do
       email.deliver_now
     end
 
+    tld = bulk ? "io" : "org"
+    from = "hello@mail.exercism.#{tld}"
```
ErikSchierboom, line 144: We seem to be using a different top-level domain for bulk emails vs transactional emails.

PR #6051, opened 2023-08-29
Allow connecting to bugsnag in CSP

ErikSchierboom:
Just something I noticed in the developer tools.


iHiD:
Nice. That's been bugging me too.

PR #6055, opened 2023-08-29
Dynamically import highlightjs

dem4ron:
#### From 10.8 MiB (with restricted import pollution) to 9.4 MiB (when highlightJs is not imported on every page load)
Screenshot 2023-08-29 at 14 30 30
Screenshot 2023-08-29 at 15 09 53

https://github.com/exercism/website/assets/66035744/7ee27a2d-ea7c-4eff-ba7b-a024f34606fb


app/javascript/packs/application.tsx
```diff
@@ -668,7 +667,10 @@ initReact(mappings)
 import { handleNavbarFocus, scrollIntoView, showSiteFooter } from '@/utils'
 
 document.addEventListener('turbo:load', () => {
-  highlightAll()
+  // only load highlightjs when a code block with lang-... classname exists in the DOM
+  if (document.querySelector('code[class^=lang-]') !== null) {
```
ErikSchierboom: The class can also be `language-`, which is used by our Markdown editor

app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,54 @@
...
+          return
+        }
+      }
+    }
+  }
+
+  checkAndHighlight()
+
+  const observer = new MutationObserver(mutationCallback)
+  observer.observe(document.body, { childList: true, subtree: true })
```
iHiD, line 46: Nice.

app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,54 @@
...
+  setTimeout(() => {
+    observer.disconnect()
+  }, 2000)
+}
+
+function hasHighlightableCodeBlock(rootNode: Document | Element): boolean {
+  return (
+    rootNode.querySelector('code[class^=lang-]') !== null ||
+    rootNode.querySelector('code[class^=language-]') !== null
+  )
```
iHiD: Can some codeblocks not have languages attached? Do we normally rely on highlightjs to auto-determine the language in these situations? Should we just check for `code`?

ErikSchierboom: > Should we just check for `code`

That's a good suggestion.

iHiD: @dem4ron I don't think there's any real downside in just simplifying this to `code`. It removes the risk of edge-cases we've not thought of.

dem4ron: okay!

iHiD (approved):

app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,46 @@
...
+        observer.disconnect()
+      })
+      .catch((e) => {
+        observer.disconnect()
+        throw new Error(e)
+      })
+  }
+
+  function checkAndHighlight(): void {
+    // only load highlightjs when a code block with lang-... or language-... classname exists in the DOM
```
iHiD:
```suggestion
    // only load highlightjs when a code block exists in the DOM
```

app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  function mutationCallback(mutationsList: MutationRecord[]): void {
+    for (const mutation of mutationsList) {
+      if (mutation.type !== 'childList') continue
+      for (const node of mutation.addedNodes) {
+        if (node.nodeType !== Node.ELEMENT_NODE) continue
+        if (hasHighlightableCodeBlock(node as Element)) {
+          checkAndHighlight()
+          return
```
iHiD: This bit is a little messy, as it calls `checkAndHighlight` which then does another check for `hasHighlightableCodeBlock` on the whole document.

I'd suggest either:
1. Passing a `force` attribute (set to `false` by default) into checkAndHighlight to avoid this; or
2. Break out L21-22 into a different function that's called here and in checkAndHighlight().

Option 2 is probably better

app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    // to avoid race conditions, set this optimistically to true before the async loading
+    highlighted = true
+    loadAndHighlight()
+  }
+
+  function mutationCallback(mutationsList: MutationRecord[]): void {
+    for (const mutation of mutationsList) {
+      if (mutation.type !== 'childList') continue
+      for (const node of mutation.addedNodes) {
```
iHiD:
```suggestion
      if (mutation.type !== 'childList') continue
      
      for (const node of mutation.addedNodes) {
```

iHiD: Always good to put a linebreak after guards.

dem4ron: Adding those!

PR #6059, opened 2023-08-30
Fix ICON 404 error

dem4ron:


test/serializers/serialize_track_for_select_test.rb
```diff
@@ -15,7 +15,7 @@ class SerializeTrackForSelectTest < ActiveSupport::TestCase
     expected = {
       slug: nil,
       title: "All Tracks",
-      icon_url: ActionController::Base.helpers.image_path("graphics/missing-track.svg")
+      icon_url: image_path("graphics/missing-track.svg")
```
iHiD, line 18: @dem4ron This is the fix you'll need to put in places where this fails.

dem4ron: Thanks, this is great!

PR #6061, opened 2023-08-30
Fix user switches pages test

dem4ron:


test/system/flows/contributing/user_views_tasks_list_test.rb
```diff
@@ -176,7 +176,7 @@ class UserViewsTasksListTest < ApplicationSystemTestCase
 
         use_capybara_host do
           visit contributing_tasks_path
-          click_on "2"
+          find('button[aria-label="Go to page 2"]').click
```
ErikSchierboom: Could it be that there is another "2" that can be clicked on? Maybe you can do something like:

```
within("XXX") { click_on "2" }
```

ErikSchierboom (approved):

test/system/flows/contributing/user_views_tasks_list_test.rb
dem4ron: yeah, probably

PR #6062, opened 2023-08-30
Use observer for highlighting

dem4ron:


app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (highlighted) return
+
+    // only load highlightjs when a code block with lang-... or language-... classname exists in the DOM
+    if (
+      document.querySelector('code[class^=lang-]') !== null ||
+      document.querySelector('code[class^=language-]') !== null
+    ) {
+      import('@/utils/highlight').then((m) => {
+        m.highlightAll()
+        highlighted = true
```
ErikSchierboom: Could there be a race condition where `checkAndHighlight` is being called more than once?

dem4ron: ~It might be called more than once, but then it would just exit early.~ Yes, you are right.
I wonder if there is a case where a code block is loaded statically, we highlight it, then another code block comes dynamically. We wouldn't highlight that, because the function would exit - and the observer is disconnected.

app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,49 @@
...
+        })
+    }
+  }
+
+  function mutationCallback(mutationsList: MutationRecord[]): void {
+    for (const mutation of mutationsList) {
+      if (mutation.type === 'childList') {
+        for (const node of mutation.addedNodes) {
+          if (node.nodeType === Node.ELEMENT_NODE) {
+            if (
```
ErikSchierboom: Maybe do some early returns to reduce nesting?

app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,49 @@
...
+  let highlighted = false
+
+  function checkAndHighlight(): void {
+    if (highlighted) return
+
+    // only load highlightjs when a code block with lang-... or language-... classname exists in the DOM
+    if (
+      document.querySelector('code[class^=lang-]') !== null ||
+      document.querySelector('code[class^=language-]') !== null
+    ) {
```
ErikSchierboom: Maybe move this logic to a separate function? I would probably also remove the logic here and put it around the `checkAndHighlight()` that handles the initial highlighting.

app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,55 @@
...
+  function loadAndHighlight() {
+    import('@/utils/highlight')
+      .then((m) => {
+        m.highlightAll()
+        observer.disconnect()
+      })
+      .catch((e) => {
+        highlighted = false
+        // eslint-disable-next-line no-console
+        console.error(e)
```
ErikSchierboom: I wonder if we should disconnect here too. If there is an error that causes highlightjs to not load, it will likely not load the second time around either

ErikSchierboom (commented):
Maybe you could add a system test for the bit that was broken (the dig deeper highlighting)?

app/javascript/utils/lazy-highlight-all.ts
dem4ron: true!

app/javascript/utils/lazy-highlight-all.ts
```diff
@@ -0,0 +1,54 @@
...
+      })
+      .catch((e) => {
+        observer.disconnect()
+        throw new Error(e)
+      })
+  }
+
+  function checkAndHighlight(): void {
+    // only load highlightjs when a code block with lang-... or language-... classname exists in the DOM
+    if (highlighted || !hasHighlightableCodeBlock(document)) return
```
ErikSchierboom, line 18: I would remove the `hasHighlightableCodeBlock` call here and move it to the calling site. This is already done in `mutationCallback`, in which case the `hasHighlightableCodeBlock` function would be called twice.



```suggestion
  function highlight(): void {
    if (highlighted) return
```

PR #6066, opened 2023-09-01
Tweak Docker layer caching

ErikSchierboom:
Tweak the Docker layer caching by moving the GeoCity IP download to the end of the Dockerfile.
We have a cache busting mechanism in place that will invalidate the cache each month, but that shouldn't cause all the other layers to be invalidated too.


iHiD:
Nice thinking.

So the only disadvantage with this is that it needs to redownload/untar on every deploy. How about we move it below `RUN yarn install` instead? That way if the dependencies haven't changed, but the code has, it won't get redownloaded.

ErikSchierboom:
Sure. Updated

iHiD (approved):
Sweet. Thanks. I'll leave you to merge and check it works etc.

PR #6067, opened 2023-09-01
Ignore more directories when building Docker image

ErikSchierboom:


.dockerignore
```diff
@@ -1,8 +1,15 @@
+.appends/
+.built-assets/
```
iHiD, line 2: The assets get compiled into this directory as part of the building process. I presume this won't affect this?
I also note that this directory doesn't actually exist in git.

ErikSchierboom: What this line will do is that when we do `COPY . ./`, it won't copy the local directory too. A couple of lines after this `COPY` command, we have `RUN bundle exec rails assets:precompile`, which creates the directory, so it was always useless I think.

iHiD (approved):
Great. I'll leave you to merge.

PR #6069, opened 2023-09-01
Migrate to React 18 and Tanstack Query v4

dem4ron:


app/javascript/components/common/CLIWalkthrough.tsx
```diff
@@ -9,15 +9,18 @@ declare global {
...
 export default ({ html }: { html: string }): JSX.Element => {
+  const hasRun = useRef(false)
+  const scrollToTop = () => {
+    window.scrollTo({ top: 0 })
+  }
   useEffect(() => {
-    window.story = new Story($('tw-storydata'))
-    window.story.start($('#main'))
+    if (!hasRun.current) {
+      window.story = new Story($('tw-storydata'))
```
ErikSchierboom, line 18: I didn't know we used jQuery on the website. It looks like this is the only place it is being used. If we ever have some spare time, it might be useful to check if we could remove that dependency.

app/helpers/react_components/mentoring/representations/with_feedback.rb
```diff
@@ -24,8 +24,7 @@ def representations_request
             endpoint: Exercism::Routes.with_feedback_api_mentoring_representations_url,
             query: representations_request_params,
             options: {
-              initial_data: representations,
-              stale_time: 5000 # milliseconds
```
ErikSchierboom, line 28: Is `stale_time` no longer a thing?

app/controllers/api/solutions_controller.rb
```diff
@@ -28,7 +28,7 @@ def show
     output = {
       solution: SerializeSolution.(@solution)
     }
-    output[:iterations] = SerializeIterations.(@solution.iterations) if sideload?(:iterations)
+    output[:iterations] = SerializeIterations.(@solution.iterations, sideload: %i[files automated_feedback]) if sideload?(:iterations)
```
ErikSchierboom: Was there data missing without the sideloading?

app/javascript/components/Editor.tsx
```diff
@@ -191,10 +190,13 @@ export default ({
             })
         }
 
-        dispatch({
-          status: EditorStatus.CREATE_SUBMISSION_FAILED,
-          error: await editorError,
-        })
+        if (editorError) {
+          const errorResult = await editorError
```
ErikSchierboom: Any particular reason why you move this to a separate line?

app/javascript/components/contributing/ContributorsList.tsx
```diff
@@ -39,14 +40,19 @@ export default function ContributorsList({
...
+  const {
+    status,
+    data: resolvedData,
+    isFetching,
+    error,
+  } = usePaginatedRequestQuery<PaginatedResult<readonly Contributor[]>>(
+    ['contributors-list', request.endpoint, request.query],
+    {
+      ...request,
+      options: { ...request.options },
```
ErikSchierboom: Can't this just be:

```suggestion
      ...request
```

app/javascript/components/dropdowns/Reputation.tsx
```diff
@@ -101,12 +101,16 @@ export default function Reputation({
...
-    })
+  const {
+    data: resolvedData,
+    error,
+    status,
+    refetch,
+  } = usePaginatedRequestQuery<APIResponse>([cacheKey], {
+    endpoint: endpoint,
+    query: { per_page: MAX_TOKENS },
+    options: {},
```
ErikSchierboom, line 112: Do you need to pass in options, or does it have a sensible default too?

app/javascript/components/editor/useSubmissionsList.ts
```diff
@@ -23,8 +23,8 @@ export const useSubmissionsList = (
 } => {
   const [list, setList] = useState(defaultList)
 
-  const [create] = useMutation<Submission, unknown, File[]>(
-    (files) => {
+  const { mutate: create } = useMutation<Submission, unknown, File[]>(
+    async (files) => {
```
ErikSchierboom, line 27: I believe the `async` modifier can be omitted here as no `await` is used.

app/javascript/components/journey/ContributionsList.tsx
```diff
@@ -45,12 +46,17 @@ export function ContributionsList({
...
+  const {
+    status,
+    data: resolvedData,
+    isFetching,
+    error,
+  } = usePaginatedRequestQuery<APIResult>(cacheKey, {
+    ...request,
+    query: removeEmpty(request.query),
+    options: { ...request.options, enabled: isEnabled },
+  })
```
ErikSchierboom, line 57: I've seen this pattern a couple of times, where empty values are removed and `enabled` is set. It could be useful to, at some point, extract this to a helper.

app/javascript/components/mentoring/automation/useAutomation.tsx
```diff
@@ -65,11 +66,27 @@ export function useAutomation(
...
-      request
-    )
+  const CACHE_KEY = [
+    'mentor-representations-list',
+    ...Object.values(request.query),
+  ]
+
+  const queryClient = useQueryClient()
+
+  queryClient.setQueryData([CACHE_KEY], request.options.initialData)
```
ErikSchierboom: Why explicitly set the initial data here? I think we use that lots in our code, but this seems to be the only place you explicitly set it.

tsconfig.json
```diff
@@ -15,6 +15,7 @@
     "noEmit": true,
     "allowSyntheticDefaultImports": true,
     "resolveJsonModule": true,
+    "noImplicitAny": false,
```
ErikSchierboom, line 18: Might be useful to try and get this to `true` in the future

app/javascript/packs/application.tsx
```diff
@@ -178,15 +178,16 @@ const PerksExternalModalButton = lazy(
...
   interface Window {
     Turbo: typeof import('@hotwired/turbo/dist/types/core/index')
-    queryCache: QueryCache
+    queryClient: QueryClient
   }
 }
-
-import { QueryCache } from 'react-query'
-window.queryCache = new QueryCache()
+// use query client by pulling it out of the provider with useQueryClient hook
```
ErikSchierboom, line 188: Is this a TODO? If so, maybe add "TODO: " at the start

app/javascript/components/track/UnlockHelpButton.tsx
```diff
@@ -39,7 +39,7 @@ export function UnlockHelpButton({
     )
   }
 
-  const [unlockHelp, { error }] = useMutation(() => UnlockHelp(), {
+  const { mutate: unlockHelp, error } = useMutation(async () => UnlockHelp(), {
```
ErikSchierboom: No `await` is used, so `async` is unneeded I think

app/javascript/components/track/Trophies.tsx
```diff
@@ -181,8 +181,8 @@ const UnrevealedTrophy = ({
   updateTrophy: (trophy: Trophy) => void
 }): JSX.Element => {
   const [showError, setShowError] = useState(false)
-  const [mutation] = useMutation(
-    () => {
+  const { mutate: mutation } = useMutation(
+    async () => {
```
ErikSchierboom, line 185: No `await` was used, so why use `async`?

app/javascript/hooks/use-latest-data.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useState } from 'react'
+
+export function useLatestData<T>(data: T): T | undefined {
```
ErikSchierboom: I'm not entirely sure what this hook is used for. Could you explain?

app/javascript/components/mentoring/session/FileViewer.tsx
```diff
@@ -20,9 +20,8 @@ export const FileViewer = ({
         data-highlight-line-numbers={true}
         data-highlight-line-number-start={1}
         style={{ tabSize: indentSize }}
-      >
-        {file.content}
-      </code>
+        dangerouslySetInnerHTML={{ __html: file.content }}
```
ErikSchierboom: Do we need to do this? `file.content` is user-submitted and not sanitized, which means we'd open ourselves up to XSS attacks.

app/javascript/components/modals/student/FinishMentorDiscussionModal.tsx
```diff
@@ -56,8 +57,15 @@ const Inner = ({
 }): JSX.Element => {
   const [currentStep, send] = useMachine(modalStepMachine)
   const [report, setReport] = useState<MentorReport | null>(null)
+  useLogger('report', report)
```
ErikSchierboom: Can this be removed?

app/javascript/components/common/CLIWalkthrough.tsx
iHiD: Yeah - and as a short-term thing, we should ensure we're only loading it on the page its needed on etc.

app/helpers/react_components/mentoring/representations/with_feedback.rb
dem4ron: It is. However, I've relocated them to the frontend. Given that the fetching occurs on the frontend, it seems logical to manage the corresponding options there rather than on the backend.

app/javascript/components/common/CLIWalkthrough.tsx
dem4ron: Yes, me neither. and this is a strange pattern, since rendering shouldn't be done by a useEffect. I didn't change this, as this has little to do with this PR, but I'd like to get rid of this dep too.

app/controllers/api/solutions_controller.rb
dem4ron: I've reverted this now, but on any refetch automated_feedback would be missing. Which raises the question whether we need to do any fetching on iterationsList (`/tracks/[track]/exercises/[exercise]/iterations`). I think we don't.

In another PR I'd like to change how this page works, handles data. Because here the data is held inside the query cache, but since we don't do any fetching, this makes little sense. The data inside the cache comes from ruby in the form of `initial_data`. 
We should just hold this initial data in a useState state. We do mutation on deleting iterations, but we can filter frontend data using the response of that. So to me this pattern feels more complex than it should be.

app/javascript/components/Editor.tsx
dem4ron: No, I was playing with it, and didn't revert. Thanks!

app/javascript/components/contributing/ContributorsList.tsx
dem4ron: Yes, this should be simply `request`. I expanded the options object, but ended up not needing any of that.

app/javascript/packs/application.tsx
dem4ron: No, this is the method you can use to access data within the queryClient, similar to how useContext is used. Consider this a note or instruction. Since this behavior is new, I thought it would be helpful to comment on it.

app/javascript/components/track/Trophies.tsx
dem4ron: Yes, these changes were made using jscodeshift. I used a script provided by TanStack Query for a simpler migration

app/javascript/hooks/use-latest-data.ts
dem4ron: in v2 there used to be a latestData returned by useQuery, since v3 it is not returned, and I've added this to not break things, but now it is safe to remove, so I just did.

ErikSchierboom (approved):

PR #6078, opened 2023-09-05
Gracefully handle discord 404s

ErikSchierboom:


iHiD:
Nice. Thanks.

Should we defer these too when someone deletes an account?

app/commands/user/reset_account.rb
```diff
@@ -20,7 +20,7 @@ def call
     )
 
     user.avatar&.destroy
-    User::RemoveRoles.(user, user.roles)
+    User::RemoveRoles.defer(user, user.roles.to_a)
```
ErikSchierboom, line 23: The `to_a` is needed as `roles` is a `Set`, which sidekiq doesn't support

iHiD: Good catch.

PR #6083, opened 2023-09-05
Fix caching 0 on reputation periods

iHiD:
Fixes this:
Screenshot 2023-08-30 at 11 30 02

I could probably refactor this to be nicer, but I think it's ok for now.

### DIFF
--- a/app/commands/user/reputation_token/calculate_contextual_data.rb
+++ b/app/commands/user/reputation_token/calculate_contextual_data.rb
@@ -111,7 +111,7 @@ class User::ReputationToken::CalculateContextualData
 
       # Sometimes we get [] calculated/cached. I don't know why.
       # But when that happens, let's not use the cached version!
-      return parsed_value if parsed_value.present?
+      return parsed_value if parsed_value.present? && parsed_value != 0
     end
 
     # Or yield and cache a new one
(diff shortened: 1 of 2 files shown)


app/commands/user/reputation_token/calculate_contextual_data.rb
```diff
@@ -109,9 +109,10 @@ def with_cache(user_id, key)
     if cached
       parsed_value = JSON.parse(cached)
 
-      # Sometimes we get [] calculated/cached. I don't know why.
-      # But when that happens, let's not use the cached version!
-      return parsed_value if parsed_value.present?
+      # Sometimes, due to someon e checking the page while the periods are
+      # being updated in their nightly cycle, this data is wrong (either
+      # empty array or 0). We shouldn't honour this.
+      return parsed_value if parsed_value.present? && parsed_value != 0
```
iHiD:
```suggestion
      # Sometimes, due to someone checking the contributors page while the reputation periods
      # are being updated in their nightly cycle, this data is wrong (either empty array or 0). 
      # We shouldn't honour this and should act as if this isn't cached.
      return parsed_value if parsed_value.present? && parsed_value != 0
```

ErikSchierboom (approved):

PR #6095, opened 2023-09-05
Add canonical urls

iHiD:


app/views/layouts/_meta_tags.html.haml
```diff
@@ -7,7 +10,7 @@
 %meta{ name: "language-server-url", content: Exercism.config.language_server_url }
 %meta{ name: "bug-reports-url", content: Exercism::Routes.api_bug_reports_url }
 %meta{ name: "parse-markdown-url", content: Exercism::Routes.api_parse_markdown_url }
-%meta{ name: "user-id", content: current_user ? current_user.id : nil }
+%meta{ name: "user-id", content: current_user&.id }
```
iHiD, line 13: This is rubocop

app/views/about/organisation_supporters.html.haml
```diff
@@ -1,5 +1,6 @@
 - content_for :meta_title, "Supporting Exercism"
 - content_for :meta_description, "Can your organisation help Exercism grow? Become rocket fuel for our mission!"
+- content_for :canonical_url, Exercism::Routes.about_partners_url
```
iHiD: This is the same as the one below.

app/views/community/videos/index.html.haml
```diff
@@ -1,3 +1,4 @@
+- content_for :canonical_url, Exercism::Routes.community_videos_url
```
iHiD: For nearly all of these, you don't need `Exercism::Routes`. You only need that really if you're outside of the view context. The only exception is `published_solution_url` which is a weird custom method.

app/views/layouts/_meta_tags.html.haml
```diff
@@ -1,3 +1,6 @@
+- if canonical_url
```
iHiD:
```suggestion
- if canonical_url.present?
```

PR #6098, opened 2023-09-06
Further CLS improvements

iHiD:
Various small tweaks to improve CLS, commented on individual lines.

### DIFF
--- a/app/css/components/makers-button.css
+++ b/app/css/components/makers-button.css
@@ -1,5 +1,5 @@
 .c-react-wrapper-track-exercise-makers-button {
-    @apply flex;
+    @apply flex min-h-[36px];
 }
 .c-makers-button {
     @apply flex items-center justify-end;
--- a/app/javascript/utils/scroll-into-view.ts
+++ b/app/javascript/utils/scroll-into-view.ts
@@ -21,10 +21,25 @@ export function scrollIntoView(): void {
 
 function scrollElementsIntoView(elements: NodeListOf<Element>): void {
   elements.forEach((element) => {
+    if (isInViewport(element)) {
+      return
+    }
+
     element.scrollIntoView({
       behavior: 'instant',
       block: 'center',
       inline: 'center',
     })
   })
+}
+
+function isInViewport(element) {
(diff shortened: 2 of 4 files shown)


app/css/components/makers-button.css
```diff
@@ -1,5 +1,5 @@
 .c-react-wrapper-track-exercise-makers-button {
-    @apply flex;
+    @apply flex min-h-[36px];
```
iHiD, line 2: This stops a CLS on exercise pages.

app/javascript/utils/scroll-into-view.ts
```diff
@@ -21,10 +21,25 @@ export function scrollIntoView(): void {
 
 function scrollElementsIntoView(elements: NodeListOf<Element>): void {
   elements.forEach((element) => {
+    if (isInViewport(element)) {
+      return
+    }
+
```
iHiD, line 27: Around the time we added this code we took a massive hit on CLS metrics with Google. Sometimes when testing, I get a weird scroll where the page scrolls down then back up again. I think this might fix it (since I've added it, I've not seen the same issue). 

I'd like to reconsider this in general if it continues to be an issue.

app/views/layouts/_font_loading.html.haml
```diff
@@ -1,12 +1,25 @@
 / We want to get the holding font rendered straight away
 %link{ rel: "preload", href: asset_path('poppins-v20-latin-regular.woff2'), as: "font", type: "font/woff2", crossorigin: :anonymous }
+%link{ rel: "preload", href: asset_path('poppins-v20-latin-600.woff2'), as: "font", type: "font/woff2", crossorigin: :anonymous }
```
iHiD, line 3: We still get a CLS on some pages due to the lack of bold font. This is now an extra step where we render a 600 weight font as 500/600/700. This will help on some pages (e.g. it stops CLS on the JS exercises). 

If we try and do the same beyond this (e.g. preload 500/600/700) we're going to start not having any value from this.

Note: `font-display: fallback` gives us 100ms to get the font loaded. The font is tiny so we should be in that window.

PR #6112, opened 2023-09-07
Use active or overdue where applicable

ErikSchierboom:


app/controllers/api/payments/active_subscriptions_controller.rb
```diff
@@ -1,5 +0,0 @@
-class API::Payments::ActiveSubscriptionsController < API::BaseController
-  def show
-    render json: AssembleActiveSubscription.(current_user)
-  end
```
ErikSchierboom, line 4: This should always have been a member on the `SubscriptionsController` class

ErikSchierboom (commented):
The main result of this is that we now acknowledge overdue subscriptions in the donation form modals

app/models/user.rb
```diff
@@ -222,16 +222,10 @@ def formatted_reputation(*args)
   end
 
   memoize
-  def active_subscription? = subscriptions.active.exists?
+  def current_subscription? = current_subscription.present?
```
iHiD, line 225: I think we still want `active_subscription` for use in `app/commands/user/insiders_status/determine_eligibility_status.rb`

iHiD: (In fact, I don't understand how this passes the tests with it removed!)

ErikSchierboom: `active_subscription` in `determine_eligibility_status` does not use the method on user :)

iHiD: Ah!

iHiD (approved):
Go for it!

PR #6117, opened 2023-09-07
Remove presentation role from images

dem4ron:


app/javascript/components/common/ExerciseIcon.tsx
```diff
@@ -22,7 +22,6 @@ export function ExerciseIcon({
       className={classNames.join(' ')}
       src={iconUrl}
       alt={title ? `Icon for exercise called ${title}` : undefined}
```
dem4ron:
```suggestion
      alt={title ? `Icon for exercise called ${title}` : ''}
```

iHiD:
@dem4ron This fails.

app/helpers/icons_helper.rb
```diff
@@ -1,6 +1,6 @@
 module IconsHelper
   def graphical_icon(icon, style: nil, css_class: nil, category: 'icons', hex: false, width: nil, height: nil)
-    icon(icon, nil, role: :presentation, style:, css_class:, category:, hex:, width:, height:)
+    icon(icon, nil, alt: '', style:, css_class:, category:, hex:, width:, height:)
```
ErikSchierboom: Isn't empty alt text bad?

iHiD: It's the correct way to do icons without semantic meaning (ie things just there for visual purpose, which is what `graphical_icon` is). 

This doesn't actually change anything as we previously passed `nil` as the second param. `role: presentation` isn't actually valid HTML or widly supported by screenreaders (which was the intention when we added it).

That said, this **is** wrong. The second param is the alt text, so this should be:
```suggestion
    icon(icon, "", style:, css_class:, category:, hex:, width:, height:)
```

PR #6123, opened 2023-09-08
Representation search UI

ErikSchierboom:
- Create exercise representation opensearch documents
- View community solutions via exercise representation


app/commands/solution/search_via_representations.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    def filter!
+      # By grouping, we force MySQL to return just one result per group
+      @solutions = @solutions.group(:published_exercise_representation_id)
+    end
+
+    def sort!
+      case order
+      when :most_popular
+        @solutions = @solutions.order(num_published_solutions: :desc, id: :asc)
```
ErikSchierboom: There is no index on this field (or combination of fields)

app/commands/solution/search_via_representations.rb
```diff
@@ -0,0 +1,148 @@
...
+      when :newest
+        @solutions = @solutions.order(id: :desc)
+      when :oldest
+        @solutions = @solutions.order(id: :asc)
+      when :fewest_loc
+        @solutions = @solutions.order(num_loc: :desc)
```
ErikSchierboom: There is no index on this field.

PR #6140, opened 2023-09-13
Destroy account fix

ErikSchierboom:
- Add test for cleaning up viewed community solutions
- Assign payments and subscriptions to ghost user when destroying account

### DIFF
--- a/app/commands/user/destroy_account.rb
+++ b/app/commands/user/destroy_account.rb
@@ -7,8 +7,8 @@ class User::DestroyAccount
     User::ResetAccount.(user)
     user.student_relationships.destroy_all
     user.mentor_relationships.destroy_all
-    user.subscriptions.update_all(user_id: User::GHOST_USER_ID)
-    user.payments.update_all(user_id: User::GHOST_USER_ID)
+    user.subscriptions.destroy_all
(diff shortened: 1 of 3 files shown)


app/commands/user/destroy_account.rb
```diff
@@ -7,6 +7,8 @@ def call
     User::ResetAccount.(user)
     user.student_relationships.destroy_all
     user.mentor_relationships.destroy_all
+    user.subscriptions.update_all(user_id: User::GHOST_USER_ID)
+    user.payments.update_all(user_id: User::GHOST_USER_ID)
```
ErikSchierboom: The association sets the `:dependent` value to `:nullify`, but that is invalid as the `user_id` column is not nullable for payments or subscriptions. So we can either:

1. Make it nullable
2. Assign it to the ghost user

iHiD: Or maybe just destroy them? We probably don't need the record in the database any more?

ErikSchierboom: I had that originally, but then saw that you have `nullify` explicitly defined so I didn't dare. Let me update

iHiD (approved):

PR #6141, opened 2023-09-13
Cancel pending subscriptions

ErikSchierboom:
- Rename SyncSponsorsJob to SyncPaymentsJob
- Change payment sync job schedule to daily
- Cancel pending subscriptions in job
- Cancel pending subscriptions

### DIFF
--- a/config/sidekiq-schedule.yml
+++ b/config/sidekiq-schedule.yml
@@ -41,6 +41,11 @@ mark_outdated_reputation_year_periods:
   queue: reputation
   args: ["year"]
 
+sync_payments:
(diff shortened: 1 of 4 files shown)


config/sidekiq-schedule.yml
```diff
@@ -41,6 +41,11 @@ mark_outdated_reputation_year_periods:
   queue: reputation
   args: ["year"]
 
+sync_payments:
+  cron: "0 5 * * *"
```
ErikSchierboom, line 45: I've changed this to run every day

iHiD (approved):

PR #6145, opened 2023-09-13
Split out page CSS

iHiD:
Proof of concept that splits out CSS into multiple files. See Loom on Slack.


app/views/layouts/application.html.haml
```diff
@@ -9,6 +9,7 @@
 
     = render "layouts/meta_tags"
 
+
```
iHiD:
```suggestion
```

PR #6161, opened 2023-09-15
Rerun when dependencies install fails

ErikSchierboom:
- Split up yarn and exercism dependencies
- Minor renaming
- Rerun job when exercism services fail

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -197,14 +197,10 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
-      ###
-      # Install yarn dependencies
       - name: Install yarn dependencies
         run: yarn install
 
-      ###
-      # Install bundler and dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -197,16 +197,16 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
-      ###
-      # Install bundler and yarn dependencies
-      - name: Install dependencies
+      - name: Install yarn dependencies
+        run: yarn install
+
+      - name: Setup Exercism services
```
ErikSchierboom, line 203: I assume that we want to re-run when either `bundle exec setup_exercism_config` or `bundle exec setup_exercism_local_aws` fails. I've excluded `yarn` because that could fail due to us messing up the `package.json` file, which would lead to endless re-running.

iHiD:
This is great. Thanks :)

PR #6203, opened 2023-09-18
Split out CSS

iHiD:
@ErikSchierboom The dockerfile failed to build with an error during deploy, so I reverted this and reopened it. Could you see if you can build it locally and work this out pls?
 
Screenshot 2023-09-18 at 15 08 35


iHiD:
(cc @dem4ron in case he has any ideas/insight!)

PR #6210, opened 2023-09-19
Add generic representer

iHiD:
I **think** this will work pretty easily.

### DIFF
--- a/app/commands/submission/representation/init.rb
+++ b/app/commands/submission/representation/init.rb
@@ -11,19 +11,27 @@ class Submission::Representation::Init
   end
 
   def call
-    ToolingJob::Create.(
-      submission,
-      :representer,
-      git_sha:,
-      run_in_background:,
-      context:
-    ).tap do
+    queue_job!.tap do
       update_status!
     end
   end
 
   private
   attr_reader :submission, :type, :git_sha, :run_in_background
+
+  def queue_job!
+    return Submission::Representation::GenerateBasic.defer(submission) unless track.has_representer?
+
+    ToolingJob::Create.(
+      submission,
+      :representer,
+      git_sha:,
+      run_in_background:,
+      context:
+    )
+  end
+
+  delegate :track, to: :submission
 
   def update_status!
     return if type == :exercise
--- a/app/commands/submission/representation/process.rb
+++ b/app/commands/submission/representation/process.rb
@@ -4,97 +4,23 @@ class Submission::Representation::Process
   initialize_with :tooling_job
 
   def call
-    if reason == :update
-      handle_update!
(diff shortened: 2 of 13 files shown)


app/commands/submission/representation/process.rb
```diff
@@ -4,99 +4,25 @@ class Submission::Representation::Process
   initialize_with :tooling_job
 
   def call
```
iHiD, line 6: All the code below this is moved into `process_results` below

app/commands/submission/representation/init.rb
```diff
@@ -11,19 +11,27 @@ def initialize(submission, type: :submission, git_sha: nil, run_in_background: f
...
+    queue_job!.tap do
+      update_status!
+    end
+  end
+
+  private
+  attr_reader :submission, :type, :git_sha, :run_in_background
+
+  def queue_job!
+    return Submission::Representation::GenerateBasic.defer(submission) unless track.has_representer?
```
iHiD, line 23: This is the change to this file.

app/commands/submission/representation/process_results.rb
```diff
@@ -0,0 +1,87 @@
+class Submission::Representation::ProcessResults
```
iHiD, line 1: This is the code that was deleted above.

app/commands/submission/representation/process.rb
iHiD: This file is now effectively responsible for taking a tooling job and extracting the relevant data to pass to `processresults`

test/commands/submission/representation/generate_basic_test.rb
```diff
@@ -0,0 +1,164 @@
...
+
+  test "normalises ast correctly" do
+    filename_1 = "start.rb"
+    code_1 = "foo eeee"
+    filename_2 = "end.rb"
+    code_2 = "bar wwww \n aaaa"
+
+    # Check things are normalised correctly
+    digest = Submission::Representation.digest_ast("#{filename_1}\n#{code_1.gsub(/\s/,
+      '')}\n\n#{filename_2}\n#{code_2.gsub(/\s/, '')}")
```
ErikSchierboom: I wonder if using a plain multiline string without the interpolated values and gsubbing, which makes reading it a bit harder I feel:

```suggestion
# Check things are normalised correctly
    expected_ast = <<~AST
      start.rb
      fooeeee

      end.rb
      barwwwwaaaa
    AST
    digest = Submission::Representation.digest_ast(expected_ast.chomp)
```

iHiD: Yep.

PR #6219, opened 2023-09-21
Extract profile css

iHiD:


iHiD:
@dem4ron Are there any tooltips or components that can also be brought into here (they need to be *only* things that show on the published profile pages)

PR #6220, opened 2023-09-21
Extract mentoring CSS

iHiD:
@dem4ron Most of the automation CSS seems to be a duplicate of the inbox CSS. Do we need both? I think this CSS could probably be tidied a bit.


iHiD:
@dem4ron Are there any tooltips or components that can also be brought into here (they need to be *only* things that show to mentors - but they could be any page under `exercism.org/mentoring` except for the external page).

dem4ron:
Yes, inbox and automation is almost the same. I copy-pasted it a year ago, it definitely can be cleaned up. Maybe create a `c-list` or `c-table-with-rows` or something for this kind of layout?

iHiD:
I added a missing file.

PR #6224, opened 2023-09-22
Update puma, grpc, and commonmarker to include CVE fixes

heliocola:
### In this PR
- [x] 1. Update  puma, grpc, and commonmarker gems to include CVE fixes


Gemfile
```diff
@@ -21,7 +21,7 @@ gem 'grpc', '>= 1.53.0'
 gem 'crawler_detect'
 
 # Serving requests
-gem 'puma', '~> 4.3'
+gem 'puma', '~> 5.6'
```
heliocola, line 24: I've updated puma to the the smallest version that has all CVEs fixed.
Puma is usually, IME, easy to upgrade so maybe this was behind for some specific and know reason.
If that is the case I can revert this upgrade the others.

heliocola:
I just noticed the PR closed and the message.
The other 2 gems (grpc and commonmarker) had only minor version updated and I was planning on letting CI run with the PR as draft.

Anyway... if this gets opened for consideration I can review in more details if these gems update requires more tests or need any further code changes.

.gitignore
```diff
@@ -50,3 +50,6 @@ dump.rdb
 
 # Ignore vendored local databases
 /vendor/*mmdb
+
+# RVM gemset
+.ruby-gemset
```
iHiD: Can you remove this pls?

heliocola: @iHiD : this file is useful for me, in my local development environment, as a rvm user.
This allows me to separate the gems installation for each project I have in my machine.

Usually, when a repo don't have neither `.ruby-gemset` itself, nor is ignored, I add it to the `.gitignore`.

Having it on `.gitignore` will prevent myself (or anybody else) to mistakenly push this file in a PR, as it will show in `git status`.

Is it a problem to have this file ignored?

Looking at the contents of `.gitignore`, this file is similar to `dump.rdb` which is a redis backup file, or `.cache` (cached used for docker).

heliocola: This is one example that ignores files for multiple ruby version managers: https://github.com/forem/forem/blob/main/.gitignore#L72-L79

Mastodon have it in the repo itself: https://github.com/mastodon/mastodon/blob/main/.ruby-gemset

Another one that ignores (that I've added): https://github.com/joemasilotti/railsdevs.com/blob/main/.gitignore#L51-L52

Anyway, if it is problematic to have this in `.gitignore` let me know and I will remove it.

iHiD: Hi @heliocola.

For local files that are not created by the code, it's generally better to include them in your own local exclusions list, rather than adding them to the repositories `.gitignore` file. This avoids having infinite different files listed for different editors, sets up. 

Things like `dump.rdb` and `.cache` are created by our codebase, so having them in a `.gitignore` file is useful as it stops everyone having to deal with the same issue. 

See [this section](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) on GitHub docs for more info on this :)

heliocola: It is not the first time I get this reply back.
I don't want to globally ignore this file because I do use it in all projects I manage.
`.gitignore` is the repo based solution to do this.

The repo already have `.gitignore`, `.dockerignore`, `.eslintignore`, `.ignore`, `.prettierignore`, so I find it really interesting that using this file for its purposes would have such a push back.

To be clear: this isn't a criticism in any shape or form. 
But I am always really surprised how much push back I get on adding this entry to `.gitignore`.

heliocola: My opinion is that having this entry in `.gitignore` (and other editors used by contributors) is a way to welcome users with all kinds of tools! :-D

Anyway, I was truly more concerned about the puma major version update than this one! :-D
I will remove it.

iHiD: When your project has 20,000 contributors that each use different IDEs, tools, etc, you quickly realise that if everyone starts configuring things to help with their usecases, you end up in chaos :)

heliocola: I have a different opinion about `.gitignore`, `projects with 20,000 contributors`, and `chaos in large OSS projects`, but I have already removed it as per your request.

heliocola: Just one final comment: this isn't the most important contribution on this PR and neither what motivated me to contribute.

iHiD: > Just one final comment: this isn't the most important contribution on this PR and neither what motivated me to contribute.

Sure. Thanks. I'm just continuing the discussion that was started above. But I'll leave it now if you're done with chatting about it :)

Gemfile
```diff
@@ -90,6 +90,8 @@ group :development, :test do
 
   # To avoid polling for changes on Windows MRI, it needs this opt-in gem
   gem 'wdm', '>= 0.1.0' if Gem.win_platform?
+
+  gem 'bundler-audit'
```
iHiD: What is the benefit of this over GitHub's Dependency Bot PRs?

.gitignore
heliocola: Just acting based on the initial reason the PR was closed due to pausing community contributions (ref: https://github.com/exercism/website/pull/6224#issuecomment-1730841435).
So maybe here isn't the proper place for these discussions. That's all.

Gemfile
heliocola: @iHiD : I don't want to advocate for neither.
I can remove it if you prefer.

heliocola: @iHiD : there is a plan to integrate `bundler-audit` with `bundler`, so I removed it.
Source: https://rubycentral.org/news/stf-announce/

iHiD: > I can remove it if you prefer.

I hadn't formed an opinion. It was a genuine question.

> There is a plan to integrate bundler-audit with bundler, so I removed it.

Interesting. Thanks for the source.

heliocola:
@iHiD : do you know if the code that uses `grpc` and `commonmarker` are covered with automated tests?
They are both c code and the code changes doesn't help me much get an idea if they will break something, but CI run and it is all green.

iHiD:
I'll defer to @ErikSchierboom to get this tested/deployed.

ErikSchierboom:
Thanks for the PR! Testing this will take some time, so I'll come back to this later.

PR #6246, opened 2023-09-26
Add CancelledMentorRequest modal

dem4ron:
https://github.com/exercism/website/assets/66035744/25c4098b-5288-4fff-8f3d-90b90da0b790


app/helpers/react_components/mentoring/session.rb
```diff
@@ -67,6 +67,7 @@ def mentor_student_relationship
       def links
         {
           mentor_dashboard: Exercism::Routes.mentoring_inbox_path,
+          mentor_queue: Exercism::Routes.mentoring_queue_path,
```
dem4ron, line 70: I don't know if we want to do this, or just simply take mentors back to inbox.

app/channels/mentor_request_channel.rb
```diff
@@ -9,10 +9,8 @@ def unsubscribed; end
 
   def self.broadcast!(request)
     broadcast_to request, {
-      mentor_request: {
```
dem4ron: I've simplified the structure. Within the scope of the channel, the extra nesting is unnecessary. It's already clear that we're dealing with a fragment of a mentor request.

app/javascript/channels/mentorRequestChannel.ts
```diff
@@ -0,0 +1,31 @@
+import consumer from '../utils/action-cable-consumer'
+import { MentorSessionRequest } from '../components/types'
+
+export type ChannelResponse = {
+  uuid: string
```
dem4ron: we don't need the `uuid`, but it is probably always nice to have an identifier around.

app/javascript/components/mentoring/session/CancelledRequestModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+      aria={{
+        modal: true,
+        describedby: 'cancelled-mentoring-request-description',
+      }}
+    >
+      <h3
+        id="cancelled-mentoring-request-description"
+        className="text-h3 mb-16 text-center"
+      >
+        The student has cancelled this mentoring request. <br />
```
dem4ron: @iHiD Please review the wording and the structure here. Perhaps it would be better to place the 'Sorry...' text inside a div and keep only 'The student..." text in the h3. Having a line break within the header looks odd. I just wanted to see how the text would appear on a new line.

app/javascript/components/mentoring/session/CancelledRequestModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+        Sorry for the inconvenience!
+      </h3>
+      <div className="flex justify-center gap-16">
+        <a href={links.mentorQueue} className="btn-m btn-primary">
+          Back to mentor requests
+        </a>
+        {isLocked && (
```
dem4ron: If the request is cancelled before it's locked, mentors might end up trying to lock a non-existent request. So, I thought it'd be best to force them to return to the dashboard.

app/serializers/serialize_mentor_session_request.rb
```diff
@@ -10,6 +10,7 @@ def call
       uuid: request.uuid,
       comment: SerializeMentorDiscussionPost.(Mentor::RequestComment.from(request), user),
       is_locked: request.locked?,
+      status: request.status,
```
dem4ron:
```suggestion
 
```

dem4ron: I've added this, but didn't end up using it. Do you think we might need it? Is there a chance mentors could accidentally open a cancelled request?

app/helpers/react_components/mentoring/session.rb
iHiD: Back to the queue is good.

app/channels/mentor_request_channel.rb
iHiD: I'm not sure about this. We may in the future want to have something like broadcasting meta info, e.g. what's changed or something, and that shouldn't be part of `mentor_request`. I prefer to have it nested.

dem4ron: okay

PR #6250, opened 2023-09-27
Improve screen reader user experience

dem4ron:
https://github.com/exercism/website/assets/66035744/e2545bef-8770-4fba-8821-bc6bfed57150


app/javascript/components/modals/PublishSolutionModal.tsx
```diff
@@ -49,10 +49,18 @@ export const PublishSolutionModal = ({
...
+        labelledby: 'publish-solution-label',
+        describedby: 'publish-solution-description',
+      }}
+      className="m-change-published-iteration"
+    >
       <form data-turbo="false" onSubmit={handleSubmit}>
-        <h3>Publish your solution</h3>
-        <p>
+        <h3 id="publish-solution-label">Publish your solution</h3>
+        <p id="publish-solution-description">
```
iHiD: Can we prefix these with `aria-{hex}`, where {hex} is a unique 5 length hex pls (e.g. `aria-8af73-publish-solution-label`)

Then we don't have to worry about accidently having the same id defined twice and trying to keep track of this globally.

dem4ron: sure, I think these could be automatically generated by a function, because we need these on all modals, and the pattern is very simple. `aria-{hex}-{modal-name}-[description|label]`. I'll play with this tomorrow.

iHiD: OK great :)

We can always get this green with this, and then have a second PR that adds them everywhere using a function?

app/javascript/components/modals/complete-exercise-modal/PublishSolutionModal.tsx
```diff
@@ -17,18 +18,23 @@ export const PublishSolutionModal = ({
   endpoint: string
   onSuccess: (data: ExerciseCompletion) => void
 }): JSX.Element => {
+  const ariaObject = generateAriaFieldIds('publish-code')
   return (
     <Modal
       cover={true}
       open={open}
+      role="dialog"
+      aria={ariaObject}
```
iHiD, line 27: 1. Are there modals where we **don't** want these? If so, we could consider moving them upstream. 
2. Consider passing `ariaId` in here, with the value of `publish-code` and move the `generateAriaFieldIds` call into `Modal` itself. That'll save having to have this same `const ariaObject = generateAriaFieldIds(...)` line of code scattered throughout the codebase.

app/javascript/components/modals/complete-exercise-modal/PublishSolutionModal.tsx
```diff
@@ -17,18 +18,23 @@ export const PublishSolutionModal = ({
   endpoint: string
   onSuccess: (data: ExerciseCompletion) => void
 }): JSX.Element => {
+  const ariaObject = generateAriaFieldIds('publish-code')
   return (
     <Modal
       cover={true}
       open={open}
+      role="dialog"
```
iHiD: Are there modals where we **don't** want this?

iHiD (approved):
If you're happy, I'm happy! :)

PR #6260, opened 2023-09-27
Refactor to allow automator role

iHiD:
This implements https://forum.exercism.org/t/maintainers-widening-access-to-representations-ui/7431/10

This PR only adds one thing - a boolean for `automator` on a track mentorship. But setting it in the right place(s) turned out to be complex, and so I refactored to reduce that. I've commented the PR with details. The actual functionality doesn't really change at all (as you can see by the fact there's almost no substantive changes to the tests)


app/assemblers/assemble_representation_context.rb
```diff
@@ -42,7 +42,7 @@ def representation_count
     def supermentored_tracks
       return Track.all if mentor.staff?
 
-      Track.where(id: mentor.track_mentorships.supermentor_frequency.select(:track_id)).order(title: :asc)
+      Track.where(id: mentor.track_mentorships.automator.select(:track_id)).order(title: :asc)
```
iHiD, line 45: We no longer use the scope that checked for frequency, but use the new scope that is the flag instead. You'll see this throughout.

app/commands/mentor/discussion/process_finished.rb
```diff
@@ -4,14 +4,16 @@ class Mentor::Discussion::ProcessFinished
   initialize_with :discussion
 
   def call
```
iHiD, line 6: There were a lot of potential race conditions in here. I've moved the logic of badges and roles into the `UpdateStats` call, as we can more firmly specify the order there.

app/commands/mentor/update_num_finished_discussions.rb
```diff
@@ -1,9 +1,8 @@
-class Mentor::Discussion::UpdateNumFinishedDiscussions
+class Mentor::UpdateNumFinishedDiscussions
```
iHiD, line 1: We want to be able to call this more generally than just on a single discussion, so we now pass in the mentor and track, rather than a discussion.

app/commands/mentor/update_stats.rb
```diff
@@ -3,9 +3,7 @@ class Mentor::UpdateStats
 
   queue_as :default
 
-  initialize_with :mentor,
-    update_counts: false,
-    update_satisfaction_rating: false
```
iHiD, line 8: We can't use these fields in the new way, but we don't actually need to, as the new code (using `sum` is more efficient, so we don't have the pain we used to).

app/commands/mentor/update_stats.rb
```diff
@@ -3,9 +3,7 @@ class Mentor::UpdateStats
...
+  initialize_with :mentor, :track
```
iHiD, line 6: We now also pass the `track` in here, which allows us to update number of finished discussions and the automator flag. This is probably the key change in the PR (or at least it was my big mental breakthrough), as it enables this command to do all the things in the correct order, without races and spagetti.

app/commands/mentor/update_stats.rb
```diff
@@ -21,10 +19,18 @@ def call
...
     new_values = old_values.clone
 
-    new_values[0] = User::ResetCache.(mentor, :num_students_mentored) if update_counts
-    new_values[1] = User::ResetCache.(mentor, :num_solutions_mentored) if update_counts
-    new_values[2] = Mentor::UpdateSatisfactionPercentage.(mentor) if update_satisfaction_rating
+    new_values[0] = User::ResetCache.(mentor, :num_students_mentored)
+    new_values[1] = User::ResetCache.(mentor, :num_solutions_mentored)
+    new_values[2] = Mentor::UpdateSatisfactionPercentage.(mentor)
 
-    User::UpdateMentorRoles.defer(mentor) if old_values != new_values
```
iHiD, line 28: This command got renamed to `UpdateSupermentorRole`, which is still called here, but just below.

app/commands/mentor/update_stats.rb
```diff
@@ -3,28 +3,26 @@ class Mentor::UpdateStats
...
   # - Triggers some extra checks if they've changed
   def call
-    # Actively use the cache to get these values as we don't want
-    # to follow the automatic setting pathway we normally use.
-    old_values = [
-      mentor.data.cache['num_students_mentored'],
-      mentor.data.cache['num_solutions_mentored'],
-      mentor.data.cache['mentor_satisfaction_percentage']
-    ]
-    new_values = old_values.clone
```
iHiD, line 22: We don't bother with the whole old vs new thing any more. It was necessary before as this command got called all the time, but now it's only called at specific relevant times, it's not worth the hassle.

app/commands/mentor/update_stats.rb
```diff
@@ -3,28 +3,26 @@ class Mentor::UpdateStats
...
-    new_values[1] = User::ResetCache.(mentor, :num_solutions_mentored) if update_counts
-    new_values[2] = Mentor::UpdateSatisfactionPercentage.(mentor) if update_satisfaction_rating
+    # Now we do the other essential checks
+    User::ResetCache.(mentor, :num_students_mentored)
+    User::ResetCache.(mentor, :num_solutions_mentored)
+    Mentor::UpdateSatisfactionPercentage.(mentor)
 
-    User::UpdateMentorRoles.defer(mentor) if old_values != new_values
+    # These can safely be defered into other jobs
+    User::UpdateSupermentorRole.defer(mentor)
```
iHiD, line 24: This is the same as `UpdateMentorRoles` - just renamed.

app/commands/user/update_mentor_roles.rb
```diff
@@ -1,14 +0,0 @@
...
-
-  def call
-    if Mentor::Supermentor.eligible?(user)
-      User::AddRoles.(user, [Mentor::Supermentor::ROLE])
-      AwardBadgeJob.perform_later(user, :supermentor)
-    else
-      User::RemoveRoles.(user, [Mentor::Supermentor::ROLE])
-    end
-  end
-end
```
iHiD, line 14: Renamed to UpdateSupermentorRole below.

app/commands/user/update_automator_role.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def eligible?
+    return false unless user.mentor?
+    return false unless mentorship
+    return false if user.mentor_satisfaction_percentage.to_i < MIN_SATISFACTION_PERCENTAGE
+    return false if mentorship.num_finished_discussions < MIN_FINISHED_MENTORING_SESSIONS
+
+    true
+  end
```
iHiD: Inlined from the old `supermentor` class.

app/commands/user/update_supermentor_role.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  end
+
+  def eligible?
+    return false unless user.mentor?
+    return false if user.mentor_satisfaction_percentage.to_i < MIN_SATISFACTION_PERCENTAGE
+    return false if user.track_mentorships.sum(:num_finished_discussions) < MIN_FINISHED_MENTORING_SESSIONS
+
+    true
+  end
```
iHiD, line 21: Inlined from the old `supermentor` class.

app/controllers/api/base_controller.rb
```diff
@@ -41,9 +41,9 @@ def ensure_onboarded!
       render_403(:not_onboarded)
     end
 
-    def ensure_supermentor!
-      return if current_user&.supermentor?
-      return if current_user&.admin? # Admins have supermentor permissions
+    def ensure_automator!
+      return if staff?
+      return if current_user&.track_mentorships&.automator&.exists?
```
iHiD, line 46: Rubocop did this horrible daisy-chaining 🤷

app/controllers/api/mentoring/representations_controller.rb
```diff
@@ -27,7 +27,7 @@ def admin
   def use_representation
     @representation = Exercise::Representation.find_by!(uuid: params[:uuid])
 
-    render_403(:not_supermentor_for_track) unless Mentor::Supermentor.for_track?(current_user, @representation.track)
+    render_403(:not_automator_for_track) unless current_user.automator?(@representation.track)
```
iHiD: As this is now much simpler (a db field basically), I've changed it to be a method on user rather than a whole extra class.

app/models/mentor/discussion.rb
```diff
@@ -111,13 +110,4 @@ def student_handle = anonymous_mode? ? "anonymous" : student.handle
...
   def student_avatar_url = anonymous_mode? ? nil : student.avatar_url
-
-  def update_stats!
-    Mentor::UpdateStats.defer(
-      mentor,
-      update_counts: previous_changes.key?('status'),
-      update_satisfaction_rating: previous_changes.key?('rating')
-    )
-    Mentor::Discussion::UpdateNumFinishedDiscussions.defer(self) if previous_changes.key?('status') && finished?
-  end
```
iHiD, line 122: We now only call all this during the finishing up process of a solution. Which makes much more sense!

app/models/mentor/supermentor.rb
```diff
@@ -1,20 +0,0 @@
...
-  end
-
-  def self.for_track?(mentor, track)
-    return true if mentor.staff? || mentor.admin?
-
-    mentor.mentor? &&
-      mentor.mentor_satisfaction_percentage.to_i >= MIN_SATISFACTION_PERCENTAGE &&
-      mentor.track_mentorships.supermentor_frequency.where(track:).exists?
-  end
-end
```
iHiD, line 20: This has been split out and inlined into the commands above.

config/locales/api/en.yml
```diff
@@ -3,7 +3,7 @@ en:
     errors:
       not_onboarded: "You have not yet onboarded on the website"
       not_supermentor: "You do not have supermentor permissions"
-      not_supermentor_for_track: "You do not have supermentor permissions for this track"
+      not_automator_for_track: "You do not have supermentor permissions for this track"
```
iHiD:
```suggestion
      not_automator_for_track: "You do not have automation permissions for this track"
```

test/commands/mentor/discussion/finish_abandoned_test.rb
```diff
@@ -16,74 +16,23 @@ class Mentor::Discussion::FinishAbandonedTest < ActiveSupport::TestCase
...
-        Mentor::Discussion::FinishAbandoned.()
-      end
-
-      assert_includes student.reload.trophies.map(&:class), Track::Trophies::MentoredTrophy
-    end
-
-    test "adds metric when abandoning from status #{status}" do
-      student = create :user
-      finished_at = Time.current.utc - 1.week
-      discussion = create(:mentor_discussion, status:, student:, finished_at:, updated_at: finished_at)
```
iHiD, line 66: All of this got moved into a dedicated test for the `ProcessFinished` command to DRY things up in the tests.

test/commands/mentor/discussion/finish_by_student_test.rb
```diff
@@ -146,68 +153,6 @@ class Mentor::Discussion::FinishByStudentTest < ActiveSupport::TestCase
...
-    perform_enqueued_jobs
-
-    assert_equal 1, Metric.count
-    metric = Metric.last
-    assert_instance_of Metrics::FinishMentoringMetric, metric
-    assert_equal discussion.finished_at, metric.occurred_at
-    assert_equal discussion.track, metric.track
-    assert_equal discussion.student, metric.user
-  end
-
```
iHiD, line 210: DRY'd up as per above.

app/commands/user/update_automator_role.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def mentorship = user.track_mentorships.find_by(track:)
+  def explicit? = automators.fetch(user.handle, []).include?(track.slug)
```
ErikSchierboom, line 24: I couldn't come up with a great name. Any ideas?

app/models/git/website_copy.rb
```diff
@@ -27,12 +27,25 @@ def mentor_notes_for_track(track_slug)
...
 
+    memoize
+    def automators = repo.read_json_blob(head_commit, "automators.json")
+
+    memoize
+    def walkthrough = repo.read_text_blob(head_commit, "walkthrough/index.html")
+
     def update!
       repo.fetch!
+      update_automator_roles!
```
ErikSchierboom, line 38: The `update!` method this is part of is automatically called when a commit is pushed to `main` for the `website-copy` repo

app/models/git/website_copy.rb
```diff
@@ -27,12 +27,25 @@ def mentor_notes_for_track(track_slug)
...
-    def walkthrough
-      repo.read_text_blob(head_commit, "walkthrough/index.html")
+    def update_automator_roles!
+      automators.each do |automator|
+        user = User.for!(automator[:username])
+        automator[:tracks].each do |track_slug|
+          track = ::Track.for!(track_slug)
+          User::UpdateAutomatorRole.defer(user, track)
+        end
+      end
```
ErikSchierboom, line 48: This is the most simple approach I could think of. It does mean that if a user is _removed_, we don't yet detect that.

app/commands/user/update_automator_role.rb
iHiD: I thought this was fine. Maybe "hardcoded" or some other things, but I'm happy with it :)

PR #6269, opened 2023-10-02
Add extra stats functionality

iHiD:
This refactors some of the code and adds the ability for us to more easily show stats.

The fundamental change is to cache the oldest and prestigious (highest rep user) solution on the actual representation, then use that data both in the creation of the search index, but also later when rendering the page. So you'll see in the code things being moved from the search document creation into the new `Recache` method (which then goes onto call the search document creation). It's important that it happens in that order: Recache first then update search. We never want the update search to be called by itself really.


test/commands/solution/search_via_representations_test.rb
```diff
@@ -88,58 +85,37 @@ class Solution::SearchViaRepresentationsTest < ActiveSupport::TestCase
   end
 
   test "filter: criteria" do
-    user = create :user, handle: 'john'
```
iHiD: - I'm removed lots of duplication and unnecessary things like this ^^. 
- I've also added a `create_solution` method that simplifies things for lots of tests. This whole test file could do with being updated to use that based on the work I've done here. 
- 
- I found there was so much noise in the setup it was hard to find the signal of what was unique about each test etc.

app/models/exercise/representation.rb
```diff
@@ -22,7 +24,8 @@ class Exercise::Representation < ApplicationRecord
   # This is too inefficient. Get the representations and then their submissions instead.
   # has_many :submission_representation_submissions, through: :submission_representations, source: :submission
 
-  has_many :published_solutions, foreign_key: "published_exercise_representation", class_name: "Solution",
+  has_many :published_solutions, -> { where(status: :published) },
+    foreign_key: "published_exercise_representation", class_name: "Solution",
```
iHiD, line 28: This was a bug that meant we counted all solutions, not just published ones in this field. So we'll need to recalculate and resync all representations to the indexes anyway.

ErikSchierboom (approved):

app/javascript/components/track/ExerciseCommunitySolutionsList.tsx
```diff
@@ -73,11 +73,7 @@ export function ExerciseCommunitySolutionsList({
   return (
     <div className="lg-container c-community-solutions-list">
       {resolvedData ? (
-        <h2>
-          {resolvedData.meta.unscopedTotal}{' '}
-          {pluralize('person', resolvedData.meta.unscopedTotal)} published
-          solutions
-        </h2>
+        <h2>Explore {resolvedData.meta.unscopedTotal} unique solutions</h2>
```
ErikSchierboom: Could `unscopedTotal` be `1`? If so, we might need to pluralize "solutions"

app/models/exercise/representation.rb
```diff
@@ -56,6 +59,9 @@ def has_feedback?
 
   def appears_frequently? = num_submissions >= APPEARS_FREQUENTLY_MIN_NUM_SUBMISSIONS
 
+  def first_submitted_at = oldest_solution.published_iterations.last.created_at
+  def max_reputation = prestigious_solution.user.reputation
+
```
ErikSchierboom, line 64: The `oldest_solution` and `prestigious_solution` fields are nullable, so in theory this could give a `nil` reference error, right?

app/commands/exercise/representation/create_search_index_document.rb
```diff
@@ -1,39 +1,30 @@
...
-  initialize_with :representation do
-    # Check a maximum of 5 solutions until we find one that's not broken.
-    # This is just an extra guard in case one or two solutions get in a weird state.
-    5.times do
-      @solution = representation.published_solutions.
-        where.not(user_id: User::GHOST_USER_ID).
-        first
-      @published_iteration = @solution&.latest_published_iteration
-      break if @published_iteration
-    end
```
ErikSchierboom, line 13: Nice to get rid of this

app/commands/exercise/representation/create_search_index_document.rb
```diff
@@ -51,15 +42,13 @@ def passes_latest_tests?
...
-      period: :forever,
-      category: :any,
-      about: :track,
-      track_id: representation.track_id,
-      user_id: representation.published_solutions.select(:user_id)
-    ).maximum(:reputation).to_i
-  end
+  def code = source_submission.files.map(&:content) || []
+  def num_loc = representation.oldest_solution.num_loc
+  def max_reputation = representation.prestigious_solution&.user&.reputation.to_i
```
ErikSchierboom: Could this be:

```suggestion
  def num_loc = oldest_solution.num_loc
  def max_reputation = prestigious_solution&.user&.reputation.to_i
```

due to the `delegate:` setup?

app/commands/solution/search_via_representations.rb
```diff
@@ -52,14 +52,16 @@ def search_body
       sort: search_sort,
 
       # Only return the solution IDs, not the entire document, to improve performance
-      _source: [:featured_solution_id],
+      _source: ["#{featured_solution_type}_solution_id"],
```
ErikSchierboom, line 55: Maybe extract this document ID logic to a separate method?

app/javascript/components/track/ExerciseCommunitySolutionsList.tsx
iHiD: Yes. It could.

app/models/exercise/representation.rb
iHiD: Correct. We should guard either here or upstream. Happy with either.

app/commands/exercise/representation/create_search_index_document.rb
iHiD: Yes. This felt like a win!

app/commands/exercise/representation/create_search_index_document.rb
```diff
@@ -51,15 +42,13 @@ def passes_latest_tests?
...
-      period: :forever,
-      category: :any,
-      about: :track,
-      track_id: representation.track_id,
-      user_id: representation.published_solutions.select(:user_id)
-    ).maximum(:reputation).to_i
-  end
+  def code = source_submission.files.map(&:content) || []
+  def num_loc = oldest_solution.num_loc
+  def max_reputation = prestigious_solution&.user&.reputation.to_i
```
iHiD:
```suggestion
  def max_reputation = prestigious_solution.user.reputation.to_i
```

app/commands/exercise/representation/recache.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def oldest_solution
+    representation.published_solutions.
+      where.not(user_id: User::GHOST_USER_ID).
+      first
+  end
+
+  def prestigious_solution
+    user_ids = representation.published_solutions.select(:user_id)
```
ErikSchierboom: You could also exclude the ghost user here if you'd like 🤷

test/commands/exercise/representation/create_search_index_document_test.rb
```diff
@@ -44,83 +44,56 @@ class Exercise::Representation::CreateSearchIndexDocumentTest < ActiveSupport::T
...
     num_published_solutions = 20
+    reputation = 1234
+    num_loc = 42
+
+    solutions = Array.new(2).map do
+      create(:practice_solution, :published, published_iteration_head_tests_status: :passed).tap do |solution|
+        submission = create(:submission, solution:)
+        create(:iteration, submission:)
+      end
+    end
```
ErikSchierboom: Maybe create a non-published solutions before this code block to sanity check that only published solutions are considered?

test/commands/exercise/representation/sync_to_search_index_test.rb
```diff
@@ -2,57 +2,57 @@
...
+    num_published_solutions = 20
+    reputation = 1234
+    num_loc = 42
+
+    solutions = Array.new(2).map do
+      create(:practice_solution, :published, published_iteration_head_tests_status: :passed).tap do |solution|
+        submission = create(:submission, solution:)
+        create(:iteration, submission:)
+      end
+    end
```
ErikSchierboom, line 17: Maybe create a non-published solutions before this code block to sanity check that only published solutions are considered?

ErikSchierboom (approved):

PR #6271, opened 2023-10-02
Show all exercises on #12in23

iHiD:
Context: http://forum.exercism.org/t/the-12in23-challenge/2213/123

This could all be tidied a lot, but I don't mind the technical debt massively here.


ErikSchierboom (commented):
This does list all 6 exercises from that month in which we had 6. Maybe we want to change that?

iHiD:
I feel like that was the worst decision we ever made :D

ErikSchierboom:
> I feel like that was the worst decision we ever made :D

It's definitely up there at the top

PR #6277, opened 2023-10-03
Optimize notifications creation

ErikSchierboom:


app/commands/user/notification/create.rb
```diff
@@ -16,6 +16,13 @@ def call
       exercise:,
       params:
     )
+
+    # Don't attempt to create a notification when there already is one
+    # This optimizes for scripts that create notifications but where
+    # the notification has usually already been created
+    existing_notification = user.notifications.find_by(uniqueness_key: notification.uniqueness_key)
+    return existing_notification if existing_notification.present?
```
ErikSchierboom, line 24: I'm not sure we depend on the notification being returned?

iHiD: Yeah, good shout.

iHiD: I wish there was a one liner for this pattern (but there's not!)

PR #6291, opened 2023-10-04
Sort Representation tracks alphabetically

dem4ron:


iHiD:
@dem4ron this is red btw.

ErikSchierboom (commented):
I think this is the wrong place to solve this. 

Take a look at:

Is this for sorting the tracks when you're staff?

```ruby
memoize
def automator_tracks
  return Track.all if mentor.staff?

  Track.where(id: mentor.track_mentorships.automator.select(:track_id)).order(title: :asc)
end
```

You can see that the non-staff path is sorted. You could easily also add the sorting to the staff-path.

ErikSchierboom (approved):

PR #6301, opened 2023-10-06
Prevent duplicate rendering of iterations from channel

dem4ron:


iHiD:
Happy to merge, but is it worth adding a test?

PR #6309, opened 2023-10-13
Add tags to representation document

ErikSchierboom:
- Read tags from new field
- Sync analysis tags into search index document

### DIFF
--- a/app/commands/exercise/representation/create_search_index_document.rb
+++ b/app/commands/exercise/representation/create_search_index_document.rb
@@ -25,6 +25,7 @@ class Exercise::Representation::CreateSearchIndexDocument
       num_solutions: representation.num_published_solutions,
       code: published_iteration.submission.files.map(&:content) || [],
       max_reputation:,
+      tags:,
       exercise: {
         id: solution.exercise.id,
         slug: solution.exercise.slug,
@@ -59,6 +60,14 @@ class Exercise::Representation::CreateSearchIndexDocument
       track_id: representation.track_id,
       user_id: representation.published_solutions.select(:user_id)
     ).maximum(:reputation).to_i
+  end
+
+  def tags
+    representation.
+      submission_representations.
+      joins(submission: :analysis).
+      where(submission: { analysis_status: :completed }).
+      last&.submission&.analysis&.tags.to_a
   end
 
   attr_reader :solution, :published_iteration
--- a/app/commands/exercise/representation/create_search_index_document.rb
+++ b/app/commands/exercise/representation/create_search_index_document.rb
@@ -63,11 +63,18 @@ class Exercise::Representation::CreateSearchIndexDocument
   end
 
   def tags
+    return [] if last_analyzed_submission_representation.nil?
+
+    last_analyzed_submission_representation.submission.analysis.tags
+  end
+
+  memoize
+  def last_analyzed_submission_representation
     representation.
       submission_representations.
       joins(submission: :analysis).
       where(submission: { analysis_status: :completed }).
-      last&.submission.analysis.tags.to_a
+      last
   end
 
   attr_reader :solution, :published_iteration
--- a/app/commands/exercise/representation/create_search_index_document.rb
+++ b/app/commands/exercise/representation/create_search_index_document.rb
@@ -67,7 +67,7 @@ class Exercise::Representation::CreateSearchIndexDocument
       submission_representations.
       joins(submission: :analysis).
       where(submission: { analysis_status: :completed }).
-      last&.submission&.analysis&.tags.to_a
+      last&.submission.analysis.tags.to_a
   end
 
   attr_reader :solution, :published_iteration
--- a/app/commands/exercise/update_tags.rb
+++ b/app/commands/exercise/update_tags.rb
@@ -7,8 +7,14 @@ class Exercise::UpdateTags
 
   private
   def tags
-    Solution::Tag.where(exercise:).pluck(:tag).map do |tag|
+    solution_tags = Solution::Tag.where(exercise:).distinct.pluck(:tag)
+    existing_tags = Exercise::Tag.where(exercise:).where(tag: solution_tags).select(:id, :tag).to_a
+    exercise_tags = existing_tags.map(&:tag)
+
+    new_tags = (solution_tags - exercise_tags).map do |tag|
       Exercise::Tag.find_or_create_by!(tag:, exercise:)
     end
+
+    existing_tags + new_tags
   end
 end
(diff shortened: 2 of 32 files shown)


db/migrate/20231013123032_add_tags_to_submission_analyses.rb
```diff
@@ -0,0 +1,7 @@
+class AddTagsToSubmissionAnalyses < ActiveRecord::Migration[7.0]
+  def change
+    return if Rails.env.production?
+
+    add_column :submission_analyses, :tags_data, :text, null: true
```
ErikSchierboom, line 5: I've named it `tags_data` for:

a) consistency
b) to allow us to have a `tags` method to get the actual tags

app/commands/exercise/representation/create_search_index_document.rb
```diff
@@ -61,5 +62,13 @@ def max_reputation
     ).maximum(:reputation).to_i
   end
 
+  def tags
+    representation.
+      submission_representations.
+      joins(submission: :analysis).
+      where(submission: { analysis_status: :completed }).
+      last&.submission&.analysis&.tags.to_a
```
iHiD: If these are ever nil we're in a quagmire and I probably want to know with an exception on bugsnag.

```suggestion
      last&.submission.analysis.tags.to_a
```

app/models/submission/analysis.rb
```diff
@@ -58,6 +59,8 @@ def summary
     data[:summary].presence
   end
 
+  def tags = tags_data[:tags].to_a
```
iHiD, line 62: Probably good to memoize this.

iHiD (approved):

ErikSchierboom:
One open question: do we want to update the tags of an exercise representation document whenever a new submission representation is created? Or only when it doesn't already have tags?

iHiD:
> One open question: do we want to update the tags of an exercise representation document whenever a new submission representation is created? Or only when it doesn't already have tags?

Do you mean whenever a new analysis is created? I think the tags to be set via this command every time, so the question is when we call the command. And I guess the extra question is "do we rerun it when an analysis is created"? Or is there something else I'm not thinking about?

app/models/exercise/tag.rb
```diff
@@ -0,0 +1,14 @@
...
+  belongs_to :exercise
+
+  memoize
+  def category = tag.split(':').first
+
+  memoize
+  def name = tag.split(':').second
+
+  memoize
+  def to_s = "#{category.titleize}: #{name.titleize}"
```
iHiD, line 13: Nice use of `to_s`

app/models/solution/tag.rb
```diff
@@ -0,0 +1,21 @@
+class Solution::Tag < ApplicationRecord
+  extend Mandate::Memoize
+
+  belongs_to :solution
+  belongs_to :exercise
+  belongs_to :user
+
+  before_validation on: :create do
+    self.exercise = solution.exercise unless exercise
+    self.user = solution.user unless user
```
iHiD: Would probably be better to set `exercise_id` and `user_id` here to avoid loading the records from the db. (Thinking if we do batch things this could add a lot of extra DB calls)

db/migrate/20231017072611_create_solution_tags.rb
```diff
@@ -0,0 +1,15 @@
+class CreateSolutionTags < ActiveRecord::Migration[7.0]
+  def change
+    return if Rails.env.production?
+
+    create_table :solution_tags do |t|
+      t.string :tag, null: false
+
+      t.references :solution, null: false, foreign_key: true
+      t.references :exercise, null: false, foreign_key: true
+      t.references :user, null: false, foreign_key: true
```
iHiD, line 8: I tend to put references first then other things later in a migration. (ie the heavy stuff at the top)

app/commands/exercise/update_tags.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  initialize_with :exercise
+
+  def call = exercise.update(tags:)
+
+  private
+  def tags
+    Solution::Tag.where(exercise:).pluck(:tag).map do |tag|
+      Exercise::Tag.find_or_create_by!(tag:, exercise:)
+    end
```
iHiD: To avoid `n+1` lookups of exercise tags, maybe consider something like:

Also note the `distinct` on the first line.

```suggestion
    solution_tags = Solution::Tag.where(exercise:).distinct.pluck(:tag)
    existing_tags = Exercise::Tag.where(exercise:).where(tag: solution_tags).select(:id, :tag).to_a
    exercise_tags = existing_tags.map(&:tag)
    
    new_tags = (solution_tags - exercise_tags).map do |tag|
      Exercise::Tag.find_or_create_by!(tag:, exercise:)
    end
    
    existing_tags + new_tags
```

app/commands/solution/update_tags.rb
```diff
@@ -0,0 +1,22 @@
...
+    Exercise::UpdateTags.(solution.exercise)
+  end
+
+  private
+  def tags
+    return [] if latest_analysis.nil?
+
+    latest_analysis.tags.map do |tag|
+      Solution::Tag.find_or_create_by(tag:, solution:)
+    end
```
iHiD, line 21: And some other n+1 avoidance strategy here.

app/commands/solution/publish_iteration.rb
```diff
@@ -6,6 +6,7 @@ class Solution::PublishIteration
   def call
     solution.update!(published_iteration: iteration)
 
+    Solution::UpdateTags.(solution)
     Solution::UpdatePublishedExerciseRepresentation.(solution)
     Solution::UpdateSnippet.(solution)
     Solution::UpdateNumLoc.(solution)
```
iHiD, line 12: I'd like to check the code path for when someone publishes a new iteration on a solution that's set to "publish all iterations". Does this command get called as part of that? My guess would be that it doesn't (but I hope I'm wrong). If not, we should change it so it does call this.

app/commands/submission/analysis/process.rb
```diff
@@ -53,4 +55,12 @@ def data
   rescue StandardError
     {}
   end
+
+  memoize
+  def tags_data
+    res = JSON.parse(tooling_job.execution_output['tags.json'])
+    res.is_a?(Hash) ? res.symbolize_keys : {}
+  rescue StandardError
```
iHiD: Worth pushing a bugsnag here so we know a representer is misbehaving?

Maybe add a guard before L61 to see if `tooling_job.execution_output['tags.json'])` is missing, and if it is, then  silently return `{}` if the file isn't there. Then add a `bugsnag.notify` to this exception block?

app/models/solution/tag.rb
ErikSchierboom: Makes sense. We should probably do that in other places too.

PR #6358, opened 2023-10-19
Verify that we don't change tables in production

ErikSchierboom:
This is completely rudimentary, but gives us _some_ protection

### DIFF
--- a/test/misc/migrations_test.rb
+++ b/test/misc/migrations_test.rb
@@ -1,0 +1,13 @@
+require "test_helper"
+
+class MigrationsTest < ActiveSupport::TestCase
+  test "don't modify tables in production" do
+    ApplicationRecord.connection.migration_context.migrations.each do |migration_proxy|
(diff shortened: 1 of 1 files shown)


test/misc/migrations_test.rb
```diff
@@ -0,0 +1,36 @@
...
+      change_column
+      change_column_null
+    ]
+
+    ApplicationRecord.connection.migration_context.migrations.each do |migration_proxy|
+      # We only started running migrations manually after the below time
+      next if migration_proxy.version <= 20231017101049 # rubocop:disable Style/NumericLiterals
+
+      migration = File.read(migration_proxy.filename)
+      next unless modification_calls.any? { |modification_call| migration.include?(modification_call) }
```
iHiD: I sort of feel like we should't run *any* code automtically in production, so maybe removing this and just hasving the line below is actually fine.

PR #6392, opened 2023-10-25
Re-sync exercise representation after num_loc changes

ErikSchierboom:
- Add tests for updating num_loc
- No-op when num-loc hasn't changed
- Resync representation when solution num_loc changes

### DIFF
--- a/app/commands/solution/update_num_loc.rb
+++ b/app/commands/solution/update_num_loc.rb
@@ -5,9 +5,10 @@ class Solution::UpdateNumLoc
 
   def call
     if iteration && iteration.num_loc.nil?
-      Solution::UpdateNumLoc.defer(solution, prereq_jobs: [
-                                     Iteration::CalculateLinesOfCode.defer(iteration)
-                                   ])
+      Solution::UpdateNumLoc.defer(solution,
+        prereq_jobs: [
+          Iteration::CalculateLinesOfCode.defer(iteration)
+        ])
       return
     end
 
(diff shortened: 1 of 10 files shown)


app/commands/solution/update_num_loc.rb
```diff
@@ -3,11 +3,23 @@ class Solution::UpdateNumLoc
...
 
   private
+  memoize
   def num_loc
     iteration = solution.latest_published_iteration || solution.latest_iteration
     iteration&.num_loc.to_i
   end
+
+  def resync_representation_to_search_index!
+    Exercise::Representation.where(oldest_solution: solution).each do |representation|
```
ErikSchierboom, line 32: I probably could have used `find_by` here. Do you want me to do that?

app/commands/solution/update_num_loc.rb
```diff
@@ -3,11 +3,32 @@ class Solution::UpdateNumLoc
 
   initialize_with :solution
 
-  def call = solution.update!(num_loc:)
+  def call
+    if iteration && iteration.num_loc.nil?
+      calc_job = Iteration::CalculateLinesOfCode.defer(iteration)
+      Solution::UpdateNumLoc.defer(prereq_jobs: [calc_job])
+      return
```
iHiD: Is this more or less readable?

```suggestion
      return Solution::UpdateNumLoc.defer(prereq_jobs: [
        Iteration::CalculateLinesOfCode.defer(iteration)
      ])
```

app/commands/solution/update_num_loc.rb
iHiD: Yeah, I considered this when I first reviewed. I think what you have is safer, just in case there are multiple with `oldest_solution` as the same (there shouldn't be, but 🤷)

iHiD (approved):

PR #6407, opened 2023-10-27
Show upload count on community solutions

iHiD:


app/javascript/components/common/CommunitySolution.tsx
```diff
@@ -19,23 +19,43 @@ const PublishDetails = ({ solution }: { solution: CommunitySolutionProps }) => {
         solution.publishedAt
       )}`}</time>
       <div className="--counts">
+        {solution.representationNumPublishedSolutions ? (
```
ErikSchierboom, line 22: Just to be clear: this is `false` if `representationNumPublishedSolutions == 0`, but that's probably what we want, right?

ErikSchierboom (approved):

app/javascript/components/common/CommunitySolution.tsx
iHiD: Yes. Thanks for checking.

PR #6433, opened 2023-11-01
Guard exercise tags

iHiD:


app/commands/exercise/update_tags.rb
```diff
@@ -17,4 +16,15 @@ def tags
...
     existing_tags + new_tags
   end
+
+  memoize
+  def solution_tags
+    Solution::Tag.joins(:solution).where(
+      exercise_id: exercise.id,
+      solution: {
+        status: Solution.statuses[:published],
+        published_iteration_head_tests_status: "passed"
```
ErikSchierboom: Why not:

```suggestion
        status: :published,
        published_iteration_head_tests_status: :passed
```

ErikSchierboom (approved):

app/commands/exercise/update_tags.rb
iHiD: Because it's `solution: status` I'm not sure if Rails is intelligent enough to understand this (because the column is an `int` in the database). It might do - I didn't check!

iHiD: I've checked. it is intellgent enough so I'll change it.

ErikSchierboom: And the "passed" => :passed bit?

PR #6434, opened 2023-11-01
Add some ML Training scaffolding

iHiD:
#### Dashboard:
Screenshot 2023-11-17 at 15 52 13

---

#### Submenu:

Screenshot 2023-11-20 at 9 58 46

I shortened the `description` because it was making the submenu too wide, although we can modify the CSS to keep it narrower.

---

#### Code-tagger:

https://github.com/exercism/website/assets/66035744/5eba8b36-d039-4a48-a8ae-426f545683d9

Basically, `solution_tags` should be a semi-copy-paste of the `mentoring/automation` controller/pages, but with reference to the Loom on Slack. 

@dem4ron tell @ErikSchierboom what you need :)


app/helpers/react_components/training_data/code_tagger.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      initialize_with :sample
+
+      def to_s
+        super("training-data-code-tagger", {
+          code: {
+            track: {
+              title: track.title,
+              icon_url: track.icon_url,
+              highlightjs_language: track.highlightjs_language
```
iHiD: I think `tags` should move here.

app/helpers/react_components/training_data/code_tagger.rb
```diff
@@ -0,0 +1,42 @@
...
+            end
+          },
+          tags: sample.tags,
+          all_tags:,
+          links: {
+            confirm_tags_api: '',
+            next_sample: Exercism::Routes.next_training_data_code_tags_samples_path(track_id: track.id, status: sample.status),
+            training_data_dashboard: ''
+          },
+          status: sample.status
```
iHiD: And we should have `sample: { status:, tags:  }` here.

app/models/user.rb
```diff
@@ -350,11 +350,13 @@ def automator?(track = nil)
...
   end
 
-  def trainer?(_track = nil)
+  def trainer?(track = nil)
     return true if staff?
-    return true if maintainer?
-    return true if supermentor?
+    return reputation_for_track(track) >= MIN_TRACK_REP_FOR_TRAINER if track.present?
 
     false
```
iHiD: I think rather than false here, this need to be "any track.rep > MIN_TRACK_REP_FOR_TRAINER`

iHiD: I'm not sure how to do that though:)

app/javascript/components/training-data/code-tagger/right-pane/RightPane.tsx
```diff
@@ -27,7 +27,7 @@ export function RightPane({
       />
       <button
         onClick={() => confirmTags()}
-        className="btn-m btn-primary mt-auto mb-64"
+        className="btn-m btn-primary mb-32 mt-auto"
```
iHiD: Out of interest, why did you swap these? I tend to prefer margins ordered like they are in CSS shorhand (top/right/bottom/left).

dem4ron: I think by pure accident :D I deleted, `mt-auto`, adjusted `mb-32`, then readded `mt-auto`

app/commands/training_data/code_tags_sample/update_tags.rb
```diff
@@ -0,0 +1,31 @@
...
+    when :machine_tagged
+      :community_checked
+    when :human_tagged
+      :community_checked
+    when :community_checked
+      :admin_checked
+    else
+      sample.status
+    end
+  end
```
ErikSchierboom: I was unsure if we wanted the command to calculate the next status, or whether the caller should do that 🤷

app/commands/training_data/code_tags_sample/retrieve.rb
```diff
@@ -0,0 +1,65 @@
+class TrainingData::CodeTagsSample::Retrieve
+  include Mandate
+
+  initialize_with :status, track: nil, criteria: nil, dataset: :training, page: 1 do
+    @status = status&.to_sym
+    raise "Invalid 'status' parameter" unless STATUSES.include?(@status)
+  end
+
+  def self.samples_per_page = SAMPLES_PER_PAGE
```
iHiD: Is there a reason to do this rather than just using the constant in the code?

app/commands/training_data/code_tags_sample/update_tags.rb
```diff
@@ -0,0 +1,31 @@
+class TrainingData::CodeTagsSample::UpdateTags
+  include Mandate
+
+  initialize_with :sample, :tags, :user
+
+  def call
+    ActiveRecord::Base.transaction do
+      sample.lock!
+
+      raise TrainingDataCodeTagsSampleLockedByAnotherUserError unless sample.lockable_by?(user)
```
iHiD: Maybe?
```suggestion
      raise TrainingDataCodeTagsSampleLockedError unless sample.lockable_by?(user)
```

app/commands/training_data/code_tags_sample/update_tags.rb
```diff
@@ -0,0 +1,31 @@
+class TrainingData::CodeTagsSample::UpdateTags
+  include Mandate
+
+  initialize_with :sample, :tags, :user
+
+  def call
+    ActiveRecord::Base.transaction do
+      sample.lock!
```
iHiD: I have a small preference for:
```suggestion
    sample.with_lock do
```

Which does the same thing.

app/commands/training_data/code_tags_sample/update_tags.rb
iHiD: Also, I don't get why this LOC is here at all.

At this stage the record should have been locked when they were editing it, so I'm not sure we need to check it again. What's your thinking?

app/commands/training_data/code_tags_sample/update_tags.rb
iHiD: Good question. Atm this never sets it to `machine_tagged` which is a problem. And also, it means someone could replay the same request twice more and move it from `untagged` to `:admin_checked` just by that replay.

So I think we probably want to do it as part of the caller instead.

app/models/training_data/code_tags_sample.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def lock_for_editing!(user)
+    with_lock do
+      raise "Already locked" if locked?
+
+      update!(
+        locked_until: Time.current + 30.minutes,
+        locked_by: user
+      )
+    end
```
iHiD, line 44: This bit here is where the locking currently happens. And where you probably want your real exception raised etc, rather than in the updating bit.

app/commands/user/reputation_token/create.rb
```diff
@@ -19,6 +19,9 @@ def call
       AwardBadgeJob.perform_later(user, :contributor, context: token)
       User::ReputationPeriod::MarkForToken.(token)
       User::ResetCache.defer(user, :has_unseen_reputation_tokens?)
+
+      user_track = token.track.present? ? UserTrack.find_by(user:, track: token.track) : nil
```
iHiD: Maybe
```suggestion
      user_track = token.track.present? && UserTrack.find_by(user:, track: token.track)
```

app/commands/user_track/update_reputation.rb
```diff
@@ -0,0 +1,15 @@
...
+  include Mandate
+
+  initialize_with :user_track
+
+  def call = user_track.update!(reputation:)
+
+  private
+  def reputation
+    User::ReputationToken.where(
+      track: user_track.track,
```
iHiD:
```suggestion
      track_id: user_track.track_id,
```

app/controllers/training_data/code_tags_samples_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def show; end
+
+  def next
+    sample = TrainingData::CodeTagsSample::RetrieveNext.(@track, params[:status])
+    return redirect_to training_data_root_path if sample.nil?
+
+    sample.lock_for_editing!(current_user)
+
+    redirect_to training_data_code_tags_sample_path(sample, status: params[:status])
```
iHiD: We need to do something like rescue and try again if `lock_for_editing` fails (maybe `redo` that 3 times (presumably getting a different sample each time) then give up?)

app/controllers/training_data/code_tags_samples_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    @track = @sample.track
+  rescue ActiveRecord::RecordNotFound
+    render_404(:sample_not_found)
+  end
+
+  def ensure_trainer!
+    return if current_user.trainer?(@track)
+
+    render_403(:not_trainer)
+  end
```
iHiD: This isn't an API controller, so these `render_xxx` things aren't valid. They need to redirect to actual UI pages.

app/controllers/training_data/dashboard_controller.rb
```diff
@@ -0,0 +1,14 @@
+class TrainingData::DashboardController < ApplicationController
+  before_action :guard_trainer!
+
+  def index
+    @training_data_dashboard_params = params.permit(:status, :order, :criteria, :page, :track_slug)
+    @statuses = %w[needs_tagging needs_checking needs_checking_admin]
+  end
+
+  def guard_trainer!
```
iHiD:
```suggestion
  private
  def guard_trainer!
```

iHiD: Maybe move this into `application_controller`

app/controllers/training_data/code_tags_samples_controller.rb
iHiD: The next controller has the correct one for `ensure_trainer` btw.

app/controllers/training_data/external_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def index
+    redirect_to training_data_root_path if current_user&.trainer?
+
+    @eligible = current_user.eligible_for_trainer?
+    @max_reputation_track = current_user.user_tracks.order(reputation: :desc).first
+  end
+
+  def become_trainer
+    User::BecomeTrainer.(current_user) unless current_user.trainer?
```
iHiD: Maybe move this guard inside `User::BecomeTrainer` (I'm guessing it's there already - so maybe just remove it here?)

app/helpers/view_components/nav_helpers/nav_submenus.rb
```diff
@@ -89,22 +89,20 @@ module NavSubmenus
...
-        #   view: :mentoring
-        # },
-
+        {
+          title: "Training Hub",
+          description: "Help train Exercism's neural network",
+          path: Exercism::Routes.training_data_root_path,
+          icon: :robot,
+          icon_filter: "textColor6"
+        },
```
iHiD: Maybe comment this back out so we can dark-deploy?

app/javascript/components/mentoring/inbox/TrackList.tsx
```diff
@@ -5,15 +5,15 @@ export type Track = {
...
+  count?: number
 }
 
 const OptionComponent = ({ option: track }: { option: Track }) => {
   return (
     <React.Fragment>
       <TrackLogo track={track} />
       <div className="title">{track.title}</div>
-      <div className="count">{track.count}</div>
+      {track.count && <div className="count">{track.count}</div>}
```
iHiD: Why this change? And should it be a different PR (as it's to do with mentoring inbox)?

app/models/training_data/code_tags_sample.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def to_param = uuid
+
+  def dataset = super.to_sym
+  def status = super.to_sym
+
+  def checked?
+    status == :human_tagged || status == :admin_tagged
+  end
```
iHiD, line 30: This looks wrong to me. What is `checked?` - I'd presume it should be at least one of the `_checked` statuses.

app/serializers/serialize_code_tags_sample.rb
```diff
@@ -0,0 +1,28 @@
...
+      track: {
+        title: track.title,
+        icon_url: track.icon_url
+      },
+      exercise: {
+        title: exercise.title,
+        icon_url: exercise.icon_url
+      },
+      created_at: sample.created_at,
+      updated_at: sample.updated_at,
```
iHiD: Are these needed? Why do we show `created_at` in the UI?

app/views/training_data/code_tags_samples/index.html.haml
```diff
@@ -0,0 +1,4 @@
+- require_stylesheet "mentoring"
+
+#mentor-inbox-page
```
iHiD: @dem4ron This class should change.

app/javascript/components/mentoring/inbox/TrackList.tsx
ErikSchierboom: CC @dem4ron

app/serializers/serialize_code_tags_sample.rb
ErikSchierboom: @dem4ron ?

app/models/training_data/code_tags_sample.rb
ErikSchierboom: Well, you tell me :) https://github.com/exercism/website/blame/5404db7a3d591b676b5a3b7ce91cee880aa6fd70/app/models/training_data/code_tags_sample.rb

app/commands/training_data/code_tags_sample/update_tags.rb
ErikSchierboom: It's just a safe measure 🤷 But maybe it's too much?

app/commands/training_data/code_tags_sample/retrieve.rb
ErikSchierboom: We use this pattern elsewhere to allow easy mocking, but it's not used here so let me update

app/controllers/training_data/code_tags_samples_controller.rb
ErikSchierboom: > This isn't an API controller, so these render_xxx things aren't valid. 

Ehm, that seems to not be true? This is in `ProfilesController`:

```ruby
def use_user
    @user = User.find_by(handle: params[:id])
    render_404
  end
```

What is true is that I shouldn't pass in an argument.

app/models/training_data/code_tags_sample.rb
iHiD: Let's rename this to `safe_to_override?` and also add `community_checked` as an `||`

app/javascript/components/training-data/dashboard/TaggableCode.tsx
```diff
@@ -0,0 +1,21 @@
...
+    <a className="--code-sample " href={code.links.edit}>
+      <TrackIcon title={code.track.title} iconUrl={code.track.iconUrl} />
+      <ExerciseIcon
+        title={code.exercise.title}
+        iconUrl={code.exercise.iconUrl}
+      />
+      <div className="--info">
+        <div className="--handle">{code.exercise.title}</div>
+      </div>
+      <time className="-updated-at">{fromNow(code.createdAt)}</time>
```
iHiD: Delete this.

app/javascript/components/training-data/dashboard/Dashboard.types.ts
```diff
@@ -0,0 +1,62 @@
...
+  track: {
+    title: string
+    iconUrl: string
+  }
+  exercise: {
+    title: string
+    iconUrl: string
+  }
+  createdAt: string
+  updatedAt: string
```
iHiD:
```suggestion
```

app/serializers/serialize_code_tags_sample.rb
iHiD:
```suggestion
```

app/javascript/components/training-data/dashboard/TaggableCode.tsx
iHiD:
```suggestion
```

app/controllers/training_data/code_tags_samples_controller.rb
iHiD: `render_404` is defined in application controller but doesn't take an argument (it just renders a 404 page). `render_403` is not defined there at all.

PR #6435, opened 2023-11-01
Update tags when head tests status changes

ErikSchierboom:
- Extract update_published_iteration_head_tests_status! from Solution to command
- Extract update_latest_iteration_head_tests_status! from Solution to command
- Use Solution::UpdatePublishedIterationHeadTestsStatus command instead of method
- Use Solution::UpdateLatestIterationHeadTestsStatus command instead of method
- Update tags when updating published_iteration_head_tests_status

### DIFF
--- a/app/commands/solution/sync_to_search_index.rb
+++ b/app/commands/solution/sync_to_search_index.rb
@@ -6,6 +6,8 @@ class Solution::SyncToSearchIndex
   initialize_with :solution
 
   def call
+    Solution::CreateSearchIndexDocument.(solution)
+
     if solution.user.ghost?
       delete_document!
     else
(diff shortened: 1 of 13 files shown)


app/models/solution.rb
```diff
@@ -102,20 +102,6 @@ def self.for(*args)
...
-    update_column(:published_iteration_head_tests_status, status)
-    Solution::SyncToSearchIndex.defer(self)
-  end
-
-  def update_latest_iteration_head_tests_status!(status)
-    return if latest_iteration_head_tests_status == status.to_sym
-
-    update_column(:latest_iteration_head_tests_status, status)
-    Solution::SyncToSearchIndex.defer(self)
-  end
```
ErikSchierboom, line 117: I've also taken the liberty to extract this logic to its own command.

app/commands/solution/sync_to_search_index.rb
```diff
@@ -6,6 +6,8 @@ class Solution::SyncToSearchIndex
   initialize_with :solution
 
   def call
+    Solution::CreateSearchIndexDocument.(solution)
```
iHiD, line 9: Ignore this.

PR #6445, opened 2023-11-02
Add code tagger page

dem4ron:
https://github.com/exercism/website/assets/66035744/c71b389e-6575-49c8-bfc5-08ca169dacef


app/controllers/solution_tagger_controller.rb
```diff
@@ -0,0 +1,2 @@
+class SolutionTaggerController < ApplicationController
+end
```
iHiD: This should be in ml_trainer directory.

app/helpers/react_components/ml_trainer/solution_tagger.rb
```diff
@@ -0,0 +1,14 @@
+module ReactComponents
+  module MLTrainer
+    class SolutionTagger < ReactComponent
+      def to_s
+        super("ml-trainer-solution-tagger", {
+          solution: SerializeSolution.(Solution.last),
```
iHiD: This should just be `code` not `solution`. But I think @ErikSchierboom  probably knows this after what we discussed yesterday.

PR #6447, opened 2023-11-02
Link submissions to approaches via tags

ErikSchierboom:
- Add tags to approaches in test track
- Create table for tag
- Add command to create Exercise::Approach::Tag
- Add exercise approach
- Add approach reference to submission
- Stub command for linking submission to approach
- Add trait for approach tags
- Link to approach after processing
- Fix model
- Fixup ActiveRecord
- Make approach optional for submission
- Link submission to matching approach
- Link submissions to approach after syncing
- Link matching submissions


app/commands/exercise/approach/link_matching_submissions.rb
```diff
@@ -0,0 +1,47 @@
+class Exercise::Approach::LinkMatchingSubmissions
+  include Mandate
+
+  initialize_with :approach
+
+  def call
+    if approach.tags.empty?
+      relink_submissions!(linked_submissions)
```
ErikSchierboom: When the approach does _not_ have any tags, I've opted to try and re-link any submissions that are currently linked to this approach. Whilst it is highly unlikely that:

a) an approach would lose it tags
b) a linked submissions would match another approach

I felt this was still the safe thing to do. Let me know what you think.

app/commands/exercise/approach/link_matching_submissions.rb
```diff
@@ -0,0 +1,47 @@
...
+    else
+      update_submissions!(linked_submissions)
+      update_submissions!(unlinked_submissions)
```
ErikSchierboom: The first call is to check if any submissions currently linked to this approach should still be linked to this approach.
The second call is to check if any currently unlinked submissions should be linked to this approach.
I've purposefully ignored checking any submissions that are linked to other approaches,

app/commands/git/sync_exercise_approach.rb
```diff
@@ -5,7 +5,11 @@ class Git::SyncExerciseApproach
 
   def call
     find_or_create!.tap do |approach|
+      old_tags = approach.tags.pluck(:tag).sort
       approach.update!(attributes_for_update(approach))
+      new_tags = approach.tags.pluck(:tag).sort
+
+      Exercise::Approach::LinkMatchingSubmissions.(approach) if old_tags != new_tags
```
ErikSchierboom: We re-check the matching submissions of the approach if its tags have changed

app/commands/submission/analysis/process.rb
```diff
@@ -41,6 +41,7 @@ def handle_ops_error!
   def handle_completed!
     submission.analysis_completed!
     Solution::UpdateTags.(submission.solution)
+    Submission::LinkToMatchingApproach.defer(submission)
```
ErikSchierboom: At this point the submission's tags will be known (could be empty of course), so try linking it to an approach

app/models/exercise/approach.rb
```diff
@@ -31,4 +41,28 @@ def git = Git::Exercise::Approach.new(slug, exercise.slug, exercise.git_type, sy
...
+  end
+
+  memoize
+  def tag_conditions
+    tags.
+      pluck(:condition_type, :tag).
+      group_by(&:first).
+      transform_keys(&:to_sym).
+      transform_values { |v| v.map(&:second) }
+  end
```
ErikSchierboom: I've opted for some simple Ruby code in favor of several queries. This is easier to read and only requires one query to fetch all tags.

app/commands/submission/link_to_matching_approach.rb
```diff
@@ -0,0 +1,17 @@
...
+  initialize_with :submission
+
+  def call
+    submission.update(approach: matching_approach)
+  end
+
+  def matching_approach
+    return nil unless submission.tags.present?
+
+    submission.exercise.approaches.includes(:tags).find do |approach|
```
ErikSchierboom: There are usually just a couple of approaches per exercise, so this is probably fine

app/models/exercise/approach/tag.rb
```diff
@@ -0,0 +1,10 @@
+class Exercise::Approach::Tag < ApplicationRecord
+  belongs_to :approach,
+    class_name: "Exercise::Approach",
+    foreign_key: :exercise_approach_id,
+    inverse_of: :tags
+
+  enum condition_type: { all: 0, any: 1, not: 2 }, _prefix: true
```
ErikSchierboom: I struggled naming this :)

app/commands/exercise/approach/link_matching_submissions.rb
```diff
@@ -0,0 +1,45 @@
+class Exercise::Approach::LinkMatchingSubmissions
+  include Mandate
+
+  initialize_with :approach
+
+  def call
+    if approach.tags.empty?
+      relink_submissions!(linked_submissions)
```
iHiD: Sounds good. You could change this to be a guard.

```suggestion
    return relink_submissions!(linked_submissions) if approach.tags.empty?
```

app/commands/exercise/approach/link_matching_submissions.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  private
+  memoize
+  def analyzed_exercise_submissions
+    Submission.includes(:analysis).
+      where(exercise: approach.exercise).
+      where(analysis_status: :completed)
+  end
```
iHiD, line 14: Can you move this method below the next two pls?

iHiD: I'm a little wary of this:
`where(analysis_status: :completed)`

Because if we generate tags through a ML approach, it probably won't be via this analyzer.

I wonder if we should materialise tags onto a submission, and then match on tags not being nil? That probably makes SQL a bit easier etc with one less hop to the analysis record?

app/models/exercise/approach/tag.rb
```diff
@@ -0,0 +1,10 @@
+class Exercise::Approach::Tag < ApplicationRecord
```
iHiD: What's the advantage of us having this as a many-to-many table, rather than materialising this as a JSON field on the approach? As later on, we just reget all the data into one object then work on it, I'm not sure what we're gaining from this.

app/models/exercise/approach/tag.rb
iHiD: `operator` or `logical_operator`?

app/models/exercise/approach.rb
iHiD: This makes me wonder if we should just store it as JSON. I've commented more below.

app/models/exercise/approach.rb
```diff
@@ -31,4 +41,28 @@ def git = Git::Exercise::Approach.new(slug, exercise.slug, exercise.git_type, sy
...
   def content_html = Markdown::Parse.(content)
+
+  def matching_tags?(check_tags)
+    return false if check_tags.empty?
+
+    all_tags = tag_conditions[:all].to_a
+    return false if all_tags.present? && (all_tags - check_tags).present?
+
+    any_tags = tag_conditions[:any].to_a
+    return false if any_tags.present? && (any_tags & check_tags).empty?
```
iHiD: So in the case where someone specifies `all` and `any`, both are honoured? That's probably right, but I just want to check :)

ErikSchierboom: Yes, that's according to the spec I think

app/models/exercise/approach/tag.rb
ErikSchierboom: I used a many-to-many table as I originally intended some fancy queries, which would not have worked as easily with a JSON field. But maybe a JSON field is easier.

db/schema.rb
```diff
@@ -373,7 +374,7 @@
   create_table "exercise_tags", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.bigint "exercise_id", null: false
     t.string "tag", null: false
-    t.boolean "filterable", default: true, null: false
+    t.boolean "filterable", default: true, null: false
```
ErikSchierboom:
```suggestion
    t.boolean "filterable", default: true, null: false
```

db/schema.rb
```diff
@@ -1006,11 +1009,6 @@
     t.index ["track_id"], name: "index_submission_representations_on_track_id"
   end
 
-  create_table "submission_tags", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-  end
```
ErikSchierboom, line 1013: I didn't see this table in production.

app/commands/exercise/approach/link_matching_submissions.rb
```diff
@@ -0,0 +1,35 @@
+class Exercise::Approach::LinkMatchingSubmissions
+  include Mandate
+
+  initialize_with :approach
+
+  delegate :exercise, to: :approach
```
iHiD: I normally have this under the `private` modifier

app/commands/exercise/approach/link_matching_submissions.rb
```diff
@@ -0,0 +1,35 @@
...
+    submissions.find_each do |submission|
+      new_approach = approach.matching_tags?(submission.tags) ? approach : nil
+      submission.update(approach: new_approach)
+    rescue StandardError => e
+      Bugsnag.notify(e)
+    end
+  end
+
+  def relink_submissions!(submissions)
+    submissions.find_each do |submission|
```
iHiD:
```suggestion
  def relink_submissions!
    linked_submissions.find_each do |submission|
```

app/commands/exercise/approach/link_matching_submissions.rb
```diff
@@ -0,0 +1,35 @@
...
+  def call
+    return relink_submissions!(linked_submissions) if approach.tags.blank?
+
+    update_submissions!(linked_submissions)
+    update_submissions!(unlinked_submissions.tagged)
+  end
+
+  private
+  def linked_submissions = Submission.where(exercise:, approach:)
+  def unlinked_submissions = Submission.where(exercise:, approach: nil)
```
iHiD: These need to just be submissions that have iterations attached to them (else we'll end up doing 20x the amount of work needed). I'd probably do that via an inner join to only get records where there's an iteration with this submission_id.

iHiD: It's probably worth adding a scope on `submissions` for (`has_iteration`) or something?

app/commands/exercise/approach/link_matching_submissions.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def update_submissions!(submissions)
+    submissions.find_each do |submission|
+      new_approach = approach.matching_tags?(submission.tags) ? approach : nil
+      submission.update(approach: new_approach)
```
iHiD: I'm not sure of this logic. I think what I'm seeing is that we find all solutions that are `nil` or this approach, and update them to either be `nil` or this approach. Is that correct? I think, if so, this could do with a comment, as I spent a while trying to work out the logic here. What confused me was whether we were overriding other approaches or not, as we **do** do that via the relink submissions.

Maybe adding this comment at the top of the file would help:

```
# This is intended to run when the tags on an approach changes.
# If the approach has no tags then we want to relink all solutions that are attached to it - this serves
# the case where we remove tags from an approach. In other cases, we want to find submission that
# either are untagged, or ones that were previously linked to this approach, and check whether they should
# be still linked to this approach. Note: this command will not reassign a submission that's linked to a different
# approach to the approach that's passed in. For that, use `Submission::LinkToMatchingApproach.defer`
```

iHiD: A further comment. I feel like if we're setting this to be `nil` here and previously it had an approach, we should instead run `Submission::LinkToMatchingApproach.defer(submission)` in case a different approach **would** match. I think that's the missing bit that's causing me confusion.

app/commands/submission/analysis/process.rb
```diff
@@ -40,7 +40,9 @@ def handle_ops_error!
 
   def handle_completed!
     submission.analysis_completed!
+    submission.update(tags:)
     Solution::UpdateTags.(submission.solution)
+    Submission::LinkToMatchingApproach.defer(submission)
```
iHiD: I think we should do this inline, not via defer. My logic for this is that once `Process` finishes, we alert the user, and it would be nice to say at that stage that there's a relevant approach.

app/commands/submission/analysis/process.rb
```diff
@@ -68,4 +70,7 @@ def tags_data
     Bugsnag.notify(e)
     {}
   end
+
+  memoize
+  def tags = tags_data[:tags]
```
iHiD, line 75: Just a sanity check as to whether `tags_data` can be nil?

app/commands/submission/link_to_matching_approach.rb
```diff
@@ -0,0 +1,19 @@
+class Submission::LinkToMatchingApproach
+  include Mandate
+
+  queue_as :solution_processing
+
+  initialize_with :submission
+
+  def call
+    submission.update(approach: matching_approach)
```
iHiD:
```suggestion
    submission.update!(approach: matching_approach)
```

app/commands/exercise/approach/link_matching_submissions.rb
iHiD: Final comment - you pretty much always want to use `update!` not `update` - else we'll get silent failures.

app/models/exercise/approach.rb
```diff
@@ -31,4 +35,19 @@ def git = Git::Exercise::Approach.new(slug, exercise.slug, exercise.git_type, sy
 
   memoize
   def content_html = Markdown::Parse.(content)
+
+  def matching_tags?(check_tags)
```
iHiD:
```suggestion
  def matches_tags?(check_tags)
```

app/commands/submission/analysis/process.rb
ErikSchierboom: I don't think that's needed:

```ruby
def tags_data
    tags_json = tooling_job.execution_output['tags.json']
    return {} if tags_json.blank?

    res = JSON.parse(tags_json)
    res.is_a?(Hash) ? res.symbolize_keys : {}
  rescue StandardError => e
    Bugsnag.notify(e)
    {}
  end
```

app/commands/exercise/approach/link_matching_submissions.rb
ErikSchierboom: I've re-thought the logic a bit and refactored the code. Let me know if this version makes more sense

PR #6452, opened 2023-11-03
Work out TrainingData Dashboard UI

dem4ron:


app/helpers/react_components/training_data/dashboard.rb
```diff
@@ -33,6 +29,9 @@ def tracks_request
           options: { stale_time: 0 }
         }
       end
+
+      DEFAULT_STATUS = :untagged
+      private_constant :DEFAULT_STATUS
```
iHiD, line 34: @dem4ron Constants tend to go at the bottom of the file if they're private and the top if they're public.

app/helpers/react_components/training_data/code_tagger.rb
```diff
@@ -6,12 +6,12 @@ def to_s
           code: {
             track: {
               title: 'Ruby',
-              iconUrl: 'ruby_icon_url',
-              highlightjsLanguage: 'ruby'
+              icon_url: 'ruby_icon_url',
+              highlightjs_language: 'ruby'
```
iHiD, line 10: Always use underscores in Ruby.

PR #6453, opened 2023-11-03
Rename command to count lines of code

ErikSchierboom:


app/commands/iteration/count_lines_of_code.rb
```diff
@@ -1,4 +1,4 @@
-class Iteration::CalculateLinesOfCode
```
ErikSchierboom, line 1: We're not really calculating as much as counting :)

PR #6475, opened 2023-11-07
Only count passed submissions for representations

iHiD:


app/commands/exercise/representation/update_num_submissions.rb
```diff
@@ -6,17 +6,13 @@ class Exercise::Representation::UpdateNumSubmissions
...
+    representation.update!(num_submissions:)
   end
 
   private
-  def num_submissions_sql
-    Arel.sql(representation.submission_representations.select("COUNT(DISTINCT(submission_id))").to_sql)
+  def num_submissions
+    representation.submission_representations.joins(:submission).
+      where(submissions: { tests_status: :passed }).
+      select(:submission_id).distinct.count
```
ErikSchierboom, line 16: The race-conditions bit does not scare you anymore?

ErikSchierboom: Is there an index on `tests_status`?

iHiD: No and no. But I very much appreciate the sanity check.

> The race-conditions bit does not scare you anymore?

It doesn't really matter if we're off by one here, and the nested condition with the `update_all` scares me more than the very rare occasion.

> Is there an index on tests_status?

No, from what I can tell it's always quicker for MySQL to loop check the rows as there's only like 3k of them, and it's on a join table, so it doesn't really work very well.

If I could make indexes across tables, I'd be a very happy boy :)

PR #6489, opened 2023-11-08
Add num_submissions cache and render approach on solutions

iHiD:


db/schema.rb
```diff
@@ -10,7 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_11_06_131251) do
+ActiveRecord::Schema[7.0].define(version: 2023_11_08_140031) do
+
```
ErikSchierboom:
```suggestion
```

PR #6495, opened 2023-11-09
Fix race condition when updating tags

ErikSchierboom:


app/commands/exercise/update_tags.rb
```diff
@@ -12,6 +12,8 @@ def tags
 
     new_tags = (solution_tags - exercise_tags).map do |tag|
       Exercise::Tag.find_or_create_by!(tag:, exercise:)
```
iHiD, line 14:
```suggestion
      Exercise::Tag.find_create_or_find_by!!(tag:, exercise:)
```

app/commands/exercise/update_tags.rb
```diff
@@ -12,6 +12,8 @@ def tags
...
+    rescue ActiveRecord::RecordNotUnique
+      Exercise::Tag.find_by!(tag:, exercise:)
```
iHiD:
```suggestion
      Exercise::Tag.find_create_or_find_by!(tag:, exercise:)
```

ErikSchierboom: Did you mean to suggest this for the original line, not for the exception?

iHiD: I think this replaces all three lines.

iHiD: The `or_find_by!` does exactly this under the hood.

ErikSchierboom: Okay good!

PR #6496, opened 2023-11-09
Add global css class for completed heading

dem4ron:
Irons this wrinkle:

Screenshot 2023-11-09 at 17 10 12

---

Screenshot 2023-11-09 at 17 11 02


app/css/components/completed-heading.css
```diff
@@ -0,0 +1,21 @@
+.c-completed-heading {
```
iHiD:
```suggestion
.c-completed-exercise-heading {
```

Can you rename this in the various places pls (filename, etc)

PR #6499, opened 2023-11-10
Add racket to languageCompartment.ts

BNAndras:
The Racket track uses the Scheme highlighting from highlight.js so it makes sense to also use the Scheme mode here.


dem4ron:
Works well. However, it seems CodeMirror's syntax highlighting for Scheme/Racket is very basic compared to highlight.js. It only highlights a couple of things, like strings, numbers, the quote symbol. 

Screenshot 2023-11-13 at 12 36 40
Screenshot 2023-11-13 at 12 38 48

dem4ron (approved):

PR #6510, opened 2023-11-14
Disable refetchOnMount for paginated query results

dem4ron:
- Fix casing for initial data on student/TracksList
- Manually reset staleTime after mounting
- Better solution
- Disable refetchOnMount generally

### DIFF
--- a/app/javascript/components/student/TracksList.tsx
+++ b/app/javascript/components/student/TracksList.tsx
@@ -43,12 +43,8 @@ export default ({
     isFetching,
   } = usePaginatedRequestQuery<APIResponse>(CACHE_KEY, {
     ...request,
-    options: { ...request.options, staleTime },
+    options: { ...request.options, refetchOnMount: false },
   })
-
-  useEffect(() => {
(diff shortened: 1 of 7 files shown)


iHiD (approved):

app/javascript/components/student/TracksList.tsx
```diff
@@ -29,6 +29,7 @@ export default ({
   tagOptions: readonly TagOption[]
   request: Request
 }): JSX.Element => {
+  const [staleTime, setStaleTime] = useState(1000)
```
iHiD: Legacy?

dem4ron: Yes

ErikSchierboom: But this isn't used?

iHiD: No. It's not needed (and not used in the code as far as I can tell).

dem4ron: I used this to try out a concept, and forgot to delete.

PR #6513, opened 2023-11-15
Don't send emails to ghost or system user

ErikSchierboom:


app/models/user.rb
```diff
@@ -297,6 +297,8 @@ def may_receive_emails?
     return false if disabled?
     return false if email.ends_with?("users.noreply.github.com")
     return false if email_status_invalid?
+    return false if system?
```
ErikSchierboom, line 300: We probably don't ever do this, but it might not hurt being safe

PR #6514, opened 2023-11-15
Add maximum length validation to user fields

ErikSchierboom:


db/schema.rb
```diff
@@ -373,9 +373,9 @@
   end
 
   create_table "exercise_tags", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.bigint "exercise_id", null: false
```
iHiD: Can you revert these pls?

iHiD (approved):

PR #6516, opened 2023-11-15
Gracefully handle discussions API status being invalid

ErikSchierboom:


app/commands/mentor/discussion/retrieve.rb
```diff
@@ -17,10 +17,7 @@ def initialize(mentor,
                  sorted: true, paginated: true,
                  exclude_uuid: nil)
 
-    # This will be a code-level exception rather than a user-level
-    # exception so we don't worry about a special class. Getting this
-    # tells us that we've got a bug.
```
ErikSchierboom, line 22: As this is directly callable from the API, I though that it might make sense to have a custom error

iHiD (approved):
Looks great. Does Aron need to do anything on the front end for this?

ErikSchierboom:
> Does Aron need to do anything on the front end for this? 

Not sure. @dem4ron could you check?

PR #6526, opened 2023-11-17
Don't register own viewed solution

ErikSchierboom:


app/commands/user_track/viewed_community_solution/create.rb
```diff
@@ -4,6 +4,8 @@ class UserTrack::ViewedCommunitySolution::Create
   initialize_with :user, :track, :solution
 
   def call
+    return if solution.user == user
```
iHiD: (Avoid an extra db lookup)

```suggestion
    return if solution.user_id == user.id
```

ErikSchierboom:
The cleanup script has run

PR #6527, opened 2023-11-20
Show correct error message

dem4ron:
https://github.com/exercism/website/assets/66035744/8f641579-f6a1-4076-8c32-cefc03b9aa99


app/javascript/utils/fetch-json.ts
```diff
@@ -1,35 +1,46 @@
...
 
-export const fetchJSON = <T extends any>(
+export class ApiError extends Error {
+  type: string
+
+  constructor(type: string, message: string) {
+    super(message)
+    this.name = this.constructor.name
+    this.type = type
+    Object.setPrototypeOf(this, ApiError.prototype)
```
ErikSchierboom: Why is this needed?

app/javascript/utils/fetch-json.ts
```diff
@@ -1,35 +1,46 @@
 import { camelizeKeys } from 'humps'
 
-export const fetchJSON = <T extends any>(
+export class ApiError extends Error {
+  type: string
+
+  constructor(type: string, message: string) {
```
ErikSchierboom: Could this be:

```suggestion
  constructor(public type: string, message: string) {
```

app/javascript/utils/fetch-json.ts
dem4ron: For Typescript. It is needed to correctly set the prototype chain when extending built-in classes. This ensures that instance methods, like instanceof, function correctly.

app/javascript/utils/fetch-json.ts
dem4ron: yes!

db/schema.rb
```diff
@@ -39,6 +39,11 @@
     t.index ["blob_id", "variation_digest"], name: "index_active_storage_variant_records_uniqueness", unique: true
   end
 
+  create_table "add_uuid_to_code_tags_samples", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
```
ErikSchierboom: I think this should probably be removed, right @iHiD

app/views/training_data/dashboard/index.html.haml
ErikSchierboom, line 1: app/views/training_data/dashboard/index.html.haml was moved to app/views/training_data/code_tags_samples/index.html.haml. Could you update?

dem4ron: then I think the whole `app/views/training_data/dashboard` folder can be deleted, right?

dem4ron: ~also `code_tags_samples` could be removed, and its content moved one level up.~

app/controllers/training_data/dashboard_controller.rb
ErikSchierboom, line 1: Could you revert this?

PR #6544, opened 2023-11-22
Add LLM tags to code tags samples

ErikSchierboom:


db/migrate/20231122082820_add_llm_tags_to_code_tags_samples.rb
```diff
@@ -0,0 +1,9 @@
+class AddLlmTagsToCodeTagsSamples < ActiveRecord::Migration[7.0]
+  def change
+    return if Rails.env.production?
+
+    add_column :training_data_code_tags_samples, :llm_tags, :text, null: true
+
+    TrainingData::CodeTagsSample.update_all('llm_tags = tags')
```
ErikSchierboom, line 7: I did this just for easy testing locally :)

app/models/training_data/code_tags_sample.rb
```diff
@@ -18,6 +19,7 @@ class TrainingData::CodeTagsSample < ApplicationRecord
 
     self.exercise_id = solution.exercise_id unless exercise_id
     self.track_id = exercise.track_id unless track_id
+    self.llm_tags = tags unless llm_tags
```
ErikSchierboom: I thought that maybe this was the simplest way of doing this 🤷

iHiD: This isn't write, because the tags might not be coming from an LLM. I think just leave this out.

iHiD:
```suggestion
```

test/models/training_data/code_tags_sample_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    sample = create(:training_data_code_tags_sample, tags: nil)
+    assert_nil sample.tags
+    assert_nil sample.llm_tags
+
+    tags = '["construct:if"]'
+    sample = create(:training_data_code_tags_sample, tags:)
+    assert_equal tags, sample.tags
+    assert_equal tags, sample.llm_tags
+  end
+
```
iHiD:
```suggestion
```

PR #6546, opened 2023-11-22
Set latest brief introduction video to C#

ErikSchierboom:


app/views/community/brief_introductions/index.html.haml
```diff
@@ -14,7 +14,7 @@
 
       .flex-shrink-0{ class: "lg:w-[400px] w-100" }
         .c-youtube-container
-          %iframe{ width: "560", height: "315", src: "https://www.youtube-nocookie.com/embed/zoGruqFThFE", title: "YouTube video player", frameborder: "0", allow: "accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture", allowfullscreen: true }
+          %iframe{ width: "560", height: "315", src: "https://www.youtube-nocookie.com/embed/xVl1NZaJYIg", title: "YouTube video player", frameborder: "0", allow: "accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture", allowfullscreen: true }
```
iHiD: I think this is incompleted.

I think the ID should be added to the controller (`app/controllers/community/brief_introductions_controller.rb`). 

It would also be nice if this was a variable in the controller (`@latest_id`) so we never have to touch the view.

ErikSchierboom: Good idea. Updated.

PR #6549, opened 2023-11-22
store llm tags when generating

ErikSchierboom:
- Store LLM tags when generating tags
- Fix typo
- Rename tuple to sample
- Fix broken test
- Update safe_to_override? logic
- Add tests for generating tags

### DIFF
--- a/app/models/training_data/code_tags_sample.rb
+++ b/app/models/training_data/code_tags_sample.rb
@@ -26,9 +26,7 @@ class TrainingData::CodeTagsSample < ApplicationRecord
   def dataset = super.to_sym
   def status = super.to_sym
 
-  def safe_to_override?
-    %i[human_tagged admin_tagged community_checked].include?(status)
-  end
+  def safe_to_override? = %i[untagged machine_tagged].include?(status)
 
   def locked? = locked_until && locked_until > Time.current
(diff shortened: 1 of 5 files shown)


app/models/training_data/code_tags_sample.rb
```diff
@@ -26,9 +28,7 @@ def to_param = uuid
   def dataset = super.to_sym
   def status = super.to_sym
 
-  def safe_to_override?
-    %i[human_tagged admin_tagged community_checked].include?(status)
-  end
```
ErikSchierboom, line 31: It looked to me that these were the statuses that were _not_ safe to override.

app/models/training_data/code_tags_sample.rb
```diff
@@ -70,4 +70,7 @@ def next_status
       status
     end
   end
+
+  memoize
+  def code = files.to_a.pluck("code").join('\n')
```
ErikSchierboom, line 75: Not sure how we want to handle multiple files, so I just concatenated them

iHiD: Yeah - I wasn't sure about this either. I feel like it will make it harder for an LLM to cope if both `c` and `h` files are concatenated but we can experiment. This is fine for now.

PR #6550, opened 2023-11-22
Improve training data tracks selector

ErikSchierboom:


app/models/training_data/code_tags_sample.rb
```diff
@@ -70,4 +70,15 @@ def next_status
       status
     end
   end
+
+  def self.sample_status(filter_status)
```
ErikSchierboom: @iHiD I couldn't come up with a good name. Let me know if you have any suggestions. I also considered adding a scope, but that would get messy I think

app/helpers/react_components/training_data/dashboard.rb
```diff
@@ -25,10 +25,14 @@ def training_data_request
       end
 
       def tracks
-        tracks = ::Track.where(id: current_user.user_tracks.trainer.select(:track_id))
-        AssembleTracksForSelect.(tracks)
+        ts = ::Track.where(id: ::TrainingData::CodeTagsSample.where(status: sample_status).select(:track_id))
+        ts = ts.where(id: current_user.user_tracks.trainer.select(:track_id)) unless current_user.staff?
```
ErikSchierboom: This does two things:

1. Only show relevant tracks, i.e. the ones for which there is data
2. Only filter the relevant tracks when the user is not staff

CC @dem4ron could you check this works?

app/models/training_data/code_tags_sample.rb
iHiD: `statuses_for_filter`?

PR #6554, opened 2023-11-23
Add december diversions badge

ErikSchierboom:


app/images/icons/badge-december-diversions.svg
ErikSchierboom, line 1: One for you @iHiD

PR #6559, opened 2023-11-24
Fix MobileIterationView

dem4ron:
Screenshot 2023-11-24 at 8 59 28
Screenshot 2023-11-24 at 8 59 13


iHiD:
@dem4ron What caused this and what could we have done to prevent it? Are we missing some tests we should have here?

PR #6579, opened 2023-11-30
Add command to fine tune model

ErikSchierboom:
- Deduplicate tags
- Test that nil tags are gracefully handled
- Move tag normalization to separate command
- Normalize tags
- Update normalized tags table
- Sort lines
- Add work on fine-tuning model
- Add ruby-openai gem
- Gracefully handle llm_tags not being present
- Add command to fine tune model


app/commands/training_data/code_tags_sample/create_file_for_model_fine_tuning.rb
```diff
@@ -0,0 +1,413 @@
...
+    file.close
+  end
+
+  def tags_with_namespace(namespace)
+    CRITERIA.keys.select { |tag| tag.starts_with?("#{namespace}:") }
+  end
+
+  SYSTEM_TEXT = 'You are a expert in EXERCISM_REPRESENTATION_TAGS'.freeze
+
+  CRITERIA = {
```
ErikSchierboom, line 133: Not sure if there is a better place for this 🤷

app/commands/training_data/code_tags_sample/normalize_tags.rb
```diff
@@ -0,0 +1,205 @@
+class TrainingData::CodeTagsSample::NormalizeTags
+  include Mandate
+
+  initialize_with :tags
+
+  def call
+    return nil if tags.nil?
+
+    tags.flat_map { |tag| MAPPING.fetch(tag, tag) }.uniq
```
ErikSchierboom, line 9: This not only normalizes the concept via the below mapping, but also removes any duplicates (we didn't do that before).

app/commands/training_data/code_tags_sample/create_file_for_model_fine_tuning.rb
iHiD: I think these should all be in the database, probably in a global tag table.

PR #6580, opened 2023-11-30
Add command to create training data

ErikSchierboom:


app/commands/training_data/code_tags_sample/create.rb
```diff
@@ -7,15 +7,17 @@ def call
...
+      attributes.dataset = dataset
+      attributes.status = :untagged
+      attributes.files = submission.files.map do |file|
         { "filename" => file.filename, "code" => file.content }
       end
     end
 
-    TrainingData::CodeTagsSample::GenerateTags.defer(sample) if sample.llm_tags.blank?
+    sample.tap do
+      TrainingData::CodeTagsSample::GenerateTags.defer(sample) if sample.llm_tags.blank?
```
iHiD: This should only run if it's `training` not `validation` I think.

ErikSchierboom: Updated

PR #6583, opened 2023-12-01
Create site tags

ErikSchierboom:
- Add site tag model
- Make tag unique for site tag
- Remove blank tags and treat empty tags as nil
- Show site tags minus disabled track tags as suggestions


app/commands/training_data/code_tags_sample/normalize_tags.rb
```diff
@@ -5,8 +5,13 @@ class TrainingData::CodeTagsSample::NormalizeTags
 
   def call
     return nil if tags.nil?
+    return nil if normalized_tags.empty?
```
ErikSchierboom, line 8: I felt that treating empty tags as `nil` makes things a bit easier (e.g. querying)

app/commands/training_data/code_tags_sample/normalize_tags.rb
```diff
@@ -5,8 +5,13 @@ class TrainingData::CodeTagsSample::NormalizeTags
...
 
-    tags.flat_map { |tag| MAPPING.fetch(tag, tag) }.uniq
+    normalized_tags
+  end
+
+  def normalized_tags
+    tags.flat_map { |tag| MAPPING.fetch(tag, tag) }.reject(&:blank?).uniq
```
ErikSchierboom, line 14: I'm now also removing blank tags

db/migrate/20231201091435_create_site_tags.rb
```diff
@@ -0,0 +1,293 @@
...
+    return if Rails.env.production?
+
+    create_table :site_tags do |t|
+      t.string :tag, null: false, index: { unique: true }
+      t.string :description, null: false
+
+      t.timestamps
+    end
+
+    [
```
ErikSchierboom, line 12: Let me know if you don't want this in the migration. It is an easy way for us to get the data locally 🤷

iHiD: It's fine.

PR #6595, opened 2023-12-04
Add site-header and track menu skeletons

dem4ron:
https://github.com/exercism/website/assets/66035744/ca801920-5670-42b5-95cc-86230158ef67

Screenshot 2023-12-04 at 11 49 29

https://github.com/exercism/website/assets/66035744/6e72b98c-20af-414f-975c-e5fd6dc6df86

https://github.com/exercism/website/assets/66035744/c13584e1-9efb-40c7-a2e2-dba81a8f99f0

https://github.com/exercism/website/assets/66035744/e597afaf-8cec-4a89-abef-1b9e8ec0844a

https://github.com/exercism/website/assets/66035744/c92fe368-8279-4de7-8e1a-df7364e5947f

Screenshot 2023-12-04 at 11 04 39

### DIFF
--- a/app/javascript/components/common/skeleton/skeletons/ReputationDropdownSkeleton.tsx
+++ b/app/javascript/components/common/skeleton/skeletons/ReputationDropdownSkeleton.tsx
@@ -1,17 +1,29 @@
 import React from 'react'
 import { SkeletonShape } from '../components/SkeletonShape'
 import { SkeletonLoader } from '../components/SkeletonLoader'
+import { useLogger } from '@/hooks'
+import { getTextWidth } from '@/utils/get-text-width'
 
-export function ReputationDropdownSkeleton() {
+// pl-70 comes from:
+// px-16 + 24px icon with mr-8 + 3px border
+// 32 + 24 + 8 + 6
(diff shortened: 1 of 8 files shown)


app/javascript/components/common/skeleton/skeletons/ReputationDropdownSkeleton.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { SkeletonLoader } from '../components/SkeletonLoader'
+
+export function ReputationDropdownSkeleton() {
+  return (
+    <SkeletonLoader>
+      <SkeletonShape
+        shape="tag"
+        style={{
+          height: '38px',
+          width: '100px',
```
iHiD: Rather than hardcoding this, how about we set the width to be `x+(nd*s)` where x is the fixed width of the left/right, `nd` is the number of digits and `s` is a fixed "space per digit" thing (e.g. 10px). We can set the same on the real component. I know it won't be perfectly correct as the characters aren't the right width, but if we just bleed the extra space into the left/right padding, by centralising the content, I don't think it'll matter.

PR #6617, opened 2024-01-03
Remove references to 12in23

dem4ron:
Screenshot 2024-01-03 at 15 05 10
Screenshot 2024-01-03 at 15 05 03
Screenshot 2024-01-03 at 15 04 58

### DIFF
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -25,8 +25,6 @@ class ChallengesController < ApplicationController
   def use_challenge_id!
     @challenge_id = params[:id]
 
-    render_404 if @challenge_id == '12in23'
-
     redirect_to root_path unless User::Challenge::CHALLENGES.include?(@challenge_id)
   end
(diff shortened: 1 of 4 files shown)


app/controllers/challenges_controller.rb
```diff
@@ -24,6 +24,9 @@ def start
   private
   def use_challenge_id!
     @challenge_id = params[:id]
+
+    render_404 if @challenge_id == '12in23'
+
```
iHiD: Could you remove this one please. I'm happy for people to still see their old progress pages. The rest looks great :)

app/controllers/challenges_controller.rb
```diff
@@ -24,6 +24,7 @@ def start
   private
   def use_challenge_id!
     @challenge_id = params[:id]
+
```
dem4ron:
```suggestion
```

iHiD (approved):

PR #6622, opened 2024-01-04
Discord mentors permission

ErikSchierboom:
- Add mentor? method to User
- Set mentor role in Discord
- Add private
- Refactor setting discord roles
- Add test to verify that discord roles are updated when updating mentored tracks

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -344,6 +344,12 @@ class User < ApplicationRecord
 
   def flair = super&.to_sym
 
+  def mentor?
+    return true if staff?
+
+    track_mentorships.exists?
+  end
+
   def automator?(track = nil)
     return true if staff?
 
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -344,6 +344,12 @@ def captcha_required? = !github_auth? && Time.current - created_at < 2.days
 
   def flair = super&.to_sym
 
+  def mentor?
+    return true if staff?
+
+    track_mentorships.exists?
```
iHiD: I think this should actually be `user.became_mentor_at`

iHiD: (There's probably a helper method on `User` already)

ErikSchierboom: > I think this should actually be user.became_mentor_at

That does imply that you don't have to be an active mentor, is that correct?

> (There's probably a helper method on User already)

Nope

iHiD: Yeah. It might be for people who want advice before starting.

There's a `user.mentor?` you can use (it's defined via roles)

iHiD:
This is red btw.

ErikSchierboom:
Green now!

PR #6644, opened 2024-01-09
Support article and approach widgets

ErikSchierboom:
https://github.com/exercism/docs/pull/481 updates the docs


ErikSchierboom (commented):
@iHiD Do we want tooltips for these links (like we do for concepts and exercises)?

PR #6645, opened 2024-01-09
Add optional chaining to users in Credits

dem4ron:
#### This PR just silences the error, but doesn't solve the bug which seems to be on the backend.

Screenshot 2024-01-09 at 11 48 42


ErikSchierboom:
I'm not sure we need to PR, as it _was_ an actual error and when we silence it, we won't learn about it.

PR #6654, opened 2024-01-10
Snippet extension

iHiD:


app/models/git/track.rb
```diff
@@ -144,8 +146,13 @@ def taught_concept_slugs
       concept_exercise_concept_slugs & concept_slugs
     end
 
+    memoize
+    def approaches_snippet_extension
+      (config[:approaches] || {})[:snippet_extension].presence || "txt"
```
ErikSchierboom: Not sure if this works, but maybe try this?

```suggestion
      config[:approaches].to_h.fetch(:snippet_extension, "txt")
```

ErikSchierboom (approved):

app/models/git/track.rb
iHiD: Maybe, also `config.dig(:approaches, :snippet_extension).presence || "txt"` works.

PR #6658, opened 2024-01-10
:bug: Fix number localization for "mentoring discussions"

homersimpsons:
This should solve https://forum.exercism.org/t/the-numbers-in-impact-does-not-share-the-same-formatting/9025

Note that I did try to setup Exercism in a GitHub codespace, but I failed !6659. So this is not tested. Testing this should be rater easy:
1. navigate to /about/impact page
2. change browser's location https://developer.chrome.com/docs/devtools/settings/locations to France/Paris
3. Check whether the 3 numbers are displaying the same (space as thousands separators, not comma)

Also it looks like most of the metrics are declared in `/app/javascript/components/impact/map.tsx`, but from what I understand there is no need to declare this one here.

### DIFF
--- a/app/views/about/impact.html.haml
+++ b/app/views/about/impact.html.haml
@@ -37,7 +37,8 @@
 
           .py-12.px-32
             %h3.label-large Mentoring Discussions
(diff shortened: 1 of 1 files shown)


iHiD:
Thanks :)

PR #6667, opened 2024-01-15
Update track name for Mac

ShrootBuck:
Mac OS X was changed to "macOS" in 2016. I think the current name would be a more appropriate term, but let me know your thoughts


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

PR #6677, opened 2024-01-16
Update 48in24.html.haml: Fix Exercise Link to use slug

highb:
The link on the 48in24 page seems to be linking to a serialized exercise object. This might fix it, if I'm following the other exercise link example correctly.


ErikSchierboom:
Good catch!

PR #6687, opened 2024-01-18
Sync prob-specs exercises

ErikSchierboom:
- Add initial setup for syncing problem-specifications
- Create model for generic exercise
- Add status field
- Add status field to generic exercise
- Add command to create or update generic exercise
- Sync exercises from prob-specs
- Use generic exercise
- Use generic exercise to determine unimplemented
- Fix build status
- Use generic exercise in exercise model
- Remove unused model
- Remove unneeded models


app/controllers/generic_exercises_controller.rb
```diff
@@ -5,7 +5,7 @@ class GenericExercisesController < ApplicationController
   skip_before_action :verify_authenticity_token, only: :start
 
   def show
-    @ps_data = Git::ProblemSpecifications::Exercise.new(params[:id])
+    @ps_data = GenericExercise.find_by!(slug: params[:id])
```
iHiD: Maybe this is `@exercise.generic_exercise`?

app/models/generic_exercise.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def status = super.to_sym
+
+  memoize
+  def git = Git::ProblemSpecifications::Exercise.new(slug)
+
+  def url = "https://github.com/exercism/problem-specifications/tree/main/exercises/#{slug}"
+  def icon_url = "#{Exercism.config.website_icons_host}/exercises/#{slug}.svg"
+
+  def self.for(slug) = find_by(slug:)
```
iHiD: I always like having a `for!` too.

```suggestion
  def self.for!(slug) = find_by!(slug:)
  def self.for(slug)
    for!(slug)
  rescue ActiveRecord::RecordNotFound
    nil
  end
```

app/views/tracks/exercises/show/_external.html.haml
```diff
@@ -7,7 +7,7 @@
     - else
       = render("tracks/exercises/show/action_box_join_exercism", exercise:, user_track:)
 
-    - deep_dive_youtube_id = exercise.problem_specifications_data.deep_dive_youtube_id
+    - deep_dive_youtube_id = exercise.generic_exercise.deep_dive_youtube_id
```
iHiD: We could delegate this in the `exercise` model maybe?

PR #6690, opened 2024-01-18
Store approach and article order

ErikSchierboom:
- Add position to approaches and articles
- Add scope for sorted articles and approaches
- Fix factories
- Sync article and approach order as found in config
- Dig deeper page shows articles and approaches in order

### DIFF
--- a/app/models/exercise/approach.rb
+++ b/app/models/exercise/approach.rb
@@ -27,7 +27,7 @@ class Exercise::Approach < ApplicationRecord
   has_one :track, through: :exercise
 
   scope :random, -> { order('RAND()') }
-  scope :sorted, -> { order(:position) }
+  default_scope { order(:position) }
 
   delegate :content, :snippet, to: :git
 
--- a/app/models/exercise/approach.rb
+++ b/app/models/exercise/approach.rb
@@ -27,6 +27,7 @@ class Exercise::Approach < ApplicationRecord
   has_one :track, through: :exercise
 
   scope :random, -> { order('RAND()') }
+  scope :sorted, -> { order(:position) }
 
   delegate :content, :snippet, to: :git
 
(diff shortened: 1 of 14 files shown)


app/models/exercise/approach.rb
```diff
@@ -27,6 +27,7 @@ class Exercise::Approach < ApplicationRecord
   has_one :track, through: :exercise
 
   scope :random, -> { order('RAND()') }
+  scope :sorted, -> { order(:position) }
```
iHiD: I'd probably set this to a "default scope" instead of having to remember to use this each time.

db/migrate/20240118102648_add_position_to_exercise_approaches.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    return if Rails.env.production?
+
+    add_column :exercise_approaches, :position, :integer, null: true, limit: 1
+    add_index :exercise_approaches, [:exercise_id, :position]
+
+    # The positions will be fixed manually by running Git::SyncTrack.()
+    Exercise::Approach.update_all(position: 0)
+
+    change_column_null :exercise_approaches, :position, false
```
iHiD, line 11: You could also do:

```suggestion
    add_column :exercise_approaches, :position, :integer, null: false, limit: 2, default: 0
    add_index :exercise_approaches, [:exercise_id, :position]
```

ErikSchierboom: Yes, but then it would silently fail if our code ever forgot to pass in the position

ErikSchierboom: I mean, I don't massively care, but 🤷

PR #6696, opened 2024-01-19
process ops errors gracefully

ErikSchierboom:
- Process representation ops errors gracefully
- Handle ops errors gracefully


app/commands/submission/analysis/process.rb
```diff
@@ -52,6 +55,8 @@ def submission
 
   memoize
   def data
+    return nil if ops_errored?
```
iHiD:
```suggestion
    return {} if ops_errored?
```

(Same as the exception below?)

app/commands/submission/analysis/process.rb
```diff
@@ -61,6 +66,8 @@ def data
 
   memoize
   def tags_data
+    return nil if ops_errored?
```
iHiD:
```suggestion
    return {} if ops_errored?
```

app/commands/submission/representation/process.rb
```diff
@@ -15,8 +15,9 @@ def call
 
   private
   def ops_status = tooling_job.execution_status.to_i
+  def ops_errored? = ops_status != 200
   def ast_digest = Submission::Representation.digest_ast(ast)
-  def representer_version = metadata[:version] || 1
+  def representer_version = metadata.present? ? metadata.fetch(:version, 1) : 1
```
iHiD:
```suggestion
  def representer_version = metadata.dig(:version) || 1
```

app/commands/submission/representation/process.rb
```diff
@@ -48,6 +53,8 @@ def mapping
 
   memoize
   def metadata
+    return nil if ops_errored?
```
iHiD:
```suggestion
    return {} if ops_errored?
```

app/commands/submission/representation/process.rb
```diff
@@ -39,6 +42,8 @@ def ast
 
   memoize
   def mapping
+    return nil if ops_errored?
```
iHiD:
```suggestion
    return {} if ops_errored?
```

app/commands/submission/analysis/process.rb
ErikSchierboom: I saw that the database column was nullable, and as we probably don't want to process errored records anyway, so I thought nil. Still want `{}`?

iHiD: I think it should be the same as the exception clause. That's my only hard rule. I think I prefer `{}` over `nil` just to avoid the sort of errors that we were getting here, but I'm less precious about that bit as long as they're both the same.

PR #6705, opened 2024-01-19
Prevent linking author with no public profile

dem4ron:


app/views/blog_posts/show.html.haml
```diff
@@ -38,7 +38,8 @@
           Found it useful?
           = render ViewComponents::Blog::SharePostLink.new(@post)
 
-        = link_to profile_path(@post.author), class: "byline shadow-sm rounded-8 py-16 px-24 flex flex-col md:flex-row md:items-center border-1 border-borderColor6" do
+        - link_path = @post.author.profile? ? profile_path(@post.author) : "javascript:void(0);"
+        = link_to link_path, class: "byline shadow-sm rounded-8 py-16 px-24 flex flex-col md:flex-row md:items-center border-1 border-borderColor6" do
```
iHiD: I think I prefer this rather than the void solution. I'm not 100% sure the method name etc is correct, but I think it's right.

```suggestion
        - t = @post.author.profile? ? :a : :div
        = content_tag t, href: profile_path(@post.author), class: "byline shadow-sm rounded-8 py-16 px-24 flex flex-col md:flex-row md:items-center border-1 border-borderColor6" do
```

PR #6712, opened 2024-01-22
Add page for 48in24 implementation status

ErikSchierboom:


config/routes.rb
```diff
@@ -237,6 +237,8 @@
     end
   end
 
+  get "challenges/48in23/implementation_status", to: 'challenges#implementation_status'
```
ErikSchierboom: I struggled a bit with Rails routing, so I just hacked this together

app/views/challenges/implementation_status.html.haml
```diff
@@ -0,0 +1,22 @@
+#challenge-48in24-page
+  %section.top-section
+    .lg-container
+      .flex.xl:flex-row.flex-col.xl:items-start.items-center
+        .xl:mr-auto{ class: 'max-w-[860px]' }
+          %table
```
ErikSchierboom: It's not particularly pretty and it definitely won't be pretty once we have 48 exercises to render

config/routes.rb
```diff
@@ -237,6 +237,8 @@
...
+
   resources :challenges, only: [:show] do
     post :start, on: :member
```
iHiD, line 241:
```suggestion
    post :start, on: :member
    get :implementation_status, on :member
```

config/routes.rb
iHiD: See comment below.

PR #6734, opened 2024-01-26
Add more robust youtube player, add deep dive video step to submit flow

dem4ron:


iHiD:
@dem4ron Let me know when this is ready btw. Would be nice to get it on in the next day or two :)

app/javascript/components/modals/realtime-feedback-modal/feedback-content/DeepDiveVideoContent.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { YouTubePlayer } from '@/components/common/YoutubePlayer'
+
+export function DeepDiveVideoContent({ exercise, onContinue }) {
+  return (
+    <>
+      <h3 className="text-h3 mb-8">Deep Dive into {exercise.title}!</h3>
+      <p className="text-p-large mb-16">
+        Take a deep dive into {exercise.title} with Jeremy and Erik, as they
+        explore the different ways this exercise can be solved and dig into some
+        interesting community solutions.
```
iHiD: If you check back in the video I suggest passing in some extra text here that's specific per exercsie.

dem4ron: I'll pass this in for now on the ruby end and replace this with `deep_dive_blurb` once that is ready.

app/helpers/react_components/editor.rb
```diff
@@ -86,7 +93,20 @@ def request
       }
     end
 
-    private
+    def show_deep_dive_video?
+      return false unless exercise.deep_dive_youtube_id
```
iHiD:
```suggestion
      return false unless exercise.deep_dive_youtube_id.present?
```

app/javascript/components/track/dig-deeper-components/DeepDiveVideo.tsx
```diff
@@ -17,7 +17,7 @@ export function DeepDiveVideo() {
         interesting community solutions.
       </p>
       <div className="w-[100%]">
-        <YoutubeEmbed id={exercise.deepDiveYoutubeId} />
+        <YouTubePlayer id={exercise.deepDiveYoutubeId} />
```
iHiD: I suggest standardising on `YouTube` or Youtube`, at least on the JS side. There's a lot of variance right now. 

(The Ruby side should be Youtube I think because of complex Rails rules, but JS might be better as `YouTube`).

PR #6743, opened 2024-01-31
fix: Pharo track slug is `pharo-smalltalk` (fix for 48-in-24 exercises)

clechasseur:
In the #48in24 [implementation status page](https://exercism.org/challenges/48in24/implementation_status), Pharo is never correctly identified for weeks where it's a featured language, because the track slug is incorrectly specified. This PR fixes that.

I found this by chance but will talk about it here: https://forum.exercism.org/t/little-typo-on-the-front-page-in-the-48in24-part/9545


ErikSchierboom (approved):
Thanks!

PR #6745, opened 2024-01-31
Sync deep dive blurb from prob-specs

ErikSchierboom:
- Add deep dive data to test prob-specs repo
- Add deep_dive_blurb column to generic exercise
- Sync deep dive blurb


db/migrate/20240131123604_add_deep_dive_blurb_to_generic_exercises.rb
```diff
@@ -0,0 +1,7 @@
+class AddDeepDiveBlurbToGenericExercises < ActiveRecord::Migration[7.0]
+  def change
+    return if Rails.env.production?
+
+    add_column :generic_exercises, :deep_dive_blurb, :text, null: true
```
ErikSchierboom: The deep dive blurb of `raindrops` didn't fit in a regular `:string` column: https://github.com/exercism/problem-specifications/blob/main/exercises/raindrops/metadata.toml#L6C20-L6C337

It is currently 317 characters long, so I set it to 350 (which we use in several places), but let me know if you think that is not enough. If it is, I'll need to update prob-specs to guard against the max being exceeded.

iHiD: OK - I'll work out the correct length based on the UI.

iHiD: The default (255?) is correct I think. The text was too long.

ErikSchierboom: Okay. I'll remove the limit

PR #6760, opened 2024-02-01
Use actual user count for percentage_awardees

ErikSchierboom:


app/commands/infrastructure/analyze_table.rb
```diff
@@ -0,0 +1,9 @@
+class Infrastructure::AnalyzeTable
```
ErikSchierboom, line 1: Not sure if this is the right module to put this in

config/sidekiq-schedule.yml
```diff
@@ -62,6 +62,11 @@ fetch_and_sync_all_pull_requests_reputation:
   class: "FetchAndSyncAllPullRequestsReputationJob"
   queue: dribble
 
+analyze_tables:
+  cron: "0 1 * * 1"
```
ErikSchierboom, line 66: This runs once a week, which is fine I think for our purposes

test/models/badge_test.rb
```diff
@@ -2,8 +2,10 @@
 
 class BadgeTest < ActiveSupport::TestCase
   test "percentage_awardees" do
-    badge = create :badge, num_awardees: 123_456
-    assert_equal 15.44, badge.percentage_awardees
+    create_list(:user, 21)
+
+    badge = create :badge, num_awardees: 4
```
ErikSchierboom: For some reason I'm now getting this error:

```shell
ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry 'Rookie' for key 'badges.index_badges_on_name'
    test/models/badge_test.rb:7:in `block in <class:BadgeTest>'

rails test test/models/badge_test.rb:4
```

I'm not exactly sure how my changes are causing this to happen. Thoughts?

app/models/badge.rb
```diff
@@ -61,8 +61,19 @@ def award_to?(_user) = raise "Implement this method in the child class"
...
+  def num_users
+    query = <<~QUERY
+      SELECT TABLE_ROWS
+      FROM INFORMATION_SCHEMA.TABLES
+      WHERE TABLE_NAME='#{User.table_name}'
+      AND TABLE_SCHEMA='#{ActiveRecord::Base.connection.current_database}'
+      LIMIT 1;
+    QUERY
+
+    ActiveRecord::Base.connection.execute(query).first[0]
```
iHiD: I suspect this would work, but it needs checking.

```suggestion
    ActiveRecord::Base.connection.select_value(query)
```

test/models/badge_test.rb
iHiD: I suspect you need to create the badge before creating the users.

app/commands/infrastructure/analyze_table.rb
iHiD: Yeah, seems good!

test/models/badge_test.rb
```diff
@@ -2,8 +2,10 @@
...
+    assert_equal 9.76, badge.percentage_awardees
```
ionFreeman: Sorry, I don't generally read Ruby. What's this test saying? That 4 is 9.76% of something? Is that something 41?

PR #6800, opened 2024-02-15
Update 48in24 exercises

ErikSchierboom:


app/commands/user/challenges/featured_exercises_progress_48_in_24.rb
```diff
@@ -12,7 +12,7 @@ class User::Challenges::FeaturedExercisesProgress48In24
     { week: 5, slug: 'protein-translation', title: 'Protein Translation', featured_tracks: %w[fsharp crystal csharp], learning_opportunity: "" },
```
iHiD, line 12: This is a fun little exercise that lets you delve into recursion, pattern matching, higher order functions, and other functional concepts. Our featured solutions also use bitshifting, regexp and more!

iHiD:
```suggestion
    { week: 5, slug: 'protein-translation', title: 'Protein Translation', featured_tracks: %w[fsharp crystal csharp], learning_opportunity: "This is a fun little exercise that lets you delve into recursion, pattern matching, higher order functions, and other functional concepts. Our featured solutions also use bitshifting, regexp and more!" },
```

PR #6805, opened 2024-02-17
Fix #48in24 language that states there's an exercise per month instead of per week

ryanplusplus:
image


ryanplusplus:
As far as I can tell I'm failing an image comparison because the changed text is included in an image used for another test. Can someone help me with the proper way to update the image(s)?

dem4ron (approved):
Thanks! I think this PR is perfectly fine.

PR #6809, opened 2024-02-19
Link exercise in mentoring session info

dem4ron:
https://github.com/exercism/website/assets/66035744/fe802e8a-fab2-4f7d-a9de-04888943752e


app/javascript/components/mentoring/Session.tsx
```diff
@@ -51,7 +51,7 @@ import { usePanelFallback } from './session/mobile-code-panel/usePanelFallback'
 export type Links = {
   mentorDashboard: string
   mentorQueue: string
-  improveNotes: string
```
ErikSchierboom, line 54: This isn't used anywhere? If not, is that something we should do @ihid?

dem4ron: No, this isn't even passed down, so I removed it.

ErikSchierboom (approved):

iHiD (approved):

PR #6813, opened 2024-02-21
Check if new RegExp features are supported before highlighting

dem4ron:


app/javascript/utils/regex-check.ts
ErikSchierboom, line 1: How did you come up with the list of features to test for? And what is the issue being solved (I guess highlighting not working :))?

iHiD: Context: https://exercism-io.slack.com/archives/C05626SPQP4/p1708523611098849?thread_ts=1708278181.399079&cid=C05626SPQP4

iHiD (approved):

app/javascript/utils/regex-check.ts
ErikSchierboom: Could the various check be merged into a single regex?

PR #6824, opened 2024-02-22
Improve code search

ErikSchierboom:
- Extract criteria less than length three to separate test
- Simplify unit tests
- Switch to match
- Add command to escape opensearch simple query string
- Improve code search

### DIFF
--- a/app/commands/solution/search_via_representations.rb
+++ b/app/commands/solution/search_via_representations.rb
@@ -14,7 +14,7 @@ class Solution::SearchViaRepresentations
     @page = page.present? && page.to_i.positive? ? page.to_i : DEFAULT_PAGE
     @per = per.present? && per.to_i.positive? ? per.to_i : self.class.default_per
     @order = order&.to_sym || :most_popular
-    @criteria = criteria&.split.to_a
+    @criteria = criteria&.strip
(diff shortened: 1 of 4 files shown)


app/commands/solution/search_via_representations.rb
```diff
@@ -103,15 +98,33 @@ def search_sort
...
+  end
+
+  memoize
+  def criteria_query_terms
+    criteria.filter_map do |value|
+      next if value.size < MIN_CRITERIA_LEN
 
-    [value].flatten
+      term = Infrastructure::EscapeOpensearchSimpleQueryStringTerm.(value)
+      "#{term}~#{CODE_TERM_FUZZINESS}"
```
ErikSchierboom, line 120: The fuzziness allows even smaller words to be different in one position. E.g. `class` and `closs` are considered a match. By default, opensearch will do fuzziness when it thinks it should, but it would not apply it to words as small as 10 characters. For us, I think this might be useful.

ErikSchierboom (commented):
See https://forum.exercism.org/t/search-by-code-doesnt-really-work-in-community-solutions/8263/3 for context.

iHiD (approved):
Go for it!

PR #6829, opened 2024-02-27
Fix crash caused by Google Translate

dem4ron:
https://github.com/exercism/website/assets/66035744/6f7d9506-2030-4633-ba2e-3226b3eae888


ErikSchierboom (approved):
Hopefully we'll have a proper fix downstream at some point :)

PR #6864, opened 2024-04-04
Upgrade dependencies

ErikSchierboom:


app/commands/submission/representation/process_results.rb
```diff
@@ -79,7 +79,7 @@ def handle_generated!
   end
 
   def create_notification!
-    return unless exercise_representation.has_feedback?
+    nil unless exercise_representation.has_feedback?
```
iHiD: I don't love that this return is removed as it's a guard. Maybe add a `rubocop#disable` here?

PR #6894, opened 2024-05-01
Use new mandate rails gem

iHiD:


iHiD:
Closing as stale: this has been superseded or gone untouched too long to be worth reviving.

PR #6914, opened 2024-05-15
Handle tracks without exercises gracefully on journey page

ErikSchierboom:


app/javascript/components/journey/types.ts
```diff
@@ -30,6 +30,8 @@ export class TrackProgress {
   progressChart: TrackProgressChart
 
   get completion(): number {
+    if (this.numExercises === 0) return 100
```
ErikSchierboom: The UI will show `0/0`, so I think showing this as `100%` completed might make sense. Let me know if you prefer to list it as `0%`.

dem4ron: I think 100% is fine.

dem4ron (approved):

app/javascript/components/journey/types.ts
ErikSchierboom:
```suggestion
    if (this.numExercises === 0) return 0
```

iHiD (approved):

PR #6935, opened 2024-06-06
Add track-specific 48in24 overview page

ErikSchierboom:


config/routes.rb
```diff
@@ -249,6 +249,7 @@
   resources :challenges, only: [:show] do
     post :start, on: :member
     get :implementation_status, on: :member
+    get "implementation_status/:track_slug" => "challenges#track_implementation_status", on: :member, as: :track_implementation_status
```
ErikSchierboom, line 252: @iHiD This is not the most elegant solution. Feel free to fix this in a follow-up PR

PR #6986, opened 2024-07-10
Display docs by position in config

ErikSchierboom:


db/migrate/20240710082636_add_position_to_docs.rb
```diff
@@ -0,0 +1,14 @@
...
+    return if Rails.env.production?
+
+    add_column :documents, :position, :integer, null: true, limit: 1
+    add_index :documents, [:track_id, :position]
+    add_index :documents, [:section, :position]
+
+    # The positions will be fixed manually by running Git::SyncMainDocss
+    Document.update_all(position: 0)
+
+    change_column_null :documents, :position, false
```
iHiD: I think we can just set default to 0 and null to false on the initial LOC

PR #6990, opened 2024-07-11
Auto update maintainership

ErikSchierboom:
Whenever we're adding people to track teams in GitHub, I always have to remember to also give them the maintainer role. This PR automates that.

Some open questions: 

- Do we want to support having maintainers that are _not_ part of a track team?
- Are only track team members eligible to be a maintainer, or are there more teams?


iHiD:
I thought I remembered this not working before, as the server doesn't have admin permissions. Am I misremembering?

- Do we want to support having maintainers that are not part of a track team?

Yeah, possibly. Maybe people responsible for docs or snippet etc? I'm not hugely fussed either way though - I suspect there's no-one in this bucket atm.

- Are only track team members eligible to be a maintainer, or are there more teams?

Is this the same question as above? It seems to be the same thing, written in the inverse. Am I misunderstanding?

ErikSchierboom:
> I thought I remembered this not working before, as the server doesn't have admin permissions. Am I misremembering?

That was different as that tried to update things in GH. We're now doing things the other way around, updating our database when GH changes. This new strategy should work well, as bootstrap a track will add them to a GH team which we get notified about. 

> Yeah, possibly. Maybe people responsible for docs or snippet etc? I'm not hugely fussed either way though - I suspect there's no-one in this bucket atm.

🤷 It's not very hard to add them later on if we decide we want that. We can start with "only" having track team members auto become maintainers.

> Is this the same question as above? It seems to be the same thing, written in the inverse. Am I misunderstanding?

I phrased this _very_ badly, sorry. What I meant was: do we want to allow for exceptions, as in: can someone who is _not_ in a track team be a maintainer? If so, we need some way of preventing the code in this PR from removing the maintainers role from that user when it runs.

iHiD (approved):

PR #6991, opened 2024-07-12
Simplify track creation

ErikSchierboom:
With this PR, one can just pass in a track's slug to create a new track (as opposed to passing in the repo url).

It also adds the track directory to the git safe directories list, and the triggers a new deploy, which is needed for the permissions to apply.

### DIFF
--- a/app/commands/track/set_file_system_permissions.rb
+++ b/app/commands/track/set_file_system_permissions.rb
@@ -21,9 +21,7 @@ class Track::SetFileSystemPermissions
     Kernel.system("git config --global --add safe.directory #{safe_directory}")
   end
 
-  def trigger_redeploy!
-    Github::DispatchWorkflow.("website", "deploy.yml")
-  end
+  def trigger_redeploy! = Infrastructure::TriggerRedeploy.()
 
   def safe_directory = "/mnt/efs/repos/#{track.slug}"
 end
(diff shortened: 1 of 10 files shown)


app/commands/track/set_file_system_permissions.rb
```diff
@@ -0,0 +1,21 @@
+class Track::SetFileSystemPermissions
+  include Mandate
+
+  initialize_with :track
+
+  def call
```
iHiD, line 6: Could we add a guard clause to this to check if a track is already safe? Just thinking of any future infinitely recursive deploys we could end up in if something goes wrong.

ErikSchierboom: Good idea. I've added the check

app/commands/track/set_file_system_permissions.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    add_safe_directory!
+
+    # A redeploy of the website is required to pick up the new safe directory
+    trigger_redeploy!
+  end
+
+  private
+  def safe_directory_already_set?
+    Kernel.system("git config --global --get safe.directory #{safe_directory}")
```
ErikSchierboom, line 17: According to the docs, `system` will return `true` when the exit code is 0, and `false` when it isn't zero. And conveniently, the git command returns 1 when the safe directory isn't set, and 0 when it is

iHiD (approved):

PR #7000, opened 2024-07-30
Fix destroying of team member not working

ErikSchierboom:


app/models/github/team_member.rb
```diff
@@ -1,8 +1,6 @@
 class Github::TeamMember < ApplicationRecord
-  has_one :user,
-    foreign_key: :uid,
-    primary_key: :user_id,
-    class_name: "User",
+  belongs_to :user,
+    primary_key: :uid,
     inverse_of: :github_team_memberships,
-    dependent: :destroy
+    optional: true
```
iHiD: There is currently a [not null constraint on the database](https://github.com/exercism/website/blob/main/db/schema.rb#L529), so this isn't actually optional. So either the DB or this line should be changed.

ErikSchierboom: I've changed things a bit, but the factory is not setup well enough apparently:

> ActiveRecord::RecordInvalid: Validation failed: User must exist

Any ideas?

test/factories/github/team_members.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :github_team_member, class: 'Github::TeamMember' do
-    user_id { SecureRandom.hex }
+    user { create(:user, uid: SecureRandom.hex) }
```
iHiD, line 3:
```suggestion
    user
```

I'm not sure why you need to specify `uid`here. I think just a normal user is fine.

test/commands/webhooks/process_membership_update_test.rb
```diff
@@ -6,6 +6,8 @@ class Webhooks::ProcessMembershipUpdateTest < ActiveSupport::TestCase
     team_name = 'team11'
     org = 'exercism'
 
+    create(:user, uid: user_id)
```
iHiD, line 9: Do you also need a `provider` here?

app/models/github/team_member.rb
iHiD: This is green now, so I guess this is an old comment?

app/models/github/team_member.rb
```diff
@@ -1,8 +1,5 @@
 class Github::TeamMember < ApplicationRecord
-  has_one :user,
-    foreign_key: :uid,
-    primary_key: :user_id,
-    class_name: "User",
-    inverse_of: :github_team_memberships,
-    dependent: :destroy
+  belongs_to :user,
+    primary_key: :uid,
+    inverse_of: :github_team_memberships
```
iHiD, line 4: I don't think this is right. Looking at the schema, there is a `user_id` column. So the primary key is the `user_id` column on this table. 

We can remove the user_id column, in which case we need to remove that column and add a `uid` column instead. But I feel like we should just be referencing a normal `user_id` instinctively.

ErikSchierboom: That would be better, as `user_id` is currently actually the `uid` (it's named user ID in github).

ErikSchierboom: The original idea was to decouple the table as we might want to query non-DB users, but that hasn't really happend. I'll do a separate PR to fix this. I'll also check the organization members.

ErikSchierboom:
Superseded by #7027

PR #7027, opened 2024-08-13
Make user_id foreign key for github team members

ErikSchierboom:


app/commands/github/team_member/create.rb
```diff
@@ -1,14 +1,16 @@
 class Github::TeamMember::Create
   include Mandate
 
-  initialize_with :user_id, :team_name
+  initialize_with :github_uid, :team_name
```
ErikSchierboom, line 4: I've made it explicit that this is the GitHub UID.

iHiD (approved):
Yeah, this looks loads better now. Thanks dude.

PR #7042, opened 2024-08-28
Add track maintainers status page

ErikSchierboom:


app/commands/track/retrieve_maintainers_status.rb
```diff
@@ -0,0 +1,90 @@
+class Track::RetrieveMaintainersStatus
```
ErikSchierboom, line 1: This is a relatively expensive command. What's the best way to cache this data? Redis?

app/commands/track/retrieve_maintainers_status.rb
```diff
@@ -0,0 +1,90 @@
...
+  include Mandate
+
+  MIN_REP_FOR_MEMBER = 50
+  MIN_REP_FOR_CANDIDATE = 100
+  LAST_NUMBER_OF_MONTHS_FOR_REP = 9
+
+  def call = tracks.index_with { |track| track_maintainers(track) }
```
iHiD: Not sure what the right frequency is. Daily? 

```suggestion
  def call 
    Rails.cache.fetch("Track::RetrieveMaintainersStatus", expires_in: 2.hours) do
      tracks.index_with { |track| track_maintainers(track) }
    end
```

app/commands/track/retrieve_maintainers_status.rb
iHiD: See below

PR #7049, opened 2024-09-01
Remove files_for_editor delegation to exercise which is overridden by method later in the file

toy:


toy:
Disregard, I overlooked `prefix: :exercise`, so the delegate gives `exercise_files_for_editor`, and not overridden by `files_for_editor` method

PR #7056, opened 2024-09-04
Add YAML highlighting for YAMLScript

m-dango:
FYI I have not yet tested this


app/javascript/components/misc/CodeMirror/languageCompartment.ts
```diff
@@ -56,6 +56,10 @@ export const loadLanguageCompartment = async (
       const { wren } = await import('@exercism/codemirror-lang-wren')
       return compartment.of(wren())
     }
+    case 'yamlscript': {
+      const { yamlScript } = await import('@codemirror/lang-yaml')
+      return compartment.of(yamlScript())
```
ErikSchierboom:
```suggestion
      const { yaml } = await import('@codemirror/lang-yaml')
      return compartment.of(yaml())
```

PR #7065, opened 2024-09-13
Add support for highlightjs-roc

ageron:
The Roc track currently uses syntax highlighting for Elm because there was no highlightjs plugin for Roc. Now [there is one](https://www.npmjs.com/package/highlightjs-roc), so this PR adds support for it. Hopefully we will also have CodeMirror support for Roc soon.


ErikSchierboom:
Works well!

Screenshot 2024-09-16 at 08 24 51

PR #7079, opened 2024-09-27
Add a begging modal

iHiD:
I've started this but it needs you to properly do it.

- This should be an on-demand modal that I can call whenever (the "calling whenever" bit is done but it needs to be an actual modal)
- Clicking the negative button (Currently "continue without donating") should close the modal.
- It needs to work in dark theme (does the mentoring flow one currently also look a little unreadable with the blue in dark theme - see images below?)
- What do I need to pass into this from the Ruby helper to make it work?

---

Screenshot 2024-09-27 at 14 20 23
Screenshot 2024-09-27 at 14 20 21

### DIFF
--- a/app/helpers/react_components/modals/beg_modal.rb
+++ b/app/helpers/react_components/modals/beg_modal.rb
@@ -2,7 +2,47 @@ module ReactComponents
   module Modals
     class BegModal < ReactComponent
       def to_s
-        super("beg-modal", {})
+        return unless show_modal?
+
+        super("beg-modal", {
+          previous_donor:,
+          request: {
+            endpoint: Exercism::Routes.current_api_payments_subscriptions_url,
+            options: {
+              initial_data: AssembleCurrentSubscription.(current_user)
+            }
+          },
+          links: {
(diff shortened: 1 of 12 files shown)


app/helpers/react_components/modals/beg_modal.rb
```diff
@@ -0,0 +1,49 @@
...
+        false
+      end
+
+      memoize
+      def previous_donor = current_user.total_donated_in_dollars.positive?
+
+      private
+      def slug
+        "beg-modal"
+      end
```
iHiD, line 46: @dem4ron This can be `def slug = "beg-modal"` in future (but I don't want to delay the CI by changing it now :))

app/javascript/components/modals/student/finish-mentor-discussion-modal/DonationStep.tsx
```diff
@@ -94,3 +94,5 @@ export function DonationStep({
     </div>
   )
 }
+
+export default DonationStep
```
iHiD, line 98: This can be removed I think.

PR #7081, opened 2024-09-30
Post merge tweaks

iHiD:


app/views/dashboard/show.html.haml
```diff
@@ -4,6 +4,7 @@
 
 #page-dashboard
   = render ReactComponents::Modals::WelcomeModal.new
+  = render ReactComponents::Modals::BegModal.new
```
iHiD, line 5: We have this twice now. It should probably JUST be under here.

app/views/tracks/show.html.haml
```diff
@@ -4,6 +4,7 @@
 #page-track-show
   = render ViewComponents::Track::Header.new(@track, :overview)
   = render ReactComponents::Modals::TrackWelcomeModal.new(@track)
```
iHiD, line 6: Let's add it here too. No harm having it in two places.

PR #7089, opened 2024-10-04
Update announcement bar logic

iHiD:
This shows the bar to people who have donated, but not in the last 35 days (before it was hidden for anyone that had ever donated).

### DIFF
--- a/app/helpers/view_components/site_header.rb
+++ b/app/helpers/view_components/site_header.rb
@@ -17,15 +17,16 @@ module ViewComponents
     end
 
     def announcement_bar
-      return tag.span("") if !user_signed_in? ||
-                             current_user.donated? ||
-                             current_user.solutions.count < 3
+      return tag.span("") unless user_signed_in?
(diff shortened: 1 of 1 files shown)


app/helpers/view_components/site_header.rb
```diff
@@ -17,15 +17,16 @@ def to_s
...
     def announcement_bar
-      return tag.span("") if !user_signed_in? ||
-                             current_user.donated? ||
-                             current_user.solutions.count < 3
+      return tag.span("") unless user_signed_in?
+      return tag.span("") if current_user.solutions.count <= 5
+      return tag.span("") if current_user.current_subscription
+      return tag.span("") if current_user.donated_in_last_35_days?
 
       link_to(Exercism::Routes.insiders_path, class: "announcement-bar md:block hidden") do
```
dem4ron, line 25: does it correctly take you to `/insiders`?

iHiD: Yep, it will do.

PR #7151, opened 2024-11-08
Add insiders settings

dem4ron:
- **Add InsiderBenefitsForm**
- **Add component**
- **Add migration**
- **Progress**
- **Sort things out better**
- **Permit editing hide thingy**
- **Pass down props**
- **Disable form for non insiders**
- **Remove unused import**
- **Create an insiders settings page**
- **Add new components**
- **Add API endpoints**


app/controllers/api/generate_bootcamp_affiliate_coupon_code.rb
```diff
@@ -0,0 +1,15 @@
+class API::GenerateBootcampAffiliateCouponCode < API::BaseController
+  skip_before_action :authenticate_user!
+  before_action :authenticate_user
+
+  def create
+    return unless current_user.insider?
+    return if current_user.bootcamp_affiliate_coupon_code.present?
```
iHiD: I think these are better places into the command below (the one is already there and I'll add the insiders guard too). I prefer keeping all the logic in one place.

Also, you can't just "return" here as you need to render json, so it should be `return render(json: {})` to be a valid response for a JSON API. You could also raise an error of some description like `raise "Not insider"` which would get caught by your block, but as we have standardised error messages in the API, that's probably not ideal.

So I would make this method:

```
code = User::GenerateBootcampAffiliateCouponCode.(current_user)
render json: { coupon_code: code }
```

or if you want to raise an error if the code is missing:
```
code = User::GenerateBootcampAffiliateCouponCode.(current_user)
if code 
  render json: { coupon_code: code }
else
  render_403(:could_not_generate_coupon_code)
```

And then you need to add translations for `could_not_generate_coupon_code` (see `config/locales/api/en.yml`)

app/commands/user/bootcamp/subscribe_to_onboarding_emails.rb
```diff
@@ -25,7 +25,10 @@ def add_subscriber!
               "has_paid": bootcamp_data.paid?,
               "package": bootcamp_data.package,
               "country_name": bootcamp_data.country_name,
-              "payment_url": bootcamp_data.payment_url
+              "payment_url": bootcamp_data.payment_url,
+              "price": bootcamp_data.price,
+              "has_discount": bootcamp_ata.has_discount?,
```
dem4ron:
```diff
- bootcamp_ata
+ bootcamp_data
```

PR #7167, opened 2024-11-22
Add uiua codemirror support

ErikSchierboom:
@dem4ron This doesn't seem to work locally :( Last time, you helped me with Arturo. Maybe you know what's wrong?


dem4ron:
This PR adds the parser.
https://github.com/exercism/codemirror-lang-uiua/pull/2
After this you need to rebuild it with rollup, then I think it should work. :)

PR #7176, opened 2024-11-29
fix: change face sorting direction

vaeng:
I made a fiddle to test the changes. 
https://jsfiddle.net/297t4rzm/

### DIFF
--- a/app/css/components/faces.css
+++ b/app/css/components/faces.css
@@ -13,6 +13,10 @@
             height: 32px;
             width: 32px;
         }
+        /* leftmose face on top */
+        &:nth-child(1) { z-index: 3; }
+        &:nth-child(2) { z-index: 2; }
+        &:nth-child(3) { z-index: 1; }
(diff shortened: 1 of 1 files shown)


iHiD:
Thanks!

Reflecting on this, I think it will look better with the faces still stacked how they were but with the order of the faces reversed. I think the right places to do this are:
- https://github.com/exercism/website/blob/main/app/helpers/react_components/track/concept_makers_button.rb#L26
- https://github.com/exercism/website/blob/main/app/helpers/react_components/track/exercise_makers_button.rb#L25
(and maybe somewhere else for the track itself)

A `.reverse` on the end of that probably works perfect.

@vaeng Do you want to do this, or shall I throw it @dem4ron's way? :)

vaeng:
If @dem4ron can do this and it is not a big job, I'm happy to have him give it a go. I would probably miss a lot of things :D

PR #7209, opened 2024-12-20
48in24 challenge page to stop counting weeks on Dec 15

glennj:
In order to avoid "Week 49 of #48in24"


glennj:
Unless you'd rather just use `week = 48` and be done with it.

PR #7409, opened 2025-01-29
Add more kinds of exercises

dem4ron:


test/repos/bootcamp_content/projects/maze/exercises/manual-solve/config.json
```diff
@@ -0,0 +1,52 @@
...
+              "name": "getGameResult()",
+              "value": "win",
+              "error_html": "You didn't reach the end of the maze."
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}
```
iHiD, line 52: @dem4ron Rather than adding these here, I **think** it would be better to just directly reference the `bootcamp_content` folder. Else the tests and the exercises and everything could get out of date.

Happy to hear your logic if you think otherwise though!

PR #7429, opened 2025-01-30
Interrupt animation on inspectedTestResult change

dem4ron:


app/javascript/components/bootcamp/SolveExercisePage/Tasks/useTasks.tsx
```diff
@@ -8,38 +8,41 @@ import {
...
-  | 'wasFinishLessonModalShown'
-  | 'setWasFinishLessonModalShown'
->) {
+export function useTasks() {
   const [isFinishModalOpen, setIsFinishModalOpen] = useState(false)
   const [nextExerciseData, setNextExerciseData] = useState<NextExercise | null>(
     null
   )
+  const [modalsShown, setModalsShown] = useLocalStorage<any[]>(
+    'finish-modals-shown',
```
iHiD: Is there a max length key here? If someone finishes 200 exercises, will this key get too large? Would it be better to have a key per exercise, or put this as part of the existing exercise key?

app/javascript/components/bootcamp/SolveExercisePage/TestResultsView/InspectedTestResultView.tsx
```diff
@@ -9,9 +9,17 @@ import {
...
+import useTaskStore from '../store/taskStore/taskStore'
 
 function _InspectedTestResultView() {
   const { result, viewContainerRef, firstExpect } = useInspectedTestResultView()
+  const { testSuiteResult } = useTestStore()
+  const { wasFinishLessonModalShown } = useTaskStore()
+
+  const isSpotlightActive = useMemo(() => {
+    if (!testSuiteResult) return false
+    return !wasFinishLessonModalShown && testSuiteResult.status === 'pass'
```
iHiD: I have a preference for not having && where possible and also not having `!x && y`. I think it's clearer to have an extra guard clause.

```suggestion
    if (!testSuiteResult) return false
    if (wasFinishLessonModalShown) return false
    
    return testSuiteResult.status === 'pass'
```

app/javascript/components/bootcamp/SolveExercisePage/TestResultsView/TestResultsButtons.tsx
```diff
@@ -1,19 +1,63 @@
...
-  const { setInformationWidgetData } = useEditorStore()
-
+  const { setInformationWidgetData, setReadonly } = useEditorStore()
+  const { wasFinishLessonModalShown } = useTaskStore()
+  const { shouldAutoplayAnimation } = useAnimationTimelineStore()
   const { shouldAnimate } = useShouldAnimate(testSuiteResult)
 
+  const isSpotlightActive = useMemo(() => {
+    if (!testSuiteResult) return false
+    return !wasFinishLessonModalShown && testSuiteResult.status === 'pass'
```
iHiD: Same comment as above.

iHiD: Should this be deduped somehow? Or is it ok to have it twice. Feels maybe a little risky to have the same condition in two places?

app/javascript/components/bootcamp/SolveExercisePage/TestResultsView/TestResultsButtons.tsx
```diff
@@ -1,19 +1,63 @@
...
+  }, [wasFinishLessonModalShown, testSuiteResult?.status])
+
+  useEffect(() => {
+    if (isSpotlightActive) {
+      setReadonly(true)
+    } else {
+      setReadonly(false)
+    }
```
iHiD: (I don't know `ifSpotlightActive` is a boolean or not. It looks like sometimes it's undefined above - so I've include the `!!` but if you guarantee it's a bool, you don't need this)
```suggestion
    setReadonly(!!isSpotlightActive)
```

app/javascript/components/bootcamp/SolveExercisePage/TestResultsView/TestResultsButtons.tsx
```diff
@@ -1,19 +1,63 @@
...
+          if (test.animationTimeline) {
+            const timeline = test.animationTimeline
+            if (test.testIndex === idx) {
+              timeline.timeline.play()
+            } else {
+              timeline.pause()
+            }
+          }
+        })
+      }
```
iHiD: Might be worth extracting this to its own function? 

```
function autoplayAnimation(idx) {
  if (!shouldAutoplayAnimation) return

  testSuiteResult.tests.forEach((test) => {
    // ...
```

PR #7453, opened 2025-02-03
at at deduped

kotp:
In the Guest List introduction text, there is a repeated "at at" that I noticed during the video today.  This corrects that by removing the second (or is it the first) "at".


iHiD:
Thanks :)

PR #7499, opened 2025-02-14
WIP bonus improvements

iHiD:


bootcamp_content/projects/maze/exercises/look-around/config.json
```diff
@@ -347,7 +347,7 @@
               "name": "getAddedLineCount(18)",
               "matcher": "toBeLessThanOrEqual",
               "value": 13,
-              "error_html": "You added more than 13 lines of code. (%actual%)"
+              "error_html": "You added more than %actual% lines of code."
```
dem4ron, line 350: I think this will be incorrect like this, because actual is what they added, and 13 is the required number of lines.

iHiD:
```suggestion
              "error_html": "You added %actual% lines of code (more than 13)"
```

PR #7575, opened 2025-03-03
Fix race condition in bootcamp seeds

dem4ron:


iHiD:
@dem4ron This needs rubocop

PR #7598, opened 2025-03-06
Update PronounsForm.tsx

mattshu:
Removed `required` flags for pronoun input fields. I haven't taken a trip down the Ruby track yet to know how to hide the `Pronouns: //` box on the public profile if omitted but some day! Thanks!


iHiD:
Thanks!

PR #7643, opened 2025-03-13
Set arity only once, using `FunctionStatement` data

dem4ron:


app/javascript/components/bootcamp/CustomFunctionEditor/useCustomFunctionEditorHandler.ts
```diff
@@ -85,6 +86,7 @@ export function useCustomFunctionEditorHandler({
       setLatestValueSnapshot(value)
 
       const evaluated = interpret(value, context)
+      console.log('eval', evaluated)
```
iHiD: Stray?

app/javascript/components/bootcamp/CustomFunctionEditor/useCustomFunctionEditorHandler.ts
```diff
@@ -98,6 +100,9 @@ export function useCustomFunctionEditorHandler({
         return
       }
 
+      const fnStatement = evaluated.meta.statements[0] as FunctionStatement
+      setArity(fnStatement.parameters.length)
```
iHiD, line 103: Bold approach. I like it. What happens if there's a syntax error and there are no statements etc?

dem4ron: It won't reach this bit of code, if there is a syntax error. It shows the error then exits. (that block is just above this line)
Do you think there is a more robust way to do this? Afaik this is guaranteed if the code can compile.

iHiD: OK, I'm fine with it if you can't break it in testing :) Shall we merge?

PR #7667, opened 2025-03-19
Remove duplicate text from Welcome Modal 

CarlosRascheVi:
Starting the AWK Track I noticed some duplicated text in the welcome modal.

I believe this quick change should fix the typo.

Kept the duplicated word "these" in the link to learning resources as I thought it was intended that way.


iHiD:
Thank you :)

PR #7774, opened 2025-04-26
Change style on active curtain or diff buttons

dem4ron:


dem4ron:
This was solved by another PR

PR #7839, opened 2025-05-15
Add GitHub syncer

dem4ron:
### Jeremy
- [x] Deploy/Reset config gem (and add key in production)
- [x] Guard settings serverside
- [x] Check GH Settings for the Syncer point to exercism, not localhost.

### Aron
- [x] Temporarily only show widgets/pages to lifetime insiders


app/views/tracks/exercises/show.html.haml
```diff
@@ -31,6 +31,9 @@
               = render "tracks/exercises/show/action_box_locked", exercise: @exercise, user_track: @user_track
             - else
               = render "tracks/exercises/show/action_box_locked_by_hello_world", exercise: @exercise, user_track: @user_track, track: @track
+        - if @solution
```
iHiD:
```suggestion && current_user.lifetime_insider?
        - if @solution
```

PR #7890, opened 2025-06-03
Add Chat

dem4ron:
- **Add a basic UI**
- **Add basic streaming utils**
- **Add AudioRecorder and tools**
- **Make visualizer a littlebit better**
- **Tweak visualizer**
- **Add variable for drama level**


app/commands/bootcamp/chat_message/trigger_llm.rb
```diff
@@ -1,7 +1,6 @@
 class Bootcamp::ChatMessage
-  class TriggerLLM
+  class TriggerLlm
```
iHiD: This was right before. You just needed to restart your server. Can you revert pls?

app/commands/bootcamp/chat_message/trigger_llm.rb
```diff
@@ -1,5 +1,5 @@
 class Bootcamp::ChatMessage
-  class TriggerLlm
+  class TriggerLLM
```
iHiD, line 2: Ha :)

PR #7911, opened 2025-06-07
Add typing/gradual to track.rb

codingthat:
Finishes implementation with a filter addition (linter + docs already done).

Cf. https://github.com/exercism/docs/pull/581 and http://forum.exercism.org/t/adding-typing-gradual-to-the-accepted-tags-list/17127/13


iHiD:
Thanks!

PR #7945, opened 2025-06-13
Persist data better

dem4ron:
- **Make a few React Components persistent**
- **Persist queryClient**


app/javascript/utils/react-bootloader.tsx
```diff
@@ -238,6 +238,7 @@ const render = (elem: HTMLElement, component: React.ReactNode) => {
     </React.StrictMode>
   )
   document.addEventListener('turbo:before-frame-render', () => {
+    if (elem.dataset.persistent === 'true') return
```
iHiD, line 241: Nice and simple!

PR #7947, opened 2025-06-14
Prepare for etag caching

iHiD:


app/controllers/application_controller.rb
```diff
@@ -203,6 +211,14 @@ def set_log_level
     end
   end
 
+  # This is used by cloudfront to ensure that we never send
+  # publically signed-out cached content to a signed-in user.
+  # If this cookie is set then they should only receive privately
+  # cached versions of pages.
+  def set_user_id_header
```
SleeplessByte, line 218: You want to add the header 
```
Vary: X-User-Id
```
To tell the browser to do the same.

Vary should be a comma separated list of all headers that compose a cache key for a url. So if you will potentially generate a different page based on the Accept-Language, then AL should also be in the Vary list.

You dont want things like User-Agent in there because... There are so many that all requests will have a high chance of cache miss.

Sidenote: cloudfront iirc does NOT respect this header and you need to manually set the header as cache key if you intend to use cloudfront for caching

app/controllers/tracks_controller.rb
```diff
@@ -3,6 +3,11 @@ class TracksController < ApplicationController
   skip_before_action :authenticate_user!, only: %i[index show about]
 
   def index
+    etag = [Track.active.count]
+    etag << current_user.user_tracks.order(updated_at: :desc).pick(:updated_at) if current_user
```
SleeplessByte, line 7: Rails has an association etagger which does max updated at + SQL query. Any reason to not rely on that?

app/controllers/application_controller.rb
iHiD: Yeah, so cloudfront is set to only use this header as its cache key.

PR #8060, opened 2025-07-22
Add Idris to Functional February badge

dancewithheart:
Forum discussion: https://forum.exercism.org/t/add-idris-to-functional-february/18359


iHiD:
Thanks :)

PR #8107, opened 2025-08-15
Add locales

iHiD:


app/css/pages/localization/index.css
```diff
@@ -0,0 +1,117 @@
+.originals-table {
```
iHiD: @dem4ron Please always make sure that anything in CSS files like this are wrapped in a `#page-...` for pages or a `c-....` for components. It's essential to avoid CSS leaking into other parts of the site. (I've done this in a later commit here :))

PR #9289, opened 2026-07-24
Auto-sync track GitHub maintained/unmaintained topic

iHiD:
## Problem

Track repos carry a `maintained` / `unmaintained` GitHub **topic**. That topic (not team membership) is what drives the automated PR comments from the `ping-cross-track-maintainers-team` workflow, including "This is an unmaintained repository." Until now the topic was a purely manual label, fully decoupled from who's on the track's GitHub team. So a track could gain a maintainer and still greet every contributor with bots saying it's unmaintained (this is currently happening on Fortran).

## Change

- **`Track::UpdateGithubMaintenanceStatus`** — reads the track repo's topics and toggles the plain `maintained` ↔ `unmaintained` pair based on whether the track has any members on its GitHub team:
  - has a maintainer + currently `unmaintained` → set `maintained`
  - no maintainers + currently `maintained` → set `unmaintained`
  - Leaves nuanced categories (`wip-track`, `maintained-autonomous`, `maintained-solitary`) and all other topics (e.g. `community-contributions-paused`) untouched. No-ops in development.
- Deferred from **`Github::TeamMember::Create` / `Destroy`** whenever a *track-team* membership actually changes (guarded on `previously_new_record?`), so it fires on both the real-time webhook and the `SyncTeamMembersJob` safety net, but not on no-op resyncs.

## Notes

- This is **promote-and-demote**: if a track's last maintainer leaves the GitHub team, the repo flips back to `unmaintained` automatically.
- It triggers on the *next* membership change, so it does not retroactively fix repos that already have maintainers (e.g. Fortran needs a one-off manual topic fix).

## Tests

New `update_github_maintenance_status_test.rb` (promote / demote / both no-op cases / nuanced-category protection / dev guard) plus updated Create/Destroy tests. All green.

### DIFF
--- a/app/commands/github/team_member/create.rb
+++ b/app/commands/github/team_member/create.rb
@@ -5,7 +5,10 @@ class Github::TeamMember::Create
 
   def call
     user.github_team_memberships.find_or_create_by!(team_name:).tap do |team_member|
-      User::UpdateMaintainer.(user) if team_member.previously_new_record? && team_member.track_id
+      next unless team_member.previously_new_record? && team_member.track_id
+
+      User::UpdateMaintainer.(user)
+      Track::UpdateGithubMaintenanceStatus.defer(team_member.track)
     end
   end
 end
--- a/app/commands/github/team_member/destroy.rb
+++ b/app/commands/github/team_member/destroy.rb
@@ -5,6 +5,9 @@ class Github::TeamMember::Destroy
 
   def call
     team_member.delete
-    User::UpdateMaintainer.(team_member.user) if team_member.track_id
+    return unless team_member.track_id
+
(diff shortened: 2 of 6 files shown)


ErikSchierboom:
Before I review this, isn't this what the maintenance script is for that I've written? The main downside being that that doesn't run automatically, but it _does_ cover all cases. Maybe that could be adapted into this script?

iHiD:
> Before I review this, isn't this what the maintenance script is for that I've written? The main downside being that that doesn't run automatically, but it does cover all cases. Maybe that could be adapted into this script?

Quite possibly! I couldn't remember where that script was and you were away on holiday so I thought I'd do this, but then decided I'd like your oversight :D

PR #9305, opened 2026-07-30
Continue i18n extraction

dem4ron:


iHiD:
Closing as superseded by #9514. All nine commits here are carried unchanged in `i18n-foundation`, with your authorship, so nothing is lost. Merging this separately would only mean resolving conflicts against an old base. Your review of #9514 before it merges would be really valuable, since it now contains this work. Thank you!

PR #9351, opened 2026-08-12
Make community solution pages long-cacheable at the CDN

iHiD:
## Why

Community solution pages are ~99% crawler traffic and ~92k origin requests/day, sitting at a **41% Cloudflare hit rate** because the TTL is 5-20 minutes. The site is crawled roughly daily, so a TTL shorter than the crawl interval buys almost nothing — each crawler arrives just as the entry expires. Hit rate for a daily requester is roughly `1 - 1/TTL_days`, so 24h gives ~0% and 30 days gives ~97%.

Raising the TTL requires that nothing in the cached HTML changes independently of the solution itself.

Reputation was the worst offender: it's rendered into every one of an author's solution pages and changes whenever they earn rep anywhere on the site, so a correct purge would fan out across all their solutions continuously. Flair, avatar URLs and profile-existence branches have the same shape, and commenter reputation/flair were baked into the comments payload with genuinely unbounded fan-out.

## What

**Move user identity out of cached HTML** so that fan-out collapses to nothing:

- `GET /api/v2/users/:handle.json` returning handle, avatar_url, flair, has_profile, and reputation (total plus per-track map). Public, no auth, **verified to set no cookies** so Cloudflare will actually cache it. Per-track reputation is one grouped query, not a SUM per track.
- React components (reputation, flair, avatar, profile link) consuming it, sharing one request per handle per page.
- Own-user data stays fresh via a client-side `?t=<minute>` cache-bust rather than a shorter TTL.
- Reputation and flair stripped from `SerializeSolutionComment` and `SerializeCommunitySolution`.

**Then cache the page:**

- `cache_public_action!` takes an `edge_ttl`, opted into only by `Tracks::CommunitySolutionsController#show`. Emits `max-age=1, public, s-maxage=<~30 days>` (jittered). One second in the browser so purges are authoritative, 30 days at the edge.
- `Cloudflare::PurgeUrls`, `Solution::InvalidateCloudflareCache`, `User::InvalidateCloudflareCaches`, with purges deferred from the write paths that actually change the page. Solutions are reachable at **two** URLs (handle and legacy UUID); both are purged.

Purge volume is ~2,800/day against a 1,500/sec limit, because the vast majority of the ~5M published solutions are inert.

**Extract the mutations into commands**, so the purge hangs off the operation rather than off the controller:

- `Solution::Comment::Update`, `Solution::Comment::Destroy`
- `Solution::AllowComments`, `Solution::DisallowComments`
- `Solution::Star::Create`, `Solution::Star::Destroy`
- `API::SolutionsController#unpublish` now calls `Solution::Unpublish`

That last one is a behaviour fix rather than a move. The endpoint had inlined the `published_at`/`published_iteration_id` update since the original publishing paradigm commit, so `Solution::Unpublish` was only ever reached via `Solution::Uncomplete` and `Iteration::Destroy`. The user-facing unpublish button skipped tags, representation recache, snippet, num_loc and the exercise counter. It no longer does.

## Before merging

- [ ] Add `cloudflare_zone_id` and `cloudflare_api_token` secrets. **The purge silently no-ops without them.**
- [ ] Apply the terraform change setting `browser_cache_ttl = 0`, so Cloudflare respects origin `max-age` instead of rewriting it to 4 hours.

## Not in this PR

- `csrf_meta_tags` writes `session[:_csrf_token]`, so full page responses likely carry `Set-Cookie`, which Cloudflare refuses to cache. May be a significant part of the current 41%, and worth chasing separately.
- `@other_solutions`, representation counters and `approach.num_solutions` are per-exercise values that still can't be purged cheaply. Same endpoint pattern applies; follow-up.

https://claude.ai/code/session_01Ht2xRGuicjz1Lqxtjew1Bu


app/controllers/api/community_solution_comments_controller.rb
```diff
@@ -23,6 +23,7 @@ def update
     return render_403(:solution_comment_not_accessible) unless comment.author == current_user
 
     if comment.update(content_markdown: params[:content])
+      Solution::InvalidateCloudflareCache.defer(@solution)
```
iHiD: Extract into `Comment::Update`

app/controllers/api/community_solution_comments_controller.rb
```diff
@@ -38,6 +39,7 @@ def destroy
     return render_403(:solution_comment_not_accessible) unless comment.author == current_user
 
     if comment.destroy
+      Solution::InvalidateCloudflareCache.defer(@solution)
```
iHiD: Extract into `Comment::Destroy`

app/controllers/api/community_solution_comments_controller.rb
```diff
@@ -48,12 +50,14 @@ def destroy
 
   def enable
     @solution.update!(allow_comments: true)
+    Solution::InvalidateCloudflareCache.defer(@solution)
```
iHiD: Extract into `Solution::AllowComments`

app/controllers/api/community_solution_comments_controller.rb
```diff
@@ -48,12 +50,14 @@ def destroy
...
 
     render json: {}
   end
 
   def disable
     @solution.update!(allow_comments: false)
+    Solution::InvalidateCloudflareCache.defer(@solution)
```
iHiD: Extract into `Solution::DisallowComments`

app/controllers/api/community_solution_stars_controller.rb
```diff
@@ -3,6 +3,7 @@ class API::CommunitySolutionStarsController < API::BaseController
 
   def create
     @solution.stars.create_or_find_by!(user: current_user)
+    Solution::InvalidateCloudflareCache.defer(@solution)
```
iHiD: Extract

app/controllers/api/community_solution_stars_controller.rb
```diff
@@ -14,6 +15,7 @@ def create
 
   def destroy
     @solution.stars.where(user: current_user).destroy_all
+    Solution::InvalidateCloudflareCache.defer(@solution)
```
iHiD: Extract

app/controllers/api/solutions_controller.rb
```diff
@@ -92,7 +92,11 @@ def published_iteration
   end
 
   def unpublish
+    # TODO: This bypasses Solution::Unpublish entirely, and so skips tags,
+    # representation recache, snippet, num_loc and the exercise counter.
+    # That's a pre-existing bug, deliberately not fixed here.
     @solution.update!(published_at: nil, published_iteration_id: nil)
+    Solution::InvalidateCloudflareCache.defer(@solution)
```
iHiD: Extract

PR #9364, opened 2026-08-13
Make user_tracks.reputation trustworthy, then read from it

iHiD:
Two reputation aggregates are amongst the most expensive queries on the Aurora cluster, whose capacity is driven almost entirely by physical reads (`ACU = 2.11 + 0.01284 × read IOPS`, ~$1.84/month per sustained read/second):

- `SELECT user_id, SUM(value) … GROUP BY user_id` — 5th by I/O wait even in a quiet 02:00–04:00 window
- `SELECT SUM(value) WHERE track_id = ? AND user_id = ?` — 3rd highest by total DB load cluster-wide

We already denormalise this into `user_tracks.reputation`. We just weren't maintaining it properly, and weren't reading it. This PR fixes both.

## Why the column couldn't be trusted

Measured across all 3,094,416 rows on production: 2,703 rows (0.087%) had drifted, 30,815 total absolute drift, worst single row 916.

| | |
|---|---|
| Understated (tokens exist, column never updated) | 1,139 |
| Overstated (column set, tokens since deleted) | 1,564 |

Two distinct gaps, matching those two numbers:

1. **Nothing decremented the column when a token was deleted.** `UserTrack::Reset` destroys a track's `PublishedSolutionToken`s, and `User::ReputationToken::Create` was the only thing that ever wrote the column — so the reputation stayed behind. Hence overstated > understated.
2. **A token could be created before the `UserTrack` existed.** `User::ReputationToken::Create` did `UserTrack.find_by(user:, track:)` and silently skipped the update when it found nothing. Any reputation awarded in that window was lost forever, which is why the understated rows are all recent.

### 1. Maintenance (`user_tracks.reputation` is now correct going forward)

`User::ReputationToken` already has an `after_commit` that keeps `users.reputation` in sync. That callback fires on **create, update and destroy**, so it's the right place for this too — it now updates the matching `user_track` as well, and the (now redundant) call in `User::ReputationToken::Create` is removed. Deletions therefore decrement.

For the race, `UserTrack` gained an `after_create_commit` that adopts any reputation already awarded for that track. Joining a track now picks up tokens earned before you joined, rather than dropping them.

**`users.reputation` needed neither fix**, and I've deliberately left it alone: its `after_commit` has no `on:` filter so it already runs on destroy, and it recomputes from *all* of the user's tokens rather than looking a row up, so it has no equivalent ordering race. Stating this explicitly since it was worth checking rather than assuming.

`UserTrack::UpdateReputation` is unchanged and still recomputes the full `SUM` to set the column. That's correct but wasteful — it could be made incremental (`+= value` on create, `-= value` on destroy). I've left that for a follow-up because it's a behaviour change to a writer that is now on more paths than before, and the write path is nothing like as hot as the read path this PR fixes.

### 2. Reads switched to the column

- **`User#reputation_for_track`** now reads `user_tracks.reputation` instead of running an unmemoised `SUM`. This also kills the SUMs behind `Exercise::Representation#max_reputation` and `Exercise::Representation::CreateSearchIndexDocument#max_reputation`, which both call it.
- **`Exercise::Representation::Recache#prestigious_solution`** — the worst offender. It runs *synchronously* from `Exercise::Representation::CreateOrUpdate` on **every analysed submission**, and it summed the entire track reputation history of every user who had published a solution with that representation. Cost scaled with how popular the representation was, so the hottest write path did the biggest aggregation.

## Read this before you touch `prestigious_solution`

**The obvious rewrite is 1,300x slower than the code it replaces.** This branch shipped that obvious rewrite first, and it had to be reverted. Please don't re-tread it.

Profiled with `EXPLAIN ANALYZE` on production, worst case: representation 942044, track 16, **64,077 published solvers**.

| Variant | Time |
|---|---|
| **The obvious rewrite** (`SELECT solutions.*` + LEFT JOIN + COALESCE ORDER BY) | **108,000ms** |
| INNER JOIN variant, same shape | 10,930ms |
| **Old production query** (token aggregation) | **1,130ms** |
| Rank-then-fetch + index `(track_id, user_id, reputation)` | 332ms |
| **Rank-then-fetch + index `(track_id, reputation, user_id)`** ← shipped | **81ms** |
| Same but without `reputation > 0` | 2,432ms |

Two root causes:

1. **`SELECT solutions.*` destroys the covering index.** The old query only ever reads `user_id`, so `index_solutions_er_lookup` serves it as a *covering* read — 64,077 rows in 36ms. Selecting full solution rows turns that into 58,728 random row fetches: 63,300ms.
2. **`user_tracks` rows are ~28 KB** — 86 GB across 3.09M rows, thanks to the `summary_data`/`objectives` blobs. Any plan that fetches a `user_tracks` row costs ~0.75ms cold. Plans must read reputation from an index, never from the row.

So the shipped version keeps the **rank-then-fetch** shape the original code had — find the winning `user_id` reading nothing but indexes, then fetch that one solution — with `user_tracks` as the reputation source instead of a token aggregation.

**`ut.reputation > 0` is load-bearing for performance, not just semantics.** Without it the optimiser abandons the reputation index and the query goes from 81ms to 2,432ms. It carries a comment saying exactly that, because it otherwise looks redundant. It also still provides the `oldest_solution` fallback.

### Migration

Adds `(track_id, reputation, user_id)` on `user_tracks` as `index_user_tracks_track_reputation_user`. **Column order matters:** reputation *second* is what allows the reverse range scan with early termination. Reputation third costs 4x (332ms). The index has already been created manually on production, so the migration uses `if_not_exists: true` and is a safe no-op there while still being correct for other environments and `schema.rb`.

**Open question for the reviewer:** should we also add `(track_id, user_id, reputation)`? It would make `reputation_for_track`'s point lookup covering, avoiding a 28 KB row fetch. But it's a second index on an 86 GB table and is untested, and the better fix is really to move `summary_data` out of the row. I've not added it.

## Two semantic decisions worth reviewing

**a) The `prestigious_solution` candidate set.** The old query's candidates were "published solvers who have at least one reputation token on this track". The new one's are "published solvers with a `user_tracks` row whose reputation is > 0". In practice these coincide: publishing a solution on a track requires having joined it, so a published solver always has a `user_tracks` row. The one genuine edge-case change is a user whose track tokens summed to exactly 0, who would previously have beaten the `oldest_solution` fallback and now doesn't — token values are positive, so this shouldn't occur. The fallback to `oldest_solution` when nobody has any track reputation is preserved and tested.

**b) `reputation_for_track` returns 0 for a track the user hasn't joined**, rather than a live sum. I checked every caller:
- both `max_reputation` callers go via `prestigious_solution`, i.e. a user who has published a solution on the track — so they have a `user_tracks` row. No behaviour change.
- `SerializePublicUser` does *not* use this method; it does its own single grouped query across all tracks at once, and still reports live sums. That's intentional — it's already one cheap indexed query, and it's the one place where showing reputation for a never-joined track would be visible.

No other callers exist.

## Run this around deploy time

The maintenance fix only corrects rows from now on, so the 2,703 already-drifted rows need one correction. This is the shape that ran in 27s on production:

```sql
UPDATE user_tracks ut
LEFT JOIN (SELECT user_id, track_id, SUM(value) AS live
           FROM user_reputation_tokens GROUP BY user_id, track_id) d
  ON d.user_id = ut.user_id AND d.track_id = ut.track_id
SET ut.reputation = COALESCE(d.live, 0)
WHERE ut.reputation <> COALESCE(d.live, 0);
```

It's idempotent and matches exactly what the callbacks now maintain (`SUM(value)` for the pair, 0 when there are no tokens), so re-running it is a no-op.

## Tests

`bin/rails test test/models test/commands test/serializers` — 3,867 runs, green.

Added coverage for the deletion decrement (including `destroy_all`), the creation race, tokens with no track, and the `prestigious_solution` candidate-set behaviour — including the "solver with track reputation but no `user_tracks` row" case and the `oldest_solution` fallback. A few existing tests were updated where they asserted the *old* broken behaviour (e.g. `UserTrack::UpdateReputationTest` set up `user_tracks` with reputations that no longer survive the token callbacks), and three search-index tests gained the `user_tracks` row their prestigious solver was missing.

https://claude.ai/code/session_01UGr9woFrDZjnneiK1XQbG3

### DIFF
--- a/db/migrate/20260813120000_add_track_reputation_user_index_to_user_tracks.rb
+++ b/db/migrate/20260813120000_add_track_reputation_user_index_to_user_tracks.rb
@@ -1,5 +1,7 @@
 class AddTrackReputationUserIndexToUserTracks < ActiveRecord::Migration[7.1]
   def change
+    return if Rails.env.production?
+    
     # Column order matters. reputation *second* is what lets
     # Exercise::Representation::Recache#prestigious_solution do a reverse
     # range scan with early termination (81ms on the worst representation).
--- a/db/migrate/20260813120000_add_track_reputation_user_index_to_user_tracks.rb
+++ b/db/migrate/20260813120000_add_track_reputation_user_index_to_user_tracks.rb
@@ -1,0 +1,13 @@
+class AddTrackReputationUserIndexToUserTracks < ActiveRecord::Migration[7.1]
+  def change
+    # Column order matters. reputation *second* is what lets
+    # Exercise::Representation::Recache#prestigious_solution do a reverse
+    # range scan with early termination (81ms on the worst representation).
+    # Putting user_id second instead costs 4x (332ms).
+    #
+    # Already created manually on production, hence if_not_exists.
+    add_index :user_tracks, %i[track_id reputation user_id],
+      name: "index_user_tracks_track_reputation_user",
+      if_not_exists: true
+  end
+end
--- a/app/commands/exercise/representation/recache.rb
+++ b/app/commands/exercise/representation/recache.rb
@@ -45,19 +45,24 @@ class Exercise::Representation::Recache
       first
   end
 
+  # Picks the published solver with the highest reputation on this track,
+  # falling back to the oldest solution when nobody has any.
+  #
+  # This reads the denormalised user_tracks.reputation column instead of
+  # aggregating the whole reputation history of every published solver.
+  # The LEFT JOIN keeps every published solver in the candidate set: anyone
+  # without a user_tracks row simply scores 0, exactly as they would if they
+  # had no reputation tokens for the track.
   def prestigious_solution
-    user_id = User::ReputationToken.where(
-      track_id: exercise.track_id,
-      user_id: representation.published_solutions.select(:user_id)
-    ).
-      group(:user_id).
-      select("user_id, SUM(value) as total").
-      order('total DESC').
-      first&.user_id
-
-    return oldest_solution unless user_id
-
-    representation.published_solutions.find_by!(user_id:)
+    representation.published_solutions.
+      joins(<<~SQL.squish).
+        LEFT JOIN user_tracks
+          ON user_tracks.user_id = solutions.user_id
+          AND user_tracks.track_id = #{exercise.track_id.to_i}
+      SQL
+      where('COALESCE(user_tracks.reputation, 0) > 0').
+      order(Arel.sql('COALESCE(user_tracks.reputation, 0) DESC'), :id).
+      first || oldest_solution
   end
 
   memoize
--- a/app/commands/exercise/representation/recache.rb
+++ b/app/commands/exercise/representation/recache.rb
@@ -48,21 +48,35 @@ class Exercise::Representation::Recache
   # Picks the published solver with the highest reputation on this track,
   # falling back to the oldest solution when nobody has any.
   #
-  # This reads the denormalised user_tracks.reputation column instead of
-  # aggregating the whole reputation history of every published solver.
-  # The LEFT JOIN keeps every published solver in the candidate set: anyone
-  # without a user_tracks row simply scores 0, exactly as they would if they
-  # had no reputation tokens for the track.
+  # This *must* stay rank-then-fetch: rank on user_ids only, then fetch the
+  # single winning solution. Ranking is served entirely from indexes - the
+  # subquery by index_solutions_er_lookup, the ordering by
+  # index_user_tracks_track_reputation_user - and never touches a solutions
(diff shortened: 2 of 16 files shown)


db/migrate/20260813120000_add_track_reputation_user_index_to_user_tracks.rb
```diff
@@ -0,0 +1,13 @@
+class AddTrackReputationUserIndexToUserTracks < ActiveRecord::Migration[7.1]
+  def change
```
iHiD, line 2:
```suggestion
  def change
    return if Rails.env.production?
    
```

PR #9418, opened 2026-08-20
fix: remove duplicated '(You can always change later.)' in track welcome modal

deepshekhardas:
## What

Fixes #6779

The track welcome modal rendered the text "(You can always change later.)" twice:

> Would you like to start the track in Learning Mode or Practice Mode? (You can always change later.) (You can always change later.)

The parenthetical is already part of the hasLearningModeStep.startTrackInLearningOrPracticeMode translation string (components-modals-track-welcome-modal-LHS-steps.ts), but HasLearningModeStep.tsx also hardcoded it after the translated span. Removed the duplicated hardcoded copy.

## Why

The sentence should only appear once.


iHiD:
@deepshekhardas Thank you! Will merge when green.

PR #9420, opened 2026-08-20
Use @codemirror/lang-go for the Go editor

Semih702:
Hi! I've been using Exercism for years and this is my first contribution to the site, so please do tell me if I've got the wrong end of the stick anywhere, or if this should have started as a forum topic instead.

I ran into this while working through the Go track in the online editor. I was filling in the usual stub:

```go
func Welcome(name string) string {
	panic("Please implement the Welcome function")
}
```

and when I started typing `return na` I expected `name` to be offered. Nothing came up. I assumed autocomplete was simply off in the editor, but then I tried the equivalent on the Python track:

```python
def welcome(name):
    return na
```

and there `name` appears straight away. So it is on — just not for Go.

## Why

Autocompletion is already enabled for every track: `misc/CodeMirror.tsx` uses `basicSetup`, which includes `autocompletion()`. The candidates themselves, though, come from whatever a language registers as `autocomplete` language data, and Go is still wired to the legacy stream mode:

```ts
case 'go': {
  const { go } = await import('@codemirror/legacy-modes/mode/go')
  return compartment.of(StreamLanguage.define(go))
}
```

`@codemirror/legacy-modes/mode/go` only declares `indentOnInput` and `commentTokens`, so Go ends up with no completion sources at all. Python and JavaScript get theirs from `lang-python` and `lang-javascript`.

As far as I can tell this isn't deliberate, just timing. `case 'go'` was added in #1005 in May 2021, when the legacy mode was the only option for Go — `@codemirror/lang-go` wasn't published until February 2024.

## What this changes

It moves Go out of the legacy block and onto `@codemirror/lang-go`, the same way `cpp`, `java`, `python` and `rust` are already wired. Julia had the same treatment in #2074 once a proper package existed for it.

That gives Go local completion of parameters, `var` and `:=` declarations, constants, types and same-file functions, plus keyword and snippet completion. It also swaps the regex-based stream mode for the Lezer parser, so highlighting, folding and indentation come from a real syntax tree rather than a token stream.

## What this isn't

I read the forum thread on [autocomplete in the editor](https://forum.exercism.org/t/autocomplete-in-editor/13370) before opening this, and I want to be clear I'm not trying to reopen that discussion. There's no language server here, no backend, and nothing extra to run or pay for. It only completes names already written in the file the student is looking at. Anything beyond that — stdlib symbols, types, cross-file lookups — would still need gopls and is out of scope.

## Cost

Two new dependencies, `@codemirror/lang-go` and its parser `@lezer/go`. That's roughly 12KB gzipped more than the legacy mode, and it stays in its own lazily-loaded chunk that only the Go track pulls in. I checked the build output to confirm nothing extra lands in the initial bundle.

## Testing

`yarn test` passes in full (162 suites). I added a small test that fails without the change:

- `name` is offered as a completion for the example above
- Go still parses without errors, so highlighting and indentation aren't affected

There were no existing tests around `languageCompartment`, so I kept it to those two — happy to drop it if you'd rather these PRs stay the usual three-file shape.

One caveat: I couldn't run rubocop or haml-lint locally as I don't have Ruby set up on this machine, but this PR doesn't touch any Ruby or HAML.

### DIFF
--- a/app/javascript/components/misc/CodeMirror/languageCompartment.ts
+++ b/app/javascript/components/misc/CodeMirror/languageCompartment.ts
@@ -27,6 +27,10 @@ export const loadLanguageCompartment = async (
       const { gleam } = await import('@exercism/codemirror-lang-gleam')
       return compartment.of(gleam())
     }
+    case 'go': {
+      const { go } = await import('@codemirror/lang-go')
+      return compartment.of(go())
+    }
     case 'javascript':
     case 'typescript': {
       const { javascript } = await import('@codemirror/lang-javascript')
@@ -157,10 +161,6 @@ export const loadLanguageCompartment = async (
     case 'gnu-apl': {
       const { apl } = await import('@codemirror/legacy-modes/mode/apl')
       return compartment.of(StreamLanguage.define(apl))
-    }
-    case 'go': {
-      const { go } = await import('@codemirror/legacy-modes/mode/go')
-      return compartment.of(StreamLanguage.define(go))
     }
     case 'groovy': {
       const { groovy } = await import('@codemirror/legacy-modes/mode/groovy')
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "@codemirror/commands": "^6.7.1",
     "@codemirror/lang-cpp": "^6.0.2",
     "@codemirror/lang-css": "^6.3.1",
+    "@codemirror/lang-go": "^6.0.1",
     "@codemirror/lang-java": "^6.0.1",
     "@codemirror/lang-javascript": "^6.2.1",
     "@codemirror/lang-php": "^6.0.1",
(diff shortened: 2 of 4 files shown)


iHiD:
@Semih702 Awesome. Thank you! I'll merge once green :)

kotp:
> Thank you! I'll merge once green :)

@iHiD  It won't turn green unless someone approves the ci/cd being checked.

Semih702:
@iHiD The failing test (`Maintaining::SubmissionsSummaryTableTest#test_visiting_the_index`) looks flaky rather than related to this PR. It passed on #9426's branch and then failed on `main` with the same code, and it was the only failure out of 38 jobs here.

I can't re-run the workflow myself, so would you mind triggering it again? Happy to rebase if that's easier.

iHiD:
Thanks. @dem4ron will check this and we'll get it live :)

dem4ron (approved):

PR #9514, opened 2026-09-18
i18n foundation: locale routing, catalog loaders, blob-id content lookup

iHiD:
Foundation for serving the site in languages other than English. Nothing is user-visible yet: no locale is in production, so every route, catalog and lookup added here serves English exactly as before until a locale is promoted in `config/locale_roster.json`.

Translations live in `exercism/i18n`, never in this repo or its database. This branch reads them:

- **Locale roster** (`config/locale_roster.json`), one list shared by Ruby and JavaScript, with each locale marked production / wip / planned. The served set is derived per environment; test serves English only.
- **Optional locale prefix on routes.** English stays naked and `/en/*` redirects to it. `config/routes/` is split into locale-scoped routes and the ones that must never be scoped (omniauth callbacks, webhooks, Discourse SSO, admin, moderation, SEO files, test runners, errors, legacy redirects, courses, `/bootcamp`). `rails routes` was diffed before and after: no route lost, 196 scoped, 449 not.
- **Ambient locale in every generated URL**, added to scoped routes and stripped from unscoped ones, so no `link_to` or React prop needs to know.
- **`LocaleSupport` / `LocaleRouting`** with a real browser-variant table (`es-CL`/`es-AR`/`es-419` resolve to `es-419`, bare `pt` to `pt-BR`, `zh-HK` to `zh-TW`), q-value aware `Accept-Language` parsing that does nothing when there is no header, and a path swap that only strips an exact roster locale. Also fixes `?locale=zz` raising `I18n::InvalidLocale` on main today.
- **SEO:** self-canonical locale pages, one reciprocal hreflang map shared by the page head and the sitemap (with `x-default`; `es-419` emitted as `es`), wip locales noindex and absent from hreflang, `<html lang dir>` from the locale.
- **`rails-i18n`**, loaded for roster locales only.
- **The `exercism/i18n` checkout** (`lib/translation_repo.rb`): the website keeps a plain working-tree checkout of `exercism/i18n` at `<efs_repositories_mount_point>/i18n` and reads the repo's own layout straight from it (`locales/<locale>/website/{backend,frontend}.json`, `locales/<locale>/content/<ab>/<cd>/<rest>.<ext>`). No mirror, no pointer files, no S3, no new config key. The checkout's HEAD sha is the translation version, read from `.git/HEAD` (through `refs/heads/main` or `packed-refs`) and re-checked every 10s, so a corrected content file is a change Rails sees.
- **Keeping it fresh:** a push to `exercism/i18n` main runs `TranslationRepo::Sync` from the push webhook, and sidekiq-scheduler runs it every ten minutes. It clones on first run (depth 1, main only) with a sparse checkout of `locales.json` plus the served locales except English, re-applies that set from the roster every run (promoting a locale widens the checkout), and pulls with `--ff-only` under a file lock so two pulls never overlap. Developers clone to the same path (AGENTS.md).
- **Backend catalog loader** (`TranslationRepo::Backend`): holds each locale's `backend.json` in memory, rebuilds the tree and swaps it atomically when the HEAD sha changes, lazily from lookups so Puma and Sidekiq behave the same. A fallback to English for a production locale is reported to Sentry, never treated as normal.
- **Frontend catalog:** Rails serves `GET /i18n/:locale/frontend-:hash.json` itself from the checkout with a one-year immutable `Cache-Control`; the hash is a digest of the file's bytes, memoised per HEAD sha, and a stale hash is a 404, never the current file. The page renders that same-origin URL plus a preload; i18next waits for it before first render, re-checks on every `turbo:load`, and reports missing keys. English stays bundled and is the fallback. Replaces the hard-pinned `lng: 'en'`.
- **Git-sourced content by blob id:** `git_filepath ... translatable: true` looks up the translation stored under the English file's git blob id (read from the tree, no English read needed). Covers exercise instructions, introduction, appends, hints, concept docs, track docs, site docs, blog posts and stories, analyzer comments. When a student's pinned version has no translation, the exercise page shows the latest version's translation under a banner offering "switch to English" or the existing update flow.
- **Cache keys** now carry the locale and the translation version (the checkout's HEAD sha): footer, signed-out header, tracks header fragment, updates article, solution-view S3 key, `Cache::GenerateEtag`. The exercise/concept parsed-HTML S3 caches are keyed by locale and text digest.
- **`rendering_data_cache`** on notifications, activities and reputation tokens is keyed by locale (legacy flat rows read as English, no migration). Also fixes it being rebuilt on every save.
- **Mailers** run the whole action in the recipient's locale, behind `LOCALISE_EMAILS = false`; mailshots forced to English.
- Cherry-picks the nine `i18n-followups` extraction commits (nav, headers, editor, mentoring UI, mailer layouts, alt text, AGENTS.md i18n section) on top.

Shipped switched off, each marked `TODO(iHiD)` in the code, pending product decisions: the signed-in redirect from a naked URL to the user's locale (`REDIRECT_SIGNED_IN_USERS = false`), localised emails, a language picker (nothing writes `user_data.locale` yet), the first-visit prompt for signed-out visitors, the API locale rule (all API output is still English), the `es-419`/`es-ES` split, RTL CSS.

Known gaps: the banner is only on the exercise page (editor and mentoring session show the latest translation without it); a pull changes the version, but nothing purges the CDN, so cached pages stay stale until TTL; Devise's failure app redirects to the English sign-in page; titles and blurbs (the metadata catalogs in `exercism/i18n`) are not wired yet.

Tests: lib, integration, initializers, controllers, helpers, models, mailers, serializers, jobs, commands and jest all run locally after the rewrite, all green; `UserTrackTest#test_active_mentoring_discussions` failed once in the models run (ordering flake on an unordered relation) and passes alone. Routing, controllers, helpers and i18n suites re-run green after merging main (#9515, remove course enrollment). System tests not run.


app/commands/cache/generate_etag.rb
```diff
@@ -16,6 +16,8 @@ def call
   def extra_args
     [
       Cache::KeyForFooter.(current_user),
+      # The locale and its catalog hashes, so a published translation fix is never a 304
```
iHiD: This comment makes no sense.

iHiD:
Superseded. This work was split into two PRs: the string extraction went in as #9551, and everything else (locale routing, the locale layer, the cache work, the Cloudflare purge and the `served` flip) is #9553.

Every line of this branch's content is in those two. The `i18n-foundation` branch stays in place for reference.

